PR #2, opened 2015-04-11
[WIP] remove components submodule

minrk:
This removes the components submodule and compiled css files, instead relying on bower to fetch the dependencies and less to build css.

Running the notebook from master will from now on require npm.

It works already, the only pieces I want to get in before merging are distutils hooks for triggering the npm, bower, and less calls, which I will mostly copy from JupyterHub where this same pattern is already working well.


takluyver:
Continuing in #8.

PR #3, opened 2015-04-12
JQuery dependency and backbone view

SylvainCorlay:
I would like to eventually allow users to write widgets using [backbone native views](https://github.com/akre54/Backbone.NativeView) or [d3 views](https://github.com/akre54/Backbone.D3View) or other drop-in replacements for `Backbone.View` using the DOM manipulation library of their choice.

These replacements for backbone views share the main `View.el` attribute, which is a native DOM element. Hence it should be used to insert the widget view in the page generically rather than `View.$el` which is specific to jquery.

Besides inserting the widget in the page, jquery is only used for some basic css styling and `$.proxy`.

In this PR, I Widget.js I removed the jquery dependency in manager.js, and Widget.js. 

In the very end, I do:
```JavaScript
var WidgetView = Backbone.View.extend(WidgetInterface);
var DOMWidgetView = WidgetView.extend(DOMWidgetInterface);
```
which I would like to move to a different file. The idea, is that anyone could define their own version of DOMWidgetView based on NativeView:
 
```JavaScript
var NativeWidgetView = Backbone.NativeView.extend(WidgetInterface);
var NativeDOMWidgetView = NativeWidgetView.extend(DOMWidgetInterface);
```

### DIFF
--- a/jupyter_notebook/static/notebook/js/codecell.js
+++ b/jupyter_notebook/static/notebook/js/codecell.js
@@ -231,7 +231,7 @@ define([
         var that = this;
         return view_promise.then(function(view) {
             that.widget_area.show();
-            dummy.replaceWith(view.$el);
+            dummy.replaceWith(view.el);
             that.widget_views.push(view);
 
             // Check the live state of the view's model.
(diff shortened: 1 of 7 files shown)


SylvainCorlay:
@jdfreder will reopen after the widget split.

PR #8, opened 2015-04-13
Remove submodules

takluyver:
Supersedes #2.


takluyver:
OK, tests are now passing. Merging this so we can work on the notebook without initialising submodules.

PR #9, opened 2015-04-13
Interim update for dev quickstart

ptone:
Just something to get a few more people started till docs get better

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,19 @@
 
 The Jupyter HTML notebook is a web-based notebook environment for interactive computing.
 
+
+Dev quickstart:
+
+* Create a virtual env (ie jupyter-dev)
+* Clone this repo and cd into it
+* pip install -r requirements.txt
+* pip install -e .
+* ensure that you have node/npm install (ie brew install node)
+* npm install -g 'less@<3.0'
+* python setup.py js
+* python setup.py css
+
 Launch with:
 
     jupyter notebook
+
--- a/README.md
+++ b/README.md
@@ -2,17 +2,16 @@
 
 The Jupyter HTML notebook is a web-based notebook environment for interactive computing.
 
-
 Dev quickstart:
 
 * Create a virtual env (ie jupyter-dev)
+* ensure that you have node/npm install (ie brew install node on OS X)
 * Clone this repo and cd into it
 * pip install -r requirements.txt
 * pip install -e .
-* ensure that you have node/npm install (ie brew install node)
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, this looks sensible. I'll let @minrk have a look before we merge.

README.md
```diff
@@ -2,6 +2,19 @@
...
 The Jupyter HTML notebook is a web-based notebook environment for interactive computing.
 
+
+Dev quickstart:
+
+* Create a virtual env (ie jupyter-dev)
+* Clone this repo and cd into it
+* pip install -r requirements.txt
+* pip install -e .
+* ensure that you have node/npm install (ie brew install node)
```
Carreau: brew only on os x, but OK :-)

README.md
```diff
@@ -2,6 +2,19 @@
...
+* npm install -g 'less@<3.0'
```
minrk: less doesn't need to be installed globally anymore. Nothing does, except npm.

requirements.txt
```diff
@@ -6,3 +6,4 @@
 -e git+https://github.com/ipython/ipython.git#egg=ipython
 -e git+https://github.com/ipython/ipython_kernel.git#egg=ipython_kernel
 -e git+https://github.com/jupyter/jupyter_nbconvert.git#egg=jupyter_nbconvert
+invoke
```
minrk: invoke isn't used anymore

README.md
minrk: and ensure you have node/npm should come *before* pip install.

drocco007:
@takluyver pointed out that for Debian/Ubuntu systems, if you're installing the system `node` you need to use the `nodejs-legacy` package and *not* the `node` package. It would be nice to mention this in the instructions as this is easy to have incorrect and almost impossible to discover form the failure messages.

ptone:
So the invoke + less stuff was needed earlier today, this is all now already a bit simpler

README.md
```diff
@@ -2,6 +2,18 @@
 
 The Jupyter HTML notebook is a web-based notebook environment for interactive computing.
 
+Dev quickstart:
+
+* Create a virtual env (ie jupyter-dev)
+* ensure that you have node/npm install (ie brew install node on OS X)
```
minrk: s/install/installed

takluyver:
@minrk this was updated, can you take another look?

minrk:
:+1:

PR #15, opened 2015-04-14
Update notebooks to use the new widget and traitlet paths

jasongrout:
This was just a quick search and replace.


minrk:
and of course, all of these will change again when the widgets are split out of this repo.

PR #17, opened 2015-04-14
use gulp to build css

Carreau:
No clue why package.json seem completely changed, I did not modify it by hand, just used --save-dev fro dependencies.


takluyver:
Looks like it got rewritten with a different indentation.

minrk:
Seems sensible enough. Is there a reason to use gulp-minify instead of clean-css?

Carreau:
> Seems sensible enough. Is there a reason to use gulp-minify instead of clean-css?

```
DEPRECATED
This package is marked as deprecated. Please use jonathanepollack/gulp-minify-css instead.
```

Carreau:
> Looks like it got rewritten with a different indentation.

Yes, I just don't know why npm would use a different indentation.

minrk:
Ha, and it's less that specifically recommends clean-css because less's own minification is deprecated. Deprecate all the things!

gulpfile.js
```diff
@@ -0,0 +1,17 @@
+var gulp = require('gulp');
+var less = require('gulp-less');
+var path = require('path');
+var minifyCSS = require('gulp-minify-css');
+var rename = require('gulp-rename');
+ 
+gulp.task('less', function () {
```
minrk: does it make more sense for the task to be 'css' instead of 'less'? It's building css that's relevant, and I wouldn't expect the command to change if we happened to switch to SCSS or something else, for whatever reason.

Carreau:
Yes it make sens, I just copied and pasted an example, thought the same, and got the habit of typing `less`. renamed.

gulpfile.js
```diff
@@ -0,0 +1,21 @@
...
+  return gulp.src('./jupyter_notebook/static/style/*.less')
+    .pipe(sourcemaps.init())
+    .pipe(less({
+      paths: [ path.join(__dirname, 'less', 'includes') ]
+    }))
+    .pipe(minifyCSS())
+    .pipe(rename({
+            suffix: '.min'
+        }))
+    //.pipe(sourcemaps.write())
```
jdfreder: If your intention is to not write the source maps, you can remove `sourcemaps.init()` above and the require.  You could also use a command line parsing library like yargs to make sourcemaps conditional.  i.e. `gulp css --dev` http://stackoverflow.com/a/23038290

minrk: The sourcemaps should be written, though

jdfreder: Then I suggest just removing this comment.

Carreau:
Commented line was a mistake. Uncommented.

minrk:
When I run `gulp css`, it doesn't build the sourcemaps. Do you know what's going on there?

Carreau:
> When I run gulp css, it doesn't build the sourcemaps. Do you know what's going on there?

No, but I figured it out. I'll tell you at lunch I want to see your face when I say it.

Carreau:
update to give an explicit path. @min, do you get external sourcemaps  now ?

minrk:
@Carreau yup.

PR #22, opened 2015-04-16
Set secure cookie by default if login handler is hit.

Carreau:
There is few chances that logged-in people do not use https connexion,
but I guess it can happened if the server is ran in front of a proxy
that does the https termination, so leave it configurable.

closes ipython/ipython#8325

### DIFF
--- a/jupyter_notebook/auth/login.py
+++ b/jupyter_notebook/auth/login.py
@@ -37,7 +37,13 @@ class LoginHandler(IPythonHandler):
         typed_password = self.get_argument('password', default=u'')
         if self.login_available(self.settings):
             if passwd_check(self.hashed_password, typed_password):
-                self.set_secure_cookie(self.cookie_name, str(uuid.uuid4()))
+                # tornado <4.2 have a bug that consider secure==True as soon as 
+                # 'secure' kwarg is passed to set_secure_cookie
+                if self.settings.get('secure_cookie', self.request.protocol == 'https'):
+                    kwargs = {'secure':True}
+                else:
+                    kwargs = {}
+                self.set_secure_cookie(self.cookie_name, str(uuid.uuid4()), **kwargs)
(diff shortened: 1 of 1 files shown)


jupyter_notebook/auth/login.py
```diff
@@ -11,12 +11,18 @@
...
+from traitlets import Bool
+
 
 class LoginHandler(IPythonHandler):
     """The basic tornado login handler
     
     authenticates with a hashed password from the configuration.
     """
+
+    secure_cookie = Bool(default_value=True, help="""set login cookie "secure" flag, so it will"""
```
minrk: handlers don't have traits

Carreau:
used settings.get, fallback on request protocol if not set.

jupyter_notebook/auth/login.py
```diff
@@ -37,7 +37,8 @@ def post(self):
         typed_password = self.get_argument('password', default=u'')
         if self.login_available(self.settings):
             if passwd_check(self.hashed_password, typed_password):
-                self.set_secure_cookie(self.cookie_name, str(uuid.uuid4()))
+                self.set_secure_cookie(self.cookie_name, str(uuid.uuid4()), 
+                        secure=self.settings.get('secure_cookie', (self.request.protocol == 'https'))
```
minrk: as [discovered on JupyterHub](https://github.com/jupyter/jupyterhub/pull/231#issuecomment-93862873), `secure=False` doesn't actually work. The `secure` kwarg can only be specified when it should be True (a bug in Tornado, fixed in 4.2)

Carreau:
shoudl I mark that as backport ?

minrk:
I don't think it's useful to mark for backport, since we can't backport automatically anymore. It probably makes sense to open a PR against 3.x, though.

PR #23, opened 2015-04-17
Clean up usage of !=/== and don't create functions in loops

ssanderson:
Fixes a number of scary warnings from jshint/jslint related to the use of `!=`/`==` and creation of functions in loops.

I'm aware that IPython's policy is generally to eschew cleanup PRs, but these are cases for which you almost never want the behavior as written, and for which there are obvious simple substitutes.

See https://jslinterrors.com/dont-make-functions-within-a-loop for an account of why creation of functions in a loop is generally a bad idea.  The TLDR is that variables are closed over by name, which can easily create nasty bugs.

### DIFF
--- a/jupyter_notebook/static/tree/js/notebooklist.js
+++ b/jupyter_notebook/static/tree/js/notebooklist.js
@@ -151,9 +151,9 @@ define([
     NotebookList.prototype.handleFilesUpload =  function(event, dropOrForm) {
         var that = this;
         var files;
-        if(dropOrForm =='drop'){
+        if(dropOrForm === 'drop'){
             files = event.originalEvent.dataTransfer.files;
-        } else 
+        } else
         {
             files = event.originalEvent.target.files;
         }
@@ -171,7 +171,7 @@ define([
             }
             var item = that.new_item(0, true);
             item.addClass('new-file');
-            that.add_name_input(f.name, item, file_ext == '.ipynb' ? 'notebook' : 'file');
+            that.add_name_input(f.name, item, file_ext === '.ipynb' ? 'notebook' : 'file');
             // Store the list item in the reader so we can use it later
             // to know which item it belongs to.
             $(reader).data('item', item);
@@ -302,7 +302,7 @@ define([
             var list_items = $('.list_item');
             for (var i=0; i<list_items.length; i++) {
                 var $list_item = $(list_items[i]);
-                if ($list_item.data('path') == item.path) {
+                if ($list_item.data('path') === item.path) {
                     $list_item.find('input[type=checkbox]').prop('checked', true);
                     break;
                 }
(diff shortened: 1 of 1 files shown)


ssanderson:
Sidenote: 

I'd **wholeheartedly** recommend running something like jshint in travis with the following config.  I think these are about as undogmatic as linter options come:
```
{
    "eqeqeq":true,
    "browser":true,
    "devel":true,
    "undef":true,
    "unused":true,
    "latedef":true,
    "strict":true,
    "predef": ["define", "require"]
}
```

minrk:
Yup, I use jshint and it's great. Analogous to pyflakes, the similarly style-free Python linter.

I don't like the idea of linting on CI, though.

minrk:
BTW, fixing things like triple-equals is totally okay. Given our relative lack of expertise in javascript, it's quite appropriate to be more strict in these things. The function in a for-loop is definitely an issue we've encountered a few times. Those particular functions didn't have any closures, so they happened to be fine, but it's a good thing to avoid, since javascript closures don't behave in sensible ways.

takluyver:
I think our policy about cleanup PRs is more about style cleanups - if there's something that's functionally wrong or problematic in our code, I don't object to fixing it. That's much more likely in the JS code because we're less experienced in writing that.

Carreau:
I'm fine with this too. At some point anyway if we use typescript we will have to fix theses things.

ssanderson:
> I don't like the idea of linting on CI, though.

I think this is defensible for PEP8-esque style-linting (I personally still use PEP8 for everything I write, but I understand that it imposes significant friction for new users for a possibly dubious gain).

For stuff like javascript `===`, though, I can't imagine a circumstance where I would intentionally merge a change that had `==` unless the author explained explicitly why they wanted the type-unsafe equality check.  This means that, as a reviewer of javascript changes, I'm always manually doing the same checks that jshint is doing.  This detracts from my ability to actually review the content of a change, because my attention is now divided between code comprehension and checking for javascript gotchas, and invariably I'm going to miss a `==` that a programmatic solution will never miss.

Basically, I think linting in CI should be used to perform checks that you would expect any competent reviewer to perform.

/endrant

Carreau:
The problem is as soon as you **want** to have an unused variable, or an `==` instead of `===` your test fails, for all PR. 

If could check only for new code, that would be better.

ssanderson:
> The problem is as soon as you want to have an unused variable, or an == instead of === your test fails, for all PR.

Every reasonable linting tool has a way of saying "disable checks for this line or this region of code".  This forces the programmer to document that they actually want "2 == '2'" in this specific chunk of code.

In the specific case of jshint, there are lots of ways to do this:
```
// Code here will be linted with JSHint.
/* jshint ignore:start */
// Code here will be ignored by JSHint.
/* jshint ignore:end */
```
or 
```
ignoreThis(); // jshint ignore:line
```

Carreau:
OK, regardless of jshint on travis, moving forward with this.

PR #25, opened 2015-04-17
Fixed, invoke css not respecting windows PATH

jdfreder:
Using check_output shell=False on Windows doesn't respect PATH.  Since the command executed isn't subject to user input, and is dev-only, it's okay to just use shell=True.

@minrk @takluyver


Carreau:
will conflict with #17

jdfreder:
#17 nullifies this - and I prefer #17 .

PR #34, opened 2015-04-20
Segregating the widgets (the Small Split)

jdfreder:
- [x] tests


ipython_widgets/install.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
+# Thanks @takluyver for your cite2c install.py.
+from os.path import dirname, abspath, join as pjoin
+from jupyter_notebook.nbextensions import install_nbextension
+from jupyter_notebook.services.config import ConfigManager
+
+print("Installing nbextension ...")
+staticdir = pjoin(dirname(abspath(__file__)), 'static')
+install_nbextension(staticdir, destination='widgets', user=False)
```
minrk: user=False shouldn't be hardcoded

jupyter_notebook/widgets.py
```diff
@@ -0,0 +1,16 @@
+"""
+Shim to maintain backwards compatibility with old IPython.html.widget imports.
+"""
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import sys
+from warnings import warn
+
+warn("The `IPython.html.widgets`/`jupyter_notebook.widgets` package has been deprecated. "
```
minrk: No need for a shim in this repo. We will want a shim in IPython, though.

jasongrout:
Async rabbit for you!  :rabbit:

jupyter_notebook/widgets.py
jdfreder: https://github.com/ipython/ipython/pull/8342

jdfreder:
This is ready for review/merge

jupyter_notebook/notebookapp.py
```diff
@@ -203,6 +241,7 @@ def init_settings(self, ipython_app, kernel_manager, contents_manager,
             config=ipython_app.config,
             jinja2_env=env,
             terminals_available=False,  # Set later if terminals are available
+            log=log,
```
minrk: add log to settings?

jupyter_notebook/notebookapp.py
```diff
@@ -120,6 +128,36 @@ def load_handlers(name):
     mod = __import__(name, fromlist=['default_handlers'])
     return mod.default_handlers
 
+class ShimHandler(FileFindHandler):
```
minrk: This can inherit straight from RequestHandler, and just override `get`.

jupyter_notebook/notebookapp.py
```diff
@@ -120,6 +128,36 @@ def load_handlers(name):
...
+        # Check if something should be prepended.
+        if self.prepend is not None:
+            self.set_header("Content-Length", str(int(self._headers["Content-Length"]) + len(self.prepend)))
+            if isinstance(content, (bytes)): # Bytes
+                content = self.prepend + content
+            else: # Iterator
+                content = itertools.chain([self.prepend], content)
+        return content
+
+def make_shim_handler(log, prepend=None, encoding='utf-8'):
```
minrk: You should only need a single ShimHandler. No need for this type-generating function.

jupyter_notebook/notebookapp.py
jdfreder: I though it looked like settings was being used as a generic budle of things, but I must have been imagining it.  I'll set it on self.

jdfreder: :heavy_check_mark:

minrk: I was mainly asking because we log everywhere, and don't need access in
settings, so I was dubious that it actually needed to be added.

On Mon, Apr 20, 2015 at 5:16 PM, Jonathan Frederic <notifications@github.com

jupyter_notebook/notebookapp.py
```diff
@@ -261,6 +301,33 @@ def init_handlers(self, settings):
...
+        """Creates a handler that shims the loading of widgets
+        JS files for the old /static/widgets/js/*.js path."""
+        # Add RESTful shim for the ipython_widgets
+        try:
+            ipythonwidgets = imp.find_module('ipython_widgets')[1]
+            widget_js = os.path.join(ipythonwidgets, 'static', 'widgets')
+
+            return [(
+                r'/deprecatedwidgets/(.*)', 
+                make_shim_handler(self.log, prepend=DEPRECATED_WIDGET_WARNING),
```
minrk: This class shouldn't need to be defined at runtime. A single class based on RequestHandler that does:

```python
class DeprecatedWidgetHandler(RequestHandler):
    def get(self, path):
        app_log.warn("widgets have moved...")
        self.redirect('/nbextensions/widgets/' + path)
```

Should be sufficient.

minrk: It also shouldn't need to be a different handler for whether widgets are present or not. The message can change, but you don't need two handlers for that.

jdfreder:
@minrk the last commit greatly simplifies the deprecation handler.

jupyter_notebook/notebookapp.py
jdfreder: I also want the error to appear in the console- redirect doesn't let me do that.

jdfreder:
I think I can still make it one handler... One second.

jupyter_notebook/notebookapp.py
```diff
@@ -261,6 +270,44 @@ def init_handlers(self, settings):
...
+            ipythonwidgets = imp.find_module('ipython_widgets')[1]
+            widget_js = os.path.join(ipythonwidgets, 'static', 'widgets')
+
+            class DeprecationHandler(IPythonHandler):
+                def get(self, url_path):
+                    self.set_header("Content-Type", 'text/javascript')
+
+                    # Verify that the file requested exists.
+                    local_file = os.path.join(widget_js, *url_path.split('/'))
+                    if not os.path.isfile(local_file):
```
minrk: Don't check for or serve the file, you can just redirect.

jdfreder: Ah

jupyter_notebook/notebookapp.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, ipython_app, kernel_manager, contents_manager,
                  config_manager, log,
                  base_url, default_url, settings_overrides, jinja_env_options):
 
+        self.log = log
```
minrk: self.log here is no longer used, I think.

jupyter_notebook/notebookapp.py
```diff
@@ -261,6 +263,22 @@ def init_handlers(self, settings):
         new_handlers.append((r'(.*)', Template404))
         return new_handlers
 
+    def _create_widget_shim_handlers(self):
+        """Creates a handler that shims the loading of widgets
+        JS files for the old /static/widgets/js/*.js path."""
+        class DeprecationHandler(IPythonHandler):
```
minrk: This class can be defined normally in the module. It doesn't need to be in a closure.

jupyter_notebook/notebookapp.py
```diff
@@ -261,6 +263,22 @@ def init_handlers(self, settings):
...
+    def _create_widget_shim_handlers(self, settings):
+        """Creates a handler that shims the loading of widgets
+        JS files for the old /static/widgets/js/*.js path."""
+        class DeprecationHandler(IPythonHandler):
+            def get(self, url_path):
+                self.set_header("Content-Type", 'text/javascript')
+                self.finish("""
+                    console.warn('`/static/widgets/js` is deprecated.  Use `/nbextensions/widgets/widgets/js` instead.');
+                    define(['%s'], function(x) { console.log('...loaded...'); return x; });
+                """ % url_path_join(settings['base_url'], 'nbextensions', 'widgets', 'widgets', url_path))
```
jdfreder: @minrk this is driving me crazy- for the case of http://localhost:8888/deprecatedwidgets/js/init.js  this yields:

```js
console.warn('`/static/widgets/js` is deprecated.  Use `/nbextensions/widgets/widgets/js` instead.');
define(['/nbextensions/widgets/widgets/js/init.js'], function(x) { console.log('...loaded...'); return x; });
```

In the notebook, when I require this file, the warning shows in the console.  After the ~5sec require js timeout, a load timeout error is shown, no 404s in the network panel of the browser debugging tools, and the "...loaded..." text is never printed.  HOWEVER, when I try to access the file manually by navigating to `http://localhost:8888/nbextensions/widgets/widgets/js/init.js` it loads just fine!

jdfreder: I have a feeling it's the last bullet here :http://requirejs.org/docs/errors.html#timeout

jdfreder:
@minrk I moved the handler class def out of the closure.  The problem causing the require js timeout was the import paths in /ipython_widgets/static/notebook/js/extension.js not matching the path(s) written by the deprecation warning.  I'm not satisfied with the current state of things, which requires a `/` prefix and `.js` suffix on the require import.  This is because using `nbextensions/...` gets interpreted as `/notebooks/CWD/nbextensions/...` - which doesn't make sense.  It's almost as if the page.html require config is just being ignored in this case.  I'm looking some more into this...

jdfreder:
/me sighs

I must have been tired last night - didn't notice the `.js` being appended to the path, which is why the `/` was required.  Everything is fixed now - ready for review or merge.

jdfreder:
I squashed the last 8 commits, which all related to the JS deprecation warning, making this PR more lean.

minrk:
@jdfreder sorry for the struggle. This is looking good, now.

minrk:
@jdfreder almost there. Can you remove `widget.less` from `ipython.less` so that building css works?

jdfreder:
Yes, that brings up a good point - after the new repository, and your history cleaning, I'll need to do something about the CSS...

minrk:
I suspect we'll want to add [require-css](https://github.com/guybedford/require-css) or similar.

jdfreder:
> Can you remove widget.less from ipython.less so that building css 

:heavy_check_mark:

minrk:
@jdfreder thanks. I've started the actual split to create the widgets repo, but I have somewhere I need to be this evening, so it probably won't finish before tomorrow.

jdfreder:
No rush, thanks for doing that! : )

PR #36, opened 2015-04-21
Add commands needed for my Ubuntu Trusty box.

CarlFK:
Tested with docker, which needs 2 more:

apt-get update
apt-get install sudo git

I don't know where build-essential came from, but we have to draw the line somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,3 +17,15 @@ Launch with:
 
     jupyter notebook
 
+For Ubuntu Trusty:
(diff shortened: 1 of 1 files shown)


Carreau:
Hum, that detailed for a specific platform seemed to be a slippery slope, but I'm fine with it.

minrk:
I'm okay with this for now.

PR #40, opened 2015-04-22
remove widgets

minrk:
this also modifies the js tests a little to make it more re-usable for running the widget tests


Carreau:
+1

PR #43, opened 2015-04-22
remove unused mathjax install from travis

minrk:
no longer needed now that mathjax is a component


minrk:
grr, accidentally deleting my branches.

PR #45, opened 2015-04-23
Use name ignored by dropbox for atomic saving temp files

takluyver:
See discussion on ipython/ipython#8322

By using a '.~' prefix for the temporary files, Dropbox should ignore them, which makes atomic saving a bit less inconvenient.

However, to do this, we need to switch back from using temporary directories (implemented in ipython/ipython#7469). This uses the other approach described in ipython/ipython#7317 instead.

### DIFF
--- a/jupyter_notebook/services/contents/fileio.py
+++ b/jupyter_notebook/services/contents/fileio.py
@@ -70,18 +70,18 @@ def atomic_writing(path, text=True, encoding='utf-8', **kwargs):
         path = os.path.join(os.path.dirname(path), os.readlink(path))
 
     dirname, basename = os.path.split(path)
-    tmp_dir = tempfile.mkdtemp(prefix=basename, dir=dirname)
-    tmp_path = os.path.join(tmp_dir, basename)
+    # The .~ prefix will be ignored by Dropbox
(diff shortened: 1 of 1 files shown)


Carreau:
+1, and do we want to try the invert process @minrk was proposing, ie: 

Copy file to tempfile, keeping permsission, writing to old file new content, remove copy ?

damianavila:
>Copy file to tempfile, keeping permission, writing to old file new content, remove copy ?

I like the idea...

jdfreder:
So if this were to fail mid operation, you'd be left with a corrupt .ipynb and a good .~ temp file?

minrk:
@jdfreder yes.

takluyver:
Assuming the 'this' in question is #67, not this PR: it depends on the failure mode. If it fails in a way that raises an error in Python (such as running out of disk space), then the temp file should be renamed back to the regular .ipynb name - the save fails, and you're left with only the old copy. If it fails more dramatically, like the system suddenly dying mid-write, then yes, the .ipynb would be corrupt and the temp file should be a good copy of the previous save.

rgbkrk:
Since it seems like #67 solved this, closing this PR.

PR #48, opened 2015-04-24
Expose JS terminal object on window

takluyver:
This is handy for interactive debugging and fiddling, but it shouldn't be documented.


rgbkrk:
Let it be known that you have now documented it in an issue.

PR #50, opened 2015-04-24
disable set_default_headers on TermSocket

minrk:
inherited from IPythonHandler, but doesn't make sense or work on websockets.


rgbkrk:
Ah, cool. Thanks @minrk.

rgbkrk:
:+1:

PR #54, opened 2015-04-26
allow OPTIONS requests for api endpoints (to enable CORS)

jaredly:
Although `NotebookApp.allow_hosts` was set correctly, the api endpoints were disallowing `OPTIONS` requests, and so the browser refused to allow my web page to access ipython's endpoints.

These changes add `OPTIONS` handlers.

I didn't see an obvious place to add tests for this -- the `services.js` test is run in casper, and so is run from the same origin.


rgbkrk:
Looking good to me.

We need to have alternate origin/host tests for other bits as well, this may be the right area to start.

PR #60, opened 2015-04-29
build main.min.js with r.js

minrk:
Makes main.min.js for each page, including dependencies.

- build all js with `gulp js`
- build single application with `gulp notebook-js` etc.
- building js on master is *required*.
  The unminified js will not be used directly (just like less).

This seems to significantly improve page load times, especially on poor connections.

Building all the js takes about a minute, which is pretty slow. Since nodejs is fundamentally single-threaded, I decided spawn subprocesses to do the js builds in parallel, cutting the time down to ~30 seconds.

- do we want to allow using the non-minified js when running from git? My original attempt did enable this, but I'm not sure it makes sense anymore now that we require building less on master.
- This currently makes one `main.min.js` for each 'application' (tree, notebook, edit, etc.).
  We could alternately make a single jumbo blob of all js, with each 'main' as a function accessible on that blob.
  This would save space and improve cache performance a bit.

cc @ellisonbg @Carreau @jdfreder

### DIFF
--- a/build-main.js
+++ b/build-main.js
@@ -1,0 +1,57 @@
+// build main.min.js
+// spawned by gulp to allow parallelism
+
+var rjs = require('requirejs').optimize;
+
+var name = process.argv[2];
+
+var rjs_config = {
+  name: name + '/js/main',
+  out: './notebook/static/' + name + '/js/main.min.js',
+  baseUrl: 'notebook/static',
+  preserveLicenseComments: false, // license comments conflict with sourcemap generation
+  generateSourceMaps: true,
+  optimize: "uglify2",
+  paths: {
+    underscore : 'components/underscore/underscore-min',
+    backbone : 'components/backbone/backbone-min',
+    jquery: 'components/jquery/jquery.min',
+    bootstrap: 'components/bootstrap/js/bootstrap.min',
+    bootstraptour: 'components/bootstrap-tour/build/js/bootstrap-tour.min',
+    jqueryui: 'components/jquery-ui/ui/minified/jquery-ui.min',
+    moment: 'components/moment/moment',
+    codemirror: 'components/codemirror',
+    termjs: 'components/term.js/src/term',
+    contents: 'empty:'
+  },
+  shim: {
+    underscore: {
+      exports: '_'
+    },
+    backbone: {
+      deps: ["underscore", "jquery"],
+      exports: "Backbone"
+    },
+    bootstrap: {
+      deps: ["jquery"],
+      exports: "bootstrap"
+    },
+    bootstraptour: {
+      deps: ["bootstrap"],
+      exports: "Tour"
+    },
+    jqueryui: {
+      deps: ["jquery"],
+      exports: "$"
+    }
+  },
+
+  exclude: [
+    "custom/custom",
+  ]
+};
+
+rjs(rjs_config, console.log, function (err) {
+  console.log("Failed to build", name, err);
+  process.exit(1);
+});
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,13 +1,15 @@
+var fork = require('child_process').fork;
+var path = require('path');
+
 var gulp = require('gulp');
 var less = require('gulp-less');
-var path = require('path');
 var minifyCSS = require('gulp-minify-css');
 var rename = require('gulp-rename');
 var sourcemaps = require('gulp-sourcemaps');
 
 // now some dev nice utilities.
 var livereload = require('gulp-livereload');
- 
+
 gulp.task('css', function () {
   return gulp.src('./notebook/static/style/*.less')
     .pipe(sourcemaps.init())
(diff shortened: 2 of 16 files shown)


build-main.js
```diff
@@ -0,0 +1,57 @@
...
+    jqueryui: {
+      deps: ["jquery"],
+      exports: "$"
+    }
+  },
+
+  exclude: [
+    "custom/custom",
+  ]
+};
```
jdfreder: Did you try unifying this with page.html's copy of it?  Do they differ in ways that don't let you use one centralized copy?  It seems a bit tedious to maintain two copies of the config in two locations.

minrk: The one in page.html is basically unused now, I think. I'll have to make sure on that, though, and see how it behaves when things like nbextensions require dependencies.

ellisonbg:
This is great! I think we should start simple and not try to get the unminimized JS to work in dev mode. If it drives us crazy we can always add that later. 

A few other comments:

* We should at least keep in mind the possibility of building minimized JS that includes code from a number of different repos (widgets, etc.).
* It isn't obvious to me if we want to do this once for each page or once for the entire app. I would pick the simplest thing for now and see how it goes.

minrk:
@ellisonbg the js shouldn't include the widget js, since they will be installed separately. I think separating packages necessarily means having at least one .js and .css per package.

minrk:
> I would pick the simplest thing for now and see how it goes.

Simplest for now is per-app, since it means we don't need to touch the js at all. We can investigate combining to a single `jupyter-notebook.min.js` at a later point.

Carreau:
How does that affect custom.js/kernel.js that requires things ?

minrk:
@Carreau I believe it should still work, but I need to test it to make sure.

Carreau:
Also have you tried `https://github.com/jlouns/gulp-requirejs-optimize` ?

minrk:
I tried a few gulp-rjs/amd plugins (there are several), and none of them really worked, but I didn't try that one. After using it for a few things, I confess that I really don't like gulp so far.

minrk:
@Carreau thanks for that plugin, it does seem to work, unlike the ones I had found and tried. Unfortunately, with that plugin it takes about 1:45 to run the minification, whereas it's more like 45 seconds with the parallel approach I did here.

If we go with a single .js file containing all the main.js as exported functions, it should be fine to use the gulp plugin, though it doesn't seem to offer any benefits over calling rjs manually.

Carreau:
You don't seem to like javascript in general :-)

minrk:
@Carreau @jdfreder I think this is good to go as a starting point, if you want to look it over again.

Things I think we should consider after this PR, but I'd rather not do here:

- build a *single* Jupyter js that exports all the main scripts as functions, so we can have one big js for all pages
- strip unneeded requirejs config from page.html. Much of it is unused and commenting it out seems fine when I test, but I don't know exactly what can be safely removed and what can't.

minrk:
This should be ready to go.

gulpfile.js
```diff
@@ -23,9 +29,84 @@ gulp.task('css', function () {
...
+    
+    fs.readdirSync(path.join(s, 'components')).map(function (c) {
+      if (c !== 'MathJax') {
+        // skip MathJax because it has tons of files and makes everything super slow
+        sources.push(path.join(s, 'components', c, '**', '*.js'));
+      }
+    });
+    
+    gulp.src(sources, {base: s})
+    .pipe(newer(dest))
```
jdfreder: Wont this fail if the changed file is one of the dependencies?  i.e.  main.js imports notebook.js, and I change notebook.js, main.js's modified timestamp doesn't change, so re-optimization wont happen because it will think nothing has changed.

package.json
```diff
@@ -14,9 +14,13 @@
...
     "gulp-livereload": "^3.8.0",
     "gulp-minify-css": "^1.0.0",
+    "gulp-newer": "^0.5.0",
     "gulp-rename": "^1.2.2",
     "gulp-sourcemaps": "^1.5.1",
     "less": "~2",
-    "source-map": "^0.4.2"
+    "requirejs": "^2.1.17",
+    "source-map": "^0.4.2",
+    "through": "^2.3.7",
```
jdfreder: You may already know this, but I found recently that you can write this as `2.x` - if you prefer, not necessarily suggesting you do that here. https://github.com/npm/node-semver

package.json
```diff
@@ -14,9 +14,13 @@
...
+    "vinyl": "^0.4.6"
```
jdfreder: 0.4.x

gulpfile.js
```diff
@@ -1,13 +1,19 @@
+var fork = require('child_process').fork;
+var fs = require('fs');
+var path = require('path');
+
+var File = require('vinyl');
```
jdfreder: Unused

jdfreder:
Couple comments inline, other than that, looking good.

package.json
minrk: `npm install --save-dev` wrote these, so I'm inclined to leave them how npm prefers them to be.

minrk:
Comments addressed, I think.

jdfreder:
Awesome, thanks for clarifying the gulp bit with those comments.  I'm easily confused by what is or isn't in the stream at any given point.  :+1:

gulpfile.js
minrk: `sources` includes all dependencies, so if any dependency changes, it will rebuild the target. You can test with `gulp js; touch notebook/static/notebook/js/notebook.js; gulp js`

PR #61, opened 2015-04-29
Pip install all-in-one

rgbkrk:


minrk:
Just a note that the circular dependency described in jupyter/jupyter_nbconvert#12 is what forces us to install everything in one command, otherwise the circular dependency goes unresolved.

PR #63, opened 2015-04-30
check FILE_ATTRIBUTE_HIDDEN on Windows

### LINKED ISSUE #62: Notebook file list shows files with "hidden" attribute set on Windows
Starting a Jupyter notebook on my home directory on Windows, I see a bunch of files that have the "hidden" attribute set on Windows (like user registry files, etc.). Given that the likelihood of needing to open these hidden files in a notebook is very small, it would be nice if they were also hidden by default, so as not to needlessly pollute the list of files in Windows.

minrk:
in is_hidden

closes #62


takluyver:
Looks OK.

PR #64, opened 2015-04-30
remove SYSTEM_NBEXTENSIONS_DIRS

minrk:
use SYSTEM_JUPYTER_PATH from core

for common logic with kernelspec installation

### DIFF
--- a/jupyter_notebook/nbextensions.py
+++ b/jupyter_notebook/nbextensions.py
@@ -20,40 +20,13 @@ except ImportError:
     from urlparse import urlparse
     from urllib import urlretrieve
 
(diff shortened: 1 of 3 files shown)


Carreau:
is that a leftover of IPython thing ? can we check if `SYSTEM_NBEXTENSIONS_DIRS ` is set and log a warning ?

minrk:
@Carreau how could it be set? It's a module constant.

Carreau:
sorry, uppercase made me though as environemnt variable. was too tired this morning.

damianavila:
LGTM

PR #66, opened 2015-05-04
Added Mind Map support using vis.js

seanodell:
Added a new Cell Type "Mind Map" which uses vis.js to render a network graph using indented syntax and a little markdown-style hinting. It's based on the same TextCell that MarkdownCell uses.

I added this because I use Jupyter Notebook at work quite a bit to document things, run python (ML) stuff and document expressions using LaTeX. In a separate application (Freespace), I create mind-maps of notes where I also stick in bits of code and LaTeX and I just got tired of keeping everything in two places.

The code works and has been usable so far, but there's a validation error I don't understand, and I can't seem to export my notebooks as HTML. I am hoping someone can either give me some hints how to fix those or maybe it would be simpler for someone else to fix?

I'm also not 100% clear on the new feature PR policy. I am out of time right now, so maybe someone can help with that, or I will give it a go later.

Example syntax for this new cell type:

    # Cardinality Estimation
        ## LogLog
        ## Super-LogLog
        ## HyperLogLog
        ## K-Minimum Values
        
    # Sets
        ## Grouping
            ## Classification
            ## Clustering
            ## Partitioning
                (Super-LogLog)
            
    Cardinality
        (Sets)
        (Cardinality Estimation)


takluyver:
We're generally resistant to adding new cell types, because that involves changing the notebook format - I think we should work out how to do this as an extension.

minrk:
The error is because this is not a notebook according to nbformat. There is a schema describing what cell types can exist and what form they may have. Adding a cell type in the UI means you are no longer writing v4 notebooks, and the schema and all other notebook-consuming applications must be updated, versions revved, and backward-compatibility warnings issued.

New cell types are major changes that have effects in many projects - nbformat, nbconvert, nbviewer, etc., and are not taken lightly. They indicate a major backward-incompatible change. I would doubt that mind-maps are likely to cross the threshold of a feature that we can support.

Carreau:
You can also most likely do the same with a `%mindmap`  magic I guess. I'm goint to close but you can continue to comment and reopen later.

seanodell:
@takluyver: That's a great idea ... I didn't even realize there was such a thing as extensions. Any hints as to where I should get started with that? I will try to refactor this as an extension.

@minrk: Thanks for the feedback ... it makes sense that this wouldn't be part of the core, knowing that there's an extension system.

@Carreau: I'll look into that. Would that go into a raw cell, or as a directive in a markdown cell?

takluyver:
There's a couple of different extension mechanisms - @Carreau was suggesting that you do it as a kernel extension ([docs here](http://ipython.org/ipython-doc/stable/config/extensions/index.html#writing-extensions)) - that way you'd write the mindmap input in a Python cell, and the mindmap would appear as output when you run it.

Neater but possibly more complex would be doing it as a frontend extension. I don't think we've really documented this yet, but there are examples here: https://github.com/ipython-contrib/IPython-notebook-extensions/wiki/Home_3x

minrk:
If you are only interested in IPython, the `%%mindmap` idea is probably the easiest to get off the ground. If you want this available in notebooks with any kernel, a raw cell with a mindmap mime-type is probably the way to go. That would be a bit of frontier development, so definitely stay in touch here or [on gitter](https://gitter.im/jupyter/jupyter_notebook).

seanodell:
After taking a look at the examples, I feel like I would need to know a lot more about this project to do this right. I didn't see anything that I could use as a starting point. I will poke around a little more this weekend, but I sense that I will end up keeping my fork and using my own build for awhile.

Thanks for the ideas. I'll be keeping my eye out for good front end examples.

PR #67, opened 2015-05-04
Alternative way to do atomic writing

takluyver:
This is a hopefully more robust way of doing atomic writing of a file.

Previously, we wrote a new temporary file on each write, and renamed it over the target file on success. This is technically neat, but constantly recreating the file causes problems with some network filesystems, and with sync tools like Dropbox.

The new approach copies the old file contents to a temporary file, overwrites the target using standard open() and write() calls, and then removes the temporary file. In case of a failure during writing, the
temporary file is renamed over the target, which should preserve the old data.

This way, we're only using a new inode in case of a write failure, which is hopefully rare, instead of on successful writes.

### DIFF
--- a/jupyter_notebook/services/contents/fileio.py
+++ b/jupyter_notebook/services/contents/fileio.py
@@ -11,7 +11,6 @@ import errno
 import io
 import os
 import shutil
-import tempfile
 
 from tornado.web import HTTPError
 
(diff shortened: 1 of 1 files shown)


takluyver:
This is an alternative to #45 .

minrk:
I think this is worth a shot.

damianavila:
>I think this is worth a shot.

Yep, I agree... and I like more than #45

Carreau:
Let's merge.

damianavila:
OK, I guess #45 has to be closed now...

PR #74, opened 2015-05-08
replace jupyter_nbformat by nbformat

Carreau:
Also replace the `import nbformat as nbformat` by `import nbformat`
and change the name of the installed egg.


Carreau:
Hopefully this will fix travis faillures.

minrk:
I think this will work after jupyter/nbconvert#19 is merged.

minrk:
poking travis now that nbconvert is updated.

Carreau:
Travis happy.

minrk:
Thanks.

PR #76, opened 2015-05-10
data_dir is not .ipython/kernel anymore.

Carreau:
On OSX, items ~/Library/Jupyter/ which screw up where things find
kernels and anyway is already at the beginning of the search paths
lists.

### DIFF
--- a/jupyter_notebook/notebookapp.py
+++ b/jupyter_notebook/notebookapp.py
@@ -788,10 +788,6 @@ class NotebookApp(JupyterApp):
         self.kernel_spec_manager = self.kernel_spec_manager_class(
             parent=self,
         )
-        # FIXME: temporarily add .ipython/kernels to the kernel search path
-        self.kernel_spec_manager.kernel_dirs.append(
-            os.path.join(self.data_dir, 'kernels'),
-        )
         self.kernel_manager = self.kernel_manager_class(
             parent=self,
             log=self.log,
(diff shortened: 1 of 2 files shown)


takluyver:
Looks like there's a problem with the kernelspec rest API tests.

Carreau:
Hum, I can see 2 things:

1) A part of the code make the assumption that all the search path for things are return by `jupyter_core.paths:jupyter_data_dir`
2) The test make the assumption that if a specific app will search it's `kernelspec.data_dir/kernels`.

The second seem wrong, as in some cases the will be the same path both as first and last directory searched.

I'll try to go and eliminate the second option.

Carreau:
OK, I think I can get it with JUPYTER_DATA_DIR, let see how travis likes it.

takluyver:
Travis appears happier about that (I just kicked one of the js tests, but I think that's a sporadic thing).

@minrk was this waiting for migration to be implemented, though?

Carreau:
I would say that this take care of that while the migration is not done.
```
 def _kernel_dirs_default(self):
        ...
        from IPython.paths import get_ipython_dir
        dirs.append(os.path.join(get_ipython_dir(), 'kernels'))
        return dirs
```

takluyver:
Ah, right, I guess we had added the compatibility code twice. This makes sense then.

minrk:
This was leftover from the piece by piece move over to Jupyter paths. The .ipython bit that should be kept until migration is somewhere else.

PR #81, opened 2015-05-11
Start working on opt-in data collection system

takluyver:
This is initial architecture for a mechanism to gather data on the tools users are using, as discussed with @njsmith.

For now, there are two kinds of data we wanted to collect: what versions of packages are in use, and how many users call particular features (especially deprecated features, to inform when they can be removed). Obviously there are many more things that might be interesting, but I think we should be taking a minimal, well justified set of data from users rather than just getting whatever we can.

My current thinking is that each of these will be a set, aggregated by month on the user's machine. So you'll see, for instance, that during the last month, a user has used numpy versions X and Y, and has called deprecated features A and B. You won't see how many times features were called - I don't think that's terribly useful anyway, because one call to a particular function inside a loop could be called millions of times. Four weeks after the last submission, or the next time after that that the server is started, it will submit and discard its data, and start collecting again.

Users who haven't previously made a choice will see a prompt of some kind in the notebook interface (not yet implemented) asking them to choose. We don't send any data until they agree, and if they say no, we don't ask again.

### Down to the technical details:

To make this mechanism readily available to all kernels, the kernels record data by sending it over a new ZMQ PUSH socket to the server. The connection file contains an extra entry `"analytics_port": N` if analytics are enabled - this shouldn't cause any problems for kernels that don't know about it. A single PULL socket on the server receives information from all kernels.

For now, the analytics messages are single-frame unsigned JSON messages, rather than routing them through our serialisation machinery. I don't think security is a big concern here - the only malicious possibility I can see is spamming the record of what's in use, and I don't see how we can prevent that given that it would ultimately be submitting to some public server.

This also requires some changes in jupyter_client (coming up) and ipykernel (not yet started). I was thinking of making a separate small package with the API to record information, so that projects wishing to use it could depend on that without depending on all the IPython/Jupyter machinery.

### DIFF
--- a/jupyter_notebook/analytics.py
+++ b/jupyter_notebook/analytics.py
@@ -38,7 +38,7 @@ class AnalyticsSender(object):
             try:
                 with open(self.control_file, 'r') as f:
                     self._control_info = json.load(f)
-            except OSError as e:
+            except IOError as e:
                 if e.errno == errno.ENOENT:
                     self._control_info = {}
                 else:
--- a/jupyter_notebook/analytics.py
+++ b/jupyter_notebook/analytics.py
@@ -1,0 +1,193 @@
+from datetime import datetime, timedelta
+import errno
+import json
+import os
+import os.path as osp
+import sys
+
+from jupyter_client.jsonutil import parse_date
+from tornado.httpclient import AsyncHTTPClient
+from tornado.ioloop import IOLoop
+from tornado import web
+import zmq
+from zmq.eventloop.zmqstream import ZMQStream
+
+from .base.handlers import IPythonHandler, json_errors
+from .utils import url_path_join as ujoin
+from ._version import __version__
+
+CONSENT_UNKNOWN = 0
+CONSENT_OK = 1
+CONSENT_DENIED = 2
+
+class AnalyticsSender(object):
+    _control_info = None
+    socket = None
+    enabled = False
+
+    def __init__(self, nbapp):
+        self.nbapp = nbapp
+        self.directory = osp.join(nbapp.data_dir, 'analytics')
+        self.control_file = osp.join(self.directory, 'control.json')
+        self.data_file = osp.join(nbapp.data_dir, 'analytics', 'data.json')
+        self.data = {}
+        self.ioloop = IOLoop.current()
+
+    def control_info(self):
+        if self._control_info is None:
+            try:
+                with open(self.control_file, 'r') as f:
+                    self._control_info = json.load(f)
+            except OSError as e:
+                if e.errno == errno.ENOENT:
+                    self._control_info = {}
+                else:
+                    raise
+
+        return self._control_info
+
+    def store_control_info(self):
+        assert self._control_info is not None
+        try:
+            os.makedirs(self.directory)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
+
+        with open(self.control_file, 'w') as f:
+            json.dump(self._control_info, f)
+
+    def read_data(self):
+        try:
+            with open(self.data_file, 'r') as f:
+                self.data = json.load(f)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+
+    def store_data(self):
+        try:
+            os.makedirs(self.directory)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
+
+        with open(self.data_file, 'w') as f:
+            json.dump(self.data, f)
+
+    def consent_status(self):
+        ci = self.control_info()
+        if 'consent' in ci:
+            return CONSENT_OK if ci['consent'] else CONSENT_DENIED
+        return CONSENT_UNKNOWN
+
+    def change_consent_status(self, status):
+        ci = self.control_info()
+        if status == CONSENT_OK:
+            if not ci.get('consent', False):
+                self.enable()
+            ci['consent'] = True
+        elif status == CONSENT_DENIED:
+            ci['consent'] = False
+        else:
+            ci.pop('consent', None)
+
+        self.store_control_info()
+
+    def last_submission(self):
+        ci = self.control_info()
+        if 'last_submission' in ci:
+            return parse_date(ci['last_submission'])
+
+        return None
+
+    def startup(self):
+        webapp = self.nbapp.web_app
+
+        consent = self.consent_status()
+        if consent == CONSENT_OK:
+            self.enable()
+        elif consent == CONSENT_UNKNOWN:
+            webapp.settings['analytics_prompt'] = True
+
+        base_url = webapp.settings['base_url']
+        webapp.add_handlers(".*$", [
+            (ujoin(base_url, r"/analytics/consent"), ConsentHandler,
+                {'sender': self}),
+        ])
+
+    def enable(self):
+        self.enabled = True
+        self.nbapp.log.info("Enabling analytics")
+        self.read_data()
+        self.record_version('jupyter_notebook', __version__)
+        self.record_version('server_python', sys.version)
+        self.schedule_send()
+
+        self.socket = zmq.Context.instance().socket(zmq.PULL)
+        port = self.socket.bind_to_random_port('tcp://127.0.0.1')
+        self.nbapp.kernel_manager.analytics_port = port
+        self.stream = ZMQStream(self.socket)
+        self.stream.on_recv(self.on_recv)
+
+    def on_recv(self, msg_parts):
+        data = json.loads(msg_parts[0])
+        msg_type = data['msg_type']
+        if msg_type == 'version':
+            self.record_version(data['project'], data['version'])
+        elif msg_type == 'feature':
+            self.record_feature(data['project'], data['feature'])
+
+    def schedule_send(self):
+        last = self.last_submission()
+        if last is None:
+            self.ioloop.add_callback(self.send)
+            return
+
+        since_last = datetime.utcnow() - self.last_submission()
+        if since_last > timedelta(days=28):
+            self.ioloop.add_callback(self.send)
+        else:
+            self.ioloop.call_later(since_last.total_seconds(), self.send)
+
+    def send(self):
+        http_client = AsyncHTTPClient()
+        # http_client.fetch('https://...', method='POST',
+        #                   body=json.dumps(self.data))
+        self.control_info()['last_submission'] = datetime.utcnow().isoformat()
+        self.store_control_info()
+
+        # Reset the data for the next month's collection
+        self.data = {}
+        self.record_version('jupyter_notebook', __version__)
+        self.record_version('server_python', sys.version)
+        self.ioloop.call_later(timedelta(days=28).total_seconds(), self.send)
+
+    def record_version(self, project, version):
+        proj_data = self.data.setdefault(project, {})
+        versions = set(proj_data.get('versions', []))
+        if version not in versions:
+            versions.add(version)
+            proj_data['versions'] = list(version)
+            self.store_data()
+
+    def record_feature(self, project, feature):
+        proj_data = self.data.setdefault(project, {})
+        features = set(proj_data.get('features', []))
+        if feature not in features:
+            features.add(feature)
+            proj_data['features'] = list(features)
+            self.store_data()
+
+class ConsentHandler(IPythonHandler):
+    def initialize(self, sender):
+        self.sender = sender
+
+    @web.authenticated
+    @json_errors
+    def post(self):
+        data = self.get_json_body()
+        if data['status']:
+            self.sender.change_consent_status(CONSENT_OK)
+        else:
+            self.sender.change_consent_status(CONSENT_DENIED)
--- a/jupyter_notebook/analytics.py
+++ b/jupyter_notebook/analytics.py
@@ -144,11 +144,12 @@ class AnalyticsSender(object):
             self.ioloop.add_callback(self.send)
             return
 
-        since_last = datetime.utcnow() - self.last_submission()
+        since_last = datetime.utcnow() - last
         if since_last > timedelta(days=28):
             self.ioloop.add_callback(self.send)
         else:
-            self.ioloop.call_later(since_last.total_seconds(), self.send)
+            to_next = (last + timedelta(days=28)) - datetime.utcnow()
+            self.ioloop.call_later(to_next.total_seconds(), self.send)
 
     def send(self):
         http_client = AsyncHTTPClient()
--- a/jupyter_notebook/notebookapp.py
+++ b/jupyter_notebook/notebookapp.py
@@ -210,6 +210,7 @@ class NotebookWebApplication(web.Application):
             config=ipython_app.config,
             jinja2_env=env,
             terminals_available=False,  # Set later if terminals are available
+            analytics_prompt=False,  # Set later if the user hasn't answered before
         )
 
         # allow custom overrides for the tornado web app.
@@ -906,6 +907,11 @@ class NotebookApp(JupyterApp):
             log = self.log.debug if sys.platform == 'win32' else self.log.warn
             log("Terminals not available (error was %s)", e)
 
+    def init_analytics(self):
+        from .analytics import AnalyticsSender
+        self.analytics = AnalyticsSender(self)
+        self.analytics.startup()
+
     def init_signal(self):
         if not sys.platform.startswith('win'):
             signal.signal(signal.SIGINT, self._handle_sigint)
@@ -1015,6 +1021,7 @@ class NotebookApp(JupyterApp):
         self.init_webapp()
         self.init_kernel_specs()
         self.init_terminals()
+        self.init_analytics()
         self.init_signal()
         self.init_server_extensions()
 


minrk:
Don't forget to document the new message type(s) in the message spec doc, and rev the spec version.

jupyter_notebook/analytics.py
```diff
@@ -0,0 +1,194 @@
+from datetime import datetime, timedelta
```
minrk: plans for any docstrings anywhere?

takluyver: In time, yes - I want to discuss what's going on before polish the code and add things like tests and docs.

minrk: ok

jupyter_notebook/analytics.py
```diff
@@ -0,0 +1,194 @@
...
+    def enable(self):
+        self.enabled = True
+        self.nbapp.log.info("Enabling analytics")
+        self.read_data()
+        self.record_version('jupyter_notebook', __version__)
+        self.record_version('server_python', sys.version)
+        self.schedule_send()
+
+        self.socket = zmq.Context.instance().socket(zmq.PULL)
+        port = self.socket.bind_to_random_port('tcp://127.0.0.1')
```
minrk: This ip and transport probably need to be configurable.

minrk:
I think the Session object should be used to send these messages, even though they are simple. I don't think there's a good reason for us to have multiple wire formats, even if one is single-frame JSON. This might as well be a new channel and message type, rather than a whole independent spec.

jupyter_notebook/analytics.py
takluyver: I was assuming so, but it wasn't immediately obvious where to get the config values from. `NotebookApp.ip` is the IP address for the HTTP server, I think, not for the ZMQ sockets. I guess it's configured on KernelManager, but fishing it out of the config seems ugly.

jupyter_notebook/analytics.py
```diff
@@ -0,0 +1,194 @@
...
+        webapp = self.nbapp.web_app
+
+        consent = self.consent_status()
+        if consent == CONSENT_OK:
+            self.enable()
+        elif consent == CONSENT_UNKNOWN:
+            webapp.settings['analytics_prompt'] = True
+
+        base_url = webapp.settings['base_url']
+        webapp.add_handlers(".*$", [
```
minrk: Presumably this should be registered with the rest of the add_handlers calls in notebookapp, rather than here.

jupyter_notebook/analytics.py
minrk: Shouldn't be ugly if AnalyticsSender is Configurable.

takluyver:
Yeah, I imagine we'll end up serialising it like the other messages, but again, I was deliberately keeping it simple until we've discussed the overall architecture. I've put it on the agenda for tomorrow's dev meeting.

jupyter_notebook/analytics.py
takluyver: I'm in two minds about that - it's nice to have all the routing information in one place, but it's also nice for things like this to be a self contained unit, so the parent app just asks it to set itself up. This is similar to what happens with terminals.

I was also thinking that there could be a data handler that could receive data from the frontend, e.g. browser name and version, and potentially data from nbextensions.

jupyter_notebook/analytics.py
takluyver: But you shouldn't be configuring those values twice, right? There should be one transport and one IP (/IPC path) setting which affect all ZMQ connections.

jupyter_notebook/analytics.py
minrk: Yeah, I have the same thoughts myself sometimes. Moving to an `service.enable(app)` pattern has a certain amount of appeal. If it is an always-on non-optional service like the others (as it is presently), it seems like it should probably be registered in a similar fashion to the others, though. The terminal one makes sense to be separate because it may not be available.

jupyter_notebook/analytics.py
minrk: This is the first port that we have where the server listens for connections from the kernels, so it's not duplicated anywhere, yet. But that does point to it maybe being attached to KernelManager for the ip of the Manager relative to the kernel, as opposed to the current KernelManager.ip, which is the kernel relative to the manager.

jupyter_notebook/analytics.py
takluyver: That makes sense. Though then that means that AnalyticsSender (which should possibly be AnalyticsManager, but I can hear Paul groaning) needs to be attached to the web application and then looked up in the handler - we do that for a bunch of other managers, but it feels decidedly ugly. Part of what I like about this is that I can pass the manager instance to the handler without having to send it round via the application.

minrk: Fair point. In that case, let's try taking this as an exercise in moving the registration logic to a given service, and see if we like it better than what we've done so far with everything else. If nothing else, it should work better if we ever decide to enable/disable services one at a time, which we have discussed in the past.

jupyter_notebook/analytics.py
takluyver: Shouldn't the ZMQ endpoint URL be the same for both directions? It looks like we use `KernelManager.ip` for connecting to the kernel as well (see `ConnectionFileMixin.connect_*`).

minrk: > It looks like we use KernelManager.ip for connecting to the kernel as well.

Yes, it's the IP of the Kernel as seen from the KernelManager. We haven't used this value to bind sockets on the server yet.

> Shouldn't the ZMQ endpoint URL be the same for both directions?

If we assume kernels and servers are always on the same machine, yes. If we want to enforce that assumption, there's really no reason for the value to ever be anything but 127.0.0.1.

takluyver: Oh, I think I see - the ZMQ endpoint is on the kernel's machine. But my understanding of ZMQ is that it largely doesn't matter which side binds and which connects - could we have all the endpoints on the server's machine and have the kernels bind to the endpoint they see that as, thus avoiding having to have two separate config values?

minrk:
What are the structure of the messages and the stored data? Is there only one message type? What keys and possible values?

jupyter_notebook/analytics.py
minrk: Right now, it's the kernel that binds, so if you don't want a new value, the thing to do is make the kernel bind for this port as well. If you do that, I do think PUB-SUB is the right pattern.

takluyver:
The message types are 'version' and 'feature'. They both contain a project key (e.g. 'IPython', 'numpy'), and respectively a 'version' and 'feature' key, each of which is a string.

https://github.com/jupyter/notebook/pull/81/files#diff-7654a22a93a632a9e1b54546c57f0215R133

Technically, they're both doing exactly the same thing - adding an element to a set. This could obviously be generalised much more, but I wanted to keep it simple and fairly minimal.

jupyter_notebook/analytics.py
takluyver: So does the endpoint have to be on the machine that binds the socket?

I'd rather keep this PUSH-PULL with a single collector if practical, it feels much neater than having a separate channel for each kernel.

jupyter_notebook/analytics.py
```diff
@@ -0,0 +1,194 @@
...
+        # http_client.fetch('https://...', method='POST',
+        #                   body=json.dumps(self.data))
+        self.control_info()['last_submission'] = datetime.utcnow().isoformat()
+        self.store_control_info()
+
+        # Reset the data for the next month's collection
+        self.data = {}
+        self.record_version('jupyter_notebook', __version__)
+        self.record_version('server_python', sys.version)
+        self.ioloop.call_later(timedelta(days=28).total_seconds(), self.send)
```
rgbkrk: Would the ioloop have to be running an entire 28 days to run this? I certainly don't keep a notebook server running that long.

rgbkrk:
I'd prefer this be `metrics` rather than `analytics`, including the `metrics_port`.

jupyter_notebook/analytics.py
takluyver: No, neither do I. It schedules a send when the notebook server starts, based on the last sent date recorded in control.json. This is just to ensure it works as expected if someone *does* keep their server running for long periods.

minrk:
Does it make more sense for this to be something standalone, rather than part of the notebook server, and added to the formal message spec?

jupyter_notebook/analytics.py
minrk: > So does the endpoint have to be on the machine that binds the socket?

Yes. It's identical to the ip used when the kernel binds, just reversing which machine is binding, and thus which machine the ip should refer to. The ip should be an IP of the server machine, and accessible from the kernel machine(s). Just like the kernels, localhost is the logical default.

> I'd rather keep this PUSH-PULL with a single collector if practical, it feels much neater than having a separate channel for each kernel.

Sure. It just depends on how much of a special case you want this to be in the message spec.

Carreau:
> Does it make more sense for this to be something standalone, rather than part of the notebook server, and added to the formal message spec?

We discussed that a bit, and we need _some_ integration with notebook server. Though having the tracking library as standalone make also sens, a first iteration/prototype using the notebook server make sens IMHO.

takluyver:
There are two important points where we're using the notebook server: the UI to ask a user for consent, and as a long-running/predictably started process that can spare the cycles to deal with this - I don't want to be trying to upload data in atexit hooks or something, and I'd rather not write a separate daemon to deal with it.

njsmith:
It might not be hard to do data collection and upload without a server, by having the metrics collection library (a) store data directly using sqlite or something that has a strategy for coordinating multiple independent readers/writers, and (b) notice when it's time to do an upload and spawn off a little background process to do it and reset the database when it succeeds. I guess we'll need most of this anyway given that we want our metrics to still work even if someone is running multiple notebook servers at the same time in the same home directory. But we definitely need help from the server for consent, so for version 0.1 we might as well do whatever's simplest.

While it's true that there are a lot of limitations to count data, I think there are some compelling reasons to still want it:

* Relative comparisons over time: even if it's meaningless to see that there are 2 million calls to `foo` in March, it's potentially meaningful to see that in April this has dropped to 100,000.
* Relative comparisons between features: there's some debate in NumPy right now about if, supposing we change how `np.dot` works in some corner cases, whether we should also change the public C API function that currently it just calls under the hood (`PyArray_MatrixProduct2`). For discussing this, it would be useful to know how many people are using the C API directly versus the Python-level API -- but every call to the Python-level API also causes a call to the C API. One way to solve this would be to count calls to both separately, and then the difference is the number of direct calls to the C API.

I'd suggest making the data collection API look like:

```
metrics.add(project_key, feature, value=None, count=1)
metrics.register_pull_function(pull_function)
```

Example:

metrics.add("numpy-5605-4de1-ae10", "version", "1.9.1")

def numpy_pull_stats():
    # For speed, `PyArray_MatrixProduct` counts how many times it is called in a C-level variable
    # which we batch up and submit to the metrics engine later
    pycount, ccount = get_stats_from_c_variables()
    stats.submit("numpy-5605-4de1-ae10", "dot.python-level", count=pycount)
    stats.submit("numpy-5605-4de1-ae10", "dot.c-level", count=ccount)

Idea:

`project_key` is a unique string identifying the group who is taking responsibility for collecting this data. The idea being that if you want to get your data back out of ipython's server, then you have to register one of these keys and affirm that you've read our privacy doc and thought about what data you'll collect.

`feature` is an arbitrary (unicode) string identifying what thing you are counting (scoped by `project_key`), by convention hierarchical by dots, b/c conventions are nice.

`value` is another arbitrary (unicode) string, or None. We count (feature, value) pairs. The split between feature and value is arbitrary (one could just encode the (feature, value) tuple as a single string), but IME in these kind of little metric systems key/value pairs are so common that it's nice to have sugar for them. An alternative would be to allow `value` to be an arbitrary JSON-encodable value.

Firefox has some interesting documents to potentially use as a model:

https://www.mozilla.org/en-US/privacy/principles/
https://www.mozilla.org/en-US/privacy/firefox/

njsmith:
Oh right, the other example I thought of and then forgot about, for why keeping a count is useful: cases where the thing being counted actually is a meaningful thing to count, not something that happens inside a loop. E.g., if you are trying to figure out how much usage the different kernels get, and there's someone who over the course of a month started up the R kernel once for 15 minutes and started up the python kernel 45 times for 2-4 hours apiece, then probably you do not want to just collapse away the numeric information and record this as "one person who uses both R and Python".

The mention of times also suggests to me that maybe we should allow the `count` variable to be a float -- hey, it works for javascript!

Also, better idea: instead of having the first argument be a random token that gets assigned, make it an email address. That nicely encodes the idea that this is some person or group who is taking responsibility here: `metrics.add("ipython-devs@ipython.org", ...)`

njsmith:
Oo, and here's yet another example of why we want counts :-). Say we're
thinking of adding some "small array" optimizations to numpy (e.g. a
customized free list). One immediately has some complicated decisions to
make about what sizes are worth optimizing, which sizes to lump into a
single bucket, etc. The data you need to make these decisions is like --
what percent of allocations were 4 bytes, what percent were 8 bytes, what
percent were 12 bytes, etc. Easy to get by instrumenting the code, but you
really want to collect these statistics on a large and diverse set of
uses...

takluyver:
The implementation certainly needs to be smarter about multiple processes that might be trying to write to it at the same time. I have an idea for this that I like better than using sqlite.

However, development of this is on hold for a bit while we work out whether we need to get human subject review on this because we're in an academic setting.

njsmith:
Anything you do will have *some* locking mess, because locking is basically
impossible to get right in all situations (remote file systems, etc.), so
oh well, not much to be done.

Re human subjects: the document you need is here:
http://cphs.berkeley.edu/review.html

The key point I believe is that you only need to do paperwork if your
project is "designed to develop or contribute to generalizable knowledge",
i.e. if your goal is to do science and learn about how things work in
general, *not* if your goal is to study a specific entity. (I think of this
as the "would a journal care about my results" test.) They give the
specific example that if you measure how well a class learned because you
want to know if your curriculum works, then this is not "human subjects
research"; if you measure how well a class learned because you want to know
how curriculums work in general, then this is human subjects research.

In our case, we don't care about writing journal articles about general
principles for software design, we just want to know what the right thing
to do is in our specific situation. Ergo, not human subjects research.

If you want to confirm this interpretation, then there's a phone number to
call and talk to an expert, and a procedure for getting the result in
writing, given here:
  http://cphs.berkeley.edu/faqs.html#a1

Carreau:
@ellisonbg told us that at calpoly it is particularly complicated as soon
as you plan on sharing the results. Which we obviously want to do with the
tracked project. And we want to be sure that we don't start working on that
too early.

Also anyway I think that the data collection should be done in a manner
which is not Jupyter-specific, and could work in a pure-python setting
(given the right configuration). So we can still work on that separately,
cause I suppose we each like to track our own usage if features. The
question of wether we share/collect this data is othogonal.

njsmith:
Well, it depends on the data, how you gathered it, what your original
proposal said, what you told the participants, etc. If you have videos of
minors in jail talking about sensitive subjects or people's health records
or something then you have a massive problem; at the other end of the scale
the IRB doesn't care at all about sharing of data that doesn't count as
human subjects data. It just depends.

I can call that number on Monday for you if you want to get a definitive
answer :-).

ellisonbg:
Nathaniel, It would be great if you wanted to reach out to the Berkeley
IRB. Eventuall, we would also have to involve the IRB at Cal Poly as well,
but this would a good start.

The other point that I found with our IRB is that if we were simply
gathering data that we would use *privately* to improve the software, that
would not meet the "generalizable knowledge critieria" and no review was
required. However, when I told them the project/software was open source
and that we should want to share the data/results with the broader
community of the project, they said "that is publication" so IRB approval
and human subject consent would be required.

If we are OK defining a small set of "researchers" that will be able to see
the data and results, but not sharing anything with the public, I think we
have a good amount of freedom. But the second we want to share the
information, I think we have to bring the IRB in. One question is this: can
we limit access to the raw data to a small set of defined "researchers" and
share summaries of the results publicly with out IRB review. Based on my
previous discussions with them, my guess is not.

The other big red flag is that we are talking about creating a general
Python API that any Python code can call to gather information. While we
may not ourselves gather sensitive human subject data, other libraries
could use those APIs to gather than information. If *we* ever deployed the
notebook with central data collection enabled (if we don't, none if this is
useful) we would need to ensure that third party python packages are not
using our API s to send us sensitive data.

Cheers,

Brian





On Sat, May 16, 2015 at 2:19 AM, Nathaniel J. Smith <

> Well, it depends on the data, how you gathered it, what your original
> proposal said, what you told the participants, etc. If you have videos of
> minors in jail talking about sensitive subjects or people's health records
> or something then you have a massive problem; at the other end of the scale
> the IRB doesn't care at all about sharing of data that doesn't count as
> human subjects data. It just depends.
>
> I can call that number on Monday for you if you want to get a definitive
> answer :-).
> On May 15, 2015 7:20 PM, "Matthias Bussonnier" <notifications@github.com>
> wrote:
>
>
> > @ellisonbg told us that at calpoly it is particularly complicated as soon
> > as you plan on sharing the results. Which we obviously want to do with
> the
> > tracked project. And we want to be sure that we don't start working on
> that
> > too early.
> >
> > Also anyway I think that the data collection should be done in a manner
> > which is not Jupyter-specific, and could work in a pure-python setting
> > (given the right configuration). So we can still work on that separately,
> > cause I suppose we each like to track our own usage if features. The
> > question of wether we share/collect this data is othogonal.
> >
> > On Fri, May 15, 2015 at 6:13 PM, Nathaniel J. Smith <
> > notifications@github.com> wrote:
> >
> > > Anything you do will have *some* locking mess, because locking is
> > basically
> > > impossible to get right in all situations (remote file systems, etc.),
> so
> > > oh well, not much to be done.
> > >
> > > Re human subjects: the document you need is here:
> > > http://cphs.berkeley.edu/review.html
> > >
> > > The key point I believe is that you only need to do paperwork if your
> > > project is "designed to develop or contribute to generalizable
> > knowledge",
> > > i.e. if your goal is to do science and learn about how things work in
> > > general, *not* if your goal is to study a specific entity. (I think of
> > this
> > > as the "would a journal care about my results" test.) They give the
> > > specific example that if you measure how well a class learned because
> you
> > > want to know if your curriculum works, then this is not "human subjects
> > > research"; if you measure how well a class learned because you want to
> > know
> > > how curriculums work in general, then this is human subjects research.
> > >
> > > In our case, we don't care about writing journal articles about general
> > > principles for software design, we just want to know what the right
> thing
> > > to do is in our specific situation. Ergo, not human subjects research.
> > >
> > > If you want to confirm this interpretation, then there's a phone number
> > to
> > > call and talk to an expert, and a procedure for getting the result in
> > > writing, given here:
> > > http://cphs.berkeley.edu/faqs.html#a1
> > > On May 15, 2015 4:51 PM, "Thomas Kluyver" <notifications@github.com>
> > > wrote:
> > >
> > > > The implementation certainly needs to be smarter about multiple
> > processes
> > > > that might be trying to write to it at the same time. I have an idea
> > for
> > > > this that I like better than using sqlite.
> > > >
> > > > However, development of this is on hold for a bit while we work out
> > > > whether we need to get human subject review on this because we're in
> an
> > > > academic setting.
> > > >
> > > > —
> > > > Reply to this email directly or view it on GitHub
> > > > <https://github.com/jupyter/notebook/pull/81#issuecomment-102549803
> >.
> > >
> > > >
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/jupyter/notebook/pull/81#issuecomment-102558889>.
> > >
> >
> > —
> > <https://github.com/jupyter/notebook/pull/81#issuecomment-102565442>.
>
> >
>
> —
> <https://github.com/jupyter/notebook/pull/81#issuecomment-102592837>.
>



-- 
Brian E. Granger

njsmith:
I did a bit of looking into how Mozilla handles this, and I'll leave some links here for reference:

They collect histograms: https://developer.mozilla.org/en-US/docs/Mozilla/Performance/Adding_a_new_Telemetry_probe

And send them back as JSON documents (~one per day): https://ci.mozilla.org/job/mozilla-central-docs/Tree_Documentation/

Public aggregated data: https://telemetry.mozilla.org

The originally included [no unique installation id](https://blog.mozilla.org/tglek/2011/05/13/firefox-telemetry/), but eventually [changed their mind](http://robertovitillo.com/2014/12/19/clientid-in-telemetry-submissions/) and started including one (except for specific data that is deemed too sensitive and thus split off into separate unlabeled submissions), because they found the [restriction was too cumbersome](https://mail.mozilla.org/pipermail/firefox-dev/2014-December/002580.html).

One thing I was surprised by, though I guess it makes sense, is that they *unconditionally* send a ping (complete with unique per-installation ID) at (1) first startup, (2) if/when a user opts *out* of telemetry. (And these pings include some basic data -- Firefox version, OS, etc.) I guess this makes sense -- there isn't really much chance of a privacy violation in knowing that "someone somewhere installed Firefox and that's all they want us to know", and it's crucial to extrapolating from the data they do get to the overall user population.

An internal planning document with lots of interesting details on e.g. metrics they want to track: https://docs.google.com/document/d/1IGpzsYGi_sq3YFQDAPyKOkU_BKvXAC95fZYA2i4ceVs/edit#

takluyver:
This machinery will now be done in https://github.com/njsmith/sempervirens/

Issue #88, opened 2015-05-13
highligting on open notebook wrong after execution.


Carreau:
See https://github.com/EntropyOrg/p5-Devel-IPerl/issues/40

Likely candidate for (painful) backport.

PR #89, opened 2015-05-13
Store codemirror config in only one location

Carreau:
Closes (once backported) EntropyOrg/p5-Devel-IPerl#40
and jupyter/notebook#88

### DIFF
--- a/jupyter_notebook/static/notebook/js/cell.js
+++ b/jupyter_notebook/static/notebook/js/cell.js
@@ -76,7 +76,7 @@ define([
         // backward compat.
         Object.defineProperty(this, 'cm_config', {
             get: function() {
-                console.warn("warning you are accessing a deprecated field")
+                console.warn("Warning: accessing Cell.cm_config directly is deprecate.")
                 return that._options.cm_config;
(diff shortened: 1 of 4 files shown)


jupyter_notebook/static/notebook/js/cell.js
```diff
@@ -73,14 +73,23 @@ define([
             }
         });
 
+        // backward compat.
+        Object.defineProperty(this, 'cm_config', {
+            get: function() {
+                console.warn("warning you are accessing a deprecated field")
```
takluyver: Does console.warn() say which file & line the warning came from? If not, this message should be more specific.

takluyver:
I got half way through converting the cell classes to the new config system, but didn't manage to finish. I must get back to that at some point, but this looks like a reasonable compromise for the moment.

jupyter_notebook/static/notebook/js/cell.js
Carreau: It depends on the browser. But usually you get the file and line-number on the left side of the developper console.

PR #102, opened 2015-05-19
Start working on removing global JS IPython

Carreau:
1) rename to jupyter.
2) Start removing this weird "we require the global namespace to add ourself to it"

Note that 2 is all or nothing, or requireJS complain that `module Xyz has not yet be loaded in current contex`.

So if we agreed to trop the IPython/Jupyter global (at least to store module) for 4.0 that would be great and simplify things a lot.

### DIFF
--- a/examples/Notebook/Custom Keyboard Shortcuts.ipynb
+++ b/examples/Notebook/Custom Keyboard Shortcuts.ipynb
@@ -38,7 +38,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The keypress `r` in command mode is by default bound to change the type of a cell ro `raw`. This action will thus be unavaillable from the keyboard now."
+    "\"By default the keypress `r`, while in command mode, changes the type of the selected cell to `raw`.  This shortcut is overridden by the code in the previous cell, and thus the action no longer be available via the keypress `r`.\""
    ]
   },
   {
--- a/examples/Notebook/Custom Keyboard Shortcuts.ipynb
+++ b/examples/Notebook/Custom Keyboard Shortcuts.ipynb
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Starting with IPython 2.0 keyboard shortcuts in command and edit mode are fully customizable. These customizations are made using the IPython JavaScript API. Here is an example that makes the `r` key available for running a cell:"
+    "Starting with IPython 2.0 keyboard shortcuts in command and edit mode are fully customizable. These customizations are made using the Jupyter JavaScript API. Here is an example that makes the `r` key available for running a cell:"
    ]
   },
   {
@@ -24,7 +24,7 @@
    "source": [
     "%%javascript\n",
     "\n",
-    "IPython.keyboard_manager.command_shortcuts.add_shortcut('r', {\n",
+    "Jupyter.keyboard_manager.command_shortcuts.add_shortcut('r', {\n",
     "    help : 'run cell',\n",
     "    help_index : 'zz',\n",
     "    handler : function (event) {\n",
@@ -32,6 +32,13 @@
     "        return false;\n",
     "    }}\n",
     ");"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The keypress `r` in command mode is by default bound to change the type of a cell ro `raw`. This action will thus be unavaillable from the keyboard now."
    ]
   },
   {
@@ -55,7 +62,7 @@
    "source": [
     "%%javascript\n",
     "\n",
-    "IPython.keyboard_manager.command_shortcuts.add_shortcut('r', function (event) {\n",
+    "Jupyter.keyboard_manager.command_shortcuts.add_shortcut('r', function (event) {\n",
     "    IPython.notebook.execute_cell();\n",
     "    return false;\n",
     "});"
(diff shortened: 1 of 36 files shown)


Carreau:
This now uses `properties` to load lazily the module/constructor and console.log a deprecation warning.
So require does not complain.

minrk:
Are we leaving the global `IPython` for backward compat, or no?

takluyver:
I think we should be leaving the IPython name for a while. I know we say the JS API isn't stable, but people are using it anyway, and breaking all code when it's easy enough to leave that reference around would be a bit rude.

damianavila:
> I think we should be leaving the IPython name for a while. I know we say the JS API isn't stable, but people are using it anyway, and breaking all code when it's easy enough to leave that reference around would be a bit rude.

I completely agree with this opinion...

Carreau:
The global is also super useful for debugging, but I'm for slowly deprecating it with a message.

Here it's how it appears as soon as I type dot : 
screen shot 2015-05-21 at 10 10 21

takluyver:
Does it do the same with instances in the `IPython` namespace, or only with classes? `require()` doesn't help you get instances.

Carreau:
Right now only with Classes, most of the instances should be attached to the object that created them IMHO, but this is another discussion. 

I have no opposition (for now) of `main.js` mutating the global namespace, I just don't want the leaves of the dependency tree to require a global and mutate it.

Carreau:
Moar commits. less globals.

jdfreder:
Along the lines of what @minrk was saying in today's meeting, you still need a way to expose the singleton instances, which are true globals.  Looking at this PR, it looks like you aren't doing that (correct me if I'm wrong).  How about making the instances require-able via main.js?  i.e. `IPython.notebook` -> `require('notebook/js/main').notebook`?

Carreau:
I don't delete the IPython, neither remove the singleton instance. They are still exposed, I won't remove them in this PR.

Carreau:
Removed `[WIP]`. 
I start to reach the maximum of things that can be removed easily, and longer will be hard for review.

Carreau:
Travis is happy.

Carreau:
Ping.

notebook/static/base/js/security.js
```diff
@@ -59,7 +57,7 @@ define([
          * sanitize just the css in style tags in a block of html
          * called by sanitize_html, if allow_css is true
          */
-        var h = $("<div/>").append(html);
+        var h = document.createElement('div').append(html);
```
minrk: I don't think this works. element.append only adds a single node, not hierarchical html text.

minrk: Plus, `h.html()` called below won't work either.

Carreau: reverted

jdfreder:
You'll want to update `examples/Custom Keyboard shortcuts.ipynb` to use Jupyter instead of IPython.  I noticed this while working on #124 , but since this isn't merged, I didn't do it there.

Carreau:
@jdfreder Done, thanks.  I also updated the instructions a bit.

examples/Notebook/Custom Keyboard Shortcuts.ipynb
```diff
@@ -38,6 +38,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "The keypress `r` in command mode is by default bound to change the type of a cell ro `raw`. This action will thus be unavaillable from the keyboard now."
```
jdfreder: cell **to** `raw`

Also, I think moving things around in the sentence makes it read better:
"By default the keypress `r`, while in command mode, changes the type of the selected cell to `raw`.  This shortcut is overridden by the code in the previous cell, and thus the action no longer be available via the keypress `r`."

notebook/static/base/js/namespace.js
```diff
@@ -1,10 +1,86 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
+//
+//
```
jdfreder: Accident?

notebook/static/base/js/namespace.js
```diff
@@ -1,10 +1,86 @@
...
 
-var IPython = IPython || {};
-define([], function(){
+
+var Jupyter = Jupyter || {};
+
+var jprop = function(name, module_path){
+    Object.defineProperty(Jupyter, name, {
+      get: function() { 
+          console.warn('accessing `'+name+'` is deprecated. Use require("'+module_path+'")');
```
jdfreder: You surround `name` in backticks, so you should surround `require("...")` in backticks too.

notebook/static/base/js/namespace.js
```diff
@@ -1,10 +1,86 @@
...
+      },
+      enumerable: true,
+      configurable: false
+    });
+}
+
+var jglobal = function(name, module_path){
+    Object.defineProperty(Jupyter, name, {
+      get: function() { 
+          console.warn('accessing `'+name+'` is deprecated. Use require("'+module_path+'").'+name);
```
jdfreder: Same here : )

jdfreder:
Few comments, otherwise this has my stamp of approval !  :ok:

examples/Notebook/Custom Keyboard Shortcuts.ipynb
Carreau: done.

notebook/static/base/js/namespace.js
Carreau: done

notebook/static/base/js/namespace.js
Carreau: done

notebook/static/base/js/namespace.js
Carreau: and odne

jdfreder:
Thanks @Carreau

PR #105, opened 2015-05-21
update Dockerfile

minrk:
with new current repo layout

cc @rgbkrk


rgbkrk:
Thank you so much @minrk. Ended up rekicking the failed part of the Travis build since it's unrelated.

We'll probably need to update our Docker Hub links too.

PR #120, opened 2015-05-27
Hardcode widgets

jdfreder:
sibling PRs:
https://github.com/ipython/ipykernel/pull/19
https://github.com/ipython/ipywidgets/pull/38

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -238,6 +238,20 @@ class NotebookWebApplication(web.Application):
         handlers.extend(load_handlers('services.nbconvert.handlers'))
         handlers.extend(load_handlers('services.kernelspecs.handlers'))
         handlers.extend(load_handlers('services.security.handlers'))
+        
+        # BEGIN HARDCODED WIDGETS HACK
+        try:
+            import ipywidgets
+            handlers.append(
+                (r"/nbextensions/widgets/(.*)", FileFindHandler, {
+                    'path': ipywidgets.find_static_assets(),
+                    'no_cache_paths': ['/'], # don't cache anything in nbextensions
+                }),
+            )
+        except:
+            self.log.warn('ipywidgets package not installed.  Widgets are unavailable.')
+        # END HARDCODED WIDGETS HACK
+        
         handlers.append(
             (r"/nbextensions/(.*)", FileFindHandler, {
                 'path': settings['nbextensions_path'],
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -142,7 +142,7 @@ class NotebookWebApplication(web.Application):
             ipython_app, kernel_manager, contents_manager, cluster_manager,
             session_manager, kernel_spec_manager, config_manager, log, base_url,
             default_url, settings_overrides, jinja_env_options)
-        handlers = self.init_handlers(settings)
+        handlers = self.init_handlers(settings, ipython_app)
 
         super(NotebookWebApplication, self).__init__(handlers, **settings)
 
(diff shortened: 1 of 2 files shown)


notebook/notebookapp.py
```diff
@@ -238,6 +238,20 @@ def init_handlers(self, settings):
...
         handlers.extend(load_handlers('services.kernelspecs.handlers'))
         handlers.extend(load_handlers('services.security.handlers'))
+        
+        # BEGIN HARDCODED WIDGETS HACK
+        try:
+            import ipywidgets
+            handlers.append(
+                (r"/nbextensions/widgets/(.*)", FileFindHandler, {
+                    'path': ipywidgets.find_static_assets(),
+                    'no_cache_paths': ['/'], # don't cache anything in nbextensions
```
takluyver: It probably is safe to cache our own widget js, right? It may be a bit annoying while developing them, but for users who install the package, it should be able to use the cache.

takluyver:
I think this approach makes sense - it seems unlikely that we will have a satisfactory solution to the packaging questions by 4.0, and it looks like this should avoid any extra complexity for users.

notebook/static/notebook/js/main.js
```diff
@@ -50,6 +50,12 @@ require([
     ) {
     "use strict";
 
+    // BEGIN HARDCODED WIDGETS HACK
+    // Don't include in above require call, avoiding r.js
+    // and crash on import failure.
+    require(['nbextensions/widgets/notebook/js/extension'], function(){});
```
minrk: Shouldn't this have an errback to log when widgets are unavailable?

notebook/notebookapp.py
jdfreder: It probably is, but is it worth the annoyance?  I have no idea what kind of performance increase it'll give.  Does the browser cache work regardless?

notebook/notebookapp.py
```diff
@@ -238,6 +238,20 @@ def init_handlers(self, settings):
...
+                }),
+            )
+        except:
+            self.log.warn('ipywidgets package not installed.  Widgets are unavailable.')
```
takluyver:
```
AttributeError: 'NotebookWebApplication' object has no attribute 'log'
```

I'm not sure what's the best way to log from the web application class.

minrk: At this point, I think the current logger is available as `self.ipython_app.log`.

jdfreder:
Sorry, I've been gone forever, the last commit should fix the log call me thinks.  \*waits...\*

jdfreder:
*bump*

notebook/notebookapp.py
jasongrout: Now that widgets are their own package, I don't think they should be treated in a special way.

jdfreder: Too late for that https://github.com/ipython/ipykernel/pull/19 .  This is just a temporary thing, until the packaging problem is solved.

notebook/notebookapp.py
```diff
@@ -238,6 +238,20 @@ def init_handlers(self, settings):
...
+        try:
+            import ipywidgets
+            handlers.append(
+                (r"/nbextensions/widgets/(.*)", FileFindHandler, {
+                    'path': ipywidgets.find_static_assets(),
+                    'no_cache_paths': ['/'], # don't cache anything in nbextensions
+                }),
+            )
+        except:
+            ipython_app.log.warn('ipywidgets package not installed.  Widgets are unavailable.')
```
minrk: I think this can be `app_log`, and you don't need to add `ipython_app`as an arg to this method.

jdfreder:
@minrk see last commit

jdfreder:
Woot!  Looks like this was the missing piece causing the widget tests to fail.

PR #121, opened 2015-05-27
remove clusters tab from notebook

### LINKED ISSUE #85: Proposal: Move clusters tab to ipyparallel
I think we should ship it as a notebook server extension in ipyparallel, rather than keep it in the notebook code.

minrk:
moving to an nbextension in ipyparallel

closes #85


Carreau:
+1, I suppose you want to merge that in sync with #11

minrk:
In the absence of ipyparallel, there is now a placeholder Clusters tab that points for ipyparallel, rather than removing the tab altogether with no hints.

Carreau:
@minrk Want to rebase ? And once test passes. I'll merge the 2 PR at the same time.

minrk:
Rebased.

PR #124, opened 2015-05-28
Docs

### LINKED ISSUE #122: Docs
@jdfreder volunteered to start this off.

There are some docs already there. We should:

- [x] Update stuff (like `ipython notebook` -> `jupyter notebook`)
- [x] Remove docs that will live elsewhere (nbformat, nbconvert)
- [x] Document config options (I have machinery for this in IPython, can duplicate it or put part in traitlets)
- [ ] API docs - I think the public Python API is pretty small, but there should at least be enough to describe:
  - Extensions adding handlers
  - Custom contents managers, checkpoint managers, etc.
  - Save hooks
  - Probably some other stuff I'm not thinking of right now
- [ ] ??? Some amount of documentation of the JS APIs?
- [x] Docs should build on RTD

jdfreder:
partially closes #122 
sibling PRs:
https://github.com/ipython/ipython/pull/8500


jdfreder:
Notes from Day 1:

```rst
ipykernel:plotting


Plotting
--------
One major feature of the Jupyter notebook is the ability to display plots that 
are the output of running code cells. The IPython kernel is designed to work 
seamlessly with the matplotlib_ plotting library to provide this functionality.

To set this up, before any plotting is performed you must execute the
``%matplotlib``  :ref:`magic command <magics_explained>`. This performs the
necessary behind-the-scenes setup for IPython to work correctly hand in hand
with ``matplotlib``; it does *not*, however, actually execute any Python
``import`` commands, that is, no names are added to the namespace.

If the ``%matplotlib`` magic is called without an argument, the
output of a plotting command is displayed using the default ``matplotlib``
backend in a separate window. Alternatively, the backend can be explicitly
requested using, for example::

  %matplotlib gtk

A particularly interesting backend, provided by IPython, is the ``inline``
backend.  This is available only for the Jupyter Notebook and the
:ref:`IPython QtConsole <qtconsole>`.  It can be invoked as follows::

  %matplotlib inline

With this backend, the output of plotting commands is displayed *inline*
within the notebook, directly below the code cell that produced it. The
resulting plots will then also be stored in the notebook document.

.. seealso::

    `Plotting with Matplotlib`_  example notebook


ipykernel:install


IPython can be installed (different python versions, virtualenv or conda environments) can
be installed as a kernel by following these steps:

* make sure that the desired python installation is active (e.g. activate the environment)
  and ipython is installed
* run once ``ipython kernelspec install-self --user`` (or ``ipython2 ...`` or ``ipython3 ...``
  if you want to install specific python versions)

The last command installs a kernel spec file for the current python installation in
``~/.jupyter/kernels/``. Kernel spec files are JSON files, which can be viewed and changed with a
normal text editor.


nbconvert:index
nbformat:format_description
ipython:ipythonzmq
```

QUESTIONS:
is the config file still ipython_notebook_config.py ? and still in .ipython ?

Carreau:
`ipython_notebook_config.py` no.... , it will be in `usr/share/jupyter` or in ~/Library/Jupyter (osx) or .jupyter IIRC.

takluyver:
The config file should be `~/.jupyter/jupyter_notebook_config.py`, I believe.

jdfreder:
> The config file should be ~/.jupyter/jupyter_notebook_config.py, I believe.

That's what min said too, thanks guys.

takluyver:
This looks like a good start. I think you need to update the `conf.py` file to configure intersphinx for the links you're making into other docs.

jdfreder:
Note to self, stopped on `JavaScript Notebook Extensions.ipynb`

jdfreder:
As of the last commit, this PR has doc changes that depend on @Carreau 's  PR #102 .

jdfreder:
@takluyver This is bananas :banana: , https://readthedocs.org/builds/jdfreder-notebook/2905895/

Something in the notebook docs depends on QtConsole apparently.  GHPages FTW? :checkered_flag:

jdfreder:
@takluyver looks like the conf.py stuff isn't working - https://readthedocs.org/builds/jdfreder-notebook/2905991/ .  I think the notebook isn't getting installed before RTD runs.

jdfreder:
Just an update, I've made some progress with RTD, but it looks like it's failing now because of npm.
Latest here: https://readthedocs.org/builds/jdfreder-notebook/2906345/

docs/source/conf.py
```diff
@@ -22,6 +22,13 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #sys.path.insert(0, os.path.abspath('.'))
```
takluyver: If you uncomment this line and change `.` to `../..` (i.e. so it points to the repo root), the notebook modules should be importable even though it fails to install. It's a bit of a hack. I'd forgotten that the notebook can't be installed from the repo without npm.

jdfreder:
Success!  This is ready.  FYI - http://jdfreder-notebook.readthedocs.org/en/latest/

docs/autogen_config.py
```diff
@@ -0,0 +1,42 @@
...
+-------
+
+This list of options can be generated by running the following and hitting 
+enter::
+
+  $ jupyter notebook --help
+
+"""
+
+destination = os.path.join(os.getcwd(), 'config.rst')
```
takluyver: This will put the file in the wrong place when you build the docs locally using `make`. To do it relative to the location of this script:

```python
destination = os.path.join(os.path.dirname(__file__), 'source/config.rst')
```

jdfreder: Doing that gives me a __file__ undefined error on RTD, unfortunately : (

docs/source/conf.py
```diff
@@ -289,5 +296,8 @@
 #texinfo_no_detailmenu = False
 
 intersphinx_mapping = {
-    'ipython': ('http://ipython.org/ipython-doc/dev/', None)
+    'ipython': ('http://ipython.org/ipython-doc/dev/', None),
+    'ipykernel': ('http://ipython.org/ipykernel-doc/dev/', None),
+    'nbconvert': ('http://jupyter.org/nbconvert-doc/dev/', None),
+    'nbformat': ('http://jupyter.org/nbformat-doc/dev/', None)
```
takluyver: nbformat docs are at http://nbformat.readthedocs.org/en/latest/ . nbconvert docs will probably be on RTD, but I don't think they're built yet. And ipykernel may not need its own docs at all.

docs/autogen_config.py
jdfreder: try once, fail to the other maybe?

docs/source/conf.py
jdfreder: Assuming we move ipython to RTD too, should I just change it here too?

docs/autogen_config.py
takluyver: Hmm, that's annoying, exec doesn't automatically define `__file__`. Yeah, then I guess it should try one and fall back to the other.

docs/source/conf.py
takluyver: Leave it for now, as it's pointing to something that works. We can always change it later.

docs/requirements.txt
```diff
@@ -0,0 +1,9 @@
+jinja2
+tornado
+-e git+https://github.com/ipython/ipython_genutils.git#egg=ipython_genutils
```
Carreau: genutils is on PyPI.

docs/source/links.txt
```diff
@@ -16,23 +16,23 @@
 
    NOTE: Some of these were taken from the nipy links compendium.
 
-.. Main IPython links
+.. Main Jupyter notebook links
 
-.. _Basic Output: notebook_p2_
-.. _notebook_p2: http://nbviewer.ipython.org/urls/raw.github.com/ipython/ipython/1.x/examples/notebooks/Part%202%20-%20Basic%20Output.ipynb
+.. _Notebook Basics: notebook_p2_
+.. _notebook_p2: https://nbviewer.jupyter.org/urls/raw.github.com/ipython/ipython/3.x/examples/Notebook/Notebook%20Basics.ipynb
```
Carreau: I though all should be redirected to nbviewer.org without jupyter ?

Carreau:
minimal comments, but not really important. Merging.

PR #135, opened 2015-06-07
use splitext to remove extension from notebook name

minrk:
extension may not be .ipynb.

cc @rossant @gregcaporaso

closes ipython/ipython#8510

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2101,8 +2101,9 @@ define(function (require) {
      * Returns the filename with the appropriate extension, appending if necessary.
      */
     Notebook.prototype.ensure_extension = function (name) {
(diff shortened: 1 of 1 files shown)


Carreau:
We have the following around line 2100.
```
    Notebook.prototype.ensure_extension = function (name) {
        if (!name.match(/\.ipynb$/)) {
            name = name + ".ipynb";
        }
        return name;
    };
```

takluyver:
So, do we merge this as is? It seems substantially more complex to do the right thing when adding the extension, so perhaps it should be left to another PR.

minrk:
We know the extension used to load the file, so we should be able to ensure the extension doesn't change.

minrk:
ensure_extension should preserve the existing extension, rather than enforcing .ipynb unconditionally

PR #141, opened 2015-06-10
shortcut: explicit exception message

nbarray:
Hi !
Let me introducte myself. My name is Nicolas and I am a member student at
the Developpement and Research Laboratory of EPITA in Paris. For my work I
have to create custom widgets with Jupyter for the Vcsn platform.

While messing with the shortcut manager I stumbled upon a weird exception
message saying that I cannot remove a non-leaf shortcut... After a long time
(an hour or two) I finally understood that this message what actually saying
'You cannot remove a shortcut that does not exists'.

I wanted to post this patch s.t. no one has to lost time in understanding that.
Here is the commit message:

In the function remove_shortcut, if the action failed, it used to
throw an 'implementation oriented' exception message:
> throws('try to delete non-leaf')
This is not a good message for the user who wants to mess with the
shortcuts.

The solution is to leave this 'implementation oriented' exception message
and add a try-catch bloc above that catches this message and throws
a new, frendliest, exception message.
I came up with this:
> throws('try to remove non-existing shortcut')

* notebook/static/base/js/keyboard.js: (remove_shortcut) add try-catch.

Signed-off-by: Nicolas Barray <nbarray@gmail.com>

### DIFF
--- a/notebook/static/base/js/keyboard.js
+++ b/notebook/static/base/js/keyboard.js
@@ -385,10 +385,19 @@ define([
         if( typeof(shortcut) === 'string'){
             shortcut = shortcut.split(',');
         }
-        this._remove_leaf(shortcut, this._shortcuts);
-        if (!suppress_help_update) {
+        /*
+         *  The shortcut error should be explicit here, because it will be
+         *  seen by users.
+         */
+        try
+        {
+          this._remove_leaf(shortcut, this._shortcuts);
+          if (!suppress_help_update) {
             // update the keyboard shortcuts notebook help
             this.events.trigger('rebuild.QuickHelp');
+          }
+        } catch (ex) {
+          throw ('try to remove non-existing shortcut');
         }
     };
 


Carreau:
Hey Nicholas, 

Well you probably one of the few people that have read the code of the shortcut manager. 
And I think you are right. 

Any reason to throw a string instead of a new `Error` instance ? I not 1000% sure if there is a difference. @jdfreder ?

N'hésite pas à demander si tu ne veux reste a nouveau bloqué sur un problème comme ça.

nbarray:
Hey !

> Any reason to throw a string instead of a new Error instance ?
I was just doing the same as the throws in the existing code. Yet it seems that
if the exception can go outside the code base, Error would be better.

> N'hésite pas à demander si tu ne veux reste a nouveau bloqué sur un problème comme ça.
Merci beaucoup, je le ferais dans ce cas :)

Bonne soirée,

Carreau:
> I was just doing the same as the throws in the existing code. Yet it seems that
> if the exception can go outside the code base, Error would be better.

Indeed, I didn't new about throwing error in Js at the time. I'll merge and add an issue to remind me to update all the throw.

PR #147, opened 2015-06-10
Die immediately on SIGINT if not in a tty

### LINKED ISSUE #146: Ctrl-C twice only in a TTY.
If non interactive should shutdown immediately.

takluyver:
Closes #146


Carreau:
+1

minrk:
And here I was about to mark this as sprint-friendly.

PR #152, opened 2015-06-11
[Test PR] Cache node

Carreau:
And make gulp/npm explicit to get an idea of why installing take 300+ seconds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 # http://travis-ci.org/#!/ipython/ipython
 language: python
+cache:
+    directories:
+        - ~/.cache/bower
 python:
     - 3.4
     - 2.7


Carreau:
Seem to speed-up the build by ~1min/matrix item.

minrk:
Makes sense that it's bower, since a huge amount of the time is spent extracting mathjax.

Carreau:
I'll rebase/squash.

Carreau:
could we have a setup.py --nojs for pure python tests ? That would shave a minute or so too.

minrk:
We could, but I'm not sure it's worth it. Install shouldn't ever be allowed to succeed without the final compiled js.

Carreau:
like a `--yes-really-i-dont-want-js-cause-i-m-running-on-travis-just-for-python-tests`

PR #156, opened 2015-06-16
test: stop caching bower

minrk:
see if this fixes our bower problems

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 # http://travis-ci.org/#!/ipython/ipython
 language: python
-cache:
-    directories:
-        - ~/.cache/bower
 python:
     - 3.4
     - 2.7
--- a/setupbase.py
+++ b/setupbase.py
@@ -334,7 +334,7 @@ class Bower(Command):
         
         try:
             check_call(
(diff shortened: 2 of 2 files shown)


minrk:
just disabling cache: nope. Trying passing `--no-force-latest`

takluyver:
Still nope :-(

SylvainCorlay:
`--force-latest false`? https://github.com/yatskevich/grunt-bower-task/issues/7

SylvainCorlay:
@minrk  `bower cache clean` should clean the cache.

Carreau:
I just cleared travis caches. There is a button for that.

minrk:
It's definitely not the Looks like @Carreau's clearing of the cache was enough. I'm not sure why disabling the caching altogether wasn't, but I'm guessing it's not something you can do in a PR maybe?

Carreau:
I reran 1 on the job, that passed. Reopenning and rerunning everything to see.

minrk:
Purging the cache seems to have fixed things. Master and downstream builds are passing. I don't think anything in this PR is relevant, though there is some chance that `--force-latest=false` will prevent this issue in the future.

PR #157, opened 2015-06-17
Allow passing a comm_id to new_comm

SylvainCorlay:
This is a minor backward compatible change.

### DIFF
--- a/notebook/static/services/kernels/comm.js
+++ b/notebook/static/services/kernels/comm.js
@@ -31,12 +31,14 @@ define([
         }
     };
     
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I found a way around it, so it might not be necessary anymore, even though it would make sense. Besides, the tests should be restarted.

jdfreder:
Hmm I'm surprised `new_comm` doesn't exist in comm/manager.py of ipykernel.  

Anyways, +1 to this change.  I think it will simplify some code in the widget persistence..

SylvainCorlay:
Yes, I want to use it in https://github.com/ipython/ipywidgets/pull/58, where I also simplify widget persistence.

PR #162, opened 2015-06-19
Add links to the REST API documentation (what there is on the wiki)

jasongrout:
I can never seem to find the docs for the REST API.  This adds links to the most likely places to look in the code to find these docs.


jasongrout:
@minrk, @rgbkrk - I think this is ready for review.

rgbkrk:
Guess that works for now. Makes me a bit sad that we don't have a more comprehensive doc, but that's on us. Sounds like Min wants to try out swagger at the very least for JupyterHub.

minrk:
ok with me

PR #164, opened 2015-06-20
Write notebooks with Unix newlines on Windows

techtonik:
This avoids unnecessary notebook linefeed conversions
when working with version control from different systems.

### DIFF
--- a/notebook/services/contents/fileio.py
+++ b/notebook/services/contents/fileio.py
@@ -71,7 +71,7 @@ def atomic_writing(path, text=True, encoding='utf-8', **kwargs):
         shutil.copy2(path, tmp_path)
 
     if text:
-        fileobj = io.open(path, 'w', encoding=encoding, **kwargs)
+        fileobj = io.open(path, 'w', encoding=encoding, newline='\n', **kwargs)
     else:
         fileobj = io.open(path, 'wb', **kwargs)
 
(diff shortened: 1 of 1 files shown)


Carreau:
I can completly see the rational, though there is a test to check that  '\r\n' delimiter is conserved when passed as a kwarg that this breaks.

Do you want to `newline = kwargs.pop('newline','\n')` to make it the default ? I would  also get some input from what others think.

minrk:
I think this makes sense with the change proposed by @Carreau (though I would do it the other way around - `kwargs.setdefault('newline', '\n')`

minrk:
I'm genuinely not sure how this will affect users on Windows, though. I can definitely imagine that there will be those who are adversely affected by the change.

techtonik:
@minrk it will affect Windows users once, but then there should be no problem. The pain point with Windows always was that default editor `notepad.exe` does not understand LF. Luckily, IPython provides much better editing capabilities.

Carreau:
@takluyver how do we test on Jenkins ?

takluyver:
We don't have notebook test jobs set up yet. I should get round to that, but it's not high on my priority list at the moment. You can set it up if you want - clone one of the existing Windows jobs, change the repo and the installation commands.

techtonik:
So, how do we go from here?

minrk:
@techtonik I'll run a couple of tests manually on Windows, then merge this if all goes well, and then we can test in the wild.

minrk:
Seems okay. Thanks.

techtonik:
:+1:

PR #166, opened 2015-06-22
Adding comm_info method to kernel

SylvainCorlay:
This depends on https://github.com/jupyter/jupyter_client/pull/34 and https://github.com/ipython/ipykernel/pull/25.

### DIFF
--- a/notebook/static/services/kernels/kernel.js
+++ b/notebook/static/services/kernels/kernel.js
@@ -634,7 +634,7 @@ define([
         return this.send_shell_message("kernel_info_request", {}, callbacks);
     };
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Rebased. I guess that we would want to make notebook depend on the next release of `jupyter_client` and `ipykernel`.

minrk:
Since the notebook doesn't strictly depend on any kernel, that won't make sense. It also doesn't use the jupyter_client Python API, so that also won't be relevant here. This method is probably going to have to be allowed to be undefined on kernels for notebook 4.x.

PR #168, opened 2015-06-24
Don't install native kernelspec on startup

takluyver:
Part of the fix for jupyter/jupyter_client#35; other part coming up soon


minrk:
I didn't realize we hadn't merged this guy. Can you rebase so we can verify that it's okay?

takluyver:
Rebased, tests passing.

PR #169, opened 2015-06-24
Add FreeBSD instructions

ckane:
I use FreeBSD and the installation is a bit different and unintuitive, but is still very similar to the Ubunty instructions. Based upon what was already here, I have added a FreeBSD install instructions section which hopefully will help some other poor soul in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,3 +28,23 @@ pip install -r requirements.txt -e .
 jupyter notebook
 ```
 
+For FreeBSD:
+```
+cd /usr/ports/www/npm
+sudo make install    # (Be sure to select the "NODE" option)
+cd /usr/ports/devel/py-pip
+sudo make install
+cd /usr/ports/devel/py-virtualenv
+sudo make install
+cd /usr/ports/shells/bash
+sudo make install
+mkdir -p ~/.virtualenvs
+python2.7 -m virtualenv ~/.virtualenvs/notebook
+bash
+source ~/.virtualenvs/notebook/bin/activate
+pip install --upgrade setuptools pip pycurl
+git clone https://github.com/jupyter/notebook.git
+cd notebook
+pip install -r requirements.txt -e .
+jupyter notebook
+```


takluyver:
We're working on proper installation documentation for the whole project at http://jupyter.readthedocs.org/en/latest/install.html . I don't want the README to become a long list of 'installation on platform X' instructions.

ckane:
Sounds good, would you prefer if I make these changes to that instead? Note that I attempted to use the "pip install jupyter" approach from RTD first, and my env couldn't find the project in pip.

takluyver:
Yeah, those instructions are being prepared for when it's released, which it's not yet. ;-)

That particular install page is also supposed to be the short, non-scary version that covers only the most common cases. We probably need some docs somewhere in this repo about getting it installed for development. I guess this may be the relevant page for that: http://jupyter-notebook.readthedocs.org/en/latest/development.html

takluyver:
I think we should also write the instructions to be interpreted a bit rather than copied verbatim. E.g. there's absolutely no requirement to use bash to work on the notebook, except that the later instructions in the same block use it. There's lots of different ways to do the setup which are all equally valid, so the important thing is the explanation of the process, so that the developer understands which bits they can use an alternative for.

ckane:
I believe bash is required to source the "activate" script from the virtualenv. Totally agree that a better description of the install instructions could be made. In my experience starting with one specific path to success and then gradually building optional paths from there ensures that the learned knowledge gets committed somewhere.

takluyver:
In a virtualenv on my system, there are `activate.csh`, `activate.fish` and `activate_this.py`, besides the bash `activate` script. And it's not uncommon to not to activate virtualenvs at all, but instead use full paths to the scripts inside them.

I think there are so many optional paths that you'd never document them all. That's why I think it's better to explain and illustrate the steps, so that developers can work out on their own where to diverge from the instructions.

Carreau:
These are still some improvement beyond no instruction on Free BSD. We can iterate if needed. 

Thanks !

PR #177, opened 2015-06-28
handle _template_path being unicode on py2

minrk:
check string_types instead of str

closes ipython/ipython#8582

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -155,7 +155,7 @@ class NotebookWebApplication(web.Application):
             "template_path",
             ipython_app.template_file_path,
         )
-        if isinstance(_template_path, str):
+        if isinstance(_template_path, py3compat.string_types):
             _template_path = (_template_path,)
         template_path = [os.path.expanduser(path) for path in _template_path]
 


notebook/notebookapp.py
```diff
@@ -155,7 +155,7 @@ def init_settings(self, ipython_app, kernel_manager, contents_manager,
             "template_path",
             ipython_app.template_file_path,
         )
-        if isinstance(_template_path, str):
+        if isinstance(_template_path, py3compat.string_types):
```
tomdz: Why not just check for whether it is already a list ?

Carreau: Probably historical reasons. it does not make much differences. Also I suppose people could use many kind of iterables, but string are iterables, so "isiterable" wouldn't work.

Carreau:
+1

notebook/notebookapp.py
minrk: Also because checking for just a list would be insufficient, as we would also need to check for tuples. In both cases, there are at least two types to check (except on Python 3, where string_types is `(str,)`.

PR #182, opened 2015-06-30
Properly render newlines etc. in raw_input() prompt

frederikhermans:
This PR addresses https://github.com/ipython/ipython/issues/8532.

This is my first PR, do let me know if I screwed up.

### DIFF
--- a/notebook/static/notebook/less/outputarea.less
+++ b/notebook/static/notebook/less/outputarea.less
@@ -158,6 +158,7 @@ div.output_javascript:empty {
 
 div.raw_input_container {
     font-family: @font-family-monospace;
+    line-height: @code_line_height;
     // for some reason, em padding doesn't compute the same for raw_input
     // that is not the first input, but px does
     padding-top: 5px;
--- a/notebook/static/notebook/less/outputarea.less
+++ b/notebook/static/notebook/less/outputarea.less
@@ -164,7 +164,7 @@ div.raw_input_container {
 }
 
 span.raw_input_prompt {
-    /* nothing needed here */
+    white-space: pre;
 }
 
 input.raw_input {
--- a/notebook/static/notebook/less/outputarea.less
+++ b/notebook/static/notebook/less/outputarea.less
@@ -164,8 +164,8 @@ div.raw_input_container {
     padding-top: 5px;
 }
 
-span.raw_input_prompt {
-    white-space: pre;
+pre.raw_input_prompt {
+    /* nothing needed here. */
 }
 
 input.raw_input {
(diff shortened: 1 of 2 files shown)


Carreau:
@frederikhermans congratulation for your first PR! You seem to have done everything correctly. 

The fix seems fine too ! I'll let someone else have a look still.

Thanks !

notebook/static/notebook/less/outputarea.less
```diff
@@ -158,13 +158,14 @@ div.output_javascript:empty {
...
     font-family: @font-family-monospace;
+    line-height: @code_line_height;
     // for some reason, em padding doesn't compute the same for raw_input
     // that is not the first input, but px does
     padding-top: 5px;
 }
 
 span.raw_input_prompt {
-    /* nothing needed here */
+    white-space: pre;
```
takluyver: Should this be pre-wrap, to allow a really long prompt text to wrap? Input prompts in the terminal will wrap when they exceed the terminal width.

https://css-tricks.com/almanac/properties/w/whitespace/

frederikhermans:
@takluyver You're absolutely right, it should be `pre-wrap` instead of just `wrap`. (Just as `white-space` is `pre-wrap` for all `<pre>` tags, as defined in notebook.less).

But even with `pre-wrap`, the prompt's wrapping behavior is slightly different from the rendered result, because the `<span>` breaks words differently than the `<pre>`. So either we add more CSS to the  `<span>` to get `<pre>`-like behavior, or we replace the `<span>` with a `<pre>` for the prompt. The latter solution seems cleaner to me.

I'll look into it and get back in a day or two.

frederikhermans:
Took another shot at it. I replaced `span.input_prompt` with a `pre.input_prompt`.

Here's what it looked like before: http://gfycat.com/WellwornCluelessIbizanhound (Note how the formatting of cell 3 changes when the user hits enter)
Here's what it looks like now: http://gfycat.com/RashIllfatedBluebottlejellyfish (Formatting doesn't change)

I've tried the patch with an ipython kernel running in Chrome 43.0.2357.130, Firefox 38.0.5, Safari 8.0.6, all on OS X 10.10.3.

Note that I also moved the explicit setting of a monospace font on the `div.raw_input_container` to `input.raw_input`. This was necessary because the `div.raw_input_container` was be much wider (in Chrome and Firefox) than the output container with which it is replaced upon user input. The change in the width would cause words to be wrapped differently during and after the prompt. I haven't figured out what the issue actually is, but I think it was due to relative spacing somewhere.

takluyver:
Thanks, I think this is fine, but I'll let someone else double check.

damianavila:
Travis failure seem spurious, can somebody restart it? 
btw... LGTM too.

interrogator:
Woohoo! Thanks guys!

PR #185, opened 2015-07-03
Fix embedding custom.css in HTML exported notebook

spenthil:
`CSSHTMLHeaderPreprocessor._generate_header` uses 
`resources['profile_dir']` to find locate the custom css file.


takluyver:
Thanks for the contribution! I don't think this will work, though - Jupyter doesn't have 'profile' directories, just a config directory.

takluyver:
I've made an updated version of this in #241 which sets `config_dir` instead of `profile_dir`.

PR #187, opened 2015-07-07
updated keyboard.js to throw errors instead of strings.

niole:


Carreau:
Seem legit. I'll wait for the test to pass.

PR #193, opened 2015-07-11
Fix front-end extension loading

jdfreder:
So that one failure doesn't bomb on everything.

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -17,9 +17,13 @@ define([
      */
     var load_extension = function (extension) {
         return new Promise(function(resolve, reject) {
-            require([extension], function(module) {
+            require(["nbextensions/" + extension], function(module) {
                 console.log("Loaded extension: " + extension);
-                resolve(module);
+                try {
+                    module.load_ipython_extension();
+                } finally {
+                    resolve(module);
+                }
             }, function(err) {
                 reject(err);
             });
@@ -37,7 +41,7 @@ define([
         var extensions = [];
         var extension_names = arguments;
         for (var i = 0; i < extension_names.length; i++) {
-            extensions.push("nbextensions/" + arguments[i]);
+            extensions.push(arguments[i]);
         }
 
         return Promise.all(extensions.map(load_extension)).catch(function(err) {
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -9,7 +9,7 @@ define([
     'codemirror/mode/meta',
 ], function($, CodeMirror, moment){
     "use strict";
-
+    
     /**
      * Load a single extension.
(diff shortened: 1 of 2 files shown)


jdfreder:
@takluyver @minrk I just tested this locally and it works.  :D

notebook/static/base/js/utils.js
```diff
@@ -214,7 +221,7 @@ define([
         "46":"ansibgcyan",
         "47":"ansibggray"
     };
-    
+
```
minrk: I see you are giving Atom a try.

minrk:
Is it feasible to add a test with a failing nbextension?

jdfreder:
@minrk last commit adds a test

jdfreder:
....and it fails

takluyver:
Looks like nothing calls `load_ipython_extension` in the new code.

It also seems strange that `load_extensions()` takes paths without `nbextensions/`, but `load_extension()` takes a path with the `nbextension/` prefix already added. They should be consistent, whichever way we prefer.

notebook/static/base/js/utils.js
```diff
@@ -10,34 +10,45 @@ define([
...
     var load_extensions = function () {
         // load one or more Jupyter notebook extensions with requirejs
-        
+
         var extensions = [];
         var extension_names = arguments;
         for (var i = 0; i < extension_names.length; i++) {
-            extensions.push("nbextensions/" + arguments[i]);
+            extensions.push(arguments[i]);
         }
```
takluyver: This loop seems a bit ridiculous now, but IIRC, `arguments` is an array like object rather than a real array, so it may be lacking methods like `.map()`. Is there a better way to get from array-like to array than looping over it?

notebook/static/base/js/utils.js
```diff
@@ -879,4 +890,4 @@ define([
     };
 
     return utils;
-}); 
+});
```
takluyver: Does your editor strip a final newline? Mine always adds them. Might get annoying if we work on the same files...

notebook/static/base/js/utils.js
minrk: I think we don't need the extensions object anymore. It's used below for `extensions.map`, but that can be `Array.prototype.map.apply(extensions)` (signature might be slightly off).

notebook/static/base/js/utils.js
jdfreder: I removed the whitespace package...  This is probably remnant from a hand-revert of whitespace changes.

notebook/static/base/js/utils.js
jdfreder: See latest commit

takluyver:
:+1:

jdfreder:
Would merge but I feel like it's bad luck to merge one's own PR :wink:

PR #196, opened 2015-07-12
Display HTML in pager if supplied

takluyver:
This is still an interim measure until the generic 'put a mimebundle on the page' machinery can easily be used here.

Ping @ccordoba12


Carreau:
+1

ccordoba12:
Thanks @takluyver, this is what I need to create an extension based in `docrepr`

minrk:
:+1:. Travis failure was unrelated, but I'll wait for it to go green, then merge.

PR #197, opened 2015-07-12
Don't uglify js

minrk:
It seems to cause problems, and we care a lot more about request count than minimizing bytes.


JanSchulz:
thank you!

PR #199, opened 2015-07-12
fix url for custom.css

minrk:
served from .jupyter/custom


Carreau:
isn't it what `static_url()` is supposed to do ?

minrk:
No. custom.js,css are not served from /static anymore, they are served from /custom.

damianavila:
LGTM

PR #200, opened 2015-07-13
Use default theme

SylvainCorlay:
The dark theme is not quite finished. We should probably use the light one by default.


minrk:
100%. I didn't realize that it had switched the default.

PR #201, opened 2015-07-13
step-forward button

SylvainCorlay:
A lot of new users expect the play button to run the entire notebook.
The [step-forward](http://fortawesome.github.io/Font-Awesome/icon/step-forward/) button probably corresponds better to what the button actually does.


ssanderson:
:+1:

rgbkrk:
That makes perfect sense to me too.

Issue #184, opened 2015-07-02
Sort (sorting) order of files in notebook list (enhancement/feature request)


temmeand:
It would be helpful to be able to change the sort order of files in the notebook list. I would like it to be alphabetical for all file types and not case sensitive. Directories can appear before files or within the files.

(Hope I got this in the correct place. Let me know if I didn't)
-Extra Info
Currently my list is sorted as follows: - Directories (starting with upper case) - directories (starting with lower case) - IPython notebooks (starting with upper case) - ipython noteooks (starting with lower case) - Other files (starting with upper case) - other files (staring with lower case)

I'm using Jupyter 3.0.0-f75fda4

Carreau:
> I'm using Jupyter 3.0.0-f75fda4

Upgrade to 3.2 which is an important security fix.

And yes we know that a better dashbord with filtering, ordering would be nice.  We just need  someone to write it.

alimanfoo:
Can I also add a feature request to be able to sort folder contents in reverse order. I name all my notebooks starting with a datestamp (e.g., 20151202_some_work.ipynb) and it would be most useful to have these sorted in reverse order because I usually reopen the most recent notebooks (e.g., to continue working on them). Currently I have to scroll all the way to the bottom to find my most recent notebooks.

gnestor:
Similar https://github.com/jupyter/notebook/issues/1267

DSLituiev:
sorting by modification date will be highly appreciated

gnestor:
This was implemented 1/16/16 by https://github.com/jupyter/notebook/pull/943 but it's marked for a 5.0 release, which may not happen for a few months. If you install notebook from master by following [these directions](https://github.com/jupyter/notebook/blob/master/CONTRIBUTING.rst#setting-up-a-development-environment), you can take advantage of the feature now.

takluyver:
Closing as this got implemented in #943.

michaelaye:
This feature is not active in master?

gnestor:
I see it in master:

![screencap](http://g.recordit.co/uZaOqraygZ.gif)

michaelaye:
so sorry. had a clashing conda package that sneaked in and overruled the GH install.

Carreau:
> so sorry. had a clashing conda package that sneaked in and overruled the GH install.

See https://github.com/fperez/conda-edev , relatively recent. Should avoid these kind of issues by providing pip-install as a conda "overlay".

[...]

PR #203, opened 2015-07-13
sort order of all files and directories is now case-insensitive

niole:
Modified notebooklist.js to make sort order of files and directories case insensitive.


feynmanliang:
Addresses some features requested in #184

niole:
What made you think to do that?

PR #216, opened 2015-07-21
markdown output should get .rendered_html class

nitind:
Simply adds the class to top-level and descendent divs during the render().

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -595,7 +595,7 @@ define([
 
     var append_markdown = function(markdown, md, element) {
         var type = 'text/markdown';
-        var toinsert = this.create_output_subarea(md, "output_markdown", type);
+        var toinsert = this.create_output_subarea(md, "output_markdown rendered_html", type);
         var text_and_math = mathjaxutils.remove_math(markdown);
         var text = text_and_math[0];
         var math = text_and_math[1];
@@ -614,7 +614,7 @@ define([
          * We just eval the JS code, element appears in the local scope.
          */
         var type = 'application/javascript';
-        var toinsert = this.create_output_subarea(md, "output_javascript", type);
+        var toinsert = this.create_output_subarea(md, "output_javascript rendered_html", type);
         this.keyboard_manager.register_events(toinsert);
         element.append(toinsert);
 


minrk:
Why is this done twice? Shouldn't it be done on the outer element?

nitind:
Good point. It was so I could handle divs in the original html jquery object and then their children, but it can be done in one step if done after the rendered version has been added to the DOM. Update pushed.

notebook/static/notebook/js/textcell.js
```diff
@@ -291,6 +291,7 @@ define([
                 // links in markdown cells should open in new tabs
                 html.find("a[href]").not('[href^="#"]').attr("target", "_blank");
                 that.set_rendered(html);
+                that.element.find('div.text_cell_render').find('div').addClass('html_rendered');
```
minrk: I still think this isn't the right place to do it, and the class is rendered_html, not html_rendered.  Look at where the html output adds the class (append_html), and there is a similar place in the markdown output (append_markdown) where it can be added, without making any jquery calls.

nitind: I appreciate the patience. My thinking was that since output areas are 
apparently just repurposed rather than destroyed and recreated, it was 
better to respond to whatever changes the user may have made and declare 
the divs as rendered_html when rendering them.

---
Regards,
Nitin Dahyabhai

takluyver:
@minrk should this be targeted for 4.0? Just going through open PRs.

minrk:
@nitind Now that I've looked again at the code, I see that the rendered_html class is already added at the top-level on the text_cell_render element. That suggests there isn't anywhere to add it. Can you describe the problem you are trying to fix?

nitind:
@minrk As the original issue reported, when cell source is MD, the HTML rendering of its divs lacks the rendered_html class. I looked at the areas you mentioned earlier, but they look like they're related to cell execution that generates MD rather than MD source the user authored.

Perhaps I misunderstood @Carreau ?

minrk:
@nitind but their parent doesn't lack it, so the .rendered_html css applies to the entire rendered cell contents. Do you see something rendering incorrectly?

Carreau:
On 3.x it does... let me try on master. 
screen shot 2015-07-23 at 11 01 13

minrk:
Then the issue is what I described earlier - there is no problem on markdown cells, only markdown output. @nitind this should be added to the `append_markdown` method, which should do it in the same way that append_html already does.

nitind:
Does append_javascript need the same change?

minrk:
Perhaps so, just to be safe.

takluyver:
@nitind , did you want to make this change to `append_javascript` as well?

nitind:
@takluyver sorry for the delay. I don't believe the CI failure is related to the change.

takluyver:
Thanks! Looks like someone kicked Travis to restart already. @minrk , anything else on this before we merge it?

minrk:
I was just waiting for Travis to be green. Merging.

PR #218, opened 2015-07-21
Env variables for server root & URL in terminals

takluyver:
This came up [on reddit](https://www.reddit.com/r/IPython/comments/3dz9r4/is_there_a_command_from_jupyters_terminal_to/): if you want to open a notebook or a file in the text editor from the terminal inside the notebook web interface, you can construct a URL to open. But to do that accurately, you need to know the root directory where the server is looking, and the first part of the URL. This exposes both of those as environment variables: `$JUPYTER_SERVER_ROOT` and `$JUPYTER_SERVER_URL`.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -889,7 +889,7 @@ class NotebookApp(JupyterApp):
     def init_terminals(self):
         try:
             from .terminal import initialize
-            initialize(self.web_app)
+            initialize(self.web_app, self.notebook_dir, self.connection_url)
             self.web_app.settings['terminals_available'] = True
         except ImportError as e:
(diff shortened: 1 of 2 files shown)


minrk:
Does it make sense to do this for kernels as well?

takluyver:
Possibly, though I'd be worried that kernel code would try to use it for 'am I in a notebook' detection. I know we're still thinking about giving kernel code more access to some kind of frontend information, but until we've worked out something coherent, I think it's preferable not to provide this kind of information.

It can probably be misused in the terminal too - e.g. the 'open a new browser window' trick only works if you're on localhost. But without HTML output, I think there's much less opportunity for mischief.

minrk:
ok

PR #219, opened 2015-07-21
remove 'name' arg from file_exists docstring

minrk:
Only path is used, not separate name, path.

closes ipython/ipython#8646


ssanderson:
:+1:

PR #221, opened 2015-07-21
remove gulp

minrk:
Follow-suit from ipywidgets and build js/css with Python/js scripts instead:

e.g.

- `npm run build`
- `setup.py js css`


jdfreder:
woot

rgbkrk:
Eeek, I'm so happy!

PR #222, opened 2015-07-22
Ask for confirmation if kernel is busy

flying-sheep:
fixes ipython/ipython#8546

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -114,6 +114,7 @@ define(function (require) {
         this.next_prompt_number = 1;
         this.session = null;
         this.kernel = null;
+        this.kernel_busy = false;
         this.clipboard = null;
         this.undelete_backup = null;
         this.undelete_index = null;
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -249,6 +250,14 @@ define(function (require) {
             var cm_mode = langinfo.codemirror_mode || langinfo.name || 'null';
             that.set_codemirror_mode(cm_mode);
         });
+        
+        this.events.on('kernel_idle.Kernel', function () {
+            that.kernel_busy = false;
+        });
+        
+        this.events.on('kernel_busy.Kernel', function () {
+            that.kernel_busy = false;
```
takluyver: true, right?

notebook/static/notebook/js/notebook.js
```diff
@@ -283,6 +292,10 @@ define(function (require) {
             if (kill_kernel) {
                 that.session.delete();
             }
+            // if the kernel is busy, prompt the user if he’s sure
+            if (that.kernel_busy) {
+                return "The Kernel is busy, changes may be lost.";
+            }
```
takluyver: I think this should be after the unsaved changes check - if there are unsaved changes and the kernel is busy, it's probably the unsaved changes that are more important.

I'd also tweak the wording of the message a bit: "... *outputs* may be lost".

notebook/static/notebook/js/notebook.js
flying-sheep: lol of course :tired_face:

notebook/static/notebook/js/notebook.js
flying-sheep: yeah… although firefox doesn’t recognize the message anyway.

takluyver:
This looks fine to me. @minrk or @Carreau , can one of you take a look and merge if you're happy with it.

Carreau:
I'm +1. @ellisonbg Might be picky on the phrasing.

PR #232, opened 2015-07-25
Modify the check_call invocation to fix break on Windows

justacec:
When I attempted to install the development version of notebook on my Windows 7 machine, I received the following error:

    >> python setup.py build

    running build
    running build_py
    running jsversion
    running css
    running jsdeps
    installing build dependencies with npm
    npm install
    rebuilding js and css failed (not a problem)
    [WinError 2] The system cannot find the file specified
    checking package data
      File "setup.py", line 190, in <module>
        main()
      File "setup.py", line 187, in main
        setup(**setup_args)
      File "C:\Anaconda3\lib\distutils\core.py", line 148, in setup
        dist.run_commands()
      File "C:\Anaconda3\lib\distutils\dist.py", line 955, in run_commands
        self.run_command(cmd)
      File "C:\Anaconda3\lib\distutils\dist.py", line 974, in run_command
        cmd_obj.run()
      File "C:\Anaconda3\lib\distutils\command\build.py", line 126, in run
        self.run_command(cmd_name)
      File "C:\Anaconda3\lib\distutils\cmd.py", line 313, in run_command
        self.distribution.run_command(command)
      File "C:\Anaconda3\lib\distutils\dist.py", line 974, in run_command
        cmd_obj.run()
      File "C:\Users\jclutter\Source\Python\IPython\t\notebook\setupbase.py", line 508, in run
        command.run(self)
      File "C:\Users\jclutter\Source\Python\IPython\t\notebook\setupbase.py", line 199, in run
        check_package_data(self.package_data)
      File "C:\Users\jclutter\Source\Python\IPython\t\notebook\setupbase.py", line 189, in check_package_data
        assert os.path.exists(path), "Missing package data: %s" % path
    AssertionError: Missing package data: notebook\static\auth\js\main.min.js

Although the bottom error highlights a missing `main.min.js`, this is a red herring.  The real issue is at the top, `[WinError 2] The system cannot find the specified`.

After some investigation, I found that the `bower` program was not being execute.  This led to an investigation of the run command.  I was unable to execute commands from the python shell unless I added the `shell = True` parameter to the `check_call` invocation.  I made this change here and was able to perform the installation of Jupyter/notebook.  

Documentation for the `subprocess` module implies that there could be security issues when using `shell = True`.

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -293,7 +293,7 @@ def mtime(path):
 def run(cmd, *args, **kwargs):
     """Echo a command before running it"""
     log.info(" ".join(cmd))
-    return check_call(cmd, *args, **kwargs)
+    return check_call(cmd, shell = True, *args, **kwargs)
 
 
 class Bower(Command):


justacec:
I see that the Travis CI builds failed.  Looking at the logs, it seems as though this fix may have broken something else.  When I make this change on my system (Windows 7, Python 3.4.1, Anaconda 2.3.0, npm 2.11.3, node 0.12.7) it works.

takluyver:
I don't think `shell=True` is the right fix here.

@minrk , when installing from source, if building the css & js fails, it logs a warning message that says 'not a problem', and allows it to continue. But if the CSS & JS aren't already built, it is a problem, and hiding the error makes it less clear what has actually gone wrong.

Of course, the trouble is that it's the same code for installing from source as for installing from an sdist, where it ideally shouldn't even try to build the CSS & JS. But I wonder if we could do something smarter so the real error is clearer in cases like this.

justacec:
Could be.  I am still not very good with Python and am learning.  I have noticed other strange issues with the install.  For instance, the installed scripts to launch Jupyter are all bash style scripts and do not support windows.  I am not sure how to fix that right now.  I can start jupyter-notebook but then after that I think there will be issues when and if the system tries to execute the nconvert module for export...  I am still playing around here.

takluyver:
You may find it easier to install the stable version of IPython while we're finishing up the release of Jupyter. http://ipython.org/install.html

justacec:
Never!  I shall never give in.   :)  I actually got things running.

takluyver:
Kudos!

minrk:
I think the not a problem is inherited from the time when we tracked these things, they just might not be up to date. It definitely should refuse to proceed if the files don't exist. We should also not even be trying to run these commands from an sdist.

Issue #239, opened 2015-07-28
could not download the complete file


tangchang3531:
Bug：
1.Open Python 3 on Jupyter

2.Coding (take this picture as example)
<img src="https://cloud.githubusercontent.com/assets/12648107/8933052/d61e7638-350e-11e5-8a9a-bd34aaaa766a.png" width="600"/>

3.Download as xxx.ipynb

4.upload to Github, then we could find the [results showed in Github is not complete.](https://github.com/tangchang3531/tangchang3531.github.io/blob/master/ipynb/problem.ipynb)
<img src="https://cloud.githubusercontent.com/assets/12648107/8933204/85a908a2-350f-11e5-92f3-599c6daa36fb.png"600"/>

Expected results：I could download complete xxx.ipynb, which will not miss last several lines.

Supplementary description： I can download complete document after several minute when I download again. I just want to know why I could not download complete document right after I finished coding.

Version: [Online version of Jupyter!](https://try.jupyter.org/)

takluyver:
That sounds like it's not saving before it offers you the document for download. We try to save, but we have to do it with a synchronous HTTP request, because downloading uses `window.open()`, which we can only call while handling a click event. You can work around it by explicitly saving (Ctrl+S, or click the save button in the toolbar) just before you download.

I see the same issue on try.jupyter.org using Firefox.

We may be able to get round this using a trick we have elsewhere: create a window while handling a click, and then set its location in a separate callback.

tangchang3531:
Exactly！Thank you so much. I thought it could be automatically saved. But I see the problem now.

PR #240, opened 2015-07-28
Use window.location trick for download as ipynb

takluyver:
To avoid problems with synchronous HTTP request from JS to save the notebook before downloading it.

Closes gh-239


takluyver:
@minrk this is the only thing currently open and tagged for 4.0, can you take a look? It should be straightforward.

PR #243, opened 2015-07-30
Make dict comprehension on one line

jasongrout:
I had to stare at this for a long time, while my brain was parsing the statement one line at a time, trying to figure out what in the world it was doing.  I then realized it was a dict comprehension.  Personally, I think this comprehension makes a lot more sense when scanning through the code when it is on one line.

I realize this is a stylistic issue, so if I'm outvoted, fine.  This comprehension took me enough time to understand (or even syntactically recognize as a dict comprehension) that I thought it was worth a PR.

### DIFF
--- a/notebook/services/contents/handlers.py
+++ b/notebook/services/contents/handlers.py
@@ -66,11 +66,7 @@ def validate_model(model, expect_content):
                 u"Keys unexpectedly None: {keys}".format(keys=errors),
             )
     else:
-        errors = {
-            key: model[key]
-            for key in maybe_none_keys
-            if model[key] is not None
-        }
+        errors = {key: model[key] for key in maybe_none_keys if model[key] is not None}
         if errors:
             raise web.HTTPError(
                 500,


ssanderson:
As the person who wrote this expression, I'm -1 on this change.

I prefer multi-line comprehensions over one giant line because I think it clearly differentiates the three operations that are happening in the comprehension:
```
{
     <expression>
     <iteration>
     <filter>
}
```
Maybe I'm just used to reading my own style, but I find this separation much clearer than having to parse the logical sections in a long line by hand.

minrk:
I don't have a strong feeling. I don't personally find a one-line comprehension easier to follow than a multi-line one, if anything it's slightly the other way around. I'm not especially fond of filtered comprehensions in general.

ssanderson:
The multiple bindings of the name `key` in the expressions here make this unusually hard to parse I think.  A totally sane alternative is also to not use a comprehension at all:
```
errors = {}
for key in maybe_none_keys:
    if model[key] is not None:
        errors[key] = model[key]
```
This feels verbose to me, but it's almost certainly the most straightforward spelling of this.

minrk:
I'd be AOK with removing the comprehension if that's helpful to people. I don't think reshaping the comprehension improves it, though.

jasongrout:
I think why I struggled to understand it were:

1. not used to dict comprehensions
2. having for and if on separate lines helped me parse them as for and if statements, which threw my brain for a loop when it just didn't make sense
3. The list comprehension a few lines above is almost the same thing, but on a single line

In general, I think comprehensions are more efficient than explicit loops, so I prefer them, so I'll close this as a style issue.

jasongrout:
P.S. It wasn't just me.  I showed it to @SylvainCorlay as well, and he stared at it blankly for a few seconds before I explained what it was.

minrk:
@jasongrout if you wanted to keep this PR and turn it into the for-loop, that would be fine. I don't think this is a case where any efficiency benefits of a comprehension will be noticeable, since it's also talking to a filesystem or external data store.

PR #247, opened 2015-07-30
Add Jinja Blocks to notebook.html and tree.html

crwilcox:
Added blocks to support customization to:
- the logo in the upper right.  Both tmpnb and notebooks.azureml.net use
this area to add logos.
- help links.  We add links to the help tab.  We also add a
rel="noreferrer" to each link.  These block enable us to make these
corrections without altering the primary template.
- provide a place in tree to give an attribute area.  We use this to
hide the clusters tab.

### DIFF
--- a/notebook/templates/notebook.html
+++ b/notebook/templates/notebook.html
@@ -267,7 +267,7 @@ data-notebook-path="{{notebook_path}}"
 
                         {% for helplinks in sections %}
                             {% for link in helplinks %}
-                                <li><a class='help_link' href="{{link[0]}}" {{'target="_blank" title="Opens in a new window"' if link[2]}}>
+                                <li><a href="{{link[0]}}" {{'target="_blank" title="Opens in a new window"' if link[2]}}>
                                 {{'<i class="fa fa-external-link menu-icon pull-right"></i>' if link[2]}}
                                 {{link[1]}}
                                 </a></li>
--- a/notebook/templates/notebook.html
+++ b/notebook/templates/notebook.html
@@ -267,7 +267,7 @@ data-notebook-path="{{notebook_path}}"
 
                         {% for helplinks in sections %}
                             {% for link in helplinks %}
-                                <li><a {% block helplink_attributes %}{% endblock %} href="{{link[0]}}" {{'target="_blank" title="Opens in a new window"' if link[2]}}>
+                                <li><a class='help_link' href="{{link[0]}}" {{'target="_blank" title="Opens in a new window"' if link[2]}}>
                                 {{'<i class="fa fa-external-link menu-icon pull-right"></i>' if link[2]}}
                                 {{link[1]}}
                                 </a></li>
--- a/notebook/templates/notebook.html
+++ b/notebook/templates/notebook.html
@@ -45,7 +45,9 @@ data-notebook-path="{{notebook_path}}"
 </span>
 
 <span id="kernel_logo_widget">
+  {% block kernel_logo_widget %}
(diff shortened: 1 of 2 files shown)


crwilcox:
We use these so we can override certain behavior.
Example:
Tree.html
```python
{% extends "templates/tree.html" %}
{% block clusters_tab_attributes %}
style="visibility:collapse"
{% endblock %}
```

and notebook.html
```python
{% extends "templates/notebook.html" %}

{% block kernel_logo_widget %}
{{ super() }}
<a class="pull-right" style="vertical-align:middle;" href="https://studio.azureml.net/" target="_blank">
    <img class="azure_ml_logo" src="/content/ipython/base/images/powered_by_azure_ml.png"/>
</a>
{% endblock %}

{% block helplink_attributes %}
rel="noreferrer"
{% endblock %}

{% block help %}
{{ super() }}
<li title="About Azure ML's Privacy and Cookies Policy"><a id="notebook_privacy" href="https://go.microsoft.com/fwlink/?LinkID=131004" target="_blank">Privacy + Cookies</a></li>
{% endblock %}
```

We also pull this trick in page, but it was already there.
```python
{% extends "templates/page.html" %}
{% block login_widget %}
{{ super() }}
<a class="pull-right" style="vertical-align:middle;" href="https://studio.azureml.net/" target="_blank">
    <img class="azure_ml_logo" src="/content/ipython/base/images/powered_by_azure_ml.png"/>
</a>
{% endblock %}
```

notebook/templates/tree.html
```diff
@@ -19,7 +19,7 @@
       <ul id="tabs" class="nav nav-tabs">
         <li class="active"><a href="#notebooks" data-toggle="tab">Files</a></li>
         <li><a href="#running" data-toggle="tab">Running</a></li>
-        <li><a href="#clusters" data-toggle="tab">Clusters</a></li>
+        <li><a href="#clusters" data-toggle="tab" {% block clusters_tab_attributes %}{% endblock %} >Clusters</a></li>
```
minrk: It doesn't make much sense to do this on clusters, but not the other tabs. Perhaps there should be a `tabs` block?

crwilcox: Do you mean to duplicate this ability on the others or to make the same attributes appear on all three?

minrk: Perhaps rather than adding a clusters_tab_attributes, adding a class on each tab (`class=clusters_tab_link`, etc.). is the way to go, then you can hide it with CSS.

notebook/templates/notebook.html
```diff
@@ -264,7 +267,7 @@
 
                         {% for helplinks in sections %}
                             {% for link in helplinks %}
-                                <li><a href="{{link[0]}}" {{'target="_blank" title="Opens in a new window"' if link[2]}}>
+                                <li><a {% block helplink_attributes %}{% endblock %} href="{{link[0]}}" {{'target="_blank" title="Opens in a new window"' if link[2]}}>
```
minrk: like below, perhaps `class=help_link` so that it can be treated with CSS rather than a block.

crwilcox:
Thanks for the feedback @minrk, I think using custom.css for those attributes makes good sense. I have made the changes to remove the two blocks and replace them css classes that can be used.

crwilcox:
I just had a chat with @DinoV and he is starting a thread on the security group about the noreferrer link.  I am just removing that change from this so it can be handled separately.  I thought we would want to customize ours to add 'rel="noreferrer"' but it seems that should probably be there in Jupyter

PR #258, opened 2015-08-04
increase completer z-index to be in front of pager

ebenolson:
This lets tab completion still be readable when an introspection pane is open.


takluyver:
That sounds reasonable, but I'd like someone more familiar with the frontend stuff to look at it.

PR #260, opened 2015-08-05
Command Palette on notebook. 

Carreau:
Mainly openig PR to start discussion. 

I would like to start working on a **command palette** (cf atom screenshot) for the notebook. Something aka Mac OS Spotlight. It's called **command palette** in atom, so let's use this name, we can bikeshed on the name later. 

This would allow to expose a much bigger number of action to end-user without binding to a keyboard shortcut, and made them easy to find.

From a prototype I did at SciPy, I think it can be done correctly in about a week. 
And by correctly I mean with respect to the current API we have for the notebook. We all know that current API is not perfect, but we have enough hook to make something useful in a short amount of time.

I think it would be nice small project with high visibility and rapid user feedback that might help a lot of user and push for 4.1 adoption. Also one of the things I enjoyed when joining the project was to get small things that help people a lot. We might also want some design work to be done on the UI. 
So I think it would be a nice project for @svurens, @jdfreder and me to collaborate on. 

Technically I am torn in between twitter [typeahead](https://github.com/twitter/typeahead.js) and jQuery [typeahead](http://www.runningcoder.org/jquerytypeahead/), twitter one have stalled since April, so I went for second one. So I have no issue if we change. 

Thought ? Would @svurens like to work on that ?

--- 

screen shot 2015-08-05 at 14 23 15

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -45,7 +45,6 @@ def _post_save_script(model, os_path, contents_manager, **kwargs):
     log = contents_manager.log
 
     base, ext = os.path.splitext(os_path)
-    py_fname = base + '.py'
     script, resources = _script_exporter.from_filename(os_path)
     script_fname = base + resources.get('output_extension', '.txt')
     log.info("Saving script /%s", to_api_path(script_fname, contents_manager.root_dir))
--- a/notebook/static/base/js/keyboard.js
+++ b/notebook/static/base/js/keyboard.js
@@ -18,7 +18,7 @@ define([
 
     /**
      * Setup global keycodes and inverse keycodes.
-     * 
+     *
      * See http://unixpapa.com/js/key.html for a complete description. The short of
      * it is that there are different keycode sets. Firefox uses the "Mozilla keycodes"
      * and Webkit/IE use the "IE keycodes". These keycode sets are mostly the same
@@ -190,7 +190,7 @@ define([
         /**
          * Clear the pending shortcut soon, and cancel previous clearing
          * that might be registered.
-         **/ 
+         **/
          var that = this;
          clearTimeout(this._cleartimeout);
          this._cleartimeout = setTimeout(function(){that.clearqueue();}, this.delay);
@@ -225,6 +225,17 @@ define([
         }
         return dct;
     };
+    
+    ShortcutManager.prototype.get_shortcut_for_action_name = function(name){
+      var ftree = flatten_shorttree(this._shortcuts);
+      var res = {};
+      for (var sht in ftree ){
+        if(ftree[sht] === name){
+          return sht
+        }
+      }
+      return undefined;
+    }
 
     ShortcutManager.prototype.help = function () {
         var help = [];
--- a/notebook/static/base/js/keyboard.js
+++ b/notebook/static/base/js/keyboard.js
@@ -226,16 +226,16 @@ define([
         return dct;
     };
     
-    ShortcutManager.prototype.get_shortcut_for_action_name = function(name){
+    ShortcutManager.prototype.get_action_shortcut = function(name){
       var ftree = flatten_shorttree(this._shortcuts);
       var res = {};
       for (var sht in ftree ){
         if(ftree[sht] === name){
-          return sht
+          return sht;
         }
       }
       return undefined;
-    }
+    };
 
     ShortcutManager.prototype.help = function () {
         var help = [];
--- a/notebook/static/base/js/keyboard.js
+++ b/notebook/static/base/js/keyboard.js
@@ -48,12 +48,12 @@ define([
     
     // These apply to Firefox and Opera
     var _mozilla_keycodes = {
-        '; :': 59, '= +': 61, '- _': 173, 'meta': 224
+        '; :': 59, '= +': 61, '- _': 173, 'meta': 224, 'minus':173
     };
     
     // This apply to Webkit and IE
     var _ie_keycodes = {
-        '; :': 186, '= +': 187, '- _': 189
+        '; :': 186, '= +': 187, '- _': 189, 'minus':189
     };
     
     var browser = utils.browser[0];
@@ -105,7 +105,7 @@ define([
         }
 
         shortcut = shortcut.toLowerCase().replace('cmd', 'meta');
-        shortcut = shortcut.replace(/-$/, '_');  // catch shortcuts using '-' key
+        shortcut = shortcut.replace(/-$/, 'minus');  // catch shortcuts using '-' key
         shortcut = shortcut.replace(/,$/, 'comma');  // catch shortcuts using '-' key
         if(shortcut.indexOf(',') !== -1){
             var sht = shortcut.split(',');
@@ -130,7 +130,7 @@ define([
          **/
         type = type || 'keydown';
         shortcut = normalize_shortcut(shortcut);
-        shortcut = shortcut.replace(/-$/, '_');  // catch shortcuts using '-' key
+        shortcut = shortcut.replace(/-$/, 'minus');  // catch shortcuts using '-' key
         var values = shortcut.split("-");
         var modifiers = values.slice(0,-1);
         var key = values[values.length-1];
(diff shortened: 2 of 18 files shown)


Carreau:
Ok, I have a rough prototype that show something. 


screen shot 2015-08-05 at 15 28 33

ellisonbg:
@Carreau thanks for posting this - I am *really* excited about this. We have needed this for a long time and I think it is a great idea of have @svurens start to work on this with you and @jdfreder.

ellisonbg:
Would type ahead help the UI side of this:

https://twitter.github.io/typeahead.js/

jdfreder:
@svurens and I are fixing some problems with his Canopy+Anaconda+standalone IPython install...

Carreau:
> Technically I am torn in between twitter typeahead and jQuery typeahead, twitter one have stalled since April, so I went for second one. So I have no issue if we change.

> Would type ahead help the UI side of this:

@ellisonbg : as I wrote in initial description Twitter-typeahead have 0 commits since april, no responses to PRs and issues. So I'm slightly negative on it. 

Also the API seem to change dramatically in between patch version so i"m not convinced we should use it. So I'm investigating jQuery typeahead, which seem : 

  - better documented 
  - have more real-world example. 
  - maintained.

jdfreder:
@Carreau , @svurens has his development install working, and he is going to open a PR against your branch for discussion : )

Carreau:
@jdfreder, @svurens cleaned up a bit, appreciate review.

notebook/services/contents/filemanager.py
```diff
@@ -45,7 +45,6 @@ def _post_save_script(model, os_path, contents_manager, **kwargs):
     log = contents_manager.log
 
     base, ext = os.path.splitext(os_path)
-    py_fname = base + '.py'
```
Carreau: This was unused, so I removed it.

notebook/static/notebook/js/quickhelp.js
```diff
@@ -168,10 +184,8 @@ define([
         var element = $('<div/>');
 
         // The documentation
-        var doc = $('<div/>').addClass('alert alert-warning');
+        var doc = $('<div/>').addClass('alert alert-info');
         doc.append(
-            $('<button/>').addClass('close').attr('data-dismiss','alert').html('&times;')
-        ).append(
```
Carreau: This made no sens as there are no persistence of the hide this thing. So removed as well.

notebook/static/notebook/js/quickhelp.js
```diff
@@ -235,41 +249,21 @@ define([
 
     QuickHelp.prototype.build_command_help = function () {
         var command_shortcuts = this.keyboard_manager.command_shortcuts.help();
-        return build_div('<h4>Command Mode (press <code>Esc</code> to enable)</h4>', command_shortcuts);
-    };
-
-    var special_case = { pageup: "PageUp", pagedown: "Page Down", 'minus': '-' };
```
Carreau: this is now baked into `humanize_sequence`.  The logic was convoluted and wrong.

notebook/static/notebook/js/notebook.js
```diff
@@ -170,7 +171,9 @@ define(function (require) {
         // i) provide a margin between the last cell and the end of the notebook
         // ii) to prevent the div from scrolling up when the last cell is being
         // edited, but is too low on the page, which browsers will do automatically.
-        var end_space = $('<div/>').addClass('end_space');
+        var end_space = $('<div/>')
+            .addClass('end_space')
+            .html('<kbd>Cmd-Shift-P</kbd> for Command palette');
```
Carreau: What do you think of having a Tips thing here ? (I should probably remove before merge)

notebook/services/contents/filemanager.py
jdfreder: :+1:

notebook/static/notebook/js/actions.js
```diff
@@ -329,10 +329,19 @@ define(function(require){
             handler : function (env) {
                 env.pager.collapse();
             }
+        },
+        'command-palette': {
+            help_index : 'aa',
```
jdfreder: index aa?

notebook/static/notebook/js/commandpalette.js
```diff
@@ -0,0 +1,168 @@
...
+      shortcut = notebook.keyboard_manager.edit_shortcuts.get_shortcut_for_action_name(name);
+      if (shortcut) {
+        return 'edit-shortcut';
+      }
+      return 'no-shortcut';
+    };
+
+    var CommandPalette = function(notebook) {
+        if(!notebook){
+          throw new Error("CommandPalette take a notebook non-null mandatory arguement");
```
jdfreder: s/take/takes/

notebook/static/notebook/js/commandpalette.js
```diff
@@ -0,0 +1,168 @@
...
+          }
+          before_close.ok = true; // avoid double call.
+        };
+
+        var mod = $('<div/>').addClass('modal').append(
+          $('<div/>').addClass('modal-dialog')
+          .append(
+            $('<div/>').addClass('modal-content').append(
+              $('<div/>').addClass('modal-body')
+              .css('padding', '7px')
```
jdfreder: You use .css in a few places, you should give the outer most div a `command-palette` class and style through less.  That will make it easier for users to add custom styling too.

notebook/static/base/js/keyboard.js
```diff
@@ -225,6 +225,17 @@ define([
         }
         return dct;
     };
+    
+    ShortcutManager.prototype.get_shortcut_for_action_name = function(name){
```
jdfreder: Can this be `getShortcut` or `get` ?  So then when it's called it looks like this `shortcutManager.getShortcut('bob')` or `shortcutManager.get('bob')` .

jdfreder: I guess the in this file we are using underscore separated words, so `get_shortcut` ... :/

notebook/static/base/less/variables.less
```diff
@@ -15,6 +15,9 @@
 @border-radius-base: 2px;
 @border-radius-large: 3px;
 @grid-gutter-width: 0px;
+@kbd-color: #888;
+@kbd-bg: transparent;
```
jdfreder: I'm guessing `kbd` stands for keyboard dialog.  Why not `@commandpalette-*` ?

notebook/static/base/js/keyboard.js
jdfreder: I prefer `get`

jdfreder:
I left a couple of in line comments, but I'd also like to leave a general comment: you don't have many comments (no pun intended) in your code, you may want to add some more so it's easier to digest.

Other than that, it's looking good!  @svurens can you review this too?

notebook/static/notebook/js/actions.js
Carreau: sure, I think it's important and should be near the top. What woudl you use ?

notebook/static/base/less/variables.less
Carreau: no, that'a a bootstrap variable.

notebook/static/base/js/keyboard.js
Carreau: the problem with get, is that you could get many things, action, action name, shortcut, it was get and i keept getting confuse for what I was getting for what.

Carreau:
Added comment. even if I'm not sure it make things clearer, there is not a lot to explain.

notebook/static/base/js/keyboard.js
svurens: I think get_shortcut or get_action_shortcut is about the shortest you can get without making this function's name too general.

Carreau: fine with me, renamed !

Carreau:
Any extra-comments ? @jdfreder I suppose you woudl prefer your commonJS PR to be merged first ?

jdfreder:
@Carreau no, no more comments.  It's okay, I can deal with the rebase on my end : )  I think I came along after you..

jdfreder:
Thanks!  People will love this!

damianavila:
Thanks! People will love this!

>I don't see any docs reference here... this is something to add later in another PR?

PR #266, opened 2015-08-10
Multi-cell selection

takluyver:
From discussion the other day: we are thinking of accepting some new features into 4.x releases, so that we can keep the user experience moving forwards while people are working on refactoring all the frontend stuff.

Multi-cell selection is perhaps the biggest thing that we've been wanting to add for a long time, so I thought I'd take a crack at it. It turned out not to be too bad (so far...).

Use: shift+j/k to select a range of cells, shift+m to merge the selected range. I haven't yet changed all the other actions which should now be able to affect a range of cells, but doing so should be mostly straightforward.

API: I've tried to keep existing API working, so stuff that assumes the selection is a single cell will act on the cell at the cursor - the most recently selected one. This makes some names a bit confusing - e.g. `Cell.unselect()` now has a `leave_selected` parameter (to move the cursor away without actually unselecting). The notebook has some very similarly named methods, like `get_selected_cell()` and `get_selected_cells()`.

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -725,6 +725,18 @@ define(function (require) {
         return true;
     };
 
+    /**
+     * Clear selection of multiple cells (except the cell at the cursor)
+     */
+    Notebook.prototype.reset_selection = function() {
+        var current_selection = this.get_selected_cells();
+        for (var i=0; i<current_selection.length; i++) {
+            if (!current_selection[i].selected) {
+                current_selection[i].unselect()
+            }
+        }
+    };
+
 
     // Edit/Command mode
 
@@ -778,6 +790,7 @@ define(function (require) {
         if (cell && this.mode !== 'edit') {
             cell.edit_mode();
             this.mode = 'edit';
+            this.reset_selection();
             this.events.trigger('edit_mode.Notebook');
             this.keyboard_manager.edit_mode();
         }
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -600,6 +600,17 @@ define(function (require) {
         return result;
     };
 
+    /**
+     * Get an array of the cells in the currently selected range
+     *
+     * @return {Array} The selected cells
+     */
+    Notebook.prototype.get_selected_cells = function () {
+        return this.get_cells().filter(function(cell) {
+            return cell.in_selection;
+        });
+    };
+
 
     // Cell selection.
 
@@ -607,9 +618,10 @@ define(function (require) {
      * Programmatically select a cell.
      * 
      * @param {integer} index - A cell's index
+     * @param {bool} add_to_selection - whether to add to the current selection
      * @return {Notebook} This notebook
      */
-    Notebook.prototype.select = function (index) {
+    Notebook.prototype.select = function (index, add_to_selection) {
         if (this.is_valid_cell_index(index)) {
             var sindex = this.get_selected_index();
             if (sindex !== null && index !== sindex) {
@@ -618,8 +630,12 @@ define(function (require) {
                 if (this.mode !== 'command') {
                     this.command_mode();
                 }
-                this.get_cell(sindex).unselect();
             }
+            var current_selection = this.get_selected_cells();
+            for (var i=0; i<current_selection.length; i++) {
+                current_selection[i].unselect(add_to_selection)
+            }
+
             var cell = this.get_cell(index);
             cell.select();
             if (cell.cell_type === 'heading') {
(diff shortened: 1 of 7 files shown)


ellisonbg:
Awesome, thanks for working on this!

I am +1 on getting features to users quickly, but let's not break the
meaning of our versions to do that. With the notebook being a separate
package, we can just release its 5.0 version when we are happy with the new
features we have added.

On Sun, Aug 9, 2015 at 6:38 PM, Thomas Kluyver <notifications@github.com>
wrote:

> From discussion the other day: we are thinking of accepting some new
> features into 4.x releases, so that we can keep the user experience moving
> forwards while people are working on refactoring all the frontend stuff.
>
> Multi-cell selection is perhaps the biggest thing that we've been wanting
> to add for a long time, so I thought I'd take a crack at it. It turned out
> not to be too bad (so far...).
>
> Use: shift+j/k to select a range of cells, shift+m to merge the selected
> range. I haven't yet changed all the other actions which should now be able
> to affect a range of cells, but doing so should be mostly straightforward.
>
> API: I've tried to keep existing API working, so stuff that assumes the
> selection is a single cell will act on the cell at the cursor - the most
> recently selected one. This makes some names a bit confusing - e.g.
> Cell.unselect() now has a leave_selected parameter (to move the cursor
> away without actually unselecting). The notebook has some very similarly
> named methods, like get_selected_cell() and get_selected_cells().
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/jupyter/notebook/pull/266
> Commit Summary
>
>    - First attempt at multi-cell selection
>    - Multi-cell selection based on a selection anchor cell
>    - Merge cells based on range selection
>    - Clear selected range on entering edit mode
>
> File Changes
>
>    - *M* notebook/static/notebook/js/actions.js
>    <https://github.com/jupyter/notebook/pull/266/files#diff-0> (36)
>    - *M* notebook/static/notebook/js/cell.js
>    <https://github.com/jupyter/notebook/pull/266/files#diff-1> (21)
>    - *M* notebook/static/notebook/js/codecell.js
>    <https://github.com/jupyter/notebook/pull/266/files#diff-2> (8)
>    - *M* notebook/static/notebook/js/keyboardmanager.js
>    <https://github.com/jupyter/notebook/pull/266/files#diff-3> (4)
>    - *M* notebook/static/notebook/js/notebook.js
>    <https://github.com/jupyter/notebook/pull/266/files#diff-4> (209)
>
> Patch Links:
>
>    - https://github.com/jupyter/notebook/pull/266.patch
>    - https://github.com/jupyter/notebook/pull/266.diff
>
> —
> <https://github.com/jupyter/notebook/pull/266>.
>



-- 
Brian E. Granger

minrk:
@ellisonbg what would be broken by calling it 4.1? It would only need to be 5.0 if there were backward-incompatible changes, right?

takluyver:
The subject of our discussion the other day was that maybe we should shift what our version numbering means a bit. +0.1 releases of software in general do introduce new features, they just avoid breaking compatibility, and that's precisely what I've aimed to do here.

ellisonbg:
I am fine if we want to change how we version our projects, but such a
discussion should take place on the main jupyter/ipython lists. I think it
is very reasonable to have 4.x releases include new features but not break
backwards compat - that follows the semantic versioning pattern. But that
is not (quite) how we have been approaching our versioning in recent
releases so I think it would be good to discuss on the main list so
everyone at least knows about the change.


On Sun, Aug 9, 2015 at 8:45 PM, Thomas Kluyver <notifications@github.com>
wrote:

> The subject of our discussion the other day was that maybe we should shift
> what our version numbering means a bit. +0.1 releases of software in
> general do introduce new features, they just avoid breaking compatibility,
> and that's precisely what I've aimed to do here.
>
> —
> <https://github.com/jupyter/notebook/pull/266#issuecomment-129289288>.
>



-- 
Brian E. Granger

minrk:
It's definitely worth discussing how the new reorg affects how we treat major and minor releases, what's master, etc., as we have already been doing to some degree. But we have been doing (small) new features in +.1 releases since we started having them, so I don't think this is actually a change to how we've treated versions since 1.0. In the last several releases, we only created the `.x` backport branch once the first backward-incompatible change was merged. As we've been moving quickly, of course, that has been relatively soon after release in most cases.

takluyver:
I've sent a message to the jupyter list.

takluyver:
So: should I adapt more actions to make use of multi-cell selection in this PR, or should we review and merge it, and adapt code to use it in further PRs?

notebook/static/notebook/js/notebook.js
```diff
@@ -1289,37 +1370,66 @@ define(function (require) {
...
 
-    /**
-     * Merge the selected cell into the cell above it.
-     */
-    Notebook.prototype.merge_cell_above = function () {
-        var index = this.get_selected_index();
-        var cell = this.get_cell(index);
-        var render = cell.rendered;
-        if (!cell.is_mergeable()) {
+    Notebook.prototype.merge_cells = function(indices, into_last) {
```
Carreau: Docs.

Carreau:
+1

From a design perspective, we could style `selected-cell:last`, `selected-cell:first` differently to not have the ugly-double-border in between. (this can be separate PR though)

Carreau:
I'm for doing more in subsequents PRs.

Potentially marking as contributor friendly.

Carreau:
Also would like @juhasch eyes on that on how that will interact wit extensions, and if he have better/oter technical or API thoughts.

takluyver:
I'd definitely like style changes to be a separate PR. The styling is not especially beautiful in this, but it's usable.

notebook/static/notebook/js/notebook.js
takluyver: Docced

Carreau:
Merging soon if no-one beats me to it. That is to say: merging next time I remember I forgot to merged.

minrk:
I think you forgot long enough. Merging!

takluyver:
:tada: :shipit:

juhasch:
Great to see this in Jupyter!
I think it works pretty well, though the looks  can be improved. Adding something like
```
.cell.selected
{
    background-color: #f4f4f4;
}
```
looks much better IMO, but is probably still not optimal.
Also, I think `shift-up` and `shift-down` in command mode is more natural than `shift-j`and `shift-k` (where do they come from ?).

Converted the copy and paste notebook extension, works nicely with the new multicell-selection.

On a side note: Always having to do a `python setup.py install` after changing a single line of Javascript code in Jupyter is painful. Luckily, I can do everything I want to do with extensions.

takluyver:
The plan is to add it to shift-up/down as well, I just wanted to hide it a bit until more things actually used the multi-cell selection. I'm working on our own cut/copy/paste code right now.

PR #267, opened 2015-08-10
use mathjax component

### LINKED ISSUE #248: use components-mathjax when it exists
MathJax should soon have a [components repo](https://github.com/mathjax/MathJax/issues/1214) which doesn't include the PNG fonts. We should switch our bower dependency to that once it is official, since it will save a big chunk of the time currently spent installing from bower.

minrk:
which lacks the png fonts

for quicker installs, etc.

closes #248


Carreau:
Sorry, ment to merge. Merging, once it re-goes green.

minrk:
Travis, he is green.

PR #280, opened 2015-08-13
Add rel=noreferrer to helplink URLs as sending the referrer link is unnecessary

crwilcox:
Currently, links that go to external sites are getting the referrer
url, which for things like tmpnb, is a secret. This avoids us sending
along unnecessary information to possibly third party sites.


crwilcox:
Closed as Travis is convinced it is causing a break.  I will issue a new pull request with one commit that should pass Travis.

PR #283, opened 2015-08-13
Add rel=noreferrer to helplink URLs as sending the referrer link is unnecessary

crwilcox:
Currently, links that go to external sites are getting the referrer
url, which for things like tmpnb, is a secret. This avoids us sending
along unnecessary information to possibly third party sites.


crwilcox:
Ping

Carreau:
Sorry slipped through. +1 to me simple enough, merging.

Thanks !

PR #296, opened 2015-08-16
Add option to not use minified javascript and ease development.

Carreau:
because I'm starting to be annoyed by the 30+s.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -125,7 +125,7 @@ class IPythonHandler(AuthenticatedHandler):
         
         Mainly use for development and avoid file recompilation
         """
-        return self.settings.get('ignore_minified_js', True)
+        return self.settings.get('ignore_minified_js', False)
 
     @property
     def config(self):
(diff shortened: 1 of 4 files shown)


minrk:
I'm fine with this, though I would reverse the logic, so the default behavior is to use the minified sources (e.g. if the variable is unspecified, the template will use main.min.js). I also think `use_bundles` is very hard to understand.

Perhaps `ignore_minified_js` better represents what's going on?

Carreau:
Done

notebook/base/handlers.py
```diff
@@ -117,7 +117,16 @@ class IPythonHandler(AuthenticatedHandler):
...
-    
+
+
+    @property
+    def ignore_minified_js(self):
+        """Wether to user bundle in template. (*.min files)
+        
+        Mainly use for development and avoid file recompilation
+        """
+        return self.settings.get('ignore_minified_js', True)
```
minrk: default should be False

minrk:
:+1: other than one True that should be False

Carreau:
Rhaa nice catch, I remembered to changed the default fro the traitlets, but not here. 

Done.

Issue #302, opened 2015-08-16
Error configuring ipyparallel for Clusters tab on Windows


therriault:
Upgrading to 4.0 today on my Windows 8 machine (coincided with a fresh install of the latest version of Anaconda), and saw that the Clusters tab of the notebook said to follow the ipyparallel instructions for enabling clusters. I installed ipyparallel and used the `c.NotebookApp.server_extensions.append('ipyparallel.nbextension')` setting in the config file as instructed, but it doesn't seem to be working. The Clusters tab hasn't changed, and the following errors arise when starting the notebook:

```
c:\Anaconda>jupyter notebook

c:\Anaconda>symlink C:\Users\TherriaultA\AppData\Roaming\jupyter\nbextensions\ipyparallel -> c:\Anaconda\lib\site-packages\ipyparall
el\nbextension\static
[W 14:32:13.122 NotebookApp] Error loading server extension ipyparallel.nbextension
    Traceback (most recent call last):
      File "c:\Anaconda\lib\site-packages\notebook\notebookapp.py", line 992, in init_server_extensions
        func(self)
      File "c:\Anaconda\lib\site-packages\ipyparallel\nbextension\handlers.py", line 72, in load_jupyter_server_extension
        install_nbextension(static, destination='ipyparallel', symlink=True, user=True)
      File "c:\Anaconda\lib\site-packages\notebook\nbextensions.py", line 200, in install_nbextension
        os.symlink(path, full_dest)
    AttributeError: 'module' object has no attribute 'symlink'
symlink C:\Users\TherriaultA\AppData\Roaming\jupyter\nbextensions\ipyparallel -> c:\Anaconda\lib\site-packages\ipyparallel\nbextensi
on\static
[W 14:32:13.122 NotebookApp] Error loading server extension ipyparallel.nbextension
    Traceback (most recent call last):
      File "c:\Anaconda\lib\site-packages\notebook\notebookapp.py", line 992, in init_server_extensions
        func(self)
      File "c:\Anaconda\lib\site-packages\ipyparallel\nbextension\handlers.py", line 72, in load_jupyter_server_extension
        install_nbextension(static, destination='ipyparallel', symlink=True, user=True)
      File "c:\Anaconda\lib\site-packages\notebook\nbextensions.py", line 200, in install_nbextension
        os.symlink(path, full_dest)
    AttributeError: 'module' object has no attribute 'symlink'
[...]

PR #303, opened 2015-08-17
Fix symlink issue with Win/Py2

therriault:
Addresses issue #302 by adding check for symlink before using. (Alternative workaround already available for symlink=False, so just sets parameter as such if os.symlink does not exist.)

### DIFF
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -148,6 +148,10 @@ def install_nbextension(path, overwrite=False, symlink=False, user=False, prefix
     # make sure nbextensions dir exists
     ensure_dir_exists(nbext)
     
+    # forcing symlink parameter to False if os.symlink does not exist (e.g., on Windows machines running python 2)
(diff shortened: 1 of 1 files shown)


notebook/nbextensions.py
```diff
@@ -148,6 +148,10 @@ def install_nbextension(path, overwrite=False, symlink=False, user=False, prefix
     # make sure nbextensions dir exists
     ensure_dir_exists(nbext)
     
+    # forcing symlink parameter to False if os.symlink does not exist (e.g., on Windows machines running python 2)
+    if 'symlink' not in dir(os):
```
minrk: I think the normal check for this would be `if not hasattr(os, 'symlink')`, which avoids computing the whole list and then searching the whole list for 'symlink'.

minrk:
Thanks! Minor comment on how to check for os.symlink, but this makes sense to me.

PR #313, opened 2015-08-19
remove unconditional IPython import from notebookapp

minrk:
IPython shouldn't ever be assumed to be present

closes ipython/ipython#8746


dartdog:
so, I should just repeat the pip install --upgrade jupyter? I did that but the Ipython notebook instance showing 3.1.0

minrk:
@dartdog to just upgrade IPython, do `pip install --upgrade ipython`. It should give 4.0.

PR #321, opened 2015-08-20
run commands with shell=True

### LINKED ISSUE #315: dev install fails on Windows
The current master branch is failing to run through setup on Windows. The issue appears to be with the commands that are being called like 'npm install', or 'lessc', etc - they're not actually being executed, they're only being output to the screen and then the next step of the build is run. So despite having node set up correctly, I'm not able to get past the point of 'npm install'. Running it manually works fine though, however it then fails on the next step with the 'bower install'.

minrk:
appears to be needed on Windows and should be harmless elsewhere.

closes #315

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -296,8 +296,10 @@ def mtime(path):
 
 def run(cmd, *args, **kwargs):
     """Echo a command before running it"""
-    log.info(" ".join(cmd))
-    kwargs.setdefault('shell', True)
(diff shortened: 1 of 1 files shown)


Carreau:
This seem to actually break install on travis for whatever reason.

minrk:
Of course it does.

minrk:
This is working now.

PR #324, opened 2015-08-24
include mathjax extensions that are dynamically loaded

Carreau:
Allow to render binomial and other things:

Closes ipython/ipython#8754


TNames:
Sorry if this is an obvious question, but how is it that I apply your solution? I've already updated ipython to the most recent version and continue with the problem. Thanks.

Carreau:
You need to update **notebook** note IPython.

TNames:
Bingo! Thank you.

PR #338, opened 2015-08-26
Fix bug merge above

Carreau:
[WIP] need to finish it, and amend the commit to link the right bug number
reminder that I have to do it.


Carreau:
Though anyone can take over for an easy bug fix.

PR #268, opened 2015-08-11
Use NPM, node ES5, and browserify.

jdfreder:
~~Gets rid of require.js~~ Leaves require.js, only for backwards compatibility.
Moves a lot of install logic out of setup base
Allows us to use NPM packages directly in our JavaScript! :zap:
No longer minify JS

Pinging interested people @minrk @Carreau @takluyver @rgbkrk .

TODO:
- [x] Get working
- [x] Get tests working
- [x] Add pre/post install commands

PR #353, opened 2015-08-27
Revert "Use NPM, node ES5, and browserify."

Carreau:
Reverts jupyter/notebook#268

Half of the functionality are broken and make chrome crash 1/2 the time. 
Command palette and search and replace are easy broken one to test. 
Need to be more tested. 

Also do not work with `ignore_minified_js` so should be either removed, undocumented, and fixed. 

Also we have much older PRs and PR from non contributor that now do not apply cleanly.

The instructions in the readme also fail if you do not know how to use npm, and do not run a `npm install` after git pull on an already dev install'ed.


Carreau:
Ping @jdfreder @rgbkrk @ellisonbg. 

Also *******HUGE -1 on 4.1******* this will break how debian and other packager build the notebook. 

We cannot just break the way all the things work in a 1 month later 4.1 release that potentially introduce incompatibility. 

Ping @fperez

ellisonbg:
Because this get's into fundamental questions of how we are going to do the JS refactoring, I have brought it up with the Steering Council. We welcome further discussion here in the meantime.

ellisonbg:
Another option is for us to move straight to a 5.0 release with these changes and continue to bump the major versions as we release refactored JS code.

But I think the fundamental question is whether we consider the JS API to be public or not. Once we decide that, we can just follow the usual semver version bumping process.

Carreau:
Ok. Let me put it another way.

5.0 or not we need a 4.1 that does not break API with bug and security fixes. 
There is no notion of following semver or not. NPM enforce it, there is no way around for example. 

The question is: do we consider that "not API breaking", then it's ok in 4.1 or is it API breaking. Then it's not.

If it's not, either we revert, or we fork 4.x from before this merge. 


Envoyé de mon iPhone

ellisonbg:
Well put!

On Thu, Aug 27, 2015 at 9:02 AM, Matthias Bussonnier <

> Ok. Let me put it another way.
>
> 5.0 or not we need a 4.1 that does not break API with bug and security
> fixes.
> There is no notion of following semver or not. NPM enforce it, there is no
> way around for example.
>
> The question is: do we consider that "not API breaking", then it's ok in
> 4.1 or is it API breaking. Then it's not.
>
> If it's not, either we revert, or we fork 4.x from before this merge.
>
>
> Envoyé de mon iPhone
>
> > Le 27 août 2015 à 17:08, "Brian E. Granger" <notifications@github.com>
> a écrit :
> >
> > Another option is for us to move straight to a 5.0 release with these
> changes and continue to bump the major versions as we release refactored JS
> code.
> >
> > But I think the fundamental question is whether we consider the JS API
> to be public or not. Once we decide that, we can just follow the usual
> semver version bumping process.
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> <https://github.com/jupyter/notebook/pull/353#issuecomment-135478642>.
>



-- 
Brian E. Granger

jdfreder:
The intention was for it to be backwards compatible - if it wasn't I would have removed require.js.  You should still be able to require.js things, which for the most part should be the same.  

Sorry for the command palette and searchandreplace breakage, they don't have tests and I wasn't able to test every feature of the notebook (there are too many).  I think it would be better to leave the code in place, and fix the things that are broken.  Also, package building shouldn't differ, if it does then that too should be fixed.

Carreau:
I know the intension was/is to be backward compatible, I don't blame you.
The command palette if fine, I know it is not (yet) tested and it's fine
that it is broken.
I'll fix it.

I'm just extremely concern that we land a huge change only a few hours
(minutes?)
after it finally works, and you almost throw away all consideration of how
other could be affected by this change on a non JS api, like what is
packaged ? Do people actually reuse our files and are suddently
completly broke. How much work is it to update downstream from 4.x to 4.1.

How would for example IBM update that ? They probably go through a security
check on their team, which lets say takes 4weeks. If the 4.0 to 4.1 patch
is a few hundreds line then the security team can probably  check that 4.1
is ok in a day or two.

With this they basically have to audit browserify too, because these tool
can generate vulnerability in code
just by minification/optimisation.

When you do a 4.0 to a 4.1 you have contract that are other than pure API,
you wouldn't change the language from python to perl, you would'nt change
the required compiler from GCC to LLVM,
you wouldn't drop support for IE8.

Rewriting all the module from AMD to common js is such a breakage IMHO. It
might work,
it might not, but it will piss people off, and people that might not
complain because they are
using that under NDA and will just think F&^% $#f and stop using the
project.


As a developper I am happy, and I would have merged and iterate on that
earlier.
My community hat is just unhappy on the condition this was merged in.

It might just be yet another symptom of the lack of cmmunication,
and sorry if it fell on one of your PRs, but someone have to speak up.






On Thu, Aug 27, 2015 at 7:13 PM, Jonathan Frederic <notifications@github.com

Carreau:
I give up, 

the renaming of files and change in the source tree mean that packaging script like the one in arch, need to get extra logcis, and broke one more time the history of file in github that make it painful to see the last modification. 

Do whatever I don't care.

damianavila:
PLEASE @Carreau reopen this one... I think you raised valid points, let's keep discussing this and revert if necessary... I know you care about this... Thanks!

Carreau:
I'll reopen My PRs once I'm out of EuroSciPy, If I keep that open I'm going to get angry spend my time responding and not listen to talks. I'll reengage in the conversations/PRs on Monday.

ellisonbg:
From talking with Jon, my impression is that this PR can have some fixes
applied that truly makes it API compatible, even for packagers. I am fine
if we want revert as part of the process to reaching that goal. But from
what Matthias is saying it sound like API breakage can't be avoided? I
apologize as I don't know any of the technical details.

Do we think this refactor can be done without *any* API breakage?




On Fri, Aug 28, 2015 at 1:54 AM, Matthias Bussonnier <

> I know the intension was/is to be backward compatible, I don't blame you.
> The command palette if fine, I know it is not (yet) tested and it's fine
> that it is broken.
> I'll fix it.
>
> I'm just extremely concern that we land a huge change only a few hours
> (minutes?)
> after it finally works, and you almost throw away all consideration of how
> other could be affected by this change on a non JS api, like what is
> packaged ? Do people actually reuse our files and are suddently
> completly broke. How much work is it to update downstream from 4.x to 4.1.
>
> How would for example IBM update that ? They probably go through a security
> check on their team, which lets say takes 4weeks. If the 4.0 to 4.1 patch
> is a few hundreds line then the security team can probably check that 4.1
> is ok in a day or two.
>
> With this they basically have to audit browserify too, because these tool
> can generate vulnerability in code
> just by minification/optimisation.
>
> When you do a 4.0 to a 4.1 you have contract that are other than pure API,
> you wouldn't change the language from python to perl, you would'nt change
> the required compiler from GCC to LLVM,
> you wouldn't drop support for IE8.
>
> Rewriting all the module from AMD to common js is such a breakage IMHO. It
> might work,
> it might not, but it will piss people off, and people that might not
> complain because they are
> using that under NDA and will just think F&^% $#f and stop using the
> project.
>
>
> As a developper I am happy, and I would have merged and iterate on that
> earlier.
> My community hat is just unhappy on the condition this was merged in.
>
> It might just be yet another symptom of the lack of cmmunication,
> and sorry if it fell on one of your PRs, but someone have to speak up.
>
>
>
>
>
>
> On Thu, Aug 27, 2015 at 7:13 PM, Jonathan Frederic <
> notifications@github.com
> > wrote:
>
> > The intention was for it to be backwards compatible - if it wasn't I
> would
> > have removed require.js. You should still be able to require.js things,
> > which for the most part should be the same.
> >
> > Sorry for the command palette and searchandreplace breakage, they don't
> > have tests and I wasn't able to test every feature of the notebook (there
> > are too many). I think it would be better to leave the code in place, and
> > fix the things that are broken. Also, package building shouldn't differ,
> if
> > it does then that too should be fixed.
> >
> > —
> > <https://github.com/jupyter/notebook/pull/353#issuecomment-135511089>.
>
> >
>
> —
> <https://github.com/jupyter/notebook/pull/353#issuecomment-135689947>.
>



-- 
Brian E. Granger

Carreau:
Again it depends what you consider API. 

If you consider API that modules are CommonJS or AMD then there are no way to avoid that.  

And if you care that people will have to go through 3 month a software validation to update from 4.0 to 4.1.

jdfreder:
@Carreau I wouldn't close this, there is an alternative we can investigate, which is to forever use requirejs and instead amdify npm things.

Carreau:
> @Carreau I wouldn't close this, there is an alternative we can investigate, which is to forever use requirejs and instead amdify npm things.

This is not a question on how the PR does things. It's a question on when it was merged. 
The code itself and how it does it is perfectly fine. It's the implication of the change in a minor release, 
especially in between a X.0 and X.1 that bother me.

fperez:
I'm afraid I haven't had the bandwidth to dig into this one, but it seems indeed, if I understand the issue correctly, like a more serious regression/change than what we want to have in a .X series, especially when we say that we're using [semantic versioning](http://semver.org/).  

@Carreau, I know you're traveling, so nobody expects you to be able to stay on top of this while at a conference, but maybe just leaving it open so we can at least explore how to make progress on it?  

Let's try to identify:

- what we need to do to keep a 4.X series that feels as such from a semantic version perspective.

- what to do on a 5.x series if needed.

If we have to open both right away, then we need to consider the resource cost of having two such branches alive at the same time, though... We've always avoided having live development in two parallel branches because of how brutally costly that becomes in terms of cross-porting of code (backporting of bugfixes and security issues is bad enough, let alone real feature development).

fperez:
Basically, I think you (@Carreau) have identified an important issue we need to find a solution for here, not just close it.  

But we know you're traveling and this is also fairly complex, so by asking to leave it open we're not expecting that this is all on you, rather that we leave it for the whole team to find a way forward...

rgbkrk:
We ended up reverting the original PR in #364.

fperez:
Ah, OK, I hadn't seen #364, tx!

PR #354, opened 2015-08-27
Fixes for browserify

Carreau:
Some fixes after merge of 
https://github.com/jupyter/notebook/pull/268

Alternative to revert #268, but I think it was a mistake to do it for 4.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Dev quickstart:
 
 * ensure that you have node/npm installed (e.g. `brew install node` on OS X)
 * Clone this repo and cd into it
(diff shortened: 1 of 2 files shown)


rgbkrk:
Ah, this fixes it for your command palette.

Let's talk this morning before we decide to revert the other PR or merge this one.

As it stands, I'd prefer to merge this one and then hand merge some PRs rather than making them rebase.

Carreau:
I'm on Appear in / gitter if you like. 

The point is not that we can apply manually, I know we can, but if we start applying we cannot revert easily.I think it was merged on a rush.

Issue #343, opened 2015-08-26
search and replace toolbar icon ?


Carreau:
Do we want a toolbar icon for search and replace ? We already have magnifying glass for command palette, but we can change.

Carreau:
cc @parleur, @captainsafia, I think that @Achifaifa also might be looking for some easy fixes/features. 
( I try to tag easy fixes with sprint-friendly)

ellisonbg:
I am +1 on that.

On Wed, Aug 26, 2015 at 8:07 AM, Matthias Bussonnier <
notifications@github.com> wrote:

> cc @parleur <https://github.com/parleur>, @captainsafia
> <https://github.com/captainsafia>, I think that @Achifaifa
> <https://github.com/Achifaifa> also might be looking for some easy
> fixes/features.
> ( I try to tag easy fixes with sprint-friendly)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jupyter/notebook/issues/343#issuecomment-135054175>.
>



-- 
Brian E. Granger
Associate Professor of Physics and Data Science
Cal Poly State University, San Luis Obispo
@ellisonbg on Twitter and GitHub
bgranger@calpoly.edu and ellisonbg@gmail.com

captainsafia:
Sounds good. Feel free to cc me on any issues I should look into.

@parleur, here is a Github search query you can use to find all sprint-friendly issues across the Jupyter ecosystem.

`is:open is:issue is:sprint-friendly user:ipython user:jupyer user:phosphorjs user:nteract`

Achifaifa:
Thanks for the ping :)

I asusme this is a GUI issue? This looks like a very interesting thing to work on, but my design skills are next to null. 

Maybe I can find an open source icon that fits (Not sure if it's a good idea?), but I'd like to know where the current ones come from so I can find something similar or from the same place.

Carreau:
There is no design skill needed.
You just have to pick an icon from
https://fortawesome.github.io/Font-Awesome/icons/

I think we could use
http://fortawesome.github.io/Font-Awesome/icon/keyboard-o/ for Command
palette, and use magnifying glass for search and replace.

Achifaifa:
Oh, cool. I'll work on it.

minrk:
closed by #358

PR #358, opened 2015-08-28
swapped glass for keyboard for command palette icon

Achifaifa:
As requested in #343, changed the command palette icon from a magnifying glass to a keyboard.

The plan was to add another button for the search and replace with the magnifying glass but the command palette is currently broken, so I'll just send this PR now and leave the other changes for later (They are already done but I can't test them until that gets fixed)

### DIFF
--- a/notebook/static-src/notebook/js/actions.js
+++ b/notebook/static-src/notebook/js/actions.js
@@ -332,7 +332,7 @@
         'command-palette': {
             help_index : 'aa',
             help: 'open the command palette',
-            icon: 'fa-search',
+            icon: 'fa-keyboard-o',
             handler : function(env){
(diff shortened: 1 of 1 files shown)


Carreau:
Awesome ! Thanks !

Carreau:
Merged as #382.

Thanks ! I rebased and merge that as #382 (leaving you as the author of course). GitHub was refusing to merge this one through the UI for whatever reason !

Achifaifa:
Oh cool, thanks. Is the command palette working? There are a few more issues related to search and replace on my to-do list but I need that working :)

Carreau:
It should be working (I need to write tests though)

PR #326, opened 2015-08-24
Basic Spell checking

tritemio:
I modified the RST file. I saw there is a corresponding .ipynb file which I didn't touch.

PR #328, opened 2015-08-25
Update JavaScript Notebook Extensions.rst

nymoorland:
Hey, I was just reading the docs to better understand the internals. This doc is really informative and interesting and I love the conversational writing style! I hope you don't mind, I made a few quick ten minute proofs/edits. My only question is if this rst doc is the right place to make the edits to populate the ipynb file? Also, in the diff I was seeing whitespace but not in the raw file. Thanks.

PR #363, opened 2015-08-29
Docs typo

Carreau:
Supercede #326


Carreau:
Rha... should integrate modif from https://github.com/jupyter/notebook/pull/328 too.

Carreau:
Ok, this now should merge most of #326 and #328 by keeping commits of original authors !

Ping @minrk that have been working on the same file.

minrk:
The rst files are there on purpose. cc @ellisonbg

minrk:
@Carreau I'm fine closing #369 if you want to finish here, or wait for this to go in and rebase that. Either way is fine with me.

Carreau:
Either are fine. @Jdfreder [said earlier](https://github.com/jupyter/notebook/pull/328#issuecomment-134745177) : 

> We probably shouldn't be checking in the .rst files because it's confusing, but I think we needed to in order for things to work with readthedocs.

Do we need for RTD ? I thought it was just for the backlink.

Carreau:
And mybad, I though #369 was merged, I would have rebased on top of yours also. 
I can merge the 3 together tomorrow.

minrk:
> Do we need them for RTD ?

Backstory, if I remember correctly:

- Committing rst was assumed necessary for RTD since we can't install pandoc via requirements.txt
  - turns out, RTD environments have pandoc
- RTD doesn't run Makefiles, it runs sphinx directly, so Make dependencies don't resolve
  - we have a hack in conf.py to trigger nb→rst even 

So we definitely don't need the rst files for RTD. After realizing it wasn't needed, I think some folks wanted to try committing rst files, to see if they made reviewing prose changes easier. But if that's not the case, I think we are out of reasons to keep them.

Carreau:
This is a combinaison of 3 PRs that make fixes to docs and on the same file, and I don't want o get yet another PR on it, so I'm going to merge, we can iterate later.

minrk:
Thanks, @Carreau!

PR #370, opened 2015-08-31
Fix flexbox on celltoolbar and safari.

Carreau:
cf https://github.com/jupyter/nbgrader/issues/394

Cannot fix in flexbox.less cf above issue.

Ping @jhamrick

### DIFF
--- a/notebook/static/notebook/less/celltoolbar.less
+++ b/notebook/static/notebook/less/celltoolbar.less
@@ -12,6 +12,10 @@
     padding-right: 4px;
     .hbox();
     .end();
+    // safari fix, we cannot use -webkit-flex on hbox
+    // and vbox either or all css get borked
+    // cf https://github.com/jupyter/nbgrader/issues/394
+    display: -webkit-flex;
     @media print{
         display: none;
     }


jhamrick:
Thanks! I'll have a look at this in just a bit.

jhamrick:
Can you let me know what the right fix for the nbgrader css is in conjunction with this? There's no change without it, so I can't really test this out otherwise.

Carreau:
> Can you let me know what the right fix for the nbgrader css is in conjunction with this? There's no change without it, so I can't really test this out otherwise.

This shouldn't need any change on your side. 

I'll try to issue a PR against nbgrader that also fix it.

notebook/static/notebook/less/celltoolbar.less
```diff
@@ -12,6 +12,10 @@
     padding-right: 4px;
     .hbox();
     .end();
+    // safari fix, we cannot use -webkit-flex on hbox
+    // and vbox either or all css get borked
+    // cf https://github.com/jupyter/nbgrader/issues/394
+    display: -webkit-flex;
```
minrk: shouldn't this be in `.hbox()` instead, if it's needed?

Carreau: Technically that what we should do, but as I stated in the
comment it is broken on safari if in hbox or vbox. 
Reposted screenshot on linked issue:

minrk: Sorry, I didn't understand the comment. Okay, then. *grumbles about CSS*

Carreau: > grumbles about CSS

indeed.

PR #385, opened 2015-09-03
Use a 401 status code for invalid login attempts. Fixes #384

### LINKED ISSUE #384: 200 status code on invalid login
An invalid login attempt on a password protected server redirects back to the login page with a 200 status code. 

It would be useful for me to be able to check against the status code instead of doing what I ended up doing in iiSeymour/nbtop#6 *(comparing `request.url` against `response.url`)*. I suggest using 401 on an invalid login attempt.

iiSeymour:


iiSeymour:
Excuse the whitespace clean up masking the change. I have` (add-hook 'before-save-hook (lambda () (delete-trailing-whitespace)))` in my `.emacs`.

rgbkrk:
LGTM

minrk:
@iiSeymour we generally prefer to avoid that, but it's okay this time.

PR #390, opened 2015-09-03
move swagger API handler off of existing `/api`

### LINKED ISSUE #387: Swagger API
In the 3.X releases `/api` served the IPython version as a JSON object, however the 4.X releases return a 404 and `master` serves [api.yaml](https://github.com/jupyter/notebook/blob/master/notebook/services/api/api.yaml) [raw](https://github.com/jupyter/notebook/blob/master/notebook/services/api/handlers.py#L18).

Are all API routes eventually being migrated to use the swagger framework and `/api` will be a discovery route or should `/api` still return the version?

minrk:
that spot was already taken

also set the content-type correctly

closes #387


Carreau:
+1

PR #401, opened 2015-09-06
Fix overzealous auto escape

### LINKED ISSUE #398: jinja autoescape is escaping things that shouldn't be
There are some literals in the notebook template (sys_info, help links) that are now being escaped that shouldn't be.

Symptoms:

In Javascript console: `Uncaught SyntaxError: Unexpected token &`

Help links:

screen shot 2015-09-06 at 17 40 19

Carreau:
closes #398

Need backport.


minrk:
This is needed on sys_info as well, which is also being corrupted by autoescape.

Carreau:
Done.

PR #404, opened 2015-09-08
bump requirejs waitSeconds to 30

minrk:
default is 7, and can timeout inappropriately on slow connections

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -64,7 +64,8 @@
               deps: ["jquery"],
               exports: "$"
             }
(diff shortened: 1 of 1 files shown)


damianavila:
LGTM, I have seen this on slow connections and I can confirm that 30 second is a good compromise...

musically-ut:
I still run into issues if I open a particularly heavy notebook on a slow VPN connection.

Is there any loss in increasing the timeout to even higher number (i.e. 300), or, even better, making it configurable?

Issue #212, opened 2015-07-19
Load extension twice in commonconfig.


Carreau:
Question for @takluyver,  with sempervierens, I activated the extension both on common and on notebook, which lead to double loading. Should this be prevented/dealt with ?

takluyver:
Yeah, it should probably avoid loading it twice just because it's in two bits of config.

jcb91:
There's a similar [issue (#285)](https://github.com/ipython-contrib/IPython-notebook-extensions/issues/285) which has come up in [ipython-contrib/IPython-notebook-extensions](https://github.com/ipython-contrib/IPython-notebook-extensions), essentially that the same extension will be happily loaded twice by utils. The discussion on that issue more or less concluded that it might be worth having utils keep a list of which extensions have been loaded, to prevent this. This would also open up potential for a user-viewable list of active extensions (say in a menu somewhere), if that seemed useful. It might be worth adding a reload option to force multiple loads of the same extension path, although I can't immediately see a use for it.

I'd be happy to start trying to implement the above, with any modifications, if you think it'd be useful?

minrk:
I think remembering which extensions have been loaded is sensible and should be enough for now. Thanks, @jcb91.

Carreau:
It would also take care of the double loading if an extension is set up for both the tree/ and notebook/ page.

jcb91:
@Carreau I'm sorry, I'm not sure I follow - aren't the `tree/` and `notebook/` entirely separate environments? I'm not sure I understand how you'd get a double load, but I'll look into it today.

J

Carreau:
Sorry that's my brain that works in weird way early in the morning. 

The way I remeber we were planning to fix the double loading was to merge the `common` and `notebook` (respectively `tree`) section before loading the  extension. 

If you do remember which extensions have been loaded, then you do not need to merge anymore as the second loading will just be no-op. 

Does it make more sens ?

jcb91:
Sure, that makes sense.
So I've altered the js utils, which seems to work for me locally (see #414).
I'm trying to figure out how to get all the node and gulp building works to get it into `main.min.js` (which seems to be where some of the load calls go for me), but I can't say I understand it yet.

jcb91:
Ok, it seems from a cursory inspection that my modified `static/base/js/utils.js` has been compiled into `notebook/js/main.min.js` appropriately, so I'm going to assume that it should work in other places, unless someone tells me otherwise.

minrk:
closed by #409

PR #409, opened 2015-09-09
Addresses #212. Avoid loading nbextensions twice…

jcb91:
… by keeping track of which have been loaded.


Carreau:
Looks good to me.

PR #416, opened 2015-09-10
Also test on Python 3.3

jakirkham:
Seems everything else is being tested on Python 3.3 except for this. So, figured it should be added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
         - ~/.cache/bower
 python:
     - 3.4
+    - 3.3
     - 2.7
 sudo: false
 env:


jakirkham:
It seems that is a large number of environment variables. Maybe this isn't worthwhile even if there are some differences between the two versions.

jakirkham:
Appears that `GROUP=js/tree` sometimes hangs resulting in build errors. I am seeing the same build error with Python 3.4, as well. ( https://travis-ci.org/jakirkham/notebook/builds/79739648 ) So, I don't believe this problem is caused by adding Python 3.3 to the list of Python environments.

Carreau:
> Seems everything else is being tested on Python 3.3 except for this. So, figured it should be added.

Sure, seem simple enough. I don't see any reason not to test on 3.3 if test passes. 

> Appears that GROUP=js/tree sometimes hangs resulting in build errors. I am seeing the same build error with Python 3.4, as well. ( https://travis-ci.org/jakirkham/notebook/builds/79739648 ) So, I don't believe this problem is caused by adding Python 3.3 to the list of Python environments.

We are aware that travis have random failure, if it happen on one other PR, you can just drop us line and ask one of us to restart the failing test without justification, we'll just do it. 

I relaunched the failing test to see if it passes. 

Thanks !

jakirkham:
Thansk @Carreau. Glad to hear the random failure is a known issue. It looks like it is passing now.

minrk:
Thanks, @jakirkham!

PR #431, opened 2015-09-15
Fix bad message break notebook

Carreau:
Would need a test, and probably a way to display the message to the user
if it ever happen.

---

Who did the WrappedError ? was it @jasongrout ? @jdfreder ?
Do any of you rely on the `.catch()` behavior ?

How would you display the error if any ? Dialog ? 
Or  red badge in notification area which on click show a dialog with the full error message ?

### DIFF
--- a/notebook/static/services/kernels/kernel.js
+++ b/notebook/static/services/kernels/kernel.js
@@ -883,7 +883,7 @@ define([
         this._msg_queue = this._msg_queue.then(function() {
             return serialize.deserialize(e.data);
         }).then(function(msg) {return that._finish_ws_message(msg);})
-        .catch(utils.reject("Couldn't process kernel message", true));
+        .catch(function(error) {console.error("Couldn't process kernel message", error) });
     };
 
     Kernel.prototype._finish_ws_message = function (msg) {


notebook/static/services/kernels/kernel.js
```diff
@@ -882,8 +882,14 @@ define([
...
         this._msg_queue = this._msg_queue.then(function() {
             return serialize.deserialize(e.data);
-        }).then(function(msg) {return that._finish_ws_message(msg);})
-        .catch(utils.reject("Couldn't process kernel message", true));
+        }).then(function(msg) {
+            try {
+                return that._finish_ws_message(msg);
+            } catch (e) {
+                console.error("Couldn't process kernel message", e)
+                return Promise.resolve()
```
minrk: Don't forget your semicolons...

Carreau: I prefer [greek question mark]9http://www.fileformat.info/info/unicode/char/037e/index.htm)

jdfreder:
I wrote the WrappedError implementation.  I never display it to the user.  It's only ever logged to the console.  It maintains a nice stack, which is lost in complex promise operations.  In the widgets we use `catch` and `then` extensively.

If you need to display an error, I'd continue logging the error to the console with the stack but additionally display a human digestible error, depending on the context.

notebook/static/services/kernels/kernel.js
```diff
@@ -882,8 +882,14 @@ define([
...
         this._msg_queue = this._msg_queue.then(function() {
             return serialize.deserialize(e.data);
-        }).then(function(msg) {return that._finish_ws_message(msg);})
-        .catch(utils.reject("Couldn't process kernel message", true));
+        }).then(function(msg) {
+            try {
+                return that._finish_ws_message(msg);
+            } catch (e) {
+                console.error("Couldn't process kernel message", e);
+                return Promise.resolve();
```
jdfreder: So the difference here is that you want to promise chain to continue, even if there is a failure?

If so, you can continue to use catch, just don't use the utils.reject function.

jdfreder: Try this
```js
        }).then(function(msg) {return that._finish_ws_message(msg);})
        .catch(function() {console.error("Couldn't process kernel message") });
```

Carreau:
Done.

jdfreder:
This looks good to me, I'm waiting to see what Mr. Travis thinks...

jdfreder:
Thanks @Carreau !

Carreau:
Thanks for review on tips !

PR #435, opened 2015-09-15
Remove bootstrap dependency in menubar.js

### LINKED ISSUE #113: menubar.js seem to require (but not use) bootstrap
Investigate if can be removed.

jdfreder:
I looked through  all of the selector references, to make sure no bootstrap injected functions were used, and I couldn't find any.  It probably doesn't matter, require.js doesn't seem to completely isolate jquery, so bootstrap, even if imported elsewhere, will be available in menu on the jquery object (this is one of the things I tackled in the npm-ify PR).

closes #113


Carreau:
Travis happy.

Issue #352, opened 2015-08-27
mathjax installation instructions are wrong


anntzer:
Starting my first jupyter notebook I get the following message:
```
>>> from Jupyter.external import mathjax; mathjax.install_mathjax()
```

Well, there is no module named `Jupyter`.  Grepping through site-packages shows that the only function called `install_mathjax` is in `IPython.external`; calling it installs `mathjax` in `~/.ipython/nbextensions` but doesn't help Jupyter notebook at all.

Carreau:
Already fixed by `https://github.com/jupyter/notebook/pull/323` You though appear to have a blend of IPython 3.x and 4.0. how did yo install ?

anntzer:
Arch Linux standard packages:
```
$ pacman -Q jupyter ipython
jupyter 4.0.4-2
ipython 4.0.0-2
```

Carreau:
Does the rest of the notebook works ?
Because mathjax should be bundled with all the other js libraries, so there is no reason of it being unavailable and not others.

anntzer:
The rest works.  Maybe some packaging issues?  I'll let you check https://www.archlinux.org/packages/community/any/jupyter/ and https://projects.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/jupyter (which is used to build the package).

Carreau:
Yep, mathjax seem optional:

https://projects.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/jupyter#n97

https://www.archlinux.org/packages/community/any/jupyter/ `mathjax (optional)`

You can propose a patch to arch, that is basically the opposite of #323 but give information on how to install it the arch way (with Yum, pacman, or I don't remember).

anntzer:
I think this should be fixed on the distro's side (perhaps with a clear notice that if they strip out mathjax then mathjax should be installed by the package manager as a dep).  See https://bugs.archlinux.org/task/45999

Carreau:
They already do what suggested in bugs, they symlink, but the deps  is just marked as optional. 
it should either be made non optional, or fix/reintroduce the warning on load but with the correct instructions. 

It will also depends on #355 that I just open to see if we make things easier for distrib to unbundle matjax or not.

anntzer:
That's why I am asking the packagers to make mathjax a required dep.  Sounds just simpler.  My parenthetical point was that the jupyter team may want to emphasize this to downstream packagers.

[...]

PR #436, opened 2015-09-15
Removed --no-mathjax flag

### LINKED ISSUE #355: remove --no-mathjax flag (or not) ?
Mathjax should be bundled, but distrib might make it optional.

Cf discussion in https://github.com/jupyter/notebook/issues/352. 

Should we 

A) re-introduce what was removed by https://github.com/jupyter/notebook/pull/323, but with a placeholder for distributions ? 

B) Remove the --no-mathjax flag ?

---

Personally +0.5 for A) and -0.5 for B)

captainsafia:
Closes #355

Let me know if anything needs to be changed and I will fix it ASAP.

giphy 3

-- 

[Edit] 
One of the question is : do we remove the `--no-mathjax` flag, that could be disable in environment that have security concern ?

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,9 @@ Jupyter notebook changelog
 A summary of changes in the Jupyter notebook.
 For more detailed information, see `GitHub <https://github.com/jupyter/notebook>`__.
 
+4.1.0
+------
+- Deprecated --no-mathjax flag
 
 4.0.2
 -----
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,7 +8,7 @@ For more detailed information, see `GitHub <https://github.com/jupyter/notebook>
 
 4.1.0
 ------
-- Deprecated --no-mathjax flag
+- Removed --no-mathjax flag
 
 4.0.2
 -----
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -321,17 +321,6 @@ flags['pylab']=(
     {'NotebookApp' : {'pylab' : 'warn'}},
     "DISABLED: use %pylab or %matplotlib in the notebook to enable matplotlib."
 )
-flags['no-mathjax']=(
-    {'NotebookApp' : {'enable_mathjax' : False}},
-    """Disable MathJax
-    
-    MathJax is the javascript library Jupyter uses to render math/LaTeX. It is
-    very large, so you may want to disable it if you have a slow internet
-    connection, or for offline use of the notebook.
-    
-    When disabled, equations etc. will appear as their untransformed TeX source.
-    """
-)
 
 # Add notebook manager flags
 flags.update(boolean_flag('script', 'FileContentsManager.save_script',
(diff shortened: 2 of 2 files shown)


Carreau:
fi you superfast you can add a line into the [changelog](https://github.com/jupyter/notebook/blob/master/docs/source/changelog.rst)  for noticible features.

docs/source/changelog.rst
```diff
@@ -6,6 +6,9 @@ Jupyter notebook changelog
 A summary of changes in the Jupyter notebook.
 For more detailed information, see `GitHub <https://github.com/jupyter/notebook>`__.
 
+4.1.0
+------
+- Deprecated --no-mathjax flag
```
Carreau: Bikeshedding, but isn't deprecated actually having something discouraged but still present ?

captainsafia:
@Carreau — you are correct! Both @willingc and this cute kitty agree with you.

8325_8_site

minrk:
I don't think we are strictly allowed to remove things in minor releases. @Carreau does `--no-mathjax` not work in 4.0?

minrk:
PS @captainsafia issue cross-linking doesn't work in titles, so can you instead reference issues in the body in the future? That way, we get auto-closing/linking of issues on the GitHub pages.

Carreau:
It's accepted i 4.0 but has in practice no effects and lead to the confusion/bug report of mathjax beeing made optional on Arch.

captainsafia:
> @captainsafia issue cross-linking doesn't work in titles

Ah yes, thank you for letting me know! I usually put it in the commit messages, but I forgot this time.

Carreau:
So @minrk, do you think we should just leave the flag for now and just issue a DeprecationWarning ?

minrk:
The main reason I see for leaving `--no-mathjax` is that there could possibly be a deployment with security concerns about mathjax js, as there have been in the past. Re: #323, #352 I don't think it should be possible to *install* the notebook without mathjax, and such an installation should rightly be considered broken and incomplete. But it is conceivable to allow it to be *used* without mathjax. I don't feel too strongly about it, though, so if we want to remove the ability for it to run without mathjax, I'm okay with that proposal (in which case Deprecation is appropriate), though I would recommend bringing that up on the list in case there is anyone relying on it now.

damianavila:
> But it is conceivable to allow it to be used without mathjax

This is a very valid point... I think we should left the option for these cases...

Carreau:
> This is a very valid point... I think we should left the option for these cases...

Ok, fait enough. 

So do we want to get rid of the flag at all ? If no let's change #355 title.

Carreau:
(bumping this PR to 5.x if we decide to remove Mathjax anyway)

jakirkham:
> if we decide to remove Mathjax anyway

Do you just mean the flag or Mathjax entirely? If the latter, what is the alternative for future reference? Thanks.

Carreau:
> Do you just mean the flag or Mathjax entirely?

No just the flag. We wont' remove Mathjax,

ellisonbg:
Yeah, this should wait until 5.0.

Carreau:
Now that we are on our way to 5.x, can we rebase than and decide wether or not we want to keep the --no-mathjax  flag ?

Carreau:
Still needs rebase.

I'm going to go with: Go with deprecation cycle, leave the flag (that does nothing) but have it print a warning. If by 5.0 nobody has complained we can re-add it.

captainsafia:
OK. That seems reasonable to me. I'll update tomorrow.

Carreau:
> OK. That seems reasonable to me. I'll update tomorrow.

@captainsafia want to take care of that once you get your new laptop? 
Or should we assign an undergrad ?

captainsafia:
Not sure when the new laptop would be coming so it might be better to give it to a new undergrad so they can learn a little bit too.

Carreau:

PR #437, opened 2015-09-15
Do not trigger events only once.

### LINKED ISSUE #256: Kernel info not updated completely till kernel restart
Quote from @jasongrout on the ipython/ipywidgets gitter channel:

> When I just do Kernel | Change Kernel to python 3, then save, the diff looks like this:
> ```diff
diff --git a/examples/Widget List.ipynb b/examples/Widget List.ipynb
index 11915b4..fc21a4c 100644
--- a/examples/Widget List.ipynb        
+++ b/examples/Widget List.ipynb        
@@ -621,9 +621,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "python2"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
```
> however, when I restart the kernel after changing the kernel (Kernel | restart | clear output and restart), then save, I get a bigger diff
```diff
diff --git a/examples/Widget Events.ipynb b/examples/Widget Events.ipynb
index e47b7f9..be47d18 100644
--- a/examples/Widget Events.ipynb      
+++ b/examples/Widget Events.ipynb      
@@ -359,21 +359,21 @@
    ]
   ],
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "python2"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
     "name": "ipython",
-    "version": 2
+    "version": 3
    },
    "file_extension": ".py",
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython2",
-   "version": "2.7.10"
+   "pygments_lexer": "ipython3",
+   "version": "3.4.3"
   }
  },
  "nbformat": 4,
```
> It appears that the language_info, the version, and the pygments_lexer aren't being appropriately updated when just doing a kernel change.
> My worry is that, for example, the codemirror isn't set, so someone opening up the notebook will see the wrong syntax highlighting

Carreau:
Kernel_ready should be triggered on kernel changed or restart.
notebook loaded can be triggered on checkpoint change.

closes #256

### DIFF
--- a/notebook/static/notebook/js/main.js
+++ b/notebook/static/notebook/js/main.js
@@ -138,7 +138,7 @@ require([
 
     page.show();
 
-    events.one('notebook_loaded.Notebook', function () {
+    events.trigger('notebook_loaded.Notebook', function () {
         var hash = document.location.hash;
         if (hash) {
             document.location.hash = '';
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -1802,7 +1802,7 @@ define(function (require) {
         function restart_and_resolve () {
             that.kernel.restart(function () {
                 // resolve when the kernel is *ready* not just started
-                that.events.one('kernel_ready.Kernel', resolve_promise);
+                that.events.trigger('kernel_ready.Kernel', resolve_promise);
             }, reject_promise);
         }
     


jdfreder:
I would say a regression test would be nice, but I'll leave that up to you, seeing that our JS tests may slowly die when we start npm installing components like jupyter-js-services which are self tested.  Let me know if you want to merge without a unit test...

notebook/static/notebook/js/main.js
```diff
@@ -138,7 +138,7 @@ require([
 
     page.show();
 
-    events.one('notebook_loaded.Notebook', function () {
+    events.trigger('notebook_loaded.Notebook', function () {
```
minrk: This is registering a callback. It shouldn't be triggering the event.

notebook/static/notebook/js/notebook.js
```diff
@@ -1802,7 +1802,7 @@ define(function (require) {
         function restart_and_resolve () {
             that.kernel.restart(function () {
                 // resolve when the kernel is *ready* not just started
-                that.events.one('kernel_ready.Kernel', resolve_promise);
+                that.events.trigger('kernel_ready.Kernel', resolve_promise);
```
minrk: Same here - this is *adding* a trigger for kernel_ready and unregistering a callback.

Carreau:
Ok, I missunderstood the `one`.

Issue #432, opened 2015-09-15
Cross origin websockets are broken for TermSocket


zischwartz:
Regardless of what `--NotebookApp.allow_origin` is set to, trying to use the the terminal/terminado websocket connection across origins is broken, and causes the error below.

Working on a fix now. I think TermSocket just needs to implement the `check_origin` [method](http://tornado.readthedocs.org/en/latest/websocket.html#tornado.websocket.WebSocketHandler.check_origin) (to satisfy tornado>4) and, confusingly `origin_check` (to override the terminado [method](https://github.com/takluyver/terminado/blob/master/terminado/websocket.py#L59)).

I'm running this from the included docker file. 

Vaguely related to  ipython/ipython#8375.

```
[I 15:33:34.190 NotebookApp] The Jupyter Notebook is running at: http://[all ip addresses on your system]:80/
[I 15:33:34.191 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[I 15:33:38.299 NotebookApp] 302 GET / (192.168.99.1) 1.14ms
[I 15:33:46.678 NotebookApp] New terminal with automatic name: 1
[I 15:41:31.373 NotebookApp] New terminal with automatic name: 2
[E 15:41:31.394 NotebookApp] Uncaught exception GET /terminals/websocket/2 (192.168.99.1)
    HTTPServerRequest(protocol='http', host='192.168.99.100', method='GET', uri='/terminals/websocket/2', version='HTTP/1.1', remote_ip='192.168.99.1', headers={'Accept-Encoding': 'gzip, deflate, sdch', 'Connection': 'Upgrade', 'Origin': 'http://0.0.0.0:8080', 'Sec-Websocket-Version': '13', 'Pragma': 'no-cache', 'Upgrade': 'websocket', 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.85 Safari/537.36', 'Host': '192.168.99.100', 'Accept-Language': 'en-US,en;q=0.8', 'Sec-Websocket-Extensions': 'permessage-deflate; client_max_window_bits', 'Cookie': '_ga=GA1.1.1035836965.1441995168', 'Cache-Control': 'no-cache', 'Sec-Websocket-Key': 'LZhtgg5WwFz5xFaDFUNpGw=='})
    Traceback (most recent call last):
[...]

PR #439, opened 2015-09-16
Added cross origin websocket handling for terminal

zischwartz:
Resolves jupyter/notebook#432

I ended up using the `check_origin(self)` code from `APIHandler`  in a new `check_origin(self, origin)` method in `TermSocket `to satisfy Tornado, which requires the `origin` argument, and adding a `origin_check` method to `TermSocket` to override Terminado's own `origin_check` function which hates cross origin websockets.

If anyone sees a more DRY way to do this, I'm all ears.

### DIFF
--- a/notebook/terminal/handlers.py
+++ b/notebook/terminal/handlers.py
@@ -22,53 +22,14 @@ class TerminalHandler(IPythonHandler):
         self.write(self.render_template('terminal.html',
                    ws_path="terminals/websocket/%s" % term_name))
 
-class TermSocket(terminado.TermSocket, IPythonHandler):
+class TermSocket(IPythonHandler, terminado.TermSocket):
 
     def set_default_headers(self):
         pass
 
-    def check_origin(self, origin):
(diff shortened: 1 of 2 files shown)


notebook/terminal/handlers.py
```diff
@@ -17,8 +23,53 @@ def get(self, term_name):
                    ws_path="terminals/websocket/%s" % term_name))
 
 class TermSocket(terminado.TermSocket, IPythonHandler):
+
     def set_default_headers(self):
         pass
+
+    def check_origin(self, origin):
```
Carreau: this seem to be identical to what is in notebook/base/handler:APIHandler.
Do you think we should move the method up the hierarchy ! class (to IPythonHandler) to avoid code duplication ?

zischwartz: Sounds good

zischwartz:
Could this also be backported to 3.X? Thanks!

PR #441, opened 2015-09-16
Fix small docs errors and add hidden index for clean doc builds

willingc:
This PR contains edits to allow `make html` execute with no errors or warnings. This will be useful for automating build and test of docs in the future.

`conf.py`  disabled html_static_path since it is not currently used
`contents.rst` fix spacing error
`index.rst` Added a hidden table of contents to prevent the docs generated from older ipynb docs from causing warnings.


Carreau:
:sparkles:

PR #447, opened 2015-09-16
Added view and logic for search and replace menu item

### LINKED ISSUE #342: Search and replace menu item.
Add action item in one menu. 

@parleur, @captainsafia should be an easy  5-10 line fix, 1 in notebook/template/notebook.html and 3 or 4 in `notebook/static/notebook/js/(notebook|main).js`, that bind the click to `action.call('name of search and replace action')`

captainsafia:
Closes #342.

Let me know how if anything needs to be changed.

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -325,6 +325,11 @@ define(function (require) {
         var x = new commandpalette.CommandPalette(this);
     };
 
+    Notebook.prototype.show_search_and_replace = function(event) {
+        var kbd = this.keyboard_manager;
+        kbd.actions.call('ipython.search-and-replace-dialog', event, kbd);
+    }
+
     /**
      * Trigger a warning dialog about missing functionality from newer minor versions
      */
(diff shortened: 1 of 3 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -325,6 +325,11 @@ define(function (require) {
         var x = new commandpalette.CommandPalette(this);
     };
 
+    Notebook.prototype.show_search_and_replace = function(event) {
+        var kbd = this.keyboard_manager;
+        kbd.actions.call('ipython.search-and-replace-dialog', event, kbd);
+    }
+
```
Carreau: That feel to me like an extra step to me, directly do `action.call(...)` in `menubar.js` and no need to pass kbd at the end.

captainsafia:
@Carreau — Yeah, that was one of those things that made sense in the moment (all the other `click` listeners invoke a function on the `Notebook` prototype, so should this one!), but the complexity is apparent in hindsight.

Carreau:
> all the other click listeners invoke a function on the Notebook prototype, so should this one!

that's changing in jupyter/notebook#329

Carreau:
Thanks !

PR #449, opened 2015-09-17
Don't crash if a filename contains invalid unicode.

### LINKED ISSUE #287: Unicode error on index page
Hey,

I've just installed Ipython and Jupyter into a Python-2.7 virtualenv on my OpenBSD machine.

When invoking `ipython notebook` the following exception is thrown:

```
...
Created new window in existing browser session.
[E 11:01:40.917 NotebookApp] Unhandled error in API request
      File "/tmp/c/lib/python2.7/site-packages/notebook/base/handlers.py", line 436, in wrapper
        result = yield gen.maybe_future(method(self, *args, **kwargs))
      File "/tmp/c/lib/python2.7/site-packages/tornado/gen.py", line 870, in run
        value = future.result()
      File "/tmp/c/lib/python2.7/site-packages/tornado/concurrent.py", line 215, in result
        raise_exc_info(self._exc_info)
      File "/tmp/c/lib/python2.7/site-packages/tornado/gen.py", line 230, in wrapper
        yielded = next(result)
      File "/tmp/c/lib/python2.7/site-packages/notebook/services/contents/handlers.py", line 129, in get
        path=path, type=type, format=format, content=content,
      File "/tmp/c/lib/python2.7/site-packages/notebook/services/contents/filemanager.py", line 346, in get
        model = self._dir_model(path, content=content)
      File "/tmp/c/lib/python2.7/site-packages/notebook/services/contents/filemanager.py", line 249, in _dir_model
        os_path = os.path.join(os_dir, name)
      File "/tmp/c/lib/python2.7/posixpath.py", line 73, in join
        path += '/' + b
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xb0 in position 26: ordinal not in range(128)
[E 11:01:40.919 NotebookApp] {
      "Accept-Language": "en-US,en;q=0.8", 
      "Accept-Encoding": "gzip, deflate, sdch", 
      "Connection": "keep-alive", 
      "Accept": "application/json, text/javascript, */*; q=0.01", 
      "User-Agent": "Mozilla/5.0 (X11; OpenBSD amd64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.132 Safari/537.36", 
      "Host": "localhost:8888", 
      "Referer": "http://localhost:8888/tree", 
      "Cookie": "__S=Q6uwuZs6NjSkTT3WOAMSt2", 
      "X-Requested-With": "XMLHttpRequest"
    }
[E 11:01:40.919 NotebookApp] 500 GET /api/contents?type=directory&_=1439546499953 (127.0.0.1) 22.39ms referer=http://localhost:8888/tree
```

The browser also displays the traceback in the space reserved for the list of notebooks.

I am however, able to create a notebook and use the notebook. Just the index page is broken.

Thanks

vext01:
Fixes #287.

I've created this pull against master. May need to cherry pick into 4.0.x also?

Cheers

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -245,7 +245,12 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
             model['content'] = contents = []
             os_dir = self._get_os_path(path)
             for name in os.listdir(os_dir):
-                os_path = os.path.join(os_dir, name)
+                try:
+                    os_path = os.path.join(os_dir, name)
+                except UnicodeDecodeError as e:
+                    self.log.warn(
+                        "failed to decode filename '%s': %s", name, e)
+                    continue
                 # skip over broken symlinks in listing
                 if not os.path.exists(os_path):
                     self.log.warn("%s doesn't exist", os_path)


minrk:
Thanks. I'm still surprised this is failing, since all the variables should be unicode strings already. Can you add a debug statement that prints

```python
print("os_dir=%r, name=%r, os_path=%r" % (os_dir, name, os_path))
```

and share the output for the failing directory?

vext01:
Here's what we print for the troublesome filename:

```
os_dir=u'/home/edd', name='.\n    set document.color\x1a\xb0\x7f\xa6w', os_path=u'/home/edd/.systrace'
```

Is the problem that os.path.join expects to build an ascii string? If you look  at the types, `os_dir` is a unicode string, whereas name is not.

```
In [25]: x
Out[25]: '.\n    set document.color\x1a\xb0\x7f\xa6w'
In [26]: os.path.join("", x)
Out[26]: '.\n    set document.color\x1a\xb0\x7f\xa6w'

In [27]: os.path.join(u"", x)
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-27-239136892ae0> in <module>()
----> 1 os.path.join(u"", x)

/usr/local/lib/python2.7/posixpath.pyc in join(a, *p)
     69             path = b
     70         elif path == '' or path.endswith('/'):
---> 71             path +=  b
     72         else:
     73             path += '/' + b

UnicodeDecodeError: 'ascii' codec can't decode byte 0xb0 in position 25: ordinal not in range(128)
```

minrk:
What surprises me is that `name` is a str and not unicode, which it should be. What do you get from `os.listdir(u'.')`? And what is the full Python version? When I use Python 2.7.10, listdir returns the same type as its input.

vext01:
```
$ python2.7
Python 2.7.10 (default, Sep  1 2015, 08:18:20) 
[GCC 4.2.1 20070719 ] on openbsd5
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.listdir(u'.')
[u'.cvsrc', u'.Xdefaults', u'.cshrc', u'.ssh', ..., '.\n    set document.color\x1a\xb0\x7f\xa6w', ...]
>>> os.listdir('.')
['.cvsrc', '.Xdefaults', '.cshrc', '.ssh', ..., '.\n    set document.color\x1a\xb0\x7f\xa6w', ...]
```

In the former case, all elements are unicode bar the problem element. Is it because the unicode codebases are invalid?

minrk:
@vext01 Thanks for investigating. In that case, I think this fix is fine as-is, but at least we know the affected case:

- Python 2 + filenames that Python cannot decode from the filesystem (e.g. non-utf8).

vext01:
Thanks so much.

PR #443, opened 2015-09-16
Update the 'Securing your server' section of public server security document

willingc:
This PR contains significant formatting and editing of the 'Securing your server' section of the public_server.rst document.

Addresses #428.

TODO: Update second half of public_server.rst. Migrate any content from ipynb file that may enhance the rst doc.

Issue #428, opened 2015-09-14
Configuring the Notebook and Server.ipynb / public_server.rst duplicate info


Carreau:
also the ipynb refer to a crt file which is not created by the openssl command.

Carreau:
@willingc that look that something you might be good at, want to have a look ?

willingc:
@Carreau Taking a look at this [public_server.rst](https://github.com/jupyter/notebook/blob/master/docs/source/public_server.rst). A couple of questions:

* Is the goal to improve the docs on creating a self-signed cert and the gotchas that you may face when creating?
* When using the notebook with a self-signed cert `--certfile=mycert.pem`, the browser will likely complain that it's an unsafe action and ask to return to safety. What is our typical response to the user on what to do?
* Are both the `.rst` file or the `.ipynb` documents being used by users to create and use a cert? Or is one doc a legacy doc and the other the doc that will be used going forward?

Carreau:
> Is the goal to improve the docs on creating a self-signed cert and the gotchas that you may face when creating?

That's one goal if possible but not obligatory. 



> When using the notebook with a self-signed cert --certfile=mycert.pem, the browser will likely complain that it's an unsafe action and ask to return to safety. What is our typical response to the user on what to do?

That we are aware, and even self signed is better than no certificate at all if you don't want people to snoop or modify what information on the wire.

> Are both the .rst file or the .ipynb documents being used by users to create and use a cert? Or is one doc a legacy doc and the other the doc that will be used going forward?

I think the rst is older. I think we can keep the `.ipynb` I just want to make sure all the info is on the one we keep whichever you think is fit. 

@rgbkrk, @minrk any other thoughts on that ?

minrk:
I think this doc makes more sense as plain rst, but I don't feel strongly about which one we keep. I suspect there's information in both that should be merged into whichever one.

rgbkrk:
> When using the notebook with a self-signed cert --certfile=mycert.pem, the browser will likely complain that it's an unsafe action and ask to return to safety. What is our typical response to the user on what to do?

We should link people to a solid resource for adding their (trusted) self signed certs to their system. This applies for Firefox, Chrome, and IE. Safari on the other hand is pretty much a non-starter. IIRC, self signed certs with Safari don't work with websockets for some reason.

minrk:
> self signed certs with Safari don't work with websockets for some reason.

[...]

PR #458, opened 2015-09-18
Add review changes to public server doc

willingc:
This PR adds changes from review comments of PR #443.

Addresses #428.

@Carreau @minrk I will submit a separate PR for removing the ipynb doc that duplicates this rst file. @ellisonbg suggested that we remove the duplicate ipynb docs that are better served by an rst file. For example, a doc like this one where the the user is working on the command line instead of in the notebook.


minrk:
Removing the duplicates is good, thanks.

PR #472, opened 2015-09-21
avoid double-encoding base_url

### LINKED ISSUE #295: mishandling when base_url needs encoding
We inconsistently include base_url in the URL encoding of API requests, making it impossible to include characters like `@` in the URL prefix.

proposal: always treat base_url as already encoded

I think this will be simpler in the way it's passed around from the server to the frontend, especially since tornado's url matching is done on the escaped URL.

minrk:
base_url should already be url-encoded (on both sides), so it shouldn't ever be passed through to urlencode.

Both the javascript and server-side tests now run with a base URL that would be dealt with incorrectly if it were encoded.

closes #295

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -564,7 +564,7 @@ class FilesRedirectHandler(IPythonHandler):
         cm = self.contents_manager
         if cm.dir_exists(path):
             # it's a *directory*, redirect to /tree
-            url = url_path_join(self.base_url, 'tree', path)
+            url = url_path_join(self.base_url, 'tree', url_escape(path))
         else:
             orig_path = path
             # otherwise, redirect to /files
@@ -580,8 +580,7 @@ class FilesRedirectHandler(IPythonHandler):
             if not cm.file_exists(path=path):
                 raise web.HTTPError(404)
 
-            url = url_path_join(self.base_url, 'files', path)
-        url = url_escape(url)
+            url = url_path_join(self.base_url, 'files', url_escape(path))
         self.log.debug("Redirecting %s to %s", self.request.path, url)
         self.redirect(url)
     
--- a/notebook/jstest.py
+++ b/notebook/jstest.py
@@ -222,6 +222,7 @@ class JSController(TestController):
         self.section = section
         self.xunit = xunit
         self.url = url
+        self.base_url = '/a@b/'
         self.slimer_failure = re.compile('^FAIL.*', flags=re.MULTILINE)
         js_test_dir = get_js_test_dir()
         includes = '--includes=' + os.path.join(js_test_dir,'util.js')
@@ -264,7 +265,7 @@ class JSController(TestController):
             self.server_port = 0
             self._init_server()
             if self.server_port:
-                self.cmd.append("--port=%i" % self.server_port)
+                self.cmd.append('--url=http://localhost:%i%s' % (self.server_port, self.base_url))
             else:
                 # don't launch tests if the server didn't start
                 self.cmd = [sys.executable, '-c', 'raise SystemExit(1)']
@@ -310,7 +311,7 @@ class JSController(TestController):
             '-m', 'notebook',
             '--no-browser',
             '--notebook-dir', self.nbdir.name,
-            '--NotebookApp.base_url=/e%40mail/',
+            '--NotebookApp.base_url=%s' % self.base_url,
         ]
         # ipc doesn't work on Windows, and darwin has crazy-long temp paths,
         # which run afoul of ipc's maximum path length.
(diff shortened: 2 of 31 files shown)


minrk:
Marked WIP, since there's some fixes to do Python-side as well.

minrk:
Took some tracking down, and got waylaid by a phantomjs bug where it helpfully unescapes websocket URLs when it shouldn't, but I think this is good to go, assuming Travis agrees.

Carreau:
Travis seem happy, will try to review post today's meeting.

jdfreder:
> waylaid

Word of the day! cool

Reviewing...

notebook/base/handlers.py
```diff
@@ -580,8 +580,7 @@ def redirect_to_files(self, path):
             if not cm.file_exists(path=path):
                 raise web.HTTPError(404)
 
-            url = url_path_join(self.base_url, 'files', path)
-        url = url_escape(url)
```
jdfreder: `base_url` doesn't need to be escaped, ever?

notebook/jstest.py
```diff
@@ -222,6 +222,7 @@ def __init__(self, section, xunit=True, engine='phantomjs', url=None):
         self.section = section
         self.xunit = xunit
         self.url = url
+        self.base_url = '/a@b/'
```
jdfreder: What's `/a@b/` mean?  Sorry if I'm missing something obvious here.  Kind of looks like a regex!  Google foo failed me: https://www.google.com/#q=%2Fa%40b%2F

jdfreder: Ah, is that short for "AB testing"?

notebook/base/handlers.py
jdfreder: Ignore that

notebook/services/kernels/tests/test_kernels_api.py
```diff
@@ -63,29 +63,33 @@ def test_default_kernel(self):
         # POST request
         r = self.kern_api._req('POST', '')
         kern1 = r.json()
-        self.assertEqual(r.headers['location'], '/api/kernels/' + kern1['id'])
+        self.assertEqual(r.headers['location'], url_path_join(self.url_prefix, 'api/kernels', kern1['id']))
```
jdfreder: Here you aren't escaping `kern1['id']`, is that just because you know the format of the kernel id doesn't need to be escaped?

jdfreder: (I see this a couple places below too)

notebook/services/sessions/handlers.py
```diff
@@ -12,7 +12,7 @@
 
 from ...base.handlers import APIHandler, json_errors
 from jupyter_client.jsonutil import date_default
-from notebook.utils import url_path_join, url_escape
```
jdfreder: cool

notebook/templates/notebook.html
```diff
@@ -26,11 +26,10 @@
 {% block params %}
 
 {{super()}}
-data-project="{{project}}"
-data-base-url="{{base_url}}"
-data-ws-url="{{ws_url}}"
-data-notebook-name="{{notebook_name}}"
-data-notebook-path="{{notebook_path}}"
+data-base-url="{{base_url | urlencode}}"
```
jdfreder: Ah!

notebook/tests/launchnotebook.py
```diff
@@ -39,11 +39,12 @@ class NotebookTestBase(TestCase):
 
     port = 12341
     config = None
+    url_prefix = '/a%40b/'
```
jdfreder: Ah - `/a@b/` is just an arbitrary unique string you picked to prefix testing URLs?

notebook/services/kernels/tests/test_kernels_api.py
jdfreder: Ah, is this the thing that waylaid you?  Phantom unescaping the URLs.

jdfreder:
Awesome

notebook/tests/launchnotebook.py
jdfreder: Ah I just read the original issue that this PR closes.  You're doing this just to test our ability to handle the @ symbol in the path name!

notebook/jstest.py
jdfreder: (figured it out, see my comment below)

notebook/tests/launchnotebook.py
jdfreder: Maybe a comment explaining that?

minrk:
@jdfreder comments added

jdfreder:
Sweet, thanks Min!

PR #488, opened 2015-09-23
Update docs build and config for clean RTD deploys

willingc:
Work in process to update config/build for RTD deployment.


willingc:
@minrk @Carreau This PR will hopefully make RTD happier, and let it build the docs without failing. If you could review and merge to see if it builds cleanly, that would be great. If the build fails, RTD will just continue to deliver the old docs. Thanks.

willingc:
I should clarify that it builds cleanly locally. It would be interesting to see if RTD's build system will build cleanly.

minrk:
Thanks! Let's see how it goes...

Issue #479, opened 2015-09-22
Docs: notebooks rendering and broken link


tritemio:
It is probably a known issue, but the [notebook examples](http://jupyter-notebook.readthedocs.org/en/latest/examples/Notebook/Examples%20and%20Tutorials%20Index.html) integrated in the Sphinx documentations are not rendering well. In particular, [images](http://jupyter-notebook.readthedocs.org/en/latest/examples/Notebook/Notebook%20Basics.html) and many [LaTeX equations](http://jupyter-notebook.readthedocs.org/en/latest/examples/Notebook/Typesetting%20Equations.html) are not rendered.

Also, there is a broken link that prevents accessing the original notebook here

http://jupyter-notebook.readthedocs.org/en/latest/examples/Notebook/Configuring%20the%20Notebook%20and%20Server.html

Finally the link "Edit on GitHub" is broken for rendered notebook.

Carreau:
Thanks, we'll try to work on that !

willingc:
@tritemio Thanks.

@carreau @minrk Would you mind checking if Read the Docs is getting our latest docs changes? Thanks.

tritemio:
@willingc, you are right. The latest RTD version is e122f061d8441618136d486edf759e50bbfe40a6, which is a commit  from August 19. All the recent build are failing:

https://readthedocs.org/projects/jupyter-notebook/builds/

RTD may have updated sphinx and other things that broke the docs building here.

Carreau:
@jdfreder @minrk or @takluyver can you make me admin on rtd for this repo ?

Carreau:
Also I think we could make @willingc admin on rtd too.

jdfreder:
I added @willingc and @Carreau : )  Let me know if you have trouble using it.

willingc:
@jdfreder Thanks :sunflower:

willingc:
I just added a PR #492 as a follow up to #488.

willingc:
Recommendation: refactor docs to use `nbgrader`'s model for autogenerating rst from ipynb. This would enhance maintainability of docs. See note #494

minrk:
@willingc good idea.

PR #492, opened 2015-09-24
Update for additional build information from Read The Docs

willingc:
Follow up on #488.

I've updated the `docs/requirements.txt` to install jupyter for docs builds on RTD and hopefully avoid RTD build from trying to drop back and run `setup.py` and `setup_base.py` which is causing build errors on RTD related to bower, npm, css/js. (Note: I'm hoping to avoid having to change either setup file.)

I've also updated `conf.py` with some debug code that should provide more info if RTD build fails.

cc/@minrk, @carreau

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,5 @@
+jupyter
+sphinx_rtd_theme
 jinja2
 tornado
 -e git+https://github.com/ipython/ipython_genutils.git#egg=ipython_genutils
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -20,12 +20,30 @@ import shlex
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('.'))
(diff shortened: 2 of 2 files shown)


Carreau:
Relaunched the tests. Mergin once green.

willingc:
@Carreau :green_book: : :green_apple: :green_heart:  Thanks for relaunching. It seems like that version 3.3/js is the one that likes to time out ;)

willingc:
@minrk @carreau Thanks. Partially fixed #479 with this PR. 

## Current Doc Status
The `latest` docs reflects the current docs (though examples/tutorials not ideal though accessible) with the doc updates on security, changelogs, etc.  https://jupyter-notebook.readthedocs.org/en/latest/

The `stable` docs reflect the old build which handles the examples/tutorials section better https://jupyter-notebook.readthedocs.org/en/stable/examples/Notebook/Examples%20and%20Tutorials%20Index.html

## Actions on RTD
This PR along with clearing the Advanced Settings in RTD (the checkbox and the requirements which was before doc/requirements.txt). Clearing the setup.py install checkbox allowed the build to complete without running `setup.py install --force` which was what was failing.

screen shot 2015-09-24 at 2 35 49 pm

## Next step
Try to update the latest build to fix the autogen of the notebooks on RTD without having to reenable the Advanced Setting checkbox.

PR #493, opened 2015-09-24
Allow all Cross-Origin Methods

blink1073:
Allows all REST API methods to be cross-origin, if explicitly requested with `--NotebookApp.allow_origin="*"`

### DIFF
--- a/notebook/services/api/handlers.py
+++ b/notebook/services/api/handlers.py
@@ -18,6 +18,13 @@ class APIHandler(web.StaticFileHandler, IPythonHandler):
         self.set_header('Content-Type', 'text/x-yaml')
         return web.StaticFileHandler.get(self, 'api.yaml')
 
+    @web.authenticated
+    def options(self, section_name):
+        self.set_header('Access-Control-Allow-Headers', 'accept, content-type')
+        self.set_header('Access-Control-Allow-Methods',
+                        'GET, PUT, POST, PATCH, DELETE, OPTIONS')
+        self.finish()
+
 default_handlers = [
     (r"/api/spec.yaml", APIHandler)
 ]
--- a/notebook/services/api/handlers.py
+++ b/notebook/services/api/handlers.py
@@ -19,7 +19,7 @@ class APIHandler(web.StaticFileHandler, IPythonHandler):
         return web.StaticFileHandler.get(self, 'api.yaml')
 
     @web.authenticated
-    def options(self, section_name):
+    def options(self, *args, **kwargs):
         self.set_header('Access-Control-Allow-Headers', 'accept, content-type')
         self.set_header('Access-Control-Allow-Methods',
                         'GET, PUT, POST, PATCH, DELETE, OPTIONS')
--- a/notebook/services/api/handlers.py
+++ b/notebook/services/api/handlers.py
@@ -18,13 +18,6 @@ class APIHandler(web.StaticFileHandler, IPythonHandler):
         self.set_header('Content-Type', 'text/x-yaml')
         return web.StaticFileHandler.get(self, 'api.yaml')
 
-    @web.authenticated
-    def options(self, *args, **kwargs):
-        self.set_header('Access-Control-Allow-Headers', 'accept, content-type')
-        self.set_header('Access-Control-Allow-Methods',
-                        'GET, PUT, POST, PATCH, DELETE, OPTIONS')
-        self.finish()
-
 default_handlers = [
     (r"/api/spec.yaml", APIHandler)
 ]
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -386,6 +386,13 @@ class APIHandler(IPythonHandler):
         self.set_header('Content-Type', 'application/json')
         return super(APIHandler, self).finish(*args, **kwargs)
 
+    @web.authenticated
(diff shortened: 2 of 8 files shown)


minrk:
This is explicitly reintroducing loads of XSS vulnerabilities, which I don't think we want.

blink1073:
Even though it is only turned available when `--NotebookApp.allow_origin="*"`?

minrk:
Ah, then maybe I misunderstand. If it only has an effect when allow_origin is already whitelisting, doesn't this do nothing at all?

blink1073:
When you send a cross-origin PATCH or DELETE request, it also sends an OPTION request.  This allows that transaction to take place.

blink1073:
I tested it on both this patch and on Kernel DELETE.

minrk:
Arg, I need to do more testing. I hate CORS more than just about anything.

If you want this to affect all API handlers, it should go on the base APIHandler class, rather than on every APIHandler subclass.

blink1073:
But not all of them support the same methods

minrk:
But CORS restrictions shouldn't be relevant to that. Is there an 'allow-methods: *'?

blink1073:
Sure, we could do that and then explicitly add OPTIONS to all of the SUPPORTED_METHODS

minrk:
By the way, I just checked and I can make cross-origin API requests with current master without this PR if I do `--allow-origin='*'`, so I'm not sure what this accomplishes.

blink1073:
With PATCH and DELETE methods?

blink1073:
It looks like `Access-Control-Allow-Methods` must be a comma separated, capitalized list.

minrk:
Aha! Now I understand, sorry. Default Access-Control-Allow-Methods is just get and post (of course, what I tried with). Adding all the verbs to APIHandler.options makes perfect sense. I understand now that this PR only makes the server behave exactly how I thought it already behaved.

blink1073:
So, add a full set of `options` to `APIHandler`, and then OPTIONS to all of the SUPPORTED_METHODS?

minrk:
@blink1073 yup. I don't think any of the current `options` handlers need to exist, and a single `options` handler on the base APIHandler should be the right thing.

minrk:
@blink1073 I think you can actually delete the `SUPPORTED_METHODS` attribute on all of the handlers, but I'll check on that.

blink1073:
I explicitly needed to add it to make it work.  We can add it to `APIHandler` instead.

minrk:
Yup, instead of adding it, you can remove SUPPORTED_METHODS from all the subclasses. The default value from RequestHandler should be fine.

blink1073:
Okay, overhaul coming right up.

notebook/services/api/handlers.py
```diff
@@ -18,6 +18,13 @@ def get(self):
         self.set_header('Content-Type', 'text/x-yaml')
         return web.StaticFileHandler.get(self, 'api.yaml')
 
+    @web.authenticated
+    def options(self, section_name):
```
minrk: I think you will need to accept `*args, **kwargs`  to deal with tornado's URL parsing

blink1073: Fixed, but I'm still getting rejected on OPTIONS requests.

blink1073:
Weird, the method doesn't get called if it is in the base class.  What kind of shenanigans is tornado pulling here?

notebook/services/api/handlers.py
```diff
@@ -18,6 +18,13 @@ def get(self):
         self.set_header('Content-Type', 'text/x-yaml')
         return web.StaticFileHandler.get(self, 'api.yaml')
 
+    @web.authenticated
+    def options(self, *args, **kwargs):
```
minrk: ah, sorry for the confusion. This is just a super-bad choice of names. This isn't be base APIHandler class, this is a weird, only partially implemented handler for providing the REST API spec. The base class is in base/handlers.py

blink1073:
Nevermind, I'm just dumb as usual...

blink1073:
Okay, this now passes my test suite that has DELETE and PATCH method calls.

blink1073:
Do you ever sleep @minrk?

minrk:
:+1: I'd love a test for it, but I think testing cross-origin behavior should be a separate project.

minrk:
(and yes, I happen to be in bed right now!)

blink1073:
good, you go to bed I'll go get dinner.  Thanks again, Min.

episodeyang:
@blink1073 thanks for fixing this! I just ran into this problem [here](https://github.com/ipython/ipython/issues/8849)

zischwartz:
The weird thing about this is the lack of `OPTIONS` should have broken Thebe (which always runs cross origin) long ago, but I'm only seeing it now.

episodeyang:
I think the only api method that needs this would be to halt the kernel. This is used in switching kernel types!

rgbkrk:
The API call that @zischwartz ran into with the new version of the notebook + thebe (built from 4.x js) was the initial creation of the kernel.

episodeyang:
I see! @zischwartz what API call is that? (method and end point). I just want to know so that I can keep it in mind

minrk:
Initial creation is POST, so should have been allowed without this patch.

PR #494, opened 2015-09-24
Temporarily add notebooks' generated rst files to version control

willingc:
Addresses #488 and #479 

This adds the autogenerated rst files to version control so that RTD can build without regenerating the rst files. I know that this is a change in workflow but I believe it will be temporary.


minrk:
Is there a reason to not reinstate building the rst docs on RTD? When did we stop doing that?

willingc:
@minrk The rst docs should be built on RTD. This PR makes the docs current for the end user while we troubleshoot why RTD was failing for the past month to build the docs correctly.

minrk:
@willingc fair enough

willingc:
Thanks @minrk. :boat: :sunrise: :ocean:

PR #495, opened 2015-09-25
Don't re-use APIHandler name for serving yaml spec

minrk:
the re-used name causes unnecessary confusion


blink1073:
:+1: as one who was confused by this.

PR #497, opened 2015-09-25
Replaced 'Search and Replace' with 'Find and Replace'

### LINKED ISSUE #452: Replace "Search and Replace" with "Find and Replace"
@ellisonbg recommends that we rename the "Search and Replace" with "Find and Replace" to maintain consistency across editors that use a similar language to describe that functionality. I can give this one a go.

cc/ @Carreau

captainsafia:
Closes #452.

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -194,7 +194,7 @@ define([
         });
              
         var id_actions_dict = {
-            '#search_and_replace' : 'ipython.search-and-replace-dialog',
+            '#find_and_replace' : 'ipython.find-and-replace-dialog',
             '#save_checkpoint': 'ipython.save-notebook',
             '#restart_kernel': 'ipython.restart-kernel',
             '#int_kernel': 'ipython.interrupt-kernel',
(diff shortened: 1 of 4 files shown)


jdfreder:
Line 185 of searchandreplace.js `.attr('id', 'search-and-replace')` should be `.attr('id', 'find-and-replace')`, and after doing that you'll need to update the less file - replace `#search-and-replace {` in `searchandreplace.less` with `#find_and_replace {`.  You can then rebuild the CSS by running `python setup.py css` in the notebook root IIRC

jdfreder:
(I would have left those as line comments, but I couldn't figure out how to do it since those lines don't show in your diff and GitHub seems to have disabled commenting on lines that don't show in the diff - lame!)

captainsafia:
@jdfreder — Solid!

damianavila:
LGTM

jdfreder:
Sweet!  Thanks @captainsafia

Jeffiy:
I think the file name ``searchandreplace.js`` ``searchandreplace.less`` should be also changed.

PR #502, opened 2015-09-26
Improve navigation to 'Jupyter' Installation docs

willingc:
Addresses #501 

Screen captures at #501 and additional detail on the current user experience and proposed changes.

* Add Jupyter to intersphinx in `conf.py` so jupyter docs can be easily referenced.
* Add URLs to 'Jupyter' RTD sections on Installation, Running the Notebook, and Migrating from IPython (note: unfortunately, Sphinx does not support using intersphinx references in the toctree)
* Update link in see also box to refer to Jupyter docs

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -3,10 +3,13 @@ The Jupyter notebook
 ====================
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 1
    :caption: User Documentation
 
    notebook
+   Installation <https://jupyter.readthedocs.org/en/latest/install.html>
+   Running the Notebook <https://jupyter.readthedocs.org/en/latest/running.html>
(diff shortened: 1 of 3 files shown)


docs/source/index.rst
```diff
@@ -3,10 +3,13 @@ The Jupyter notebook
 ====================
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 1
    :caption: User Documentation
 
    notebook
+   Installation <https://jupyter.readthedocs.org/en/latest/install.html>
```
minrk: Can intersphinx links work for this instead of hardcoded URLs?

willingc: @minrk Unfortunately, a limitation of Sphinx is that intersphinx links are not supported in toctrees.

willingc:
@minrk Thanks. :octopus:

Issue #395, opened 2015-09-05
List of running notebooks should indicate the kernel type


mjbright:
On the page which lists running Terminals and Notebooks, each running notebook is displayed with it's relative path.

It would be nice if the running kernel was indicated - a time saver if any "Untitled" notebooks are hanging around, for example.

hwine:
works for me in 4.1.0 - time to close?

jtpio:
Closing as fixed.

For reference this is what it looks like in Notebook 7:

![image](https://github.com/jupyter/notebook/assets/591645/c86405a5-3d7c-4629-a765-83d68373280b)

PR #503, opened 2015-09-28
Adds kernel name to running notebooks list

julienr:
This adds the kernel name to the running notebooks list, as requested in #395 .

I had to change the parameters passed to the `sessions_loaded.Dashboard` event to turn it from just the id to an object so it can store the id + kernel name (and maybe other things in the future). It *seems* like this does not break things, but if someone familiar with the js side can review this, that would be great.

jupyter_kernel_name

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -103,8 +103,10 @@ class MappingKernelManager(MultiKernelManager):
         """Return a dictionary of kernel information described in the
         JSON standard model."""
         self._check_kernel_id(kernel_id)
+        kernel_spec = self._kernels[kernel_id].kernel_spec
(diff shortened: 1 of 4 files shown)


mjbright:
Looks great !

Just a thought - and if "Untiltled.ipynb" notebooks were created as
"Untitled_<kernel>.ipynb" ?

minrk:
Nice! It seems like it should be the display_name not the short kernel name, though. Can you make that change?

@mjbright there are some technical reasons that would make the kernel name affecting the new notebook name not quite as simple as it sounds (the kernel name is not available when the notebook name is chosen), so I'd leave that for now.

julienr:
@minrk Changed that, looks better now :

jupyter_kernel_display_name

minrk:
@julienr thanks!

PR #505, opened 2015-09-28
[WIP] Added ConfigSection object to MarkdownCell.config

### LINKED ISSUE #454: CodeCell.config is config section MarkdownCell.config is a config object
This make it impossible to config on MD cell, and is inconsistent.

captainsafia:
Should close #454.

I'm opening this @Carreau because I've got a couple of questions. Can you explain to me the purpose of `mergeopt` in this? I understand that in this particular case, it is just returning the `default_options` associated with the `Markdown` class, but why is this done?

Otherwise, I would just do:

```
var config = options.config;
this.class_config = new configmod.ConfigWithDefaults(options.config, {}, 'MarkdownCell');
```

As is done in the other Cell classes.

Thanks!

### DIFF
--- a/notebook/static/notebook/js/textcell.js
+++ b/notebook/static/notebook/js/textcell.js
@@ -227,7 +227,8 @@ define([
          *          notebook: Notebook instance
          */
         options = options || {};
-        var config = utils.mergeopt(MarkdownCell, {});
+        var config = new configmod.ConfigSection('markdown',
+                utils.mergeopt(MarkdownCell, options.config, {}));
         this.class_config = new configmod.ConfigWithDefaults(options.config,
                                             {}, 'MarkdownCell');
         TextCell.apply(this, [$.extend({}, options, {config: config})]);
--- a/notebook/static/notebook/js/textcell.js
+++ b/notebook/static/notebook/js/textcell.js
@@ -227,8 +227,7 @@ define([
          *          notebook: Notebook instance
          */
         options = options || {};
-        var config = new configmod.ConfigSection('markdown',
-                utils.mergeopt(MarkdownCell, options.config, {}));
+        var config = new configmod.ConfigSection('markdown');
         this.class_config = new configmod.ConfigWithDefaults(options.config,
                                             {}, 'MarkdownCell');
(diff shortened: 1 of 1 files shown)


Carreau:
I'm running to a workshop so I might not be able to give you feedback before Thursday. 

Mergeopt basically allow to merge the configuration with the default of the the parent class, so that if you set the option on `Cell`, it affects `MarkdownCell`, `CodeCell`, `RawCell`, but if you **also** set the config on `MarkdownCell` the config in MarkdownCell override the value set in `Cell`, the the option dict bubble down to `Cell`, which  set an attribute on the instance which is the use.

This is highly historical behavior has we had many more cell types before, like HtmlCell and header 1-6. 
Such an abstraction layer might not be needed anymore.

captainsafia:
@Carreau, I think I get it. So should we also make sure that the config in `RawCell` is also a `ConfigSection` object?

Also, I think Travis needs to be kicked.

captainsafia:
Any updates on this @Carreau?

Carreau:
I'm still in conference until Thursday. Haven't had time to site an code.

notebook/static/notebook/js/textcell.js
```diff
@@ -227,7 +227,8 @@ define([
          *          notebook: Notebook instance
          */
         options = options || {};
-        var config = utils.mergeopt(MarkdownCell, {});
+        var config = new configmod.ConfigSection('markdown',
+                utils.mergeopt(MarkdownCell, options.config, {}));
```
takluyver: The second argument to ConfigSection is just for the `{base_url: '...'}` parameter. It doesn't take any defaults (that's what `ConfigWithDefaults` does).

I didn't get round to properly converting the cell options to the new config system, because it looked like it was going to be fiddly. As @Carreau mentioned, we also need to decide whether we still need config that the different cell types can inherit and override;  it makes things easier if we don't.

captainsafia:
@takluyver — Thanks for the feedback! Greatly appreciated.

I'll leave this here in case we decide to move forward with the config system.

Carreau:

PR #511, opened 2015-09-30
get kernelspec info from kernelspecs

minrk:
client-side, instead of modifying the sessions model to include display_name

This fixes recently-introduced test failures in master (my fault) from #503


minrk:
If Travis passes, I'll merge this so that master is back to working order. If need be, we can make any further changes in a new PR.

PR #512, opened 2015-09-30
shrink travis matrix

minrk:
we don't need to run all the js tests against all Python versions


blink1073:
Yeah, we're in *real* trouble if that becomes a concern...

damianavila:
>Yeah, we're in real trouble if that becomes a concern...

:stuck_out_tongue_winking_eye:

PR #521, opened 2015-10-01
Expose the default port of 8888

rgbkrk:
Seems like a reasonable default. We probably held back from doing this to not give people an unopinionated form of the original `ipython/ipython` image, but this is the raw notebook. With an `EXPOSE`d port, tmpnb can reason about it directly.

For Mr. @zischwartz


zischwartz:
:+1:

PR #524, opened 2015-10-03
Support TLS Client Authentication

### LINKED ISSUE #410: TLS Client Authentication
It would be nice to use TLS authentication instead of a password for authentication.

It should be a very simple change.  I'm not a python developer, but I got it working in an older version ipython and ported the change to today's master [1].  I'm submitting an issue instead of a pull request because I don't know how to install the package behind a firewall, so I don't know if it works.

[1] https://github.com/adambrewster/notebook/commit/bb4f614c44b2b3765e026463b40ceba8c4cf13b7

minrk:
patch provided in #410, but no PR was opened.

closes #410.


rgbkrk:
Thanks @minrk.

PR #527, opened 2015-10-05
Move cell toolbar selection to View menu

minrk:
It's not used enough to justify taking up a huge amount of space in the toolbar.

it's by far the biggest element in the toolbar, and rarely used. We should have a pretty high bar for putting any text-based controls in the toolbar, and I think cell type selection is probably the only thing that clears it at the moment.

### DIFF
--- a/notebook/static/notebook/js/maintoolbar.js
+++ b/notebook/static/notebook/js/maintoolbar.js
@@ -55,7 +55,6 @@ define([
             ],
             'run_int'],
          ['<add_celltype_list>'],
-         ['<add_celltoolbar_list>'],
          [['ipython.command-palette']]
         ];
         this.construct(grps);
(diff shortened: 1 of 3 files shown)


minrk:
cc @ellisonbg since it's a UI change, and @jhamrick and @damianavila as users of cell toolbars.

dsblank:
+1 from this user :+1:  I was even thinking that if you renamed "Raw NBConvert" to just "Raw" we could save some pixels. Need more room for extension buttons.

minrk:
I think it used to be `Raw`, but people are consistently confused when we call it just `raw`, since they don't understand that `raw` refers specifically to "raw nbconvert output". In the Menu, the pixels are no longer so precious, so I think it's okay to be a bit more verbose.

ellisonbg:
+1 from me

I don't think it is used enough to be in the toolbar.

jhamrick:
I think this will break some of the tests in nbgrader, but that's fine, I can fix them.

A more important comment though -- does this mean the event `preset_activated.CellToolbar` is going away? I definitely use that event in my extension, and so having that go away will break things there unless there is an alternative I'm forgetting about.

damianavila:
> I don't think it is used enough to be in the toolbar.

I think it is used... a lot of people are using slides and need this... but I am OK to have it in the menu... it is not a big deal for slides users to go to the menu instead of the toolbar and I agree that we need to clean the space and save pixel in the toolbar...

ellisonbg:
This behaves well in simple usability testing. But it sounds like we still need to understand its affect on the `preset_activated.CellToolbar` event.

minrk:
Sorry, I didn't notice the question the first time around. This has no effect on the preset_activated action. The only change is that I removed a *listener* from the preset_activated event because I didn't add an indicator for the currently selected item, since it's a bit harder to do in a menu than a select control, which always has a clearly indicated current selection.

ellisonbg:
Seems like this is ready for merge then?

jhamrick:
@minrk Oh, you're right, so you did. My mistake. Looks good then!

PR #528, opened 2015-10-05
optimize Dockerfile for reduced size

funkyfuture:
this is a spin-off of https://github.com/ipython/ipython/pull/8856

please see the commits about the changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,9 @@ RUN ln -s /usr/src/jupyter-notebook/scripts/lxc-launcher.sh /launch.sh \
  && python2 -m ipykernel.kernelspec \
  && python3 -m ipykernel.kernelspec
 
+VOLUME /notebooks
+WORKDIR /notebooks
+
 ENTRYPOINT /launch.sh
 
 EXPOSE 8888
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,6 @@ FROM ubuntu:14.04
 
 MAINTAINER Project Jupyter <jupyter@googlegroups.com>
 
-ENV DEBIAN_FRONTEND noninteractive
-
 # Not essential, but wise to set the lang
 # Note: Users with other languages should set this in their derivative image
 ENV LANGUAGE en_US.UTF-8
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -6,55 +6,69 @@ FROM ubuntu:14.04
...
+        texlive-latex-base \
+        texlive-latex-extra \
+        zlib1g-dev \
+ && rm -rf /var/lib/apt/lists/* \
+ \
+ `# Install the recent pip release` \
+ && curl -O https://bootstrap.pypa.io/get-pip.py \
+ && python2 get-pip.py \
+ && python3 get-pip.py \
+ && rm get-pip.py \
```
minrk: All these `&&` making the entire dockerfile one or two big steps makes building and debugging a *whole lot* more annoying. Can you quantify the savings of joining these lines, since the experience is so much worse?

funkyfuture:
re: https://github.com/jupyter/notebook/pull/528#discussion_r41237181

> All these && making the entire dockerfile one or two big steps makes building and debugging a whole lot more annoying. Can you quantify the savings of joining these lines, since the experience is so much worse?

yes, this is certainly annoying while testing. but for this, you can change to one layer per command. (regex substitution)

the guide on [Dockerfile best practices](https://docs.docker.com/articles/dockerfile_best-practices/#minimize-the-number-of-layers) doesn't actually give any explanation why it is a good practice. but you'll find it in every official image. ([example](https://github.com/docker-library/python/blob/15798abb6cfb145344462a345db4b572227fb859/3.4/Dockerfile))

the savings is less network-traffic and that depends on the connection. i guess Docker's overhead is slightly reduced when dealing with less layers.

alltogether the amount of layers is halfed, and that includes new layers like `VOLUME`. i haven't compared the layer's sizes.

i did this change from a user's perspective, seeking as few overhead when pulling and running on this image. and that should be achieved with images that are supposed to be used by the public.

minrk:
@funkyfuture if building it in one big step is really preferable, it seems a one-line Dockerfile that's just `bash build.sh` and then a less mangled `build.sh` script would be cleaner.

willingc:
@minrk @funkyfuture I don't believe the intent of the Dockerfile is to put all on one-line. It's to minimize issues with user introduced configuration errors (related to cache, updating/installing in same command line to avoid silent fails, and configuration maintainability)  see: [RUN documentation](https://docs.docker.com/reference/builder/#run) and RUN section of best practices (https://docs.docker.com/articles/dockerfile_best-practices/) .

Here's a file from one of the Docker devs as best practice: https://github.com/jfrazelle/docker/blob/master/Dockerfile

scripts/lxc-launcher.sh
```diff
@@ -0,0 +1,21 @@
...
+}
+
+trap _terminate SIGTERM
+trap _terminate SIGINT
+
+jupyter notebook --port 8888 --ip=* "${@}" &
+PID=$!
+
+wait $PID
+exit $?
```
rgbkrk: Certainly a fan of this. We're using `tini` in the docker stacks repo.

minrk:
@funkyfuture thanks for the PR and explanation! This does seem like an improvement, and we can iterate to find that balance discussed in the best practices.

willingc:
@minrk :surfer: :ocean: :octopus: Good decision. My explanation is feeling unloved. Just kidding :P :cake:

minrk:
@willingc thanks for your explanation as well!

funkyfuture:
sure, there is always a degree of gusto when it comes to optimizations.

especially the notebook installation should be kept in on `RUN` line as it keeps the build- and test-dependencies out of the image. they are removed from the fs before that layer gets commited.

scripts/lxc-launcher.sh
funkyfuture: i  tried to figure out what `tini` is w/o success. can you give me hint?

PR #529, opened 2015-10-06
Moved scrollbar on output area to .output_wrapper

### LINKED ISSUE #254: Vertical and horizontal scrollbar not both visible in toggled output
When displaying a large pandas DataFrame, you get some scrollbars. And starting from pandas 0.16.2, we leave the overflowing fully to the notebook itself (we removed `style="max-width:1500px;overflow:auto;"` from the html output, see https://github.com/pydata/pandas/pull/10232).

But I noticed another issue: when having a wide frame that does not fit the notebook width, you get horizontal scrollbar at the bottom of the output. When you also use the 'scroll output' feature, you get a vertical scrollbar as well. But the horizontal is only visible when you first scroll down to the bottom. To illustrate:

nb2

nb1

So if you want to see all columns of the first rows, you first have to scroll down, to be able to scroll right, then scroll up again to see the rows.

I suppose the one scrollbar belongs to the output itself, and the other to the toggled ouput area, but would it be possible to have them both on the toggled output area?

captainsafia:
Closes #254.

Tested on Safari 9, Firefox 41, and Chrome 42 on my Mac.

### DIFF
--- a/notebook/static/notebook/less/outputarea.less
+++ b/notebook/static/notebook/less/outputarea.less
@@ -4,6 +4,7 @@ div.output_wrapper {
     .vbox();
     // avoid scrolled output overlaying input in some strange circumstances
     z-index: 1;
+    overflow-x: auto;
 }
 
 /* class for the output area when it should be height-limited */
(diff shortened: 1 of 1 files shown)


captainsafia:
Closed because issue solved by #535.

PR #537, opened 2015-10-07
Docker image change and README update

funkyfuture:
this includes a change that makes local builds of a docker image faster if there are no changes to the prequisites.

i added instructions for using the Docker image and placed them rather on the top as it should be seen [at the image description](https://hub.docker.com/r/jupyter/notebook/).

i would also suggest that the docker registry repo is bound to another source repo as it is expected there to provide only stable releases or at least a stable release as default. atm only a reflection of the master's tip is available.
it would also allow to design the `Dockerfile` here for development purposes and to add image variants (i could make use of an onbuild-variant to include packages).
disclaimer: if this didn't change recently, the docker registry repo must be deleted and be created again to link it to another source repo. as a sideeffect all repo statistics go to void.

regarding `scripts/lxc-launcher.sh`, this still seems to need some work.

while a kill invoked with `docker kill` doesn't reach the timeout, it stops immediately w/o shutting down the kernels.
the wrapped process returns an exit code 137.

if this is tolerable, the script should transform that 137 to a 0 for the docker host.

furthermore i noticed that when it's running in a docker-compose setup and one is attached to their stdout, the container again exceeds the timeout when invoking a shutdown with <ctrl + c> (this sends a TERM to all containers).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,16 @@ RUN apt-get update -qq \
  && rm get-pip.py \
  \
  && pip2 --no-cache-dir install ipykernel \
- && pip3 --no-cache-dir install ipykernel
+ && pip3 --no-cache-dir install ipykernel \
+ \
+ && python2 -m ipykernel.kernelspec \
+ && python3 -m ipykernel.kernelspec \
+ \
+ && pip2 install --no-cache-dir mock nose requests testpath \
+ && pip3 install --no-cache-dir nose requests testpath \
+ && iptest2 && iptest3 \
+ && pip2 uninstall -y funcsigs mock nose pbr requests six testpath \
+ && pip3 uninstall -y nose requests testpath
 
 ADD . /usr/src/jupyter-notebook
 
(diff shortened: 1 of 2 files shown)


jakirkham:
What are your thoughts on running the container in interactive mode `-d` --> `-it`? This makes it very similar to running `ipython notebook` from the command line. This can be combined with `--rm` to ensure the container is removed after it is stopped (nice for beginners). With this sort of change, `--name` is less useful, but could still be used.

Also, what are your thought on publishing all exposed ports instead of providing the explicit mapping `-p 8888:8888` --> `-P`? As the `EXPOSE 8888` is already present, `-P` will do this mapping for us.

In short, I wonder if it wouldn't be nice to show a usage like this `docker run --rm -itP -v "$(pwd):/notebooks" jupyter/notebook` for local use. It is a nice analogue to `ipython notebook` (minus opening the browser for you).

Certainly there are use cases for all options already mentioned. I guess I am wondering about this as this is the notebook repo and for users wanting to get started with the notebook (missing needed dependencies) using `docker` like this allows for short startup time while avoiding many of the nitty gritty details about how `docker` works.

PR #538, opened 2015-10-07
lower case keyboard shortcuts for clarity

8leggedunicorn:
and cosmetic changes to the code documentation of static/notebook/js/quickhelp.js

The keyboard shortcuts are case sensitive, however each entry in the keyboard shortcut help list, found by navigating to `Help > Keyboard Shortcuts` are presented in capital letters.  This pull request is to display the help on these keyboard shortcuts with their proper case.  I realize that capital letters may have been consciously chosen for cosmetic appeal however for the sake of clarity I'd ask that the help reflect to the best degree possible the keybindings.

### DIFF
--- a/notebook/static/notebook/js/quickhelp.js
+++ b/notebook/static/notebook/js/quickhelp.js
@@ -30,7 +30,7 @@ define([
     var platform_specific;
 
     if (platform === 'MacOS') {
-        // Mac OS X specific
+        // Mac OS X (MacOS) specific
         cmd_ctrl = 'Cmd-';
         platform_specific = [
             { shortcut: "Cmd-Up",     help:"go to cell start"  },
@@ -66,10 +66,10 @@ define([
     ].concat( platform_specific );
     
     var mac_humanize_map = {
-        // all these are unicode, will probably display badly on anything except macs.
-        // these are the standard symbol that are used in MacOS native menus
-        // cf http://apple.stackexchange.com/questions/55727/
-        // for htmlentities and/or unicode value
+        // These are the standard MacOS native symbols for menus cf.
+        // http://apple.stackexchange.com/questions/55727/ for a map of HTML
+        // entities to their unicode values.  As these are unicode they will
+        // likely display poorly on all but MacOS.
         'cmd':'⌘',
         'shift':'⇧',
         'alt':'⌥',
@@ -127,7 +127,7 @@ define([
     
     function humanize_key(key){
         if (key.length === 1){
-            return key.toUpperCase();
+            return key;
         }
 
         key = humanize_map[key.toLowerCase()]||key;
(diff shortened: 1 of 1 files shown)


8leggedunicorn:
Some cosmetic changes to the code documentation in the modest hope to increase clarity were made in the pull request but if this is creating too much noise I'm more than willing to resubmit removing those changes.  The only relevant change is to line 130 of notebook/static/notebook/js/quickhelp.js:

https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/quickhelp.js#L130

minrk:
It's totally fine to make improvements to if you see them. cc @ellisonbg, @willingc for UI help change.

notebook/static/notebook/js/quickhelp.js
```diff
@@ -30,7 +30,7 @@ define([
     var platform_specific;
 
     if (platform === 'MacOS') {
-        // Mac OS X specific
+        // Mac OS X (MacOS) specific
```
minrk: Let's avoid MacOS. Spaces should always be between Mac, OS, X, and if we want a short name, OS X or Mac would be the right choice. It hasn't officially been "Mac OS X" since 2012.

notebook/static/notebook/js/quickhelp.js
```diff
@@ -127,7 +127,7 @@ define([
     
     function humanize_key(key){
         if (key.length === 1){
-            return key.toUpperCase();
+            return key;
```
minrk: If we aren't casting to upper case, we should probably cast to lower case. It's probably important that it be consistent.

8leggedunicorn:
@ellisonbg @willingc I'm requesting a small UI change in this pull request that the keyboard shortcuts help dialogue use lower case, to be consistant with the actual values, please see above.  If it makes sense please commit.  Cheers.

sccolbert:
IMHO bindings should be (have to be) case-insentive. 

For example, there is no way to type capital letter `A` with a single unmodified keystroke. What you do is type `shift-a`. That is to say, the case of a letter is a product of its numeric key code and a modifier. 

This means that the key binding `shift-+` is invalid because there is no way to type the `+` key without a modifier. That binding (on a US keyboard) would be expressed as `shift-=`. 

The [phosphor-keymap](https://github.com/phosphorjs/phosphor-keymap) repo went live at the end of last week, and @blink1073 is working towards a PR which uses that to implement these shortcuts. It supports arbitrary length chords as well as scoping via CSS selectors (ala Atom). It treats bindings in a case-insensitive manner. It is uses `+` as the separator though, since `+` is not a valid key character like `-` is.

With that technical stuff out of the way, I definitely think they should be presented to the user in the UI in a readable form. My personal preference for this is cap-cased modifiers, and capitalized primary keys.

`Ctrl-Shift-A`, `Alt-3`, `D D`, `Ctrl-Alt-F5`, etc...

8leggedunicorn:
@sccolbert I suspect one can use a caps lock key for what you were describing, although tbh I haven't used one of those in years.  You are right however that using the `Shift` modifier makes the case usage in the help dialogue irrelevant.  I'll close this PR.

sccolbert:
(Un)fortunately, caps lock does not effect the [keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) which is used to determine what key was actually pressed\*. But even if it did, it would then be impossible to distinguish `shift+a` from `shift+A`; one of those would have to be an invalid binding.

\* That's a deprecated API, but it's the only one which currently across all browsers. The keyboard API is probably the worst area for browser compatibility at the moment. The [DOM 3 standard](http://www.w3.org/TR/DOM-Level-3-Events/#widl-KeyboardEvent-key) (which fixes most of the issues) is not yet supported by Chrome.

8leggedunicorn:
On the first part I absolutely agreed which is why I closed the PR.   I'm really not too well versed in UI/UX but it's interesting.  I know how to change the code and I know if something initially rubs me the wrong way - I was too case sensitive this time, pun intended.  I expected there was a likely reason and there certainly is for using a consistent case, and agreed the `Shift` is and should be viewed as a modifier.

I remember wrestling with keyboard scancodes and what not in Linux (other *nix) and from that experience I'm certainly glad to leave it to others to work out the details for modern day browsers, the DOM 3 standard sounds promising.  Working on the computer has mostly meant the command line for me, but well adopting Jupyter into my workflow is and likely will continue to be a unique and beneficial experience.

PR #542, opened 2015-10-07
Try to clean up the Dockerfile a bit

jakirkham:
* Includes ( https://github.com/jupyter/notebook/pull/537 ).
* Put `&&` at the end of the previous line to avoid being too distracting.
* Clean up after the first use of `apt-get update` to, hopefully, cut down layer size.
* Break out installation of `pip` into a separate layer.
* Install some dependencies before adding the repo. (should help avoid cache invalidation until later)
* Breaks up installation and testing layers.
* Sum total layer size is not noticeably changed.
* Layers are better organized by steps.
* Appears to make a small speedup in build time (~30s).
* Maintains the same underlying behavior without adding any new features.
* Show how to start notebook with `docker` the same way one would without `docker`.
* Fixes an encoding issue that occurs with Python2 when there is no terminal connected to `sys.stdout` (e.g. building the image).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,8 @@ ENV LC_ALL en_US.UTF-8
 ENV PYTHONIOENCODING UTF-8
 
 # Python binary and source dependencies
-RUN apt-get update -qq \
- && DEBIAN_FRONTEND=noninteractive apt-get install -yq --no-install-recommends \
+RUN apt-get update -qq && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -yq --no-install-recommends \
         build-essential \
         ca-certificates \
         curl \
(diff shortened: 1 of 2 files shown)


jakirkham:
This is ready to go and could be included in a minor release, if desired.

funkyfuture:
> Break out installation of pip into a separate layer.

>Breaks up installation and testing layers.

what's the motivation for this?

> Clean up after the first use of apt-get update to, hopefully, cut down layer size.

this has no effect

jakirkham:
>> Break out installation of pip into a separate layer.
>> Breaks up installation and testing layers.

> what's the motivation for this?

Not to be mean, but I had a hard time figuring out what we were trying to do when first looking at the `Dockerfile`. I thought it would be better to have a couple more layers that are a bit clearer in their intent just so it could be understood more easily. Additionally, this should make diagnosing failure at a step a bit simpler. As the total image size ended up being the same and the build time improved, this doesn't seem to hurt.

Also, each layer is a point that someone could start a container from. If testing failed, one could try using the post-installation layer to startup and play around in the image to figure out what is causing the failure.

>> Clean up after the first use of apt-get update to, hopefully, cut down layer size.

> this has no effect

You're right. I hoped it would help the layer size. It didn't end up mattering.

funkyfuture:
then again, [i propose](https://github.com/jupyter/notebook/pull/537#issue-110129811) to maintain an extra repository for 'consumer' images. shall i give it a try?

jakirkham:
When you are referring to consumer images, do you mean like these ( https://github.com/jupyter/docker-stacks ) or something different? How else would you like to parameterize them?

Also, you had mentioned versioning. I suppose builds could be done for each tagged release on Docker Hub, as well. This would let one do `docker pull jupyter/notebook:4.0.0`, which does seem nice. The setup of this feature is less automated than we might like, but it may improve in the future. ( https://github.com/docker/hub-feedback/issues/165 )

funkyfuture:
oh, i wasn't aware of them. (initially i just went for `ipython/ipython` on the Docker hub and was then directed to `jupyter/notebook`.)

yes, that's conceptually what i mean. but these are not suited for me. i would just need pip and do the rest in a derived image.

PR #544, opened 2015-10-07
Nicer logging of error messages

jasongrout:
This lets us have clickable stacktraces on Chrome, for example.

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -756,7 +756,9 @@ define([
          */
         return function(error) { 
             var wrapped_error = new WrappedError(message, error);
-            if (log) console.error(wrapped_error); 
+            if (log) {
+                console.error(message, " -- ", error);
+            }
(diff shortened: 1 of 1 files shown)


jdfreder:
Awesome, it always helps to have better debugging capabilities.

Too bad we can't set custom traces in the Error, then we could get rid of the WrappedError class.

notebook/static/base/js/utils.js
```diff
@@ -756,7 +756,9 @@ define([
          */
         return function(error) { 
             var wrapped_error = new WrappedError(message, error);
-            if (log) console.error(wrapped_error); 
+            if (log) {
+                console.error(message, " -- ", error);
```
jdfreder: Is the `--` just for log aesthetics?

jasongrout:
-- is for aesthetics

jdfreder:
Tested this, it seems good.  Merging

Issue #546, opened 2015-10-08
Dockerfile CI support


jakirkham:
It would be nice if building of the Dockerfile was tested with CI. Potentially, this could include some usage tests of a container.

Due to the build time required, this probably should happen outside of the normal build matrix. It could also be done on a different CI service.

rgbkrk:
I enjoy that it's outside the normal build matrix. That's the other reason we have the webhooks to docker in place, as well as running iptest (though those have since been taken out of the image itself).

PR #547, opened 2015-10-08
Docker CI

jakirkham:
Fixes https://github.com/jupyter/notebook/issues/546.

Tries to setup CI with Circle CI for the Dockerfile only.

This is a start, but is a bit broken. I could use some help diagnosing a test failure on Circle CI. Appears to be due to an ASCII/Unicode issue.


minrk:
Makes sense to test these things. @rgbkrk can you look at this one?

rgbkrk:
Seems fine. I've started using circleCI in other projects, it's pretty good. Do you have a link to your current builds?

jakirkham:
Sure, here is the most recent build ( https://circleci.com/gh/jakirkham/notebook/9 ).

Here is the relevant error snippet. Looking at the test, what it does makes sense.

    Test group: utils
    ...............................................................................S........SSS...S.....S.S..S.E........S..S...............................
    ======================================================================
    ERROR: IPython.utils.tests.test_path.test_ensure_dir_exists
      File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
        self.test(*self.arg)
      File "/usr/local/lib/python2.7/dist-packages/IPython/utils/tests/test_path.py", line 434, in test_ensure_dir_exists
        path.ensure_dir_exists(d) # create it
      File "/usr/local/lib/python2.7/dist-packages/IPython/utils/path.py", line 440, in ensure_dir_exists
        if not os.path.exists(path):
      File "/usr/lib/python2.7/genericpath.py", line 18, in exists
        os.stat(path)
    UnicodeEncodeError: 'ascii' codec can't encode character u'\u2202' in position 25: ordinal not in range(128)

jakirkham:
There use to be more of these issues, but I added this line ( https://github.com/jupyter/notebook/blob/master/Dockerfile#L14 ), which sets `$PYTHONIOENCODING` to `UTF-8` and that removed all but this issue.

rgbkrk:
I'm a fan of merging once it passes, otherwise this will sit in error for a long time. We should only be testing the notebook here and I'm not sure why Python2 is being tested here as well.

jakirkham:
That sounds reasonable. If anybody is able to reproduce this failure outside of CircleCI, please let me know. At present, I can't seem to reproduce it.

jakirkham:
@rgbkrk, just so I follow, are you suggesting we drop Python 2 testing entirely?

I'd be ok with keeping it, but maybe we could just skip this test at present. It doesn't seem to affect building locally or on Docker Hub. It also completes successfully in a running container. Further, all the other python2 tests pass.

jakirkham:
Ha! I think I fixed it.

We will need to update iPython though ( https://github.com/ipython/ipython/pull/8886 ). I have included a patch in this PR, which is applied in the container for the interim.

If we need to keep the patch, it probably should be moved and renamed somehow.

rgbkrk:
If you rebase and rekick your build, does it bring in the freshly merged patch instead?

jakirkham:
I think it uses `pip` to install `ipython`. I could be wrong. Did you do a new release?

jakirkham:
I suppose we could install it via `git` instead.

jakirkham:
It failed without the patch.

So, at present I believe it is `pip` install `ipython`. I could try `pip` install using `git` `master` from `ipython` instead or perhaps even this commit.

What would you like me to do?

jakirkham:
So, `pip` installing was working. However, I am having some issues installing packages on Circle CI now.

jakirkham:
I suspect something is broken with the docker image `ubuntu:14.04`. There was a push about an hour ago.

jakirkham:
It turns out the there was a minimal (non-development) python installed. As a consequence, we were unable install the development dependencies that are required. I added a step that removes this development version of python first, which seems to resolve this issue.

jakirkham:
Build now passes ( https://circleci.com/gh/jakirkham/notebook/55 ).

jakirkham:
Build now passes ( https://circleci.com/gh/jakirkham/notebook/60 ).

jakirkham:
Based on feedback from @minrk, I have come up with an alternate fix for `ipython` in this PR ( https://github.com/ipython/ipython/pull/8887 ).

jakirkham:
Side note: I'm not sure whether you were aware of this, but repositories can be linked on Docker Hub (under Build Settings). This allows one to link it to the repository they are using as `FROM`. That way, when the upstream repository is pushed to or rebuilt it will trigger a rebuild of yours. This can be very handy in quickly diagnosing issues like the one I ran into last night. As the build errors, there will not be an image available to pull so you don't need to worry about somebody getting a broken image.

jakirkham:
I'd like to try and explain what is going on with the encoding issues.

On a working build, we are able to easily change the locale with environment variables and see the following.

    RUN python3 -c "import sys; print(sys.getfilesystemencoding())"
    utf-8

However, on Circle CI we are running into this bug ( https://bugs.launchpad.net/ubuntu/+source/langpack-locales/+bug/931717 ). As a result, we see this.

    RUN python3 -c "import sys; print(sys.getfilesystemencoding())"
    ascii

There is a workaround proposed here ( https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/969299/comments/6 ). Though, I don't believe we can take advantage of this. When talking with someone from Circle CI, they stated that the best thing to do would be to have an image with the locale settings workaround done. I went ahead and did that and put it in the image here ( https://github.com/jakirkham/docker_ubuntu_14_04_locale_fix ). This fixes the problem and makes the patch unnecessary.

With these changes the build will pass. ( https://circleci.com/gh/jakirkham/notebook/75 )

If you wish to take this repo and license it as you see fit, I have no problem. I only have the two constraints. It needs to be open sourced (I don't care BSD or whatever) and I would like credit for the work I did. Once the logistics of how this should be setup are addressed, I will cobble these changes back together in this PR.

rgbkrk:
Sorry to totally siderail, but does any of this issue come up with Ubuntu 15.04?

jakirkham:
I think I did a poor job of explaining. Sorry about that. Here's the TLDR:

The problem is an LXC AppArmor configuration issue on Circle CI, which blocks this encoding change.

The workaround is to do that step not on Circle CI (e.g. Docker Hub) and use it as our new base image.

jakirkham:
I should add building locally and on docker still result in working builds. This was built from my branch `docker_ci_3` ( https://hub.docker.com/r/jakirkham/notebook ).

jakirkham:
Maybe it would be ok if I merged this ( https://github.com/jakirkham/docker_ubuntu_14_04_locale_fix ) into ( https://github.com/jupyter/docker-stacks ). Thoughts @parente?

parente:
@jakirkham Sure. Can you send it as a PR against docker-stacks?

jakirkham:
Sorry for the long delay. I have added this PR ( https://github.com/jupyter/docker-stacks/pull/108 ). So that we can have a base image to use for CI here.

rgbkrk:
Ok cool. Now I see why we should just merge the other PR.

Dockerfile
```diff
@@ -62,7 +62,10 @@ RUN curl -O https://bootstrap.pypa.io/get-pip.py && \
     pip3 --no-cache-dir install requests[security]
 
 # Install some dependencies.
-RUN pip2 --no-cache-dir install ipykernel && \
+RUN pip2 --no-cache-dir install git+git://github.com/ipython/ipython.git@a58ab37122285adf8f79bd76ded718a777ae3a7c && \
+    pip3 --no-cache-dir install git+git://github.com/ipython/ipython.git@a58ab37122285adf8f79bd76ded718a777ae3a7c && \
+    \
+    pip2 --no-cache-dir install ipykernel && \
```
minrk: This should no longer need any changes to the Dockerfile, right?

jakirkham: Probably right. Did that commit end up in a release?

jakirkham:
Yeah, I think we couldn't think of anywhere else to put it. Unfortunately, we really need the locale to be UTF8, but CircleCI wasn't having it. So, setting it in a base image did the trick and allowed this image to behave the same with the added bonus of CI.

Dockerfile
minrk: Doesn't matter because the IPython tests aren't run in the Notebook dockerfile anymore

jakirkham: Oh, nvm. It has been awhile since I have looked at this. This was the commit that we reverted. So, yeah, definitely doesn't matter any more. Will remove.

jakirkham: Removed.

jakirkham:
Still needs the base image. Though, should we switch the base image to debian since this is already scheduled for 5.0? That way it can be the same as the other docker images. Maybe that should be another PR though as there is already a bit to keep track of here.

jakirkham:
Tests run for this branch ( https://circleci.com/gh/jakirkham/notebook/179 ) using the new base image in docker-stack.

Is this `TestLexers` failure just an issue caused by the new version of `pygments`?

minrk:
It's just because this isn't based on master. The IPython tests aren't run in this Dockerfile, but yes it's an inconsequential change in pygments causing an IPython lexer test to fail.

jakirkham:
Ah right, I forgot the tests got removed from the Dockerfile in master. Great. Thanks everyone.

minrk:
Thanks for your patience on this one.

rgbkrk:
:smile: :tada:

jakirkham:
Of course. Hope this helps. Probably should come up with something similar for docker stacks.

PR #530, opened 2015-10-06
Use `tini` in Dockerfile for zombie process reaping

jakirkham:
Follows from here ( https://github.com/jupyter/notebook/issues/334 ).

Contains this ( https://github.com/jupyter/notebook/pull/542 ).

PR #557, opened 2015-10-08
DOC: README docker example port option fix

jakirkham:
Switches back to the old form of port mapping as it appears the current form, unfortunately, maps ports randomly within a range. ( https://docs.docker.com/articles/networking/#binding-ports )


jakirkham:
@minrk, I made a small mistake here ( https://github.com/jupyter/notebook/pull/530 ). I have verified this does work as expected via the command line. Do you mind if we merge this?

PR #567, opened 2015-10-09
Fix tooltip and move replace in selected cells button.

### LINKED ISSUE #558: Search and Replace issues
First, the search and replace is _very_ nice!  I love the immediate feedback, especially.
In reviewing the search and replace, here are some issues I'm seeing:

Bugs/changes:
- [x] The joke in the regexp tooltip ("Now you have N+1 Problems") is funny to me, but not appropriate for this interface, I think.  On the other hand, it would be extremely helpful if the regexp tooltip said it was using javascript regexp syntax
- [x] selecting multiple cells, and then clicking Edit | Find and Replace deselects all but one cell.  Thus the "Replace in selected cells" is much less useful
- [x] "Replace in selected cells" checkbox actually also affects the find results, and so should be up with the case insensitive button (thanks to @ellisonbg for pointing this out)
- [x] ~~number of matches moved down next to the replace all button~~

Feature requests:
- [ ] I think it would be really useful to have filters for replacing only in code cells or only in markdown cells.

Visual style:
- [ ] Remove the blue highlight on the number of matches.
- [ ] Make the vertical spacing between the different elements more consistent. Rigth now there are at least 3-4 different spacings used.

jdfreder:
partially closes #558

### DIFF
--- a/notebook/static/notebook/js/searchandreplace.js
+++ b/notebook/static/notebook/js/searchandreplace.js
@@ -69,7 +69,6 @@ define(function(require){
       body.append($('<p/>').addClass('bg-warning').text("Warning, too many matches ("+html.length+"+), some changes might not be shown or applied"));
     } else {
       body.append($('<p/>').addClass('bg-info').text(html.length+" match"+(html.length==1?'':'es')));
-
     }
     for(var rindex=0; rindex<html.length; rindex++){
       var pre = $('<pre/>')
@@ -135,7 +134,7 @@ define(function(require){
         res.push([match.index, match.index+match[0].length]);
         escape_hatch++;
         if(escape_hatch > 100){
-          console.warn("More than  100 matches, aborting");
+          console.warn("More than 100 matches, aborting");
           abort = true;
           break;
         }
@@ -156,7 +155,7 @@ define(function(require){
       .attr('id', 'isreg')
       .addClass("btn btn-default")
       .attr('data-toggle','button')
-      .attr('title', 'use regular expression (now you have N+1 problems)')
+      .attr('title', 'Use regex (JavaScript regex syntax)')
       .text('.*');
 
     var onlySelectedButton = $('<button/>')
@@ -166,14 +165,14 @@ define(function(require){
         .addClass("fa fa-check-square-o")
       )
       .attr('data-toggle','button')
-      .attr('title', 'replace only in selected cells');
+      .attr('title', 'Replace only in selected cells');
 
     var isCaseSensitiveButton = $('<button/>')
       .attr('type', 'button')
       .addClass("btn btn-default")
       .attr('data-toggle','button')
       .attr('tabindex', '0')
-      .attr('title', 'is search case sensitive')
+      .attr('title', 'Case sensitive')
       .text('a≠A');
 
     var repl = $("<input/>")
@@ -188,20 +187,19 @@ define(function(require){
           $('<div/>').addClass('input-group')
           .append(
             $('<div/>').addClass('input-group-btn')
-              .append(isCaseSensitiveButton)
-              .append(isRegExpButton)
+                .append(isCaseSensitiveButton)
+                .append(isRegExpButton)
+                .append(onlySelectedButton)
           )
           .append(search)
         )
       )
       .append($('<div/>').addClass('form-group')
         .append(
-          $('<div/>').addClass('input-group')
-          .append(repl)
-          .append(
-            $('<div/>').addClass('input-group-btn')
-              .append(onlySelectedButton)
-          )
+          $('<div/>')
+              .addClass('input-group')
+              .css('width', '100%')
+              .append(repl)
         )
       )
       .append(body);
--- a/notebook/static/notebook/js/searchandreplace.js
+++ b/notebook/static/notebook/js/searchandreplace.js
@@ -195,12 +195,7 @@ define(function(require){
         )
       )
       .append($('<div/>').addClass('form-group')
-        .append(
-          $('<div/>')
-              .addClass('input-group')
-              .css('width', '100%')
-              .append(repl)
-        )
+        .append(repl)
       )
       .append(body);
 
--- a/notebook/static/notebook/less/searchandreplace.less
+++ b/notebook/static/notebook/less/searchandreplace.less
@@ -28,9 +28,4 @@
     max-height: 60vh;
     overflow: auto;
   }
-  
-  & input:last-child {
-    border-left:none;
-  }
-  
 }


notebook/static/notebook/js/searchandreplace.js
```diff
@@ -188,20 +187,19 @@ define(function(require){
...
       .append($('<div/>').addClass('form-group')
         .append(
-          $('<div/>').addClass('input-group')
-          .append(repl)
-          .append(
-            $('<div/>').addClass('input-group-btn')
-              .append(onlySelectedButton)
-          )
+          $('<div/>')
+              .addClass('input-group')
```
jasongrout: why not just append the repl, rather than appending the input group div, and then putting repl inside that?

jdfreder: That causes it to render incorrectly.

ellisonbg:
This is a good start to that. A few points:

After playing with it for a while, I find a couple of borders are missing. Here is a screenshot:

screen shot 2015-10-12 at 6 03 33 pm

The left border of the "Replace" text area and the right border of the selection checkbox.

jasongrout:
@jdfreder - I put in a PR to your PR.

@ellisonbg - I also saw that and I fixed that in my PR to @jdfreder.

jasongrout:
@jdfreder, I'm also seeing a 1-2 pixel difference between the buttons and the input area.  The buttons I think are sized to 32px, but the input area isn't, so there's a dissonance between the two.

notebook/static/notebook/js/searchandreplace.js
jasongrout: see my PR to yours, which does what I was suggesting (though I may have not been clear) and seems to render fine.

notebook/static/notebook/js/searchandreplace.js
```diff
@@ -166,14 +165,14 @@ define(function(require){
         .addClass("fa fa-check-square-o")
       )
       .attr('data-toggle','button')
```
ellisonbg: Let's also replace the faux checkbox button with a real checkbox!

jdfreder:
@jasongrout ah, that missing border is what I thought was a consequence of simplifying the border.  I didn't notice it was gone even with the code as is.  I merged your PR

notebook/static/notebook/js/searchandreplace.js
jasongrout: or change the icon to fa-square-o (to indicated a "selected" border)

jdfreder: Because the button is already a toggle button, I don't think it makes sense to add a checkbox too.  I'd either need to sync the check state with the button state or keep the checkbox static.  Not to mention, on some operating systems/browsers, check boxes aren't that pretty.

The `fa-square-o` idea is easy enough though, @ellisonbg would you like that?

Carreau:
-1 to move the checkbox. The re and case sensitive change the meaning of the text which is in the search field. The check box apply only to where it will apply. After using the SnR for some time, especially with the keyboard, I find out that I was selecting the "replace only in selected" only **after** entering the replace text. and having **after** the replace filed made it much more convenient to reach with tab. 

I was at the beginning with the other 2 button, and it was annoying. I moved it after **on purpose**.

ellisonbg:
Shouldn't the checkbox also affect which cells the search is done in. If
you want to do a replace, why would you search more widely that you know
you will replace on. Also, I think it would be very useful to be able to
"just search" on selected cells.

On Wed, Oct 14, 2015 at 2:39 PM, Matthias Bussonnier <

> -1 to move the checkbox. The re and case sensitive change the meaning of
> the text which is in the search field. The check box apply only to where it
> will apply. After using the SnR for some time, especially with the
> keyboard, I find out that I was selecting the "replace only in selected"
> only *after* entering the replace text. and having *after* the replace
> filed made it much more convenient to reach with tab.
>
> I was at the beginning with the other 2 button, and it was annoying. I
> moved it after *on purpose*.
>
> —
> <https://github.com/jupyter/notebook/pull/567#issuecomment-148151120>.
>



-- 
Brian E. Granger

Carreau:
Yes for "Just search", though harder to do that SnR, not sure about the UI for that. 

I think the browser Search is better for just search. 

From my personal use, I know I want to search and replace, it just append to replace in places where I am not expecting to find what I am searching so my workflow is the following: 

   - search for `Foo`
   - replace by `Bar`
   - oh it also replace in place I am not expecting. 
   - deselect where I do not want it to replace. 

Technically I would like a checkbox next to each preview, and be able to individually deselect some results (à la textmate). 

Hence why I placed the checkbox after as it is the last action I before validating.

ellisonbg:
I should clarify my feelings about the select button:

* First, let's make it a real checkbox rather than a
faux-checkbox-button--pretending-to-be-a-checkbox.
* Second, make sure it has a tolltip (probably does)
* Possibly move to the beginning of the replace so it has symmetry with the
other buttons being on the left.
* With these changes and your description, I think it would be fine to
leave below with replace.

On Wed, Oct 14, 2015 at 3:03 PM, Matthias Bussonnier <

> Yes for "Just search", though harder to do that SnR, not sure about the UI
> for that.
>
> I think the browser Search is better for just search.
>
> From my personal use, I know I want to search and replace, it just append
> to replace in places where I am not expecting to find what I am searching
> so my workflow is the following:
>
>    - search for Foo
>    - replace by Bar
>    - oh it also replace in place I am not expecting.
>    - deselect where I do not want it to replace.
>
> Technically I would like a checkbox next to each preview, and be able to
> individually deselect some results (à la textmate).
>
> Hence why I placed the checkbox after as it is the last action I before
> validating.
>
> —
> <https://github.com/jupyter/notebook/pull/567#issuecomment-148158592>.
>



-- 
Brian E. Granger

Carreau:
> * First, let's make it a real checkbox rather than a
> faux-checkbox-button--pretending-to-be-a-checkbox.

No objection. Design-wise, it might be less pretty, but fair enough. 



> * Possibly move to the beginning of the replace so it has symmetry with the
> other buttons being on the left.

it make keyboad usage a bit harder. I would prefer `tab`/`shift tab` to move quickly between `search` and `replace` field, if we put it before, it will be 2 press. We can change the tabindex, but then the tab selection order might be confusing. 


No objection to **try**, though I want to keep the keyboard usability in mind.

jdfreder:
> * First, let's make it a real checkbox rather than a
> faux-checkbox-button--pretending-to-be-a-checkbox.

I'm still slightly minus one on this for the reasons I stated above, which you may have missed:

Because the button is already a toggle button, I don't think it makes sense to add a checkbox too. I'd either need to sync the check state with the button state or keep the checkbox static. Not to mention, on some operating systems/browsers, check boxes aren't that pretty.

Carreau:
Merging to test it.

PR #572, opened 2015-10-09
Fixes the build on Docker Hub

jakirkham:
Broken out from my work here ( https://github.com/jupyter/notebook/pull/547 ). I found this was necessary to get building on the new `ubuntu` images given this issue ( https://github.com/docker-library/official-images/issues/1110 ). There appears to be a broken `python` install present, which blocks our first `apt-get` step and consequently every other step. This will simply remove it before proceeding.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,15 @@ ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
 ENV PYTHONIOENCODING UTF-8
 
+# Remove preinstalled copy of python that blocks are ability to install development python.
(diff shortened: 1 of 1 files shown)


jakirkham:
Verified building locally fixes the problem.

jakirkham:
attn: @minrk @rgbkrk

jakirkham:
Thoughts on merging this?

It's pretty straightforward. This does the bare minimum to fix the build on Docker Hub.

jakirkham:
It looks like the issue, which required this workaround has been resolved. I'll look into reverting this and see if the build still works.

PR #575, opened 2015-10-09
Docker add notebook profile

jakirkham:
Creates a default notebook profile.

Includes this ( https://github.com/jupyter/notebook/pull/573 ) and this ( https://github.com/jupyter/notebook/pull/572 ).


minrk:
What is the goal of writing an empty, unused config file here?

jakirkham:
I was going to add the `ip` argument to the profile. I have a `sed` command that does that in another branch. Is there a better way to do this? Also, what are your thoughts about adding this argument to the profile?

minrk:
`--generate-config` shouldn't be used in the Dockerfile. If we want a config file, we can write it directly.

PR #576, opened 2015-10-09
Specify ip via config file instead of command-line

jakirkham:
Ensures the `jupyter` notebook profile has the `ip` argument as `*` and drops it from the arguments in `CMD`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,10 +92,14 @@ RUN pip2 install --no-cache-dir mock nose requests testpath && \
     pip2 uninstall -y funcsigs mock nose pbr requests six testpath && \
     pip3 uninstall -y nose requests testpath
 
+# Add a notebook profile.
+RUN mkdir -p -m 700 /root/.jupyter/ && \
+    echo "c.NotebookApp.ip = '*'" >> /root/.jupyter/jupyter_notebook_config.py
+
 VOLUME /notebooks
 WORKDIR /notebooks
 
 EXPOSE 8888
 
 ENTRYPOINT ["tini", "--"]
-CMD ["jupyter", "notebook", "--ip=*"]
+CMD ["jupyter", "notebook"]


Dockerfile
```diff
@@ -80,10 +89,14 @@ RUN pip2 install --no-cache-dir mock nose requests testpath && \
     pip2 uninstall -y funcsigs mock nose pbr requests six testpath && \
     pip3 uninstall -y nose requests testpath
 
+# Add a notebook profile.
+RUN jupyter notebook --generate-config && \
```
minrk: This isn't the way to do do this kind of thing. If we want a config file, it should be written in its entirety, either with `echo ... >> path` or `ADD`. There is no need to do generate-config and then modify in-place.

jakirkham: Ok, so now I simply overwrite the file with echo. I still use `jupyter notebook --generate-config` to create the initial directory and file. Is that ok or would you like me to `mkdir` and the like instead?

jakirkham:
I think I have addressed the criticism I got here. Please let me know if I haven't.

jakirkham:
Travis failure is unrelated.

minrk:
`jupyter notebook --generate-config` should not appear in the Dockerfile. If you want to write the file directly, do so. There is no need to generate a config file and then delete it.

minrk:
You can use

    mkdir -p ~/.jupyter && echo 'line' >> ~/.jupyter/jupyter_notebook_config.py

without generating and truncating unused files.

jakirkham:
Ok, I think that fixes it then. Please let me know if there are more issues.

minrk:
@jakirkham can you squash the commits in this PR, perhaps?

jakirkham:
Sure, done. Anything else?

minrk:
Nope, that should do it. Thanks.

PR #588, opened 2015-10-11
Resolve some `InsecurePlatformWarning` when installing with `pip`

jakirkham:
Installs `requests[security]` to silence some `InsecurePlatformWarning`s. This won't work when installing `pip` via `get-pip.py` as that is due to a different issue ( https://github.com/pypa/pip/issues/2681 ).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ RUN apt-get update -qq && \
         git \
         language-pack-en \
         libcurl4-openssl-dev \
+        libffi-dev \
         libsqlite3-dev \
         libzmq3-dev \
         pandoc \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -55,7 +56,9 @@ RUN curl -L https://github.com/krallin/tini/releases/download/v0.6.0/tini > tini
 RUN curl -O https://bootstrap.pypa.io/get-pip.py && \
     python2 get-pip.py && \
     python3 get-pip.py && \
-    rm get-pip.py
+    rm get-pip.py && \
+    pip2 --no-cache-dir install requests && pip2 --no-cache-dir install requests[security] && \
```
minrk: This should be only `requests[security]`, you don't need to install the package and then the same package again.

jakirkham: You're right. I was having some issue where I was only seeing requests in `python3` not `python2`. Regardless, this had nothing to do with it. I will remove it.

jakirkham:
Fixed. Please let me know if you need anything else. Builds through locally.

PR #614, opened 2015-10-16
Fix #581 to prevent focusing on a code cell with double click.

### LINKED ISSUE #581: Double clicking cell enters edit mode
If I double click a cell output, I enter edit mode, with focus in the cell input. This only works the first time. Subsequent double clicks just select the word in the output.

Is this the intended behavior?

I always try to select a word in the output with double clicks. It seems kind of silly to enter edit mode with a double click when you can just reach above a bit and single click.

thachmai:
The default behavior of a cell is to focus to the editor after a double click. This fix adds a simple check to disable the behavior for code cell.

Would it be better to move the focus behavior explicitly into text_cell instead?

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -201,8 +201,9 @@ define([
             if (that.selected === false) {
                 this.events.trigger('select.Cell', {'cell':that});
             }
+
             var cont = that.unrender();
-            if (cont) {
+            if (cont && that.cell_type !== 'code') {
                 that.focus_editor();
             }
         });


jdfreder:
I think it would make more sense to move the logic into text_cell, because it probably shouldn't fire for raw_cell either.

@ellisonbg @Carreau @minrk was this an intentional behavior for the code cell?  I agree with @thachmai and prefer it without this behavior.

thachmai:
I submitted another pull request that implemented this behavior explicitly in MarkdownCell instead: https://github.com/jupyter/notebook/pull/616

jdfreder:
Closing in favor of #616

PR #628, opened 2015-10-21
Bump upload size to 25MB

blink1073:
Tested with a 24MB file in Firefox and Chrome on OSX.

screen shot 2015-10-21 at 9 12 00 am

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -46,6 +46,7 @@ define([
                 function(e, d) { that.sessions_loaded(d); });
         }
         this.selected = [];
+        this._max_upload_size_mb = 25;
     };
 
     NotebookList.prototype.style = function () {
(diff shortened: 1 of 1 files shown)


minrk:
+1

dsblank:
Should this limit be configurable? I could imagine that one might want to make the value larger, or smaller, for a set of users.

blink1073:
This is a temporary work-around until we implement streaming binary uploads.

dsblank:
Ok. But even in the future (Oh, it is Back to the Future just now!), I could imagine that the system has no set limit, but it could be configured so that there were one.

minrk:
@blink1073 could you put the constant as a class-level private attribute, just so that it's a bit more easily overridable in a pinch?

blink1073:
Better?

minrk:
:+1: thanks!

dsblank:
Cool! Thanks for this!

PR #639, opened 2015-10-21
revert style change

Carreau:
Also make the marked color a variable.

### DIFF
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -24,11 +24,9 @@ div.cell {
     }
     
     &.marked {
-        border-left-color: #009AF5;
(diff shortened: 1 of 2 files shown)


jdfreder:
If we are going to revert the input color change suggested by @cameronoelsen, we may want him to instead pick a darker marker color.  cc @ellisonbg

Carreau:
I'm not sure it is worth bikeshedding on a color that will be changed if we redo the color scheme for 5.0

cameronoelsen:
Yeah I think thats probably a good idea by @Carreau unless we want a new color. I want to, in a future iteration, get a better color palette that works well in the input cells and the website itself (along with the orange in the jupyter logo)

jdfreder:
Makes sense

PR #643, opened 2015-10-22
Force pager to draw scrollbars on FireFox

jdfreder:
Here's a less horrible hack, that looks uglier, which fixes the pager scrollbar issue in FireFox.

closes ipython/ipython#8853

@ellisonbg


Carreau:
works as advertise and does not seem to break anything on Chrome. 

Thanks !

jdfreder:
Thanks @Carreau

PR #663, opened 2015-10-26
Interpret mathjax_url relative to base_url

jdemeyer:
It should be possible to give a `mathjax_url` relative to the current `base_url`. With this pull request, this can be done by specifying a relative path for `mathjax_url`. Given that relative paths are currently useless, this is unlikely to break existing usage.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -32,7 +32,7 @@ from ipython_genutils.path import filefind
 from ipython_genutils.py3compat import string_types
 
 import notebook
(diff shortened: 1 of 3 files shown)


minrk:
@jdemeyer what relative URLs are valid?

minrk:
(Merging, since this is the right thing to do anyway, I'm just wondering what prompted it).

jdemeyer:
For SageMath, we need a ''full'' version of MathJax (Jupyter's version is not sufficient). With this branch, we can use `nbextensions` for this.

minrk:
ok, thanks.

PR #668, opened 2015-10-26
call fixConsole on raw_input prompt

minrk:
handles ANSI escapes


jdfreder:
Hmm, I see the color coding on the prompt, but it disappears once I submit.

Action shot:

colorinput

jdfreder:
If fixing the problem I mentioned above is too much effort for 4.1, we should just merge this as-is as a partial fix.

PR #669, opened 2015-10-27
UI prototype to show hidden state

Carreau:
a prototype to have less hidden state when using marked cell. 

Change the color of the **selection** when cell are marked:

ui-select-prototype

ping @ellisonbg @jdfreder @jhamrick

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -299,6 +299,7 @@ define([
                     this.element.removeClass('marked');
                 }
             }
+            this.events.trigger('cell:marking-changed', {cell:this, value:value});
         }
     });
 
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -186,6 +186,11 @@ define(function (require) {
     Notebook.prototype.bind_events = function () {
         var that = this;
 
+        this.events.on('cell:marking-changed', function(){
+            console.log('trigger');
+            that.update_marked_status();
+        });
+
         this.events.on('set_next_input.Notebook', function (event, data) {
             if (data.replace) {
                 data.cell.set_text(data.text);
@@ -621,6 +626,7 @@ define(function (require) {
     Notebook.prototype.toggle_cells_marked = function(cells) {
         cells = cells || this.get_cells();
         cells.forEach(function(cell) { cell.marked = !cell.marked; });
+        this.update_marked_status();
     };
     
     /**
@@ -630,6 +636,7 @@ define(function (require) {
     Notebook.prototype.mark_all_cells = function(cells) {
         cells = cells || this.get_cells();
         cells.forEach(function(cell) { cell.marked = true; });
+        this.update_marked_status();
     };
     
     /**
@@ -638,6 +645,7 @@ define(function (require) {
      */
     Notebook.prototype.unmark_all_cells = function(cells) {
         this.get_marked_cells(cells).forEach(function(cell) { cell.marked = false; });
+        this.update_marked_status();
     };
     
     /**
@@ -728,7 +736,17 @@ define(function (require) {
         var selectedIndex = Math.min(Math.max(this.get_selected_index() + offset, 0), this.ncells()-1);
         this.select(selectedIndex);
         this.get_selected_cell().marked = true;
+        this.update_marked_status()
     };
+
+    Notebook.prototype.update_marked_status = function(){
+        var indicies = this.get_marked_indices();
+        if ( indicies.length === 0){
+            this.element.removeClass('jp-marking');
+        } else {
+            this.element.addClass('jp-marking');
+        }
+    }
 
     // Cell selection.
 
@@ -1018,6 +1036,7 @@ define(function (require) {
         this.set_dirty(true);
 
         return this;
+        this.update_marked_status();
     };
 
     /**
@@ -1061,6 +1080,7 @@ define(function (require) {
             this.undelete_index = null;
         }
         $('#undelete_cell').addClass('disabled');
+        this.update_marked_status();
     };
 
     /**
@@ -1484,6 +1504,7 @@ define(function (require) {
             new_cell.set_text(texta);
             new_cell.marked = cell.marked;
         }
+        this.update_marked_status();
     };
 
     /**
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -39,6 +39,19 @@ div.cell {
 
     }
 
+    .command_mode  &.selected {
+
+        ._marked_style(3px);
+
+        &.marked {
+          ._marked_style(4px);
+        }
+
+
+
+    }
+
+
 
     .edit_mode &.selected {
         border-color: green;
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -14,6 +14,16 @@ div.cell {
         }
     }
 
+    .command_mode .jp-marking &.selected {
+        border-color: @marked_border_color;
+
+        /* Don't border the cells when printing */
+        @media print {
+            border-color: transparent;
+        }
+    }
+
+
     .edit_mode &.selected {
         border-color: green;
 
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -1,3 +1,10 @@
+._marked_style(@n) {
+
+    border-left-color: @marked_border_color;
+    border-left-width: @n;
+    padding-left: 6px-@n;
+}
+
 div.cell {
     border: 1px solid transparent;
     .vbox();
@@ -21,6 +28,15 @@ div.cell {
         @media print {
             border-color: transparent;
         }
+
+        ._marked_style(3px);
+
+        &.marked {
+          ._marked_style(4px);
+        }
+
+
+
     }
 
 
@@ -34,9 +50,7 @@ div.cell {
     }
     
     &.marked {
-        border-left-color: @marked_border_color;
-        border-left-width: 3px;
-        padding-left: 3px;
+        ._marked_style(4px);
     }
 
     width: 100%;
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -14,7 +14,7 @@ div.cell {
     border-style: solid;
 
     &.selected {
-        border-color: @border_color;
+        border-color: transparent;
         /* Don't border the cells when printing */
         @media print {
             border-color: transparent;
@@ -28,11 +28,11 @@ div.cell {
             border-color: transparent;
         }
 
-        ._marked_style(3px);
+        ._marked_style(2px);
 
         &.marked {
-          border-color: @marked_border_color;
-          ._marked_style(3px);
+          border-left-color: @marked_border_color;
+          ._marked_style(4px);
         }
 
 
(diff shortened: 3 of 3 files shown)


jhamrick:
I like this. Here's another thought: what if we also actually made a "mark" on the left side for any cell that is selected, if there are other selected cells? So that it "looks" like it's marked as well? It wouldn't actually be marked -- e.g. if you moved the selection up or down, it would move the mark (unless the cell were legitimately marked). But the benefit of doing this would be to allow us to operate on the union of marked + selected cells, and it wouldn't be surprising to users because the selected cell looks like it's marked.

jdfreder:
Related to the discussion in #667


> Here's another thought: what if we also actually made a "mark" on the left side for any cell that is selected, if there are other selected cells? So that it "looks" like it's marked as well? It wouldn't actually be marked -- e.g. if you moved the selection up or down, it would move the mark (unless the cell were legitimately marked). But the benefit of doing this would be to allow us to operate on the union of marked + selected cells, and it wouldn't be surprising to users because the selected cell looks like it's marked.

I like this idea a lot +10

jhamrick:
Actually, now that I think of it even more, why not just always have the selected cell look like it's marked, even if there aren't other marked cells? Then it's less surprising to see additional marked cells when you do things like multi-cell select, and it's more consistent (when you delete a cell, you always delete all marked cells, which might mean multiple cells or just the currently selected cell).

jdfreder:
> why not just always have the selected cell look like it's marked, even if there aren't other marked cells?

+1 , you're on a roll, I like these ideas

jhamrick:
Thanks! :smile:

jdfreder:
Question, what styling would we use?  The left border alone?  Or the full gray border?  I should say, if we do the full gray border, it looks ugly with multiple consecutive cells marked.

ellisonbg:
I also like these ideas. But if we visually style the selected cell like a
marked cell, won't that be confusing for actions that only work on
selected? (I am not sure, just asking)

On Tue, Oct 27, 2015 at 1:35 PM, Jonathan Frederic <notifications@github.com
> wrote:

> why not just always have the selected cell look like it's marked, even if
> there aren't other marked cells?
>
> +1 , you're on a roll, I like these ideas
>
> —
> <https://github.com/jupyter/notebook/pull/669#issuecomment-151580958>.
>



-- 
Brian E. Granger

jdfreder:
I think the idea is that no actions would only work on selected

jhamrick:
I think the left border alone is fine. I like Matthias' gif and basically am thinking that the selector would still have the gray border but additionally the left blue border.

jhamrick:
Are there currently any actions that only work on selected cells and not marked cells?

Carreau:
> Are there currently any actions that only work on selected cells and not marked cells?

Run cell. 

I'll update the CSS to have the cursor look like marked cell.

jhamrick:
Ah, hmm. I could see it being useful though to say "run this set of cells that I have marked" rather than just run the current cell, so maybe that's not a problem to have run cell run all the marked cells...

Carreau:
mark-2

Carreau:
I introduced a small difference between `selected` and `selected + marked`

jhamrick:
Nice! I hadn't thought about the difference between selected and selected+marked, but yeah, I think that's important. I wonder if the way you have it might be too subtle, though. What about having:

1. Always a gray border for the selection, *except* it's blue if the cell is truly marked
2. Always have the blue bar on the left with the selection, even if there are no other marked cells

Carreau:
I like a small difference when selection is the only marked cell, as it give you a quick indication wether there **migt be** off screen marked cell. 

I'll finish my version, and play with your 1) and 2)

jdfreder:
I would like to see \#2 especially

Carreau:
you can check latest commit to try #2

willingc:
A gentle suggestion for accessibility would be to make sure the indicator does not solely rely on color. I like 2.

Carreau:
I've pushed proto1 and proto4

proto4 rely on border width:
 - selected : 2px
 - marked : 3px
 - selected + marked : 4px

proto4

jhamrick:
* :-1: for proto1 (can't tell the difference between selected and marked+selected)
* :+1: for proto2
* :+1: for proto4 (I actually quite like it without the border)

I do think we are going to need an additional indicator besides width, though, as I still think it's too subtle. I'm not sure what would be best, though. Maybe a combination of color and width (would that also address the stuff you're thinking of, @willingc ?) -- what about

* selected, not marked: thin blue
* not selected, marked: medium gray
* selected, marked: thick blue

or something along those lines?

jhamrick:
Here's an example of what I mean, using light blue rather than gray:

jhamrick:
The more I play around with the version that doesn't have borders, the more I like it. I think this will have the added benefit of making it much, much clearer when you're in command mode vs edit mode. So +100 from me for something along the lines of proto4!

jdfreder:
I've never liked the gray border, so I second @jhamrick 's +1(00)

willingc:
Another +1 for proto4. @jhamrick Yes, color plus width works for me.

Carreau:
I like the no border too, Though there is also the "edit-mode" green, which has boder. 
and I don not want to disturb too much people with changing UI habits. 

So while I'm +1 to removing border for 5.0 I'm -1 to remove it for 4.1

michaelpacer:
I'm a fan of the different color from the selection and the marked cells. 

One thing I just talked about with @Carreau was that the real problem that is still not adequately addressed is knowing which of the two (selected vs. marked cells) is going to have actions applied to it. 

One possibility is to flip the problem on its head and rather than thinking of the mark as a state, selected as a state and marked & selected as a state, is to recognize that marks+selected structure is a set of binary variables applied to the complete set of cells. 

If you instead think of the selected as an implicit mark (which @Carreau said was already part of the discussion) you can treat the entire problem differently and address the "when you *do* something, what are the things that something is done to" problem by putting a priority queue on the sets of binary variables. The set at the top of the queue is given its own special UI indicator (I was suggesting something that contrasts with the actual selection colors like red, though I know that might have other bad connotations), and it is that set of cells to which the function will be applied. 

The idea let's you choose to step through the queue by using some kind of toggle/step-through key (based on gmail's mechanism for stepping through different inboxes I suggested  <code>`</code> , but then @Carreau let me know that that won't work for non-US keyboards :-1:), and that will toggle which cells will have actions applied to them as a separate idea from the status of marked, selected or marked and selected.

sidenote: I'm assuming that actions like merging are just going to be disabled for noncontiguous blocks of cells? Otherwise the semantics of what it is to merge get wonky…

jhamrick:
I'm not sure I understand the idea of having a priority queue. The way that I've been thinking about it is:

* With the exception of switching into edit mode, and marking/unmarking cells, all other actions are only ever applied to marked cells. The state of being marked is indicated by a bar to the left of the cell.
* In order to choose a cell that you want to edit, or to mark/unmark individual cells, you have a cursor. If the cursor is on a cell, then that cell is counted as marked and thus has a bar to the left of it as well.
* However, when the cursor is on a cell, while that cell is counted as marked, it is not a "sticky" mark, e.g. it will go away if you focus on another cell. So you can explicitly mark the cell, but then you need a way of visually distinguishing cells that are focused that have do or do not have a sticky mark.

So, essentially, it should be possible to determine from a screenshot (without knowing the history of actions) which of the following categories each cell falls under:

* Focused/selected (with a "sticky" mark)
* Focused/selected (without a "sticky" mark)
* Not selected and marked
* Not selected and unmarked

jhamrick:
Also, it may be helpful here to note the similarity to how selection in emacs works. I will make this quick because I have to run to class in a minute...

* There is a cursor (overlaps a single character)
* There is a region (overlaps multiple characters and includes the cursor)
* All actions operate on the region
* You can tell the difference between the cursor and the region because they are usually different colors, and the cursor usually blinks

jdfreder:
> I like the no border too, Though there is also the "edit-mode" green, which has boder. 
> and I don not want to disturb too much people with changing UI habits.
> 
> So while I'm +1 to removing border for 5.0 I'm -1 to remove it for 4.1

Sorry @Carreau , here I disagree with you and I say away with it for 4.1! :)  (not the edit mode one)

blink1073:
Having the width of the side denote the different states is confusing to me.  What if the `selected` cell actually did blink its border?

michaelpacer:
@jhamrick, I think we're talking past each other, because that division works for distinguishing between editing and command modes. But before what was discussed was the notion of selection being an implicit mark that is differentiated from the notion of what I think you're calling a sticky mark. 

I thought that focused was the term for when you are in edit mode in a particular cell.

In the previous release there was the notion of selecting multiple cells, and so something could be selected and not focused (because you can select multiple cells, but you cannot be _focused_ on multiple cells for the purposes of editing those multiple cells, unless — in the emacs analogy — it's allowing multiple cursors). 

So there are actually 8 states:

Focused/in edit mode (with "sticky" mark)
Focused/in edit mode (without "sticky" mark)
Selected/in command mode (with "sticky" mark)
Selected/in command mode (without "sticky" mark)
Not selected and not focused, in command mode (with "sticky" mark)
Not selected and not focused, in command mode (without "sticky" mark)
Not selected and not focused, in edit mode in another cell (with "sticky" mark)
Not selected and not focused, in edit mode in another cell (without "sticky" mark)

What I was taking as given was that operations could only be applied while in command mode (I think this is implicit in all the discussion, and I have no quibble with that). 

But when in command mode, one could have a selection on many contiguous cells and one could have marks on any subset of the cells. What I was suggesting is the ability to (at least) be able to toggle between the selected cells as the target of actions and the marked cells as the target of actions without losing either the marks or the selection when that toggle occurs(I.e., you can toggle back and forth). The UI indicator for the currently active target set is an open question (and I am open minded as to what that should look like). 

Basically, what's inevitably going to happen in the examples @jhamrick gives in an issue elsewhere(https://github.com/jupyter/notebook/issues/665#issuecomment-151339333) will be cases someone applies an operation to the wrong target set, and it should be possible to undo that operation, to and to immediately return the set of focuses to the place they were before, easily toggle to the correct target set and then apply the operation again. If you don't have some way of encoding that information so as to make it easy to undo to the state where both the marks and selections are preserved, switch targets, and redo; people(read: this includes at least me) will be less likely to use marks as opposed to the contiguous selection mechanism because then I'll at least know where all of the changes were made when misfires occur and apply to the wrong target set (which currently is being implicitly decided by…I'm not sure what?).

Carreau:
I give up on UI. Prototypes are here feel free to take inspiration. 

I shouldn't have gotten into this discussion and I should have left the selected/marked things in the state it was.

jdfreder:
> Having the width of the side denote the different states is confusing to me. What if the selected cell actually did blink its border?

That's a cool idea, like a nod to the design of the common cursor.  Just thinking out loud, it may get noisy seeing both the cell side-bar blink and the cursor in the editor.

blink1073:
Don't despair @Carreau!, thanks for the prototypes.

jdfreder:
> Don't despair @Carreau!, thanks for the prototypes.

+1 , I think we are getting close!

blink1073:
Maybe only blink in command mode?

jdfreder:
Yeah I think that would work

ellisonbg:
-1 on any blinking ;-) except of eyes

On Tue, Oct 27, 2015 at 6:03 PM, Jonathan Frederic <notifications@github.com
> wrote:

> Yeah I think that would work
>
> —
> <https://github.com/jupyter/notebook/pull/669#issuecomment-151658751>.
>



-- 
Brian E. Granger

willingc:
FYI on blinking and accessibility: https://pages.18f.gov/accessibility/flashing/

PR #675, opened 2015-10-28
Marked cells styling

jhamrick:
Continuing from #669 (which @Carreau did an awesome job on!)

Ok, first let me summarize the discussion from #669 and #667. I am also going to define a few terms:

* **marked** cells are cells that will have actions applied to them in command mode. If a cell is marked, it will have the action applied to it; if it is not marked, it will not have any action applied to it.
* The **selected** cell is the cell that the user has currently selected (using the arrow keys, or `j` and `k`). The selected cell is always counted as a marked cell while it is selected. However, if the user changes the selection to a different cell, the original selected cell will no longer be counted as marked unless the user explicitly said they wanted to mark it.
* A **sticky mark** is one that the user has explicitly applied to a cell.
* An **implicit mark** is the one that is applied to the currently selected cell.

In 99% of cases, the user will just be operating on cells by selecting one at a time (i.e. using the implicit mark). In other cases, they will operate on multiple cells, which are marked. So, it is necessary to visually distinguish:

* cells that are marked, but not selected (medium weight, light blue border)
* cells that are unmarked, and not selected (no border)
* cells that are selected, but which do not have a sticky mark (light weight, blue border)
* cells that are selected and also which have a sticky mark (heavy weight, blue border)

To clarify which actions apply to the selected cell and which apply to the marked cells: all actions apply to marked cells. I think it is more useful to think of the selected cell as a cursor, rather than as a different type of cell that can be operated on -- actions do not apply to the selected cell, they apply to marked cells, and it just is the case that the selected cell is also always a marked cell.

Note that this prototype does *not* get around the issue of having marked cells off screen. I have an idea for how to address this though and will submit another PR for that.

Here is a gif of how this prototype looks:

@jdfreder @ellisonbg @willingc @michaelpacer @blink1073

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -186,6 +186,11 @@ define(function (require) {
     Notebook.prototype.bind_events = function () {
         var that = this;
 
+        this.events.on('cell:marking-changed', function(){
+            console.log('trigger');
+            that.update_marked_status();
+        });
+
         this.events.on('set_next_input.Notebook', function (event, data) {
             if (data.replace) {
                 data.cell.set_text(data.text);
@@ -621,6 +626,7 @@ define(function (require) {
     Notebook.prototype.toggle_cells_marked = function(cells) {
         cells = cells || this.get_cells();
         cells.forEach(function(cell) { cell.marked = !cell.marked; });
+        this.update_marked_status();
     };
     
     /**
@@ -630,6 +636,7 @@ define(function (require) {
     Notebook.prototype.mark_all_cells = function(cells) {
         cells = cells || this.get_cells();
         cells.forEach(function(cell) { cell.marked = true; });
+        this.update_marked_status();
     };
     
     /**
@@ -638,6 +645,7 @@ define(function (require) {
      */
     Notebook.prototype.unmark_all_cells = function(cells) {
         this.get_marked_cells(cells).forEach(function(cell) { cell.marked = false; });
+        this.update_marked_status();
     };
     
     /**
@@ -728,9 +736,18 @@ define(function (require) {
         var selectedIndex = Math.min(Math.max(this.get_selected_index() + offset, 0), this.ncells()-1);
         this.select(selectedIndex);
         this.get_selected_cell().marked = true;
-
         this.ensure_focused();
+        this.update_marked_status()
     };
+
+    Notebook.prototype.update_marked_status = function(){
+        var indicies = this.get_marked_indices();
+        if ( indicies.length === 0){
+            this.element.removeClass('jp-marking');
+        } else {
+            this.element.addClass('jp-marking');
+        }
+    }
 
     // Cell selection.
 
@@ -1020,6 +1037,7 @@ define(function (require) {
         this.set_dirty(true);
 
         return this;
+        this.update_marked_status();
     };
 
     /**
@@ -1063,6 +1081,7 @@ define(function (require) {
             this.undelete_index = null;
         }
         $('#undelete_cell').addClass('disabled');
+        this.update_marked_status();
     };
 
     /**
@@ -1486,6 +1505,7 @@ define(function (require) {
             new_cell.set_text(texta);
             new_cell.marked = cell.marked;
         }
+        this.update_marked_status();
     };
 
     /**
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -186,7 +186,7 @@ define(function (require) {
     Notebook.prototype.bind_events = function () {
         var that = this;
 
-        this.events.on('cell:marking-changed', function(){
+        this.events.on('marked_changed.Cell', function(){
             that.update_marked_status();
         });
 
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -186,10 +186,6 @@ define(function (require) {
     Notebook.prototype.bind_events = function () {
         var that = this;
 
-        this.events.on('marked_changed.Cell', function(){
-            that.update_marked_status();
-        });
-
         this.events.on('set_next_input.Notebook', function (event, data) {
             if (data.replace) {
                 data.cell.set_text(data.text);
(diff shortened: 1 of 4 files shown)


jdfreder:
Transferring opinion from #669 to here, I prefer it without the gray border.

However, removing the gray border may not be within our policy for minor releases.

jhamrick:
Yeah, I also prefer it without the border, but I think @Carreau is right -- the marking is already a fairly big difference UI-wise, so I would say let's take it one step at a time and leave the border for now. We can remove it in the next release.

blink1073:
I am also good with this as-is.  Having three different widths was confusing, but I like this, with an eventual goal to remove the gray border.

blink1073:
Or are there still three widths and I'm just blind?

ellisonbg:
Cool, I will test now...

On Wed, Oct 28, 2015 at 3:41 PM, Steven Silvester <notifications@github.com>
wrote:

> Or are there still three widths and I'm just blind?
>
> —
> <https://github.com/jupyter/notebook/pull/675#issuecomment-151968112>.
>



-- 
Brian E. Granger

jhamrick:
@blink1073 there are still three widths, though having the border definitely makes it harder to tell.

ellisonbg:
Just played with it - I like it a lot!!!

One thing we should change. Right now there is an action that is called "unmark or close pager". Once we have commands scoped by CSS selectors we will be able to cleanly separate those. But it would be nice to not have to rename that command. What about this:

* Define separate command for unmark all cells and close pager.
* Have both of the actual logic for those commands do the same thing (see if pager is open, etc.).

That way, we won't have to change any command names once we get the scoped selectors.

But I am +1 on the behavior of this PR otherwise.

blink1073:
As I play with it locally, I see that the third width does help distinguish that the current cell is both selected and marked, so +1.

ellisonbg:
Jess, did you try to use slightly wider borders, maybe instead of 1/2/4 try
1/3/5?

It might help distinguish the three states visually - especially in
settings where color can't be resolved accurately.

On Wed, Oct 28, 2015 at 3:56 PM, Steven Silvester <notifications@github.com>
wrote:

> As I play with it locally, I see that the third width does help
> distinguish that the current cell is both selected and marked, so +1.
>
> —
> <https://github.com/jupyter/notebook/pull/675#issuecomment-151971739>.
>



-- 
Brian E. Granger

jhamrick:
The borders are currently 2/3/4. I'll try it out with 1/3/5 and see how that looks.

ellisonbg:
Shouldn't the smallest width match the regular border state?

On Wed, Oct 28, 2015 at 4:03 PM, Jessica B. Hamrick <

> The borders are currently 2/3/4. I'll try it out with 1/3/5 and see how
> that looks.
>
> —
> <https://github.com/jupyter/notebook/pull/675#issuecomment-151974097>.
>



-- 
Brian E. Granger

jhamrick:
I find it difficult to see that the border is blue if it's only 1px. Here's a gif of 2/4/6, in which (to me) it is fairly clear that the border is blue and that there are three different widths:

jhamrick:
However, the downside to having the thicknesses be more distinct is that I'm a bit worried now people will think the selected cell is something different from the marked cells...

blink1073:
^ That was my original concern, but it felt intuitive when I actually used it.

jhamrick:
@blink1073 Yeah, I 2/3/4 felt fine to me as well, even though the difference is subtle. I think maybe once we remove the gray border it might be necessary to emphasize the widths more and do e.g. 2/4/6 but for now I think it's probably fine? @ellisonbg what do you think?

willingc:
@jhamrick Excellent write up and analysis. No strong preference on the widths.

ellisonbg:
Hmm, looks like this needs rebase. Not sure which merge tripped this up. Might mean that all in the series needs to be rebased...

notebook/static/notebook/js/notebook.js
```diff
@@ -728,8 +736,18 @@ define(function (require) {
         var selectedIndex = Math.min(Math.max(this.get_selected_index() + offset, 0), this.ncells()-1);
         this.select(selectedIndex);
         this.get_selected_cell().marked = true;
+        this.update_marked_status()
     };
 
+    Notebook.prototype.update_marked_status = function(){
```
ellisonbg: Is there any way to get out of needing the notebook to track this information?

jhamrick: This function gets changed substantially in #683 anyway... why don't I remove it from here for now and we can figure out what to do with it in #683.

jhamrick:
Rebased.

jhamrick:
Changed `cell:marking-changed` to `marked_changed.Cell`, will rebase other PRs momentarily.

notebook/static/notebook/less/cell.less
```diff
@@ -1,3 +1,8 @@
+._marked_style(@n) {
+    border-left-width: @n;
+    padding-left: 6px-@n;
```
Carreau: One small tiny nitpicking I'm not sure about:  The 6px here should likely be a variable used somewhere else, when cells are both not marked and not selected. 

Likely because line 11 `border-width:thin` == `1px` and line 49: padding is 5px :-P

ellisonbg:
Checked the behavior of this PR and I really like it. Looks like @Carreau made a comment about a CSS var, anything else still underway with the code?

Carreau:
~~Well the comment about css can be made in a subsequent PR if the merger open an issue to remember~~

Fixinx that with @jhamrick live

jhamrick:
Ok, addressed @Carreau 's commends. I am done making changes here.

ellisonbg:
Will merge when travis finishes.

Carreau:
> Will merge when travis finishes.

+1 IIF travis is happy. Because travis can finish angry.

ellisonbg:
It is angry...kicking it

jhamrick:
Travis looks happy now!

Carreau:
Github is purple.

jhamrick:
Thanks! :tada: :confetti_ball:

ellisonbg:
Great work!

On Thu, Oct 29, 2015 at 2:31 PM, Jessica B. Hamrick <

> Thanks! [image: :tada:] [image: :confetti_ball:]
>
> —
> <https://github.com/jupyter/notebook/pull/675#issuecomment-152277385>.
>



-- 
Brian E. Granger

PR #682, opened 2015-10-28
Ensure focus is kept when extending the marked region

jhamrick:
Currently, if you use `shift-j` and `shift-k` to expand the marked region, focus does not follow the selected cell. This PR ensures that the focus is kept on the selected cell after expanding the marked region.


jhamrick:
Thanks!

PR #696, opened 2015-10-31
Rate limit scrolling callback

### LINKED ISSUE #694: debouce/throttle scroll listener?
In `notebook.js` it feel like the following slow scroll a lot:

```
        this.scroll_manager.element.scroll(function () {
            that.update_marked_status();
        });
```

It might be just an impression or smth else, but scrolling seem slow today. 

Investigate if we can debounce, or optimize, or if it's smth else.

jhamrick:
Fixes #694 

@Carreau can you see if this makes it better?

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -297,9 +297,9 @@ define(function (require) {
             expand_time(time);
         });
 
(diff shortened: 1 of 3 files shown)


Carreau:
+1 Ping @ellisonbg

PR #711, opened 2015-11-09
allow some KernelManager methods to be async

minrk:
Allows custom KernelManagers to do slower async operations as part of kernel startup/shutdown.

Related: jupyter-incubator/kernel_gateway#18


takluyver:
This all looks OK to me. Do you want to do anything else with it before it gets merged?

minrk:
Nope, should be ready.

PR #718, opened 2015-11-12
Fix closing ansi handling.

Carreau:
The following was wrong:

    IPython.utils.ansispan("\x1b[38;5;28;01mtry\x1b[39;00m")
    "<span  style="color: rgb(0,102,0);" class="ansibold">try< span>"

---- 
Opening to remember to write a test,  well it's default foreground color but it's used a lot to mean reset right ? 49 is background and should likely be handled. 

It comes up in pygments that use `[39;00m` for reset, which screw up tracebacks.

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -308,12 +308,12 @@ define([
         }
     }
 
-    function ansispan(str) {
+    function _ansispan(str) {
         // ansispan function adapted from github.com/mmalecki/ansispan (MIT License)
         // regular ansi escapes (using the table above)
         var is_open = false;
         return str.replace(/\033\[(0?[01]|22|39)?([;\d]+)?m/g, function(match, prefix, pattern) {
-            if (!pattern) {
+            if (!pattern || prefix === '39') {
                 // [(01|22|39|)m close spans
                 if (is_open) {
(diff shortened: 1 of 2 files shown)


notebook/static/base/js/utils.js
```diff
@@ -313,7 +313,7 @@ define([
         // regular ansi escapes (using the table above)
         var is_open = false;
         return str.replace(/\033\[(0?[01]|22|39)?([;\d]+)?m/g, function(match, prefix, pattern) {
-            if (!pattern) {
+             if (!pattern || prefix === '39') {
```
minrk: Did you mean to increase the indent to 5 spaces?

Carreau: No I did not mean too, I'm remove it when I write the tests for it.

Carreau:
Added test, and remove the extra-space. I also  made `ansispan` private by adding leading underscore, and exported it from the module (to test from js console).

Carreau:
Travis is happy.

captainsafia:
:+1: for tests!

Maybe add a comment before line 316 describing why that change was made or referencing the appropriate PR? I can imagine looking at this in a couple of months and being confused as to why that code is there.

Issue #484, opened 2015-09-22
notebook changed on disk message


dschneiderch:
When I'm working in a notebook jupyter 4.0.6 in Safari, OSX 10.9.5 I periodically get a message saying the file has changed from disk, do I want to reload or overwrite.  I"m certain there are no other kernels or notebooks running. I usually just say overwrite but why does this happen?
One more bit of info: I'm running the notebook as a publicly accessible server (so I can access it from my labtop) and all the files live on a locally mounted server.
here is the terminal output:
iMac:~ $ jupyter notebook
[I 15:17:16.737 NotebookApp] Serving notebooks from local directory: /Users/dosc3612
[I 15:17:16.737 NotebookApp] 0 active kernels
[I 15:17:16.737 NotebookApp] The IPython Notebook is running at: https://[all ip addresses on your system]:8888/
[I 15:17:16.738 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[I 15:28:57.760 NotebookApp] Kernel started: ec591d51-c03c-44fc-86bf-a221c0e1b0a7

> IRkernel::main()
> [1] "Got unhandled msg_type:" "comm_open"
> [I 15:30:57.046 NotebookApp] Saving file at /Documents/snotel-regression_project/src/validate_surveygridded.ipynb
> [I 15:37:04.133 NotebookApp] Saving file at /Documents/snotel-regression_project/src/validate_surveygridded.ipynb
> [I 15:37:05.369 NotebookApp] Saving file at /Documents/snotel-regression_project/src/validate_surveygridded.ipynb

![screen shot 2015-09-22 at 15 36 07](https://cloud.githubusercontent.com/assets/7461221/10032628/7f4274cc-6141-11e5-8bd8-6099883fe372.png)

Carreau:
> I"m certain there are no other kernels or notebooks running

Could there be something else accessing the file ? If you get this message it means that there is something else that have modified the file since last save. Could there be a dropbox syncing ? Or could the dates on the mounted filesystem be wrong ?

dschneiderch:
The dates and times look in sync.  There is no dropbox syncing on this
folder. I also don't remember getting this while using ipython3
(pre-jupyter).
[image: Inline image 1]

Dominik Schneider
o 303.735.6296 | c 518.956.3978

lithces:
Same situation.

I am running ipython notebook in a Fedora 22 box. The box is running as a vmware guest hosted on M$ windows. 

[...]

PR #729, opened 2015-11-17
Add debug statements for notebook changed on disk

Carreau:
I have a **vague** impression that #484 (Notebook changed on disk) might be due to a mishandling of dates in JS.

 The server-side do write the model and stat the file on disk for it's modified date, and send that back to the browser that store it for next comparison, so unless the modified date on disk  **actually change** I don't know what can cause that. 

Hopefully this would give us more hints as to why the dates are compared wrong. 

Thoughts ?

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -58,6 +58,8 @@ define(function (require) {
         this.ws_url = options.ws_url;
         this._session_starting = false;
         this.last_modified = null;
+        // debug 484
+        this._last_modified = 'init';
 
         //  Create default scroll manager.
         this.scroll_manager = new scrollmanager.ScrollManager(this);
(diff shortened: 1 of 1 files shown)


jdfreder:
When we merge this, we should open another issue to remind us to revert the debug statements one day when we figure out what the problem was.

minrk:
:+1:

Carreau:
> When we merge this, we should open another issue to remind us to revert the debug statements one day when we figure out what the problem was.

you can open the issue now :-)

captainsafia:
@Carreau — done.

captainsafia:
Do we want to go ahead and merge this just to start knocking this off the milestone?

minrk:
Yup, thanks.

notebook/static/notebook/js/notebook.js
```diff
@@ -2513,6 +2519,8 @@ define(function (require) {
                 that.notebook_name = json.name;
                 that.notebook_path = json.path;
                 that.last_modified = new Date(json.last_modified);
+                // debug 484
+                _that.last_modified = json.last_modified;
```
captainsafia: Should this be `that._last_modified` instead of `_that.last_modified`?

jdfreder: Yes

captainsafia: Yep! @Carreau fixed it in #761.

PR #731, opened 2015-11-17
BUG: added urllib to encode URL for browser

### LINKED ISSUE #728: BUG: Opening files with spaces fails
Just installed the latest Jupyter on Python 3.4.1 with:
```bash
$ pip install jupyter
$ pip freeze | grep jupyter
jupyter==1.0.0
jupyter-client==4.1.1
jupyter-console==4.0.3
jupyter-core==4.0.6
```
I'm having an issue opening files that contain spaces in their names. The notebook launches, but it tires to open two notebooks instead of a single notebook.
```
$ jupyter notebook Awesome\ Notebook.ipynb                                      
[I 10:48:28.080 NotebookApp] The port 8888 is already in use, trying another random port.
[I 10:48:28.085 NotebookApp] Serving notebooks from local directory: /home/user/
[I 10:48:28.085 NotebookApp] 0 active kernels 
[I 10:48:28.085 NotebookApp] The IPython Notebook is running at: http://localhost:8889/
[I 10:48:28.085 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[W 10:48:28.671 NotebookApp] 404 GET /notebooks/Awesome (::1): No such file or directory: Awesome
[W 10:48:28.717 NotebookApp] 404 GET /notebooks/Awesome (::1) 48.08ms referer=None
```
This also occurs with:
```
$ jupyter notebook "Awesome Notebook.ipynb"
```
Both should work.

I think the bug is in `parse_command_line` in [notebookapp.py](https://github.com/jupyter/notebook/blob/4578c34b0f999735ee49e1492be3dd5941951551/notebook/notebookapp.py#L788), but haven't had a chance to dig deeper yet. In previous IPython versions I have not had this issue, so it seems like a regression.

Are there tests for this issue?

cjermain:
Fixes #728 by encoding the URL for the browser, to replace spaces with `%20` using `urllib.quote`.

Should a test be written?


cjermain:
Looks like Python 3 does not have `urllib.quote`, so that needs to be made compatible.

captainsafia:
Looks great, @cjermain! Thanks for submitting the PR!

Can you add some tests under `notebook/tests/test_notebookapp.py` to confirm the behavior?

cjermain:
The fix is now using the `utils.url_escape` method which has tests in `tests_utils.py`.

captainsafia:
This looks good to merge.

cc/ @minrk @Carreau

PR #747, opened 2015-11-23
Allow disabling atomic writing

### LINKED ISSUE #739: Make atomic writing configurable.
discussed with @ellisonbg @fperez @jdfreder

parleur:
closes #739


Carreau:
We are working on that through skype. 

In Skype comment  for posterity: Need rebase, Configure Atom not to strip whitespace, set up github ssh key :-)

notebook/services/contents/fileio.py
```diff
@@ -130,8 +183,12 @@ def open(self, os_path, *args, **kwargs):
     def atomic_writing(self, os_path, *args, **kwargs):
         """wrapper around atomic_writing that turns permission errors to 403"""
         with self.perm_to_403(os_path):
-            with atomic_writing(os_path, *args, log=self.log, **kwargs) as f:
-                yield f
+            if self.use_atomic_writing:
+                with atomic_writing(os_path, *args, log=self.log, **kwargs) as f:
+                    yield f
+            else:
+                with simple_writing(os_path, *args, log=self.log, **kwargs) as f:
```
Carreau: Missing underscore.

Carreau:
Watch out, you vim or atom have "remove trailing whitespace on save" turned on, 
Also feel free to drop a message when you push a change like test, we are not warned on changes on PRs.

Carreau:
Yyou might also want to setup your git or github so that commits appear yours.

takluyver:
I believe the key for Github to recognise you is to set your email address in git to an address you have registered with Github. You can do that like this:

```
git config --global user.email "x@y.fr"
```

Unfortunately this branch has merge conflicts, so we'll ask you to rebase it. @Carreau may be able to talk you through that if you're not familiar with git.

notebook/services/contents/fileio.py
```diff
@@ -119,6 +171,9 @@ class FileManagerMixin(object):
     log : logging.Logger
     """
 
+    use_atomic_writing = Bool(True, config=True, help="""flag to deactivate
+            atomic writing on slow (like networked) file system.""")
```
takluyver: The help for this shouldn't assume users know what atomic writing is - it should mention what having it enabled means, as well as why you might want to disable it.

It also needn't start "flag to...". Have a look at some of our other config help strings to see how we tend to write them.

notebook/services/contents/fileio.py
```diff
@@ -100,7 +102,57 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
...
+        yield fileobj
+    except:
+        fileobj.close()
+        raise
+
+    # Flush to disk
+    fileobj.flush()
+    if log:
+        log.debug( path + " saved using simple fs writing")
+    os.fsync(fileobj.fileno())
```
takluyver: The calls to flush and fsync in the atomic writing are an attempt to ensure that the file is genuinely written to disk (as opposed to stored in an OS cache) before we remove the previous copy of the file. With atomic writing disabled, I would take them out and just allow the OS to do whatever caching it wants to.

notebook/services/contents/fileio.py
```diff
@@ -130,8 +185,12 @@ def open(self, os_path, *args, **kwargs):
     def atomic_writing(self, os_path, *args, **kwargs):
         """wrapper around atomic_writing that turns permission errors to 403"""
```
takluyver: This docstring should be updated now that the method may not actually do atomic_writing depending on configuration.

Also, ideally we'd rename the method to reflect what it now does. But I'm not sure whether the method counts as public API; if it does, we shouldn't rename it.

notebook/services/contents/tests/test_manager.py
```diff
@@ -528,3 +528,34 @@ def test_check_and_sign(self):
...
+        self.td = self._temp_dir.name
+        self.contents_manager = FileContentsManager(
+            root_dir = self.td,
+        )
+        self.contents_manager.use_atomic_writing = False
+
+    def tearDown(self):
+        self._temp_dir.cleanup()
+
+    def new_notebook_no_atomic(self):
```
takluyver: I think tests are only run if they have test in the name, like `test_save_notebook_no_atomic`. I might be wrong about that, but double check that the test is actually being executed.

notebook/services/contents/tests/test_manager.py
```diff
@@ -528,3 +528,34 @@ def test_check_and_sign(self):
         cm.mark_trusted_cells(nb, path)
         cm.check_and_sign(nb, path)
         assert cm.notary.check_signature(nb)
+
+
+class TestContentsManagerNoAtomic(TestCase):
+
+    def setUp(self):
+        self._temp_dir = TemporaryDirectory()
```
takluyver: If you do `self.addCleanup(temp_dir.cleanup)`, then you don't need a  `tearDown` method, and you don't need to attach the temporary directory object to `self`.

notebook/services/contents/tests/test_manager.py
```diff
@@ -528,3 +528,34 @@ def test_check_and_sign(self):
...
+        name = model['name']
+        path = model['path']
+
+        full_model = cm.get(path)
+        nb = full_model['content']
+        nb['metadata']['counter'] = int(1e6 * time.time())
+        self.add_code_cell(nb)
+
+        cm.save(full_model, path)
+        return nb, name, path
```
takluyver: This is a 'smoketest' - a test that runs the code and will fail if it raises an error, but doesn't check what the code actually did.

Sometimes a smoke test is all we can do, but in this case I think we should go one better, and check that the file it should have created really exists.

It may be overkill for this case, but I have a package called `testpath` with functions for checking this kind of thing - e.g. [testpath.assert_isfile](http://testpath.readthedocs.org/en/latest/fsasserts.html#testpath.assert_isfile).

notebook/services/contents/tests/test_manager.py
parleur: You were right.

notebook/services/contents/tests/test_manager.py
parleur: I am not sure my new solution would please you, as I relaunch the exact same set of test in non atomic writing case than in atomic writing case (this time I checked the tests are eventally done).
But I am sure of the exact way the orignal set of test check the presence of file on disk after a save.

Carreau:
@parleur Need rebase.

parleur:
You are ingredibly reactive ...

Carreau:
@ellisonbg you've seen issues wit atomic writing, can you test this ?

Carreau:
I'm going to merge this for it to be in 4.1, it is anyway disabled by default, but at least if we have it released we can have user testing it (and fix potential issues in a patch version).

Thanks All !

PR #750, opened 2015-11-24
Updated function call to match API changes in jupyter_client

captainsafia:
If [jupyter/jupyter_client#93](https://github.com/jupyter/jupyter_client/pull/93) is merged, this should be merged in order to reflect the API changes in that PR.

### DIFF
--- a/notebook/services/kernelspecs/handlers.py
+++ b/notebook/services/kernelspecs/handlers.py
@@ -53,7 +53,7 @@ class MainKernelSpecHandler(APIHandler):
         model = {}
         model['default'] = km.default_kernel_name
         model['kernelspecs'] = specs = {}
-        for kernel_name in ksm.find_kernel_specs():
+        for kernel_name in ksm.find_all_specs():
             try:
                 d = kernelspec_model(self, kernel_name)
             except Exception:


minrk:
@captainsafia that PR doesn't seem to remove the API here, and removing that API would force a major-revision release on jupyter_client. I don't see why this API would be used, since it causes strictly more work to be done than the original `find_kernel_specs`.

captainsafia:
@minrk — Did you mean to put this in [jupyter/jupyter_client#93](https://github.com/jupyter/jupyter_client/pull/93)? I'm a little bit confused as to what you mean.

minrk:
No, I meant it to be here. This PR changes the dependency on jupyter_client to an unreleased version, and as far as I can tell from the jupyter_client PR, no benefit, since iterating through `find_all_specs` and `find_kernel_specs` has the same result and `find_all_specs` does strictly more work than `find_kernel_specs`. Perhaps there is something missing in this PR to actually make use of the difference between the two.

minrk:
This can make sense if we *also* change the `kernel_model` call to allow passing the model from `get_all_specs`. But as it is, it doesn't make any use of the computed specs, it just discards them.

Carreau:

PR #760, opened 2015-11-26
refactor secure notebook server script

Carreau:
break-up into  pieces and write docstring.

### DIFF
--- a/tools/secure_notebook.py
+++ b/tools/secure_notebook.py
@@ -16,15 +16,13 @@ from traitlets.config import Config
 
 from contextlib import contextmanager
 
-from OpenSSL import crypto, SSL
-from socket import gethostname
-from pprint import pprint
-from time import gmtime, mktime
+from OpenSSL import crypto
 from os.path import exists, join
 
 import io
 import os
 import json
+import traceback
 
 
 def create_self_signed_cert(cert_dir, keyfile, certfile):
@@ -34,9 +32,9 @@ def create_self_signed_cert(cert_dir, keyfile, certfile):
     Abort if one of the keyfile of certfile exist.
     """
 
-    if not exists(join(cert_dir, certfile)) \
-            or not exists(join(cert_dir, keyfile)):
-
+    if exists(join(cert_dir, certfile))  or exists(join(cert_dir, keyfile)):
+        raise FileExistsError('{} or {} already exist in {}. Aborting.'.format(keyfile, certfile, cert_dir))
+    else:
         # create a key pair
         k = crypto.PKey()
         k.generate_key(crypto.TYPE_RSA, 1024)
@@ -60,16 +58,15 @@ def create_self_signed_cert(cert_dir, keyfile, certfile):
             f.write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert).decode('utf8'))
         with io.open(join(cert_dir, keyfile), "wt") as f:
             f.write(crypto.dump_privatekey(crypto.FILETYPE_PEM, k).decode('utf8'))
-    else :
-        raise FileExistsError('{} or {} already exist in {}. Aborting.'.format(keyfile, certfile, certdir))
 
 
 
 @contextmanager
-def persist_config():
-    """Conext manager that can be use to modify a config object
+def persist_config(mode=0o600):
+    """Context manager that can be use to modify a config object
 
-    on exit of the context manager, the config will be written back to disk. 
+    On exit of the context manager, the config will be written back to disk, 
+    by defauld with 600 permissions.
     """
 
     loader = JSONFileConfigLoader('jupyter_notebook_config.json', jupyter_config_dir())
@@ -80,27 +77,34 @@ def persist_config():
 
     yield config
 
-    with io.open(os.path.join(jupyter_config_dir(), 'jupyter_notebook_config.json'), 'w') as f:
+    filepath = os.path.join(jupyter_config_dir(), 'jupyter_notebook_config.json')
+    with io.open(filepath, 'w') as f:
         f.write(six.u(json.dumps(config, indent=2)))
+    try:
+        os.chmod(filepath, mode)
+    except Exception:
+        traceback.print_exc()
+
+        print("Something went wrong changing file permissions")
 
 
 def set_password():
     """Ask user for password, store it in notebook json configuration file"""
 
-    print("first choose a password.")
-    pw = passwd()
+    print("First choose a password.")
+    hashedpw = passwd()
     print("We will store your password encrypted in the notebook configuration file: ")
-    print(pw)
+    print(hashedpw)
 
     with persist_config() as config:
-        config.NotebookApp.password = pw
+        config.NotebookApp.password = hashedpw
 
     print('... done\n')
 
 
 def set_certifs():
     """
-    generate certificate to run notebook over ssl and set up the notebook config.
+    Generate certificate to run notebook over ssl and set up the notebook config.
     """
     print("Let's generate self-signed certificates to secure your connexion.")
     print("where should the certificate live?")
@@ -127,6 +131,6 @@ def set_certifs():
 
 
 if __name__ == '__main__':
-    print("This guide you into securing your notebook server.")
+    print("This will guide you through the steps towards securing your notebook server.")
     set_password()
     set_certifs()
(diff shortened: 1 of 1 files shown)


minrk:
I like it. cc @fperez who requested the changes.

Carreau:
Also https://github.com/ipython/traitlets/pull/137 might simplify this once we require traitlets > 4.1 :-P

tools/secure_notebook.py
```diff
@@ -19,13 +27,14 @@
...
 
 
-def create_self_signed_cert(cert_dir, keyfile, certfiile):
+def create_self_signed_cert(cert_dir, keyfile, certfile):
     """
-    If datacard.crt and datacard.key don't exist in cert_dir, create a new
-    self-signed cert and keypair and write them into that directory.
+    Create a self-signed `keyfile` and `certfile` in `cert_dir`
+
+    Abort if one of the
```
takluyver: of the...

tools/secure_notebook.py
```diff
@@ -51,14 +60,17 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
 
-if __name__ == '__main__':
-    print("This guide you into securing your notebook server")
-    print("first choose a password.")
-    pw = passwd()
-    print("We will store your password encrypted in the notebook configuration file: ")
-    print(pw)
+
+@contextmanager
+def load_config():
```
takluyver: This could be more clearly named to indicate that it's both loading and saving - maybe `persistent_config`?

tools/secure_notebook.py
```diff
@@ -66,16 +78,33 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
     except ConfigFileNotFound:
         config = Config()
 
-    config.NotebookApp.password = pw
+    yield config
 
     with io.open(os.path.join(jupyter_config_dir(), 'jupyter_notebook_config.json'), 'w') as f:
```
takluyver: If we're storing passwords, should we be trying to ensure that the config file is only user readable (i.e. chmod 600 on it)?

tools/secure_notebook.py
```diff
@@ -66,16 +78,33 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
+        config.NotebookApp.password = pw
+
+    print('... done\n')
+
+
+def set_certifs():
+    """
+    generate certificate to run notebook over ssl and set up the notebook config.
+    """
+    print("Let's generate self-signed certificates to secure your connexion.")
```
takluyver: Will it be practical to add an option here to generate a certificate with letsencrypt, once it's generally available, rather than a self-signed one?

tools/secure_notebook.py
Carreau: That usually is handled at jupyter_ore level that config dir should be 0o700, the password is sha-hashed, so not **that** sensible.

tools/secure_notebook.py
Carreau: Likely, in the end this is in tools, we could migrate it to the notebook itself later, but it does not replace a real audited way of having a SSL certificate and secure connexion.

tools/secure_notebook.py
```diff
@@ -66,16 +78,33 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
         f.write(six.u(json.dumps(config, indent=2)))
 
-    print('... done')
-    print()
 
-    print("Now let's generate self-signed certificates to secure your connexion.")
+def set_password():
+    """Ask user for password, store it in notebook json configuration file"""
+
+    print("first choose a password.")
```
fperez: Capitalize "First"

tools/secure_notebook.py
```diff
@@ -66,16 +78,33 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
+
+    with persist_config() as config:
+        config.NotebookApp.password = pw
+
+    print('... done\n')
+
+
+def set_certifs():
+    """
+    generate certificate to run notebook over ssl and set up the notebook config.
```
fperez: Capitalize "Generate"

tools/secure_notebook.py
```diff
@@ -90,12 +119,14 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
-    config.NotebookApp.keyfile = fullkey
-
-    with io.open(os.path.join(jupyter_config_dir(), 'jupyter_notebook_config.json'), 'w') as f:
-        f.write(six.u(json.dumps(config, indent=2)))
-
+    print('done.\n')
 
 
+if __name__ == '__main__':
+    print("This guide you into securing your notebook server.")
```
fperez: I'd reword the msg as `This will guide you through the steps towards securing your notebook server.` Though a review by one of our native speakers, @takluyver or @minrk would be even better...

tools/secure_notebook.py
```diff
@@ -19,13 +27,14 @@
...
+def create_self_signed_cert(cert_dir, keyfile, certfile):
     """
-    If datacard.crt and datacard.key don't exist in cert_dir, create a new
-    self-signed cert and keypair and write them into that directory.
+    Create a self-signed `keyfile` and `certfile` in `cert_dir`
+
+    Abort if one of the keyfile of certfile exist.
     """
 
-    if not exists(join(cert_dir, certfiile)) \
```
fperez: I would write the logic in the opposite order: `if exists: raise, else: write()`. Less composite negatives make the code much easier to follow, I think, even if the exception goes first, which is a bit unusual.

But just my take on it, OK if you disagree.

tools/secure_notebook.py
```diff
@@ -51,14 +60,17 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
-    print("first choose a password.")
-    pw = passwd()
-    print("We will store your password encrypted in the notebook configuration file: ")
-    print(pw)
+
+@contextmanager
+def persist_config():
+    """Conext manager that can be use to modify a config object
+
+    on exit of the context manager, the config will be written back to disk.
```
fperez: Capitalize `On` at start of sentence.

tools/secure_notebook.py
```diff
@@ -51,14 +60,17 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
-if __name__ == '__main__':
-    print("This guide you into securing your notebook server")
-    print("first choose a password.")
-    pw = passwd()
-    print("We will store your password encrypted in the notebook configuration file: ")
-    print(pw)
+
+@contextmanager
+def persist_config():
+    """Conext manager that can be use to modify a config object
```
fperez: Typo: `Context`

tools/secure_notebook.py
fperez: I think though that standard practice for these kinds of files is to still store them with tight permissions, even if they are in a 700 dir, in case they get temporarily/accidentally moved out.

tools/secure_notebook.py
```diff
@@ -66,16 +78,33 @@ def create_self_signed_cert(cert_dir, keyfile, certfiile):
...
-    print()
 
-    print("Now let's generate self-signed certificates to secure your connexion.")
+def set_password():
+    """Ask user for password, store it in notebook json configuration file"""
+
+    print("first choose a password.")
+    pw = passwd()
+    print("We will store your password encrypted in the notebook configuration file: ")
+    print(pw)
```
fperez: Are we echoing passwords in plaintext back to the screen?

tools/secure_notebook.py
Carreau: Done

tools/secure_notebook.py
Carreau: Done

tools/secure_notebook.py
Carreau: Done

tools/secure_notebook.py
Carreau: persist config default to 0o600 now.

tools/secure_notebook.py
Carreau: Done.

tools/secure_notebook.py
Carreau: No hashed password, I changed the name of the variable.

tools/secure_notebook.py
Carreau: Done

tools/secure_notebook.py
Carreau: Done.

fperez:
OK, I suggest we merge this now. It's a tool, so not in the critical path, and it's seen by now enough review/iteration that I think we can call it good enough for now.

Any objection?

fperez:
OK, nobody objected, merging to move the queue forward.

PR #775, opened 2015-11-30
new feature: NotebookApp.password_required forces user to have a password

maartenbreddels:
When the setting is set to True, it forces users to use a password for the Notebook server. This is useful in a multi user environment, for instance when everybody in the LAN can access each other's machine though ssh (which happens to be the case in our institute). In such a case, server the notebook server on localhost is not secure since any user can connect to the notebook server via ssh.

A sysadmin can enable this setting for instance by running:
```
echo "NotebookApp.password_required.default_value = True" >> `python -c "import notebook.notebookapp as mod, os; print(os.path.splitext(mod.__file__)[0]+'.py')"`
```
But maybe there are more elegant solutions to set this default (comments are welcome)

### DIFF
--- a/notebook/auth/__main__.py
+++ b/notebook/auth/__main__.py
@@ -30,10 +30,10 @@ def set_password(args):
 def main(argv):
 	parser = argparse.ArgumentParser(argv[0])
 	subparsers = parser.add_subparsers()
-	parser_password = subparsers.add_parser('password', help='create soneira peebles dataset')
+	parser_password = subparsers.add_parser('password', help='sets a password for your notebook server')
 	parser_password.add_argument("password", help="password to set, if not given, a password will be queried for (NOTE: this may not be safe)",
 			nargs="?")
-	parser_password.add_argument("--quiet", help="Supress messages", action="store_true")
+	parser_password.add_argument("--quiet", help="suppress messages", action="store_true")
 	parser_password.set_defaults(function=set_password)
 	args = parser.parse_args(argv[1:])
 	args.function(args)
--- a/notebook/auth/__main__.py
+++ b/notebook/auth/__main__.py
@@ -1,0 +1,42 @@
+from notebook.auth import passwd
+from getpass import getpass
+from traitlets.config.manager import BaseJSONConfigManager
+from jupyter_core.paths import jupyter_config_dir
+import argparse
+import sys
+
+def set_password(args):
+	password = args.password
+	while not password  :
+		password1 = getpass("" if args.quiet else "Provide password: ")
+		password_repeat = getpass("" if args.quiet else "Repeat password:  ")
+		if password1 != password_repeat:
+			print("Passwords do not match, try again")
+		elif len(password1) < 4:
+			print("Please provide at least 4 characters")
+		else:
+			password = password1
+
+	password_hash = passwd(password)
+	cfg = BaseJSONConfigManager(config_dir=jupyter_config_dir())
+	cfg.update('jupyter_notebook_config', {
+		'NotebookApp': {
+			'password': password_hash,
+		}
+	})
+	if not args.quiet:
+		print("password stored in config dir: %s" % jupyter_config_dir())
+
+def main(argv):
+	parser = argparse.ArgumentParser(argv[0])
+	subparsers = parser.add_subparsers()
+	parser_password = subparsers.add_parser('password', help='create soneira peebles dataset')
+	parser_password.add_argument("password", help="password to set, if not given, a password will be queried for (NOTE: this may not be safe)",
+			nargs="?")
+	parser_password.add_argument("--quiet", help="Supress messages", action="store_true")
+	parser_password.set_defaults(function=set_password)
+	args = parser.parse_args(argv[1:])
+	args.function(args)
+	
+if __name__ == "__main__":
+	main(sys.argv)
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -58,6 +58,15 @@ from notebook import (
     DEFAULT_TEMPLATE_PATH_LIST,
     __version__,
 )
+from .auth import passwd
+from getpass import getpass
+
+# py23 compatibility
+try:
+    raw_input = raw_input
+except NameError:
+    raw_input = input
+
 from .base.handlers import Template404
 from .log import log_request
 from .services.kernels.kernelmanager import MappingKernelManager
@@ -570,6 +579,17 @@ class NotebookApp(JupyterApp):
                       """
     )
 
+    password_required = Bool(False, config=True,
+                      help="""Forces users to use a password for the Notebook server.
+                      This is useful in a multi user environment, for instance when
+                      everybody in the LAN can access each other's machine though ssh.
+
+                      In such a case, server the notebook server on localhost is not secure
+                      since any user can connect to the notebook server via ssh.
+
+                      """
+    )
+
     open_browser = Bool(True, config=True,
                         help="""Whether to open in a browser after starting.
                         The specific browser used is platform dependent and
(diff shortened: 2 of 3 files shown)


takluyver:
One measure I've been thinking about for a while is having a form of one-time-password for the notebook. This would be randomly generated, and when it launched the browser to view the notebook, it would supply it in the URL, where it would be exchanged for a more persistent access token (via a cookie?).

This would make it more palatable to enable security measures like this by default, because it doesn't require any extra effort from the user.

Of course, it still adds some inconvenience if the user wants to e.g. open up the notebook in a different browser. We could generate a code for them to copy and paste in that case. But that's inconveniencing 5% of use, rather than requiring a password typed in and inconveniencing all use.

maartenbreddels:
Indeed, it something that I had in mind as well, but that situation makes it difficult to say start the notebook once (at work) and connect to it later via an ssh tunnel. In any case, this feature was easier to implement and will only take effect if the sysadmin enforces it.

minrk:
The more elegant version of setting it by default is:

    mkdir -p /etc/jupyter
    echo 'c.NotebookApp.password_required = True' >> /etc/jupyter/jupyter_notebook_config.py

@takluyver We could also alleviate the always-on password inconvenience by adding an nbopen-style CLI entrypoint that sets login in the same way as the first-launch open (the token should added to the existing nbinfo file).

Carreau:
Better would be to write that in the json config file, which is machine readable. 
and if password we enforce, we likely want to enforce ssl. 

I'm also not a huge fan of having the logic of creating a password into the main notebook app itself. It's likely fine as an side tool that can be reused, like improving [this](https://github.com/jupyter/notebook/blob/master/tools/secure_notebook.py).

minrk:
> Better would be to write that in the json config file, which is machine readable.

I disagree, since it's humans who need to read/write the config.

Carreau:
> I disagree, since it's humans who need to read/write the config.

But in `/etc/jupyter/jupyter_notebook_config.json|py`  you likely want to make the check programmatically when you deploy that  `<configfile>` is not here, and `NotebookApp.password_required` not set or had he wrong the value.

maartenbreddels:
> and if password we enforce, we likely want to enforce ssl.

I don't agree fully, the use case for this PR is when you want to run your server bound to localhost, but other people can access you machine (such as at our institute). This isn't really a hostile environment, no direct need for SSL.
In any case, I do think it makes sense to add an additional check, for instance
```python
if not self.ip and ssl_options is None and not self.password:
  # warning that you are using a public ip, with a password that will be send plain text,
  # or simply refuse to start up
```

Note that there already are some of these checks, but this hasn't much to do this this PR.

>I'm also not a huge fan of having the logic of creating a password into the main notebook app itself. It's likely fine as an side tool that can be reused, like improving this.

Well, an alternative would be to have this flag (password_required), but refuse to start up when there is no password setup, and print out what to do to fix this, for instance
```
# You are running without a password, please run 
# $ python -m notebook.secure_notebook
```
(this would require moving this script to a notebook module)

minrk:
> But in /etc/jupyter/jupyter_notebook_config.json|py you likely want to make the check programmatically when you deploy that <configfile> is not here, and NotebookApp.password_required not set or had he wrong the value.

I should clarify - if *a script* is writing the config file, JSON is sensible. Humans should never read nor write JSON config files.

Carreau:
> I should clarify - if a script is writing the config file, JSON is sensible. Humans should never read nor write JSON config files.

Right, in the context of the issues, I assumed this was ment for script.

Carreau:
Need rebase. 

What is the state of this ?

maartenbreddels:
Is my last proposal of moving the logic of creating the password by invoking:
```
# You are running without a password, please run 
# $ python -m notebook.secure_notebook
```
better than the original code, or shall I rebase the original PR?

takluyver:
I think I slightly prefer that proposal to this PR (interactively prompting for a password). If the notebook is launched by a script or an init system, rather than by a user at a terminal, trying to prompt for any input might make the server hang in a useless state, or crash with a confusing error message about failing to read stdin.

minrk:
I also think the explicit command is a good idea. I've been bitten by prompts during startup looking like hangs when the terminal is either remote or nonexistent.

The secure_notebook function can write a JSON config file via Python API without having to manually deal with files:


```python
from jupyter_core.paths import jupyter_config_dir
from traitlets.config.manager import BaseJSONConfigManager

def secure_notebook():
    hashed_password = get_password(...)
    cfg = BaseJSONConfigManager(config_dir=jupyter_config_dir())
    cfg.update('jupyter_notebook_config', {
        'NotebookApp': {
            'password': hashed_password,
        }
    })
```

Feel free to rebase or clobber this PR, or open a new one. Whatever you feel comfortable with.

maartenbreddels:
I think an unrelated error causes the build failure right?
So basically, if password_required is true, and a no password is given, it refuses to start up and spits out:
```
[C 14:52:02.752 NotebookApp] Notebook servers are configured to only be run with a password.
[C 14:52:02.752 NotebookApp] Hint: run the following command to set a password
[C 14:52:02.752 NotebookApp]    $ python -m notebook.auth password
```

The notebook.auth.__main__ is script will write to ~/.jupyter/jupyter_notebook_config.json with the code suggestion above. This script is set up with argparse, and it would be nice to also included the ssl setup like in https://github.com/jupyter/notebook/blob/master/tools/secure_notebook.py in this (but I don't want to clutter this PR)

I think the behaviour should also be, if using a non-localhost ip, and password is set, it should refuse to start up without ssl, and spit something out like:
```
[C 14:52:02.752 NotebookApp] Notebook servers are configured with password protection, running on a non-localhost ip, but without ssl/https, this is unsave.
[C 14:52:02.752 NotebookApp] Hint: run the following command to setup https
[C 14:52:02.752 NotebookApp]    $ python -m notebook.auth https
```
(and thus the https/ssl setup should also be in the notebook.auth.__main__ script. )

If agreed, I might put in a new PR in the future for this.

notebook/auth/__main__.py
```diff
@@ -0,0 +1,42 @@
...
+			'password': password_hash,
+		}
+	})
+	if not args.quiet:
+		print("password stored in config dir: %s" % jupyter_config_dir())
+
+def main(argv):
+	parser = argparse.ArgumentParser(argv[0])
+	subparsers = parser.add_subparsers()
+	parser_password = subparsers.add_parser('password', help='create soneira peebles dataset')
```
minrk: copy/paste? soneira peebles dataset?

notebook/auth/__main__.py
```diff
@@ -0,0 +1,42 @@
...
+	parser_password.add_argument("password", help="password to set, if not given, a password will be queried for (NOTE: this may not be safe)",
+			nargs="?")
+	parser_password.add_argument("--quiet", help="Supress messages", action="store_true")
```
minrk: s/Supress/Suppress/

minrk:
@maartenbreddels thanks, this looks very nice, now. You are right that the failure was a gremlin. I've poked Travis to get the gremlins out of there. I found a couple of small typos, but it should be good to go after that.

minrk:
As for the SSL, thanks for not putting it here; It probably does make sense as an alternate PR. It shouldn't refuse to start outright, because running the notebook without SSL doesn't mean that there is *no* ssl, it could be terminated at another level, such as nginx. But we recently made a similar change [to JupyterHub](https://github.com/jupyter/jupyterhub/pull/417), where running without SSL requires additional confirmation.

maartenbreddels:
Thx for finding those sloppy typos, and yes that was a copy paste :) (from own code btw).
Would be nice to have sth similar to the notebook server (the --no-ssl to confirm you really want passwords send over non-ssl).

Btw, do you know if the same assumptions are made with Jupyterhub about localhost being 'safe'. For instance, if I run the default jupyterhub on machine 'foo', and log in, a single user jupyter notebook is launched with my uid right? The proxy does the forwarding, but is there any security involved into this? Of with the right knowledge, and ssh access to machine 'foo', will I be able to connect to other people's notebook?

minrk:
@maartenbreddels in JupyterHub, authentication is handled by single-user servers. The proxy only forwards the connection, it doesn't perform the authentication. So if you bypass the proxy and connect directly to other single-user servers you still can't execute code because you won't be authenticated with that server.

minrk:
Thanks for your patience on this one!

Russell-Jones-OxPhys:
Hi, when will/what is required for this be included in a release? It was merged into master in March with https://github.com/jupyter/notebook/commit/3feea618027713a53e540b6f7eb1e3abef57dc03 but I can't find __main__.py in https://github.com/jupyter/notebook/tree/4.2.1/notebook/auth or password-required in https://github.com/jupyter/notebook/blob/4.2.1/notebook/notebookapp.py

minrk:
@Russell-Jones-OxPhys this will be released with notebook-5.0, which is still probably some months away.

Russell-Jones-OxPhys:
Thanks for the quick response. Useful to know.

PR #782, opened 2015-12-01
More info on front-end extensions

jcb91:
in docs/extending/frontend_extensions:
* Add some description of actions
* beautify javascript examples
* fix inline literals, which in RST start and end with ``, not just `
* fix some typos & misspellings, some improvements to phrasing

### DIFF
--- a/docs/source/extending/frontend_extensions.rst
+++ b/docs/source/extending/frontend_extensions.rst
@@ -14,8 +14,9 @@ The structure of a front-end extension
     to a lot of changes. Any extension written for the current notebook is
     almost guaranteed to break in the next release.
 
-A front-end extension is a JavaScript file that defines an
-[AMD module](https://en.wikipedia.org/wiki/Asynchronous_module_definition)
+.. _AMD module: https://en.wikipedia.org/wiki/Asynchronous_module_definition
+
+A front-end extension is a JavaScript file that defines an `AMD module`_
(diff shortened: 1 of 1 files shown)


docs/source/extending/frontend_extensions.rst
```diff
@@ -5,125 +5,210 @@ This describes the basic steps to write a JavaScript extension for the Jupyter
...
+of what is available:
 
 .. code:: javascript
 
-    Object.keys(IPython.actions._actions)
+    Object.keys(require('base/js/namespace').actions._actions);
 
-In this example, we changed keyboard shortcuts in **command mode**; you can also
-customise keyboard shortcuts in edit mode.
+In this example, we changed the keyboard shortcut in the **command mode**; you
```
takluyver: '...changed *a* keyboard shortcut in command mode' (both instances of 'the' read wrong here)

takluyver:
Thanks, I think this looks good.

In some of our docs, we set `default_role = 'literal'` in the config, so that we can do inline code using single backticks, as we're used to in Markdown. But we haven't actually done that in the notebook docs yet, so you're right to double up the backticks.

docs/source/extending/frontend_extensions.rst
jcb91: good spot, done!

takluyver:
Thanks!

PR #765, opened 2015-11-27
Fixing backspace on multiple selection bug

marceloramires:
The bug could be reproduced in two ways:

**1) Block selection**
a) Alt + select a block containing more than one line, and more than one character per line.
b) Hit backspace

Expected result: all the selected content is deleted
Actual result: only the first line of the selection is deleted

**2) Multiple selection**
a) Select any portion of text
b) Ctrl + select any other portion of text
c) Hit backspace

Expected result: all the selections' contents are deleted
Actual result: only the first selection is deleted

The fix
-------------
The portion of code that took care of selection deletion did not take into account multiple selections. The proposed fix simply gets the list of selections and removes them all from last to first.

closes #766

PR #783, opened 2015-12-02
Fixing way in which selection ranges are deleted

marceloramires:
I had made this fix: https://github.com/jupyter/notebook/pull/765

Which does work, but not exactly how it should -- It raised an exception which would make the control behave normally (which would delete the selections).

I've noticed this now, and made this fix, which tested it a lot, and now it works properly. Hope it's not too late for 4.1.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -53,8 +53,9 @@ define([
          if (sel) { 
             var ranges = cm.listSelections();
             for (var i = ranges.length - 1; i >= 0; i--) {
-              var cur = ranges[i].head;
-              cm.replaceRange("", Pos(cur.line, cur.ch - 1), Pos(cur.line, cur.ch + 1));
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/codecell.js
```diff
@@ -53,8 +53,9 @@ define([
          if (sel) { 
             var ranges = cm.listSelections();
             for (var i = ranges.length - 1; i >= 0; i--) {
-              var cur = ranges[i].head;
-              cm.replaceRange("", Pos(cur.line, cur.ch - 1), Pos(cur.line, cur.ch + 1));
+              var head = ranges[i].head;
+              var anchor = ranges[i].anchor;
+              cm.replaceRange("", Pos(head.line, head.ch), CodeMirror.Pos(anchor.line, anchor.ch));
```
marceloramires: >replaceRange [](start = 17, length = 12)

test comment

PR #786, opened 2015-12-04
4.1 changelog and docs

minrk:
- note some more changes and add screenshots
- regen notebook->rst, and some related fixes


docs/source/examples/Notebook/rstversions/JavaScript Notebook Extensions.rst
```diff
@@ -80,29 +80,27 @@ in the behavior of the notebook.
 custom.js with others.
 
 Back to theory
-==============
+''''''''''''''
```
willingc: Descending to the fifth level in Dante's documentation circles. There's not anything wrong with this line. Leaving this as a reminder to myself to add the project's RST levels to contributor docs ``( =, -, ~, ^, ''')`` since I recently cleaned up a mess of formatting on another project. :fire:

willingc:
@minrk The changes in the source look good to me. I will try to run locally and do a visual check as well though no need to hold up the merge on it. Go ahead and merge.

docs/source/examples/Notebook/rstversions/Notebook Basics.rst
```diff
@@ -174,7 +174,7 @@ Enter command mode by pressing ``Esc`` or using the mouse to click
    </div>
 
 Mouse navigation
-================
+----------------
```
willingc: Not changed in this PR. Though the green buttons in this section are a little confusing when running locally (unless I am missing something locally):

screen shot 2015-12-04 at 8 30 14 am

docs/source/examples/Notebook/rstversions/Typesetting Equations.rst
```diff
@@ -50,7 +50,7 @@ Source
     \end{equation*}
 
 Display
--------
+~~~~~~~
```
willingc: FYI. The raw:: latex in the Display sections are not displaying locally. Not sure if it is a dependency issue locally on my system or something else.

takluyver: rst raw directives are ignored if you're producing anything other than that format (like raw notebook cells). We should make a Sphinx rst template that uses the `.. math::` directive for one of the available [math extensions](http://sphinx-doc.org/ext/math.html). But if this is being produced by pandoc, it might be harder to customise...

willingc: @takluyver I just saw your note here https://github.com/jupyter/notebook/commit/884ef1dc28329741f62978cbd7c1fa5d3c00bde9 before the note on this issue. I think this rst doc is being autogenerated by pandoc. Perhaps a post pandoc processing script to search and replace the ``.. raw::latex`` with ``.. math::``. Hack but might do the trick.

takluyver: Yeah, I think that will probably work.

willingc: @takluyver @minrk OK. It looks like nbconvert has a postprocesser option which would likely be the cleanest place to do the postprocessing and then update the makefiles for the notebook docs to use the postprocessor. Busy today but could take a look at next week or if someone wants to do a quicker fix, please do. Other than this which is not a new issue, this PR is good to merge.

takluyver: I think probably the best place to do it would be in the markdown2rst filter:
https://github.com/jupyter/nbconvert/blob/ca9094c375dfab7adf694d1f6405ea7e17937c43/nbconvert/filters/markdown.py#L71

takluyver: Although the math directive is a Sphinx extension thing, not a normal part of rst. So maybe we should split out a `markdown2sphinx_rst` filter for that. In practical use, though, I think Sphinx is the main reason people want rst.

minrk: Thanks, we can refine the rst output in another PR, but I don't want to block 4.1 with that.

PR #787, opened 2015-12-04
Fix ConfigWithDefaults "this" context

parente:
Within the promise callback, refer to the ConfigWithDefaults instance, not the Promise


parente:
Various cells are using this get() function to fetch custom config, which means their custom config probably isn't working at the moment.

takluyver:
Thanks, I'll merge this once Travis has finished.

parente:
Thanks @takluyver

Issue #738, opened 2015-11-20
Cell selection and commands


Carreau:
Meta issues to list commands that we might want to apply on selection/marked cell. 

- Clear cell output
- change to code/markdown

Carreau:
PIng @ellisonbg @fperez

fperez:
* Copy/Cut/Delete

Carreau:
@parleur Still waiting for your PR. I know it was working.

Carreau:
see #802 #803 #804 #792 #793 #794 #795 #796 #797

PR #790, opened 2015-12-07
Adapt js "to-code" to handle the multi cell selection 

parleur:
For now, only handle cells conversion to codecells. Not perfect with the main menu toolbar.

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -1185,9 +1185,9 @@ define(function (require) {
         this.untocode_backup =[];
         
         for (var i=0; i <indices.length; i++){
-            this.to_code(indices[i])
+            this.to_code(indices[i]);
         }
-    }
+    };
     /**
      * Turn a cell into a code cell.
      * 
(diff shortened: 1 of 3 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -1176,7 +1176,17 @@ define(function (require) {
         var len = this.ncells();
         return this.insert_cell_below(type,len-1);
     };
-
+    
+
```
captainsafia: Can you add a documentation string before the function to describe what it does? You can look at the documentation string above `to_code` for guidance.

captainsafia:
Great PR! Thanks for submitting. I made a quick note about adding documentation.

minrk:
ping @ellisonbg, since this appears to be an action not doing the right thing in multi-select context.

Carreau:
That's part of the various actions that don't apply (yet) to multiple selection (see #738 )

Carreau:
closes #802

parente:
The commits here went in with #809. This can be closed.

Issue #801, opened 2015-12-08
Multi-selection and clear cell output


Carreau:
Should act on selection.

Carreau:
closed by #806

Carreau:
my bad, by #816

Issue #800, opened 2015-12-08
Multi-selection and toggle cell output.


Carreau:
Should act on selection.

Carreau:
affect collapse and scrolled.

Carreau:
Closed by #816

PR #806, opened 2015-12-08
Toggle scroll, collapse and clear cell now apply on selection.

Carreau:
Still need tests.

And as discussed a long time ago, the concept of toggling scrolling and
Collapsed is **not** the right one, it leads to inconsistencies.

The cell output has 3 states:
 - expanded
 - scrolled
 - collapsed

And the menu action should decide which one of the 3 user want.

Otherwise you get into inconsistencies like what is the difference
between collapsed & scrolled /vs/ collapsed & not scrolled.

Partially addresses #800 and #801 (I think)

### DIFF
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -307,22 +307,33 @@ define(function(require){
             }
         },
         'toggle-cell-output-collapsed' : {
-            help    : 'toggle output',
+            help    : 'toggle output of selected cells',
             help_index : 'gb',
             handler : function (env) {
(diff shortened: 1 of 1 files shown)


Carreau:
This made me realize that : https://github.com/jupyter/jupyter-js-output-area/issues/25.

Carreau:
I'll appreciate if someone write some test for this one while I sleep.

captainsafia:
On it. Catch some Zs!

captainsafia:
OK. OMG. Working on this now.

PR #809, opened 2015-12-09
Support multiple cell selections in markdown, raw, and code conversions

parente:
Builds atop PR #790 for issue #802. Addresses the missing function comment. Also handles multiselect in the to_markdown and to_raw functions to fix #803 and #804.

Manually vetted. I can look into adding some tests.

/cc @parleur @Carreau @captainsafia

PR #816, opened 2015-12-10
Add tests for multiselect toggles

captainsafia:
Addresses @Carreau's request for tests for functionality added in #809.

Also refactors the code he added to use the neat function separation that @parleur used.

Not sure if we want all the multiselect tests in one file but I can probably refactor and make that happen.


Carreau:
Thanks !

PR #821, opened 2015-12-10
Fix tests on merge-able cells.

### LINKED ISSUE #820: dualmode_merge test are wrong.
They can't possibly be right as shift-m merge selected cells, and tests never select multiple cells.

Also once I reintroduces the shift-m merge below (again) test are still passing, while they should not.

Carreau:
Add test to check that shift-m does actually do nothing on only 1 cell
selected.

make sure multiple cells get selected when trying to merge a lock cell
(or it's useless test)

Add a test that check 2 cells can be merged.

Closes #820


jdfreder:
moar test moar betta

PR #822, opened 2015-12-10
have shift-m merge below if one cell selected

Carreau:
Need #821 to be merge first, and rebased then to fix the test that #821 fixes that will fail once this is merged.


Carreau:
rebased, tests should fail.

titipata:
Thank you @Carreau! I wait for long for this functionality to be back.

PR #825, opened 2015-12-10
Fix the selectbox celltype behavior when multiselected 

parleur:
PR on top of @parente 's #809 (ping @Carreau, @jdfreder )
Basically when several cells are selected, the selectbox enter in a state '-' hidden otherwise. It allows to select any other cell type and flip all cell in selection to it. 

Thanks to @carreau for help on it.

### DIFF
--- a/notebook/static/notebook/js/maintoolbar.js
+++ b/notebook/static/notebook/js/maintoolbar.js
@@ -95,7 +95,8 @@ define([
             .append($('<option/>').attr('value','code').text('Code'))
             .append($('<option/>').attr('value','markdown').text('Markdown'))
             .append($('<option/>').attr('value','raw').text('Raw NBConvert'))
-            .append($('<option/>').attr('value','heading').text('Heading'));
+            .append($('<option/>').attr('value','heading').text('Heading'))
+            .append($('<option/>').attr('value','multiselect').text('-'));
         this.notebook.keyboard_manager.register_events(sel);
         this.events.on('selected_cell_type_changed.Notebook', function (event, data) {
             if (data.cell_type === 'heading') {
@@ -120,6 +121,8 @@ define([
                 that.notebook._warn_heading();
                 that.notebook.to_heading();
                 sel.val('markdown');
+                break;
+            case 'multiselect':
                 break;
             default:
                 console.log("unrecognized cell type:", cell_type);
--- a/notebook/static/notebook/js/maintoolbar.js
+++ b/notebook/static/notebook/js/maintoolbar.js
@@ -90,6 +90,7 @@ define([
     MainToolBar.prototype._pseudo_actions.add_celltype_list = function () {
         var that = this;
         var multiselect = $('<option/>').attr('value','multiselect').attr('disabled','').text('-');
+        multiselect.hide();
         var sel = $('<select/>')
             .attr('id','cell_type')
             .addClass('form-control select-xs')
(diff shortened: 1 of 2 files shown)


notebook/static/notebook/js/maintoolbar.js
```diff
@@ -89,19 +89,27 @@ define([
...
+            .append($('<option/>').attr('value','heading').text('Heading'))
+            .append(multiselect);
         this.notebook.keyboard_manager.register_events(sel);
         this.events.on('selected_cell_type_changed.Notebook', function (event, data) {
-            if (data.cell_type === 'heading') {
-                sel.val('Markdown');
+            if ( that.notebook.get_selected_cells_indices().length > 1) {
+                multiselect.show();
+                sel.val('multiselect');
+                multiselect.hide();
```
Carreau: I think you want the `multiselect.hide()` in the else part of the branch.

notebook/static/notebook/js/maintoolbar.js
```diff
@@ -89,19 +89,27 @@ define([
...
             } else {
-                sel.val(data.cell_type);
+                if (data.cell_type === 'heading') {
```
jdfreder: @Carreau can correct me if I'm wrong, but I don't think this cell type exists anymore.

Carreau: This is just indented from before, heading technically still exists with keyboard shortcut, it switched to markdown and insert the correct amount of `#`

jdfreder: Got it

notebook/static/notebook/js/maintoolbar.js
parleur: Whaou. I am surprised it works.  You can programatically select a hidden option, it is shown and back hidden when unselected... good to known.

notebook/static/notebook/js/maintoolbar.js
```diff
@@ -89,19 +89,26 @@ define([
     // encountered when building a toolbar.
     MainToolBar.prototype._pseudo_actions.add_celltype_list = function () {
         var that = this;
+        var multiselect = $('<option/>').attr('value','multiselect').attr('disabled','').text('-');
+        multiselect.hide();
```
Carreau: No, not here. Or i is never visible.

notebook/static/notebook/js/maintoolbar.js
```diff
@@ -89,19 +89,26 @@ define([
...
             .append($('<option/>').attr('value','raw').text('Raw NBConvert'))
-            .append($('<option/>').attr('value','heading').text('Heading'));
+            .append($('<option/>').attr('value','heading').text('Heading'))
+            .append(multiselect);
         this.notebook.keyboard_manager.register_events(sel);
         this.events.on('selected_cell_type_changed.Notebook', function (event, data) {
-            if (data.cell_type === 'heading') {
-                sel.val('Markdown');
+            if ( that.notebook.get_selected_cells_indices().length > 1) {
+                sel.val('multiselect');
```
Carreau: put show() line 106,

notebook/static/notebook/js/maintoolbar.js
```diff
@@ -89,19 +89,26 @@ define([
...
             } else {
-                sel.val(data.cell_type);
+                if (data.cell_type === 'heading') {
```
Carreau: and hide() somwhere in this block :-)

Carreau:
Perfect thanks, we can refine later to have a better logic that does the same things only if all the selected cells do not have the same type !

PR #828, opened 2015-12-10
contract selection on click

### LINKED ISSUE #823: Multi-selection, click on cell should contract selection
From Gitter, @SylvainCorlay wrote:

> I don't know if it is something that you have noticed already or is already being worked on: In the new UX for the multiple selection, if you select cells cells from A to B by selecting A as an anchor, pressing shift and selecting B, then we cannot set the selection to be only B by clicking on B only.
I think that the expected behavior is that whatever the current selection is, if you click on a cell without any key modifier, the selection should be cleared and this cell should be the only one selected.

minrk:
During multi-select, clicking on the current cell collapses the selection back to just that cell.

closes #823

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -160,11 +160,14 @@ define([
      * if just codemirror focus (so edit mode).
      * We **might** be able to move that to notebook `handle_edit_mode`.
      */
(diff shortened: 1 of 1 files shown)


jdfreder:
Awesome, tested/works- thanks

PR #830, opened 2015-12-11
Fix move selection up/down

### LINKED ISSUE #793: Move cell up and down inconsistant.

Carreau:
Move the selection by detach/attach previous/next cell, and reselect the
moved selection.

Closes #793 (once test written).

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -586,7 +586,7 @@ define(function (require) {
 
 
     Notebook.prototype.get_selected_cells = function () {
-        return this.get_cells().filter(function(cell, index){ return cell.selected || soft_selected(cell) || cell.anchor})
+        return this.get_cells().filter(function(cell, index){ return cell.selected || soft_selected(cell) || cell.anchor});
     };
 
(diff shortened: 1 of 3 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -840,6 +840,53 @@ define(function (require) {
...
+     * No op if the selection is at the beginning of the notebook
+     */
+    Notebook.prototype.move_selection_up = function(){
+        // actually will move the cell before the selection, after the selection
+        var indices = this.get_selected_cells_indices();
+        var first = indices[0];
+        var last = indices[indices.length-1];
+
+        var i1 = this.get_selected_index();
+        var i2 = this.get_anchor_index();
```
captainsafia: Can these variable names be a little bit more descriptive? Currently, the code is a little bit hard to read.

minrk:
Tested and works as advertised. Great job on 4.1, @Carreau!

PR #832, opened 2015-12-11
Add announce beta on ML. 

Carreau:
How does that sound ? Has anyone anything else to say ?

(also putting this here in case someone else do the release and announce and because  otherwise I'll lose it)


jdfreder:
:+1:

Issue #507, opened 2015-09-29
Error when trying to use certificate: SSLError: [SSL] PEM lib (_ssl.c:2599) 


elanmart:
Hello, following [the docs](http://jupyter-notebook.readthedocs.org/en/latest/public_server.html) I was trying to create a public IPy server. 

However, after creating certificate with 
`sudo openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout mycert.key -out mycert.pem`
and running
`jupyter notebook --certfile=mycert.pem`

The notebook doesn't start and I can't find any information about how to fix it.

The error message is:
```
[E 14:08:49.825 NotebookApp] Exception in callback (<socket._socketobject object at 0x7f021f33e980>, <function null_wrapper at 0x7f021f2cdf50>)
    Traceback (most recent call last):
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/ioloop.py", line 866, in start
        handler_func(fd_obj, events)
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/stack_context.py", line 275, in null_wrapper
        return fn(*args, **kwargs)
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/netutil.py", line 265, in accept_handler
        callback(connection, address)
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/tcpserver.py", line 239, in _handle_connection
        do_handshake_on_connect=False)
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/netutil.py", line 501, in ssl_wrap_socket
        context = ssl_options_to_context(ssl_options)
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/netutil.py", line 478, in ssl_options_to_context
        context.load_cert_chain(ssl_options['certfile'], ssl_options.get('keyfile', None))
    SSLError: [SSL] PEM lib (_ssl.c:2599)
[E 14:08:49.827 NotebookApp] Exception in callback (<socket._socketobject object at 0x7f021f33e980>, <function null_wrapper at 0x7f021f2cdf50>)
    Traceback (most recent call last):
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/ioloop.py", line 866, in start
        handler_func(fd_obj, events)
      File "/home/marcin/anaconda/lib/python2.7/site-packages/tornado/stack_context.py", line 275, in null_wrapper
        return fn(*args, **kwargs)
[...]

PR #844, opened 2015-12-11
Improve message about https in public server doc

willingc:
Relocated and made more prominent the message about using https when using SSL. The issue came up several times recently here #507.

On another note, yay!!!! Nice to see the beta out :balloon: :tada:


willingc:
@Carreau @minrk Seems like this should be in 4.1 docs.

minrk:
An improvement, thanks!

willingc:
Thanks @minrk.

PR #858, opened 2015-12-13
Add contact info more prominently on sidebar and in contents

willingc:
Thinking ahead to the 4.1 beta and future release, we will likely have more user questions on how to use the new features. I have added links to the end of the doc's Table of Contents and sidebar to help users more easily navigate to the googlegroup and jupyter website.

screen shot 2015-12-13 at 1 57 36 pm


captainsafia:
:+1:

willingc:
Promoted the "Installing Javascript Machinery" one level so it is visible on the sidebar.
screen shot 2015-12-13 at 2 22 20 pm

takluyver:
If we're pointing to ways to get support, do we also want a link to http://stackoverflow.com/questions/tagged/jupyter-notebook ?

In an earlier discussion (about links from the IPython website), we decided that pointing people to Stackoverflow was a good idea, because:

- Its design makes it easier to find similar previous questions than a mailing list does, reducing the number of people needing to ask questions.
- Because there's an impression that we 'own' the mailing list, which there isn't for the SO tag, there are more people outside the core team who will answer questions posted on SO, so less of a bottleneck on core developers responding.

willingc:
@takluyver Sure, I think that's a good suggestion. I'm going to drop it in and we can always take it back out if there are objections.

minrk:
Both good additions, thanks!

PR #860, opened 2015-12-14
Test for Pep440 non-Complaint version numbers

Carreau:
test directly in the version number file, as the version number is
likely to be committed before running tests, so we want **really quick
failure**.

---

Among other this broke the release of the sdist, that was on PyPI appearing as more-recent than the wheel as `0.b1` is bigger than `0b1` and the wheel builder auto update `0.b1` to `0b1`

### DIFF
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -1,2 +1,28 @@
-version_info = (4, 1, 0, 'b1')
+"""
+store the current version info of the notebook.
+
+"""
+
+# Downstream maintainer, when running `python.setup.py jsversion`,
+# the version string is propagated to the JavaScript files,  do not forget to
+# patch the JavaScript files in `.postN` release done by distributions. 
+
+# Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**. 
+
+version_info = (4, 1, '0b1')
 __version__ = '.'.join(map(str, version_info))
+
+import re
+
+pep440re = re.compile('^(\d+)\.(\d+)\.(\d+((a|b|rc)\d+)?)(\.post\d+)?(\.dev\d+)?$')
+
+
+
+def raise_on_bad_version(version):
+    if not pep440re.match(version):
+        raise ValueError("Versions String does apparently not match Pep 440 specification, "
+                         "which might lead to sdist and wheel being seen as 2 different release. "
+                         "E.g: do not use dots for beta/alpha/rc markers.")
+        
+
+raise_on_bad_version(__version__)
--- a/notebook/tests/test_notebookapp.py
+++ b/notebook/tests/test_notebookapp.py
@@ -15,6 +15,8 @@ from traitlets import TraitError
 from notebook import notebookapp
 NotebookApp = notebookapp.NotebookApp
 
+from notebook._version import raise_on_bad_version
+
 
 def test_help_output():
     """ipython notebook --help-all works"""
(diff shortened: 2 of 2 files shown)


notebook/_version.py
```diff
@@ -1,2 +1,28 @@
...
+"""
+store the current version info of the notebook.
+
+"""
+
+# Downstream maintainer, when running `python.setup.py jsversion`,
+# the version string is propagated to the JavaScript files,  do not forget to
+# patch the JavaScript files in `.postN` release done by distributions. 
+
+# Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**.
```
takluyver: I see what you mean, but this is rather confusing, since `X.Y.ZbN` does contain dots. Maybe phrase it as "without a dot before the 'b'"

notebook/_version.py
```diff
@@ -1,2 +1,28 @@
...
+version_info = (4, 1, '0b1')
 __version__ = '.'.join(map(str, version_info))
+
+import re
+
+pep440re = re.compile('^(\d+)\.(\d+)\.(\d+((a|b|rc)\d+)?)(\.post\d+)?(\.dev\d+)?$')
+
+
+
+def raise_on_bad_version(version):
```
takluyver: I understand where you're coming from, but my vote would be for this to be in the tests, and encourage people to run them as the very last step before uploading a release, rather than doing the check on every import.

notebook/_version.py
```diff
@@ -1,2 +1,28 @@
...
+
+"""
+
+# Downstream maintainer, when running `python.setup.py jsversion`,
+# the version string is propagated to the JavaScript files,  do not forget to
+# patch the JavaScript files in `.postN` release done by distributions. 
+
+# Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**. 
+
+version_info = (4, 1, '0b1')
```
minrk: This is an invalid version_info. It **must** be three numbers. For the reason why, try:

```python
(4, 1, '0b1') < (4, 1, 0)
```

If there's a fix to be made, it's in how we turn `version_info` into `__version__`, not in the content of `version_info` itself, and frankly I think the fix should be in setup.py, not in specifying the version here. Essentially, the only fix to care about is that setuptools-free sdist upload uses the same string for prereleases as the coerced/normalized version bdist_wheel gets from setuptools.

Carreau: > Essentially, the only fix to care about is that setuptools-free sdist upload uses the same string for prereleases as the coerced/normalized version bdist_wheel gets from setuptools.

I would still like the version from PyPI to match the notebook --version. I can make `__version__` match and keep version_info a 4 tuple.

minrk: Sure. I opened #862 which implements this in setupbase. If you take the same function, you can move it to `_version.py` and everything will be consistent.

notebook/_version.py
```diff
@@ -1,2 +1,28 @@
...
+
+
+def raise_on_bad_version(version):
+    if not pep440re.match(version):
+        raise ValueError("Versions String does apparently not match Pep 440 specification, "
+                         "which might lead to sdist and wheel being seen as 2 different release. "
+                         "E.g: do not use dots for beta/alpha/rc markers.")
+        
+
+raise_on_bad_version(__version__)
```
minrk: this test should be part of the test suite, not run on import

Carreau:
closed in favor of #866

PR #862, opened 2015-12-14
normalize version in setup.py

minrk:
avoids mismatch between wheel/sdist upload on PyPI due to setuptools normalization of prerelease tags.

alternative to #860

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ from distutils.core import setup
 # Our own imports
 
 from setupbase import (
-    version,
+    pkg_version,
     find_packages,
     find_package_data,
     check_package_data_first,
@@ -75,7 +75,7 @@ languages, sharing, and interactive widgets.
 Read `the documentation <https://jupyter-notebook.readthedocs.org>`_
 for more information.
     """,
-    version         = version,
+    version         = pkg_version,
     scripts         = glob(pjoin('scripts', '*')),
     packages        = find_packages(),
     package_data     = find_package_data(),
--- a/setupbase.py
+++ b/setupbase.py
@@ -65,8 +65,27 @@ name = 'notebook'
 version_ns = {}
 execfile(pjoin(repo_root, name, '_version.py'), version_ns)
 
-version = version_ns['__version__']
+def normalized_version(version_info):
+    """Normalize version string
+    
+    Avoids prerelease uploads getting different versions from sdist and bdist,
+    due to setuptools normalization.
+    """
+    v = '.'.join(map(str, version_info[:3]))
+    if len(version_info) == 3:
+        return v
+    extra = version_info[3]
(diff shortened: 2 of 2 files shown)


takluyver:
I like the extra complexity going into setup code rather than the actual modules, but my preference would be that we use PEP whatever compliant version numbers directly, and add some code to assert that they are correct rather than normalising them.

minrk:
@takluyver From #860, I think @Carreau feels the same, which should amount to moving this same function to `_version.py` and producing `__version__` with it.

I don't like the idea that this is forcing us to internally change perfectly valid version numbers that *PEP 440 officially allows and supports*.

Carreau:
I'll have @parleur merge the 2 PRs in one later today, and will re-examine the Pep.

Carreau:
> I don't like the idea that this is forcing us to internally change perfectly valid version numbers that PEP 440 officially allows and supports.

Also I'm not sure I get that : 

> The canonical public version identifiers MUST comply with the following scheme:
>[N!]N(.N)*[{a|b|rc}N][.postN][.devN]

I don't see any way to have a dot followed by a `a`, `b` or `rc`

But I may be missing a part of the Pep, hence why I'll be re-reading.

Carreau:
if you are referring to [this section about prerelease separator](https://www.python.org/dev/peps/pep-0440/#pre-release-separators) I understand it as applying only to software that have already been released, and so should not be used by newer release.

minrk:
We shouldn't do any checking at runtime, since that's an unnecessary waste at import time for extremely static code. The check should be made in setup.py at install/dist time.

If we are going to embed PEP440 normalization in our version strings, , what I'd propose is we put the '.' in the suffix, as the PEP describes them:

```python
version_info = (4, 1, 0, '.dev0')
__version__ = '.'.join(version_info[:3]) + ''.join(version_info[3:]
```

Carreau:
One of the issues with `version_info = (4, 1, 0, '.dev0')` is it will most of the time be `(4,1,0)` which will lead to us forgetting the `.` on next release. 

Would you version plus a normalisation in setup.py and a unit-test  be an acceptable compromise? 

Also the `'.'.join`can be seen as an extra-expensive runtime computation :-P

minrk:
> One of the issues with version_info = (4, 1, 0, '.dev0') is it will most of the time be (4,1,0) which will lead to us forgetting the . on next release.

No, 99.9% of the time it will be `version_info = (4, 1, 0, '.dev0')`. It's only `(4, 1, 0)` for one commit, and the next commit is always putting it back to `'.dev0'`.

minrk:
> Would you version plus a normalisation in setup.py and a unit-test be an acceptable compromise?

I think testing in tests is perfect. I only mean that we shouldn't be running tests on import, which is currently done in #860.

PR #866, opened 2015-12-14
Use PEP 440 prerelease tags

### LINKED ISSUE #842: beta sdist filename with extra dot.
sdist beta filename : notebook-4.1.0.b1.tar.gz

according to pep404 there should be no dot between `0` and `b1`. (does this affect upload on PyPI ?

parleur:
Fusion between #860 & #862. Test the version number correctness in unit test, normalize it when exec setup.py

closes #842

### DIFF
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -11,14 +11,3 @@ store the current version info of the notebook.
 
 version_info = (4, 1, 0, 'b1')
 __version__ = '.'.join(map(str, version_info[:3])) + ''.join(version_info[3:])
-
-import re
-
-pep440re = re.compile('^(\d+)\.(\d+)\.(\d+((a|b|rc)\d+)?)(\.post\d+)?(\.dev\d+)?$')
-
-def raise_on_bad_version(version):
-    if not pep440re.match(version):
-        raise ValueError("Versions String does apparently not match Pep 440 specification, "
-                         "which might lead to sdist and wheel being seen as 2 different release. "
-                         "E.g: do not use dots for beta/alpha/rc markers.")
-        
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -1,2 +1,28 @@
-version_info = (4, 1, 0, 'b1')
+"""
+store the current version info of the notebook.
+
+"""
+
+# Downstream maintainer, when running `python.setup.py jsversion`,
+# the version string is propagated to the JavaScript files,  do not forget to
+# patch the JavaScript files in `.postN` release done by distributions. 
+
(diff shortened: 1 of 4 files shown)


notebook/_version.py
```diff
@@ -1,2 +1,25 @@
...
+
+"""
+
+# Downstream maintainer, when running `python.setup.py jsversion`,
+# the version string is propagated to the JavaScript files,  do not forget to
+# patch the JavaScript files in `.postN` release done by distributions. 
+
+# Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**. 
+
+version_info = (4, 1, '0b1')
```
minrk: version_info should be `(4, 1, 0, 'b1')`

notebook/_version.py
```diff
@@ -1,2 +1,25 @@
...
+__version__ = '.'.join(map(str, version_info[:3])) + ''.join(version_info[3:])
+import re
```
minrk: from this line down should move to the test file

setupbase.py
```diff
@@ -65,8 +65,27 @@ def execfile(fname, globs, locs=None):
 version_ns = {}
 execfile(pjoin(repo_root, name, '_version.py'), version_ns)
 
-version = version_ns['__version__']
```
minrk: The changes in this file are no longer required, since `_version.py` is setting `__version__` to a normalized value.

Carreau:
I took care of a few of the lasts comments.

notebook/_version.py
```diff
@@ -1,2 +1,24 @@
...
+# Downstream maintainer, when running `python.setup.py jsversion`,
+# the version string is propagated to the JavaScript files,  do not forget to
+# patch the JavaScript files in `.postN` release done by distributions. 
+
+# Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**. 
+
 version_info = (4, 1, 0, 'b1')
-__version__ = '.'.join(map(str, version_info))
+__version__ = '.'.join(map(str, version_info[:3])) + ''.join(version_info[3:])
+
```
minrk: from this line down can be removed

minrk:
One more note about lingering unused code in `_version.py`, but otherwise +1.

notebook/_version.py
Carreau: Oops. I took care of it too.

Carreau:
> One more note about lingering unused code in _version.py, but otherwise +1.

Thanks for your patience !

minrk:
Thanks!

PR #868, opened 2015-12-14
Disable prompt number selection

SylvainCorlay:
When using multiple selection, I always end up selecting the prompt number by clicking on it, which looks ugly.

I don't think that we want it to be selectable, even when not using multiple selection.

### DIFF
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -63,13 +63,26 @@ div.cell {
 .prompt {
     /* This needs to be wide enough for 3 digit prompt numbers: In[100]: */
     min-width: 14ex;
+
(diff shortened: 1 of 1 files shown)


Carreau:
Fair enough, but you will still select markdown cell with shift-click, I did not found a **nice** way to avoid that.

SylvainCorlay:
Note, I also now use the default mouse pointer instead of the text selection pointer.

@Carreau I only disable selection on the prompt number where we tend to click for the multiple selection...

Carreau:
> I only disable selection on the prompt number where we tend to click for the multiple selection...

The comment was meant to be just a remark that we might want to fix later. 
I'm going to merge this. Thanks.

PR #876, opened 2015-12-16
Multi-select: avoid adding a css class to body element

SylvainCorlay:
This should not change the behavior for the user. It preserves the current behavior without resorting to adding a class to the top `body` element.

cc @Carreau @ellisonbg


Carreau:
+!

PR #877, opened 2015-12-17
Hide vertical scrollbar on rendered content by default.

### LINKED ISSUE #735: Markdown mode - Inoperable scrollbars
When I type in markdown mode, I sometimes get the scrollbar on the right, which can be disturbing, i.e.:

    $$A_{zvar} =\sum{A_{i, zvar}}$$

returns the following:

scrollbar

Is this a bug or can I avoid this somehow?

jdfreder:
Content in the y direction should cause the rendered content to grow,
the overflow-x: auto causes chrome to assume the same of y, so we need
to tell it explicitly otherwise.

closes #735

### DIFF
--- a/notebook/static/notebook/less/textcell.less
+++ b/notebook/static/notebook/less/textcell.less
@@ -37,6 +37,11 @@ h1,h2,h3,h4,h5,h6 {
 
 .text_cell.rendered .rendered_html {
     overflow-x: auto;
+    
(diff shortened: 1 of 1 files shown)


captainsafia:
Works on Safari 9.0.2 and Chrome 47 on Mac. :+1:

Cookie for @jdfreder, :cookie:!

Carreau:
Seem legit.

PR #883, opened 2015-12-18
include both sdist formats

minrk:
when making releases


willingc:
+1 @minrk I will loop back later today and add this to the making a release docs.

Carreau:
Want to add `xz` for lzma ?

minrk:
> Want to add xz for lzma?

I don't.

PR #896, opened 2015-12-23
Restore old green for edit mode

Carreau:
Mitigate #892

Ping @ellisonbg

### DIFF
--- a/notebook/static/notebook/less/variables.less
+++ b/notebook/static/notebook/less/variables.less
@@ -22,6 +22,6 @@
 @soft_select_color:             #E3F2FD;
 
 
-@edit_mode_border_color:        #66BB6A;
+@edit_mode_border_color:        green;
 @cell_padding:                  6px;
 @cell_border_width:             1px;


ellisonbg:
I am strongly opposed to this - if makes our bad green even more prominent with the thick left border. It has the same problems as master, just a worse green, that doesn't match the visual design of the blues we are using.

The original design we agreed upon took a ton of work to find consensus. Please implement that consensus for now. Then if you want to propose something else and get everyone on board you can.

We *are* changing the visual design of cells in 4.1 and I don't see any way to avoid that. For example we and everyone else says "command mode has a Grey border" but that is no longer true either. 

Bad visual design at this level is a bug and in most cases we should fix it in in minor releases

Keep in mind that none of these major new UI features should have been done in 4.1, so all of this is a compromise at this point.

Carreau:
> The original design we agreed upon took a ton of work to find consensus. Please implement that consensus for now. Then if you want to propose something else and get everyone on board you can.

I'm sorry I was trying to address what you said in #892 : 

> We have changed the green to a slightly brighter color to match the Material Design blues used in the multiple select UI.
> The new design has "even more green" in the left border.
> Between the green 5px left border and brighter green, it is just too much for a UI element that gets shown to the user that often.

I understood that as : "The old green was better", thus I thought this would please you and did that only for that reason. Apologies if I misunderstood. 

> The original design we agreed upon took a ton of work to find consensus. Please implement that consensus for now.

For me the **current** style is the consensus, and you ask in #892 to remove the green border for the cell. So your proposal in #892 is changing the style we agreed upon. 

(Sidenote: I, as a person, would prefer a notebook style with only the left indicator and no border. Though **not changing that** is what we agreed upon. We agreed to **only add a left indicator** for 4.1)


> We *are* changing the visual design of cells in 4.1 and I don't see any way to avoid that. For example we and everyone else says "command mode has a Grey border" but that is no longer true either.

Hum, my cells on 4.1rc still have grey border. 
The blue indicator on the left is in the background:
```
div.cell.selected {
    border-color: #ababab; // this is gray. 
    // this is blue and then transparent. 
    background: linear-gradient(to right, #42A5F5 -40px, #42A5F5 5px, transparent 5px, transparent 100%);
}
```

> Bad visual design at this level is a bug and in most cases we should fix it in in minor releases

Agreed, but not at the expense of breaking user habits, all tutorials, and books published. E.g. @jdfreder videos. 

> Keep in mind that none of these major new UI features should have been done in 4.1, so all of this is a compromise at this point.

Yes, and 4.1 have been slow changing for a few weeks, had a **beta** and a **rc** cycle which were already including the compromise style we agreed upon. And you are asking major UI changes 24 to 48 hours from a potential release.

I'm open to accept **some changes**, because I deeply trust you, and you are someone sensible to design. But as a maintainer it's really hard lang huge change that close to reelease.

ellisonbg:
Please do not release over a holiday - there are too many people away from the computer right now. There may not even be enough people around next week...

Carreau:
> Please do not release over a holiday - there are too many people away from the computer right now. There may not even be enough people around next week...

Then please, stop pushing people toward designing things for 4.2 and 5.0.

PR #897, opened 2015-12-23
Improve warning on bad JavaScript API usage.

Carreau:
Mitigate #891 (likely close it to, depending on what others thinks)

Separator between prefix and name for commands/action used to be `dot`,
but is internal detail. It is now (as of 4.1) a semicolon (`:`)

If someone try to access an action by name which does not have `:` warn
that they might be using the wrong API and give some hints


minrk:
Thanks for your deprecation messages.

PR #899, opened 2015-12-24
Do not violently remove `config_dir` kwarg of config manager.

Carreau:
Pr #882 removed `config_dir` kwarg as it was ignored.

Though, #893 argues that we might need it and that it will be complex to
reintroduce it.

This use a convoluted way to :

 - warn that this kwarg might not be given to config managers in future version
 - but do not warn on default installs with default config.
 - warn that the keyword is ignored when people use subclasses of out
   config manager, and pass the keyword to super().

Thus this actually advertise that the keyword **might** be removed,
by still allowing us to keep it if in the end it appears that we need
it.

Should help with #893 without un-fixing #882.

-- 
It annoys me to go to that much convolution and weird things that close to final release, but if it's the cost of actually moving forward so be it. 

Would that please both @ellisonbg  and @damianavila ?

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -23,6 +23,7 @@ import ssl
 import sys
 import threading
 import webbrowser
+import warnings
 
 
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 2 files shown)


damianavila:
>Pr #882 removed config_dir kwarg as it was ignored.

It was remove primarily because it was "forcing" the `config_dir` at the notebookapp level...

>Though, #893 argues that we might need it and that it will be complex to
reintroduce it.

We don't need the `config_dir` option at the ConfigManager class, we just need to avoid the enforcement at the notebookapp level. I guess nobody was using the ConfigManager.config_dir option before because it was not working... but we don't need that option

My believe is that the `config_dir` should be settled at the ConfigManager level, not passing that one from the notebookapp level... this is why I submitted 882.
My comment on https://github.com/jupyter/notebook/issues/893#issuecomment-167099432 should not also resolve the issue?

notebook/notebookapp.py
```diff
@@ -825,10 +826,28 @@ def init_configurables(self):
...
+            # In case it is a custom ConfigManager,  
+            # for potentially not breaking future backward 
+            # compatibility, will still pass it. 
+            if issubclass(self.config_manager_class, ConfigManager):
+                warnings.warn('Object `%s` will be passed a `config_dir` kwarg, which might be removed'
+                              'in future Notebook versions. Please make sure to handle this case.' % (self.config_manager_class),)
+            self.config_manager = self.config_manager_class(
+                parent=self,
+                log=self.log,
+                config_dir=os.path.join(self.config_dir, 'nbconfig'),
```
damianavila: This is the whole thing... we **don't** need the `config_dir` from the notebookapp for the custom ConfigManager... this is my simple implementation of a CustomConfig manager: https://github.com/Anaconda-Server/nb_config_manager/blob/master/nb_config_manager/nb_config_manager.py

Since the `config_dir` is "calculated" inside the ConfigManager class (by the `_config_dir_default`), my subclass take that info and make further modifications from there... so, we are happy with just the if branch above, where the `config_dir` is not harcoded at the notebook level... maybe I am not clear enough in my explanations... sorry about that.

Check my comment in https://github.com/jupyter/notebook/issues/893#issuecomment-167099432, with that diff you keep the things as now (I mean, not config option at the ConfigManager and the default config pointing to the user space), but you "hardcode" the config_dir at the ConfigManager level, so we can then subclass and just setup the config_dir to our desires without touching the notebookapp at all...

Issue #916, opened 2015-12-31
File save hooks—docs & question


khaeru:
`notebook/docs/source/extending/savehooks.rst` gives an example of a post-save hook for exporting a script, one that I find very useful! Two issues and a question about this:

1. As far as I can tell, this is code from elsewhere in the code base that could be just imported, e.g. a user can set the hook in `jupyter_notebook_config.py` with:

   ```python
   from notebook.services.contents.filemanager import _post_save_script
   c.FileContentsManager.post_save_hook = _post_save_script
   ```
2. The example in `savehooks.rst` is out of date—flake8 warns that `py_fname` is defined but not used.
3. In using this code, the `ScriptExporter` that's created seems to eventually create a `PythonExporter` (at least, for IPython notebooks). It seems not possible to configure this Exporter through the normal configuration mechanism. For instance adding any of the following lines to `jupyter_notebook_config.py` seems to have no effect:

   ```python
   c.Exporter.template_file = 'custom.tpl'
   c.TemplateExporter.template_file = 'custom.tpl'
   c.PythonExporter.template_file = 'custom.tpl'
   ```

   Some of these appear in the [nbconvert docs](http://nbconvert.readthedocs.org/en/latest/config_options.html), but that is for `jupyter_nbconvert_config.py`, not loaded by the notebook. I've also tried hacking `_post_save_script` to pass or set the template_file or template_path options, and looked in vain for documentation about configuring custom exporters for `ScriptExporter`. (My ultimate goal here is to use a custom .tpl file for this export so that I can suppress prompt numbers.)

Carreau:
the post_save_hook is any import string or callable, which mean that in your `notebook_config.py` you can import and instantiate the ScriptExporter yourself, as well as pass a config object (for custom templates) yourself: 

```
from nbconvert.exporters.script import ScriptExporter
from traitlets.config import Config
c = Config()
ScriptExporter(config=c)
...
```

The `post_save_hook` is not passed any config instance, though it is passed `contents_manager=self` so you should be able to use `c.FileContentsManager.(Script)Exporter.template_file = 'custom.tpl'` (`Script` beeing optional) in `jupyter_notebook_config.py` (note the extra `FileContentsManager.(Script)Exporter` nesting. But we should check that.

michaelaye:
Could someone confirm/comment if point 1. made above is a safe way to implement the post_save_hook?

Carreau:
[...]

PR #925, opened 2016-01-05
Be more explicit about deprecation.

Carreau:
Cf #916. Despite the fact that the function is private and meant
to replace an old behavior that will be removed, people explicitly
import it and register it as a post-save-hook.
Hope that explicit version removal will convey the "Don't use" in a more
obvious manner.

-- 

I'm ok if this is not in 4.1

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -7,6 +7,7 @@
 import io
 import os
 import shutil
+import warnings
 import mimetypes
 import nbformat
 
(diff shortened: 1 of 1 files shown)


willingc:
Reminder to myself to add deprecation notice to docs after merge.

Carreau:
if it's deprecated it should not be in the docs anyway. 

Sphinx should also have (apparently) a `deprected-remove` directive that deals with marking things for deprecation and removal after some time.

willingc:
Agreed. If it is in the docs already, we should mark it deprecated and remove after some time.

PR #931, opened 2016-01-06
Add simple iopub message rate limiter

jdfreder:
This is designed to alleviate notebook crashes due to an excess of messages passed to the frontend.  This is a small part of a solution to the bigger problem.  I mainly opened this PR to help students in @ellisonbg 's class this quarter avoid crashing their notebooks and losing data.


ellisonbg:
Thanks! Can you describe the approach you are using to limit things? Some possibilities:

* Just raw number of messages (not per time)
* Message number rate (per time)
* Aggregate message size
* Aggregate message size rate (per time)

minrk:
Also see #506

Carreau:
I also prefer #506. Most of the crashing are often also related to browser freeze, and we have already had POC of JS extension that would delay appending to the DOM in case of too many messages with a UI like: "Are you sure you like to append ? `append`/`discard`".

Plus a client side fixing would solve the problem at loading time when a notebook file is too big.

minrk:
The server side and client-side pieces of #506 are completely unrelated to each other, so we can have both: js patches from #506 and Python patches from #931, if they are preferred to #506.

ellisonbg:
In a classroom/tutorial setting, where I have a bunch of people on a
semi-fragile network, stopping things at the server side is pretty
important. The way these things go, all users tend to do this at around the
same time as they work on the same thing.

Also, I find that this stuff is *super* easy to trigger, even for
experienced people who are aware of it. I crashed by browser 3-4 times this
weekend getting ready for class. In *all* of those cases, server side
truncation would have been the most useful thing - there was too much data
to actually look at reasonably.

On Wed, Jan 6, 2016 at 5:30 AM, Matthias Bussonnier <

> I also prefer #506 <https://github.com/jupyter/notebook/pull/506>. Most
> of the crashing are often also related to browser freeze, and we have
> already had POC of JS extension that would delay appending to the DOM in
> case of too many messages with a UI like: "Are you sure you like to append
> ? append/discard".
>
> Plus a client side fixing would solve the problem at loading time when a
> notebook file is too big.
>
> —
> <https://github.com/jupyter/notebook/pull/931#issuecomment-169322970>.
>



-- 
Brian E. Granger

minrk:
@ellisonbg it's most likely that it was putting outputs on the page that was the real cost (especially if the symptom was a hung browser) rather than the network traffic, since the way we append stream output is *extremely* inefficient. This is the primary issue addressed in the js side of #506. If the network is slow or clogged, that should actually make it *harder* to crash the browser.

I think the scalable approach is to dump excess output to files on the server side, but that adds some peculiarities, since output is part of the kernel API, not necessarily related to notebooks (e.g. thebe has no file or path associated with the output messages). Because of this, it seems to me like overflow truncation should look like:

1. part of the Kernels service
2. overflowed output is associated with kernels, with a file per parent (i.e. output area)
3. overflowed output is only accessible while the kernel exists, and is deleted on shutdown of the kernel and/or server
4. On overflow, OutputArea presents a control that allows one-time, explicit "Yes I want this" fetch of overflowed output into the notebook.

If we want this really scalable, the only *true* answer is for notebooks to actually be multiple files, so we don't have to read the whole thing at once, and can scroll within outputs, having only a limited amount of content available in memory at a time. I don't think this is feasible with the current notebook format, though.

ellisonbg:
In the stuff I was doing there were two usability issues:

1. The fact that the browser crashed.
2. The fact that even if the browser didn't crash, the amount of data that would end up in the notebook wouldn't have been useful to me as a human in that form. When I get into this situation, it would be much more helpful to have the data in a file I could open in a text editors, tail, cat, etc.

@minrk I like the ideas you list above for truncation and saving things to files.

minrk:
It's the case that you described about saving long output to a *persistent* file that I think will be hard to fit into the APIs we have. Where would you put it? When, if ever, would you clean it up? If it's associated with a notebook (it isn't in the APIs), then this isn't so hard, because it can be cleaned up on clear-output and should reside next to the notebook. Since notebooks are really single files, when we have side files associated with notebooks (as opposed to kernels) like this, we need to add logic for handling them to file delete/rename/copy/checkpoint/restore/etc.

But I will reiterate that it's putting a large amount of output *on the page* that ~always causes the problem. Its mere existence, either in the notebook json or in memory, has a *much* higher bar for causing trouble if it isn't being displayed.

ellisonbg:
I would be fine with limiting the lifetime of those files to that of the kernel

jdfreder:
> Also see #506

Sorry, I should learn to use the search.  Brian had mentioned you had something open, but implied it was opened a while ago so I assumed it had fallen through the cracks.

Pinging @williamstein since he is interested in both this PR and #506.

dsblank:
Related topics from more than a year ago:

* https://github.com/ipython/ipython/issues/6516
* https://github.com/ipython/ipython/issues/6771

This is really a big issue in the classroom.

notebook/notebookapp.py
```diff
@@ -785,9 +790,20 @@ def _notebook_dir_changed(self, name, old, new):
         help="Reraise exceptions encountered loading server extensions?",
     )
 
+    iopub_msg_rate_limit = Float(config=True, allow_none=True, help="""(msg/sec)
+        Maximum rate at which messages can be sent on iopub before they are
+        limitted.""")
```
ellisonbg: sp. limited

jdfreder:
This will now print a message to stderr so the user can see when they exceed a limit.  @ellisonbg this is ready for you to test.  I've changed the default interval to 0.1 sec, and have not set any default msg or data rate limits.  However, in testing, it seems like the data rate limit is more valuable than the msg rate limit.  Setting `--NotebookApp.iopub_data_rate_limit=13000` (13k characters/sec) seems to work well.

notebook/services/kernels/handlers.py
```diff
@@ -182,7 +195,14 @@ def initialize(self):
...
         self.kernel_info_channel = None
         self._kernel_info_future = Future()
-    
+
+        # Rate limiting code
+        self._iopub_msg_count = 0
+        self._iopub_byte_count = 0
+        self._iopub_msgs_exceeded = False
+        self._iopub_data_exceeded = False
+        self._last_limit_check = time.time()
```
minrk: Should use `IOLoop.current().time()` for all of these timestamps

notebook/notebookapp.py
```diff
@@ -785,9 +790,20 @@ def _notebook_dir_changed(self, name, old, new):
         help="Reraise exceptions encountered loading server extensions?",
     )
 
+    iopub_msg_rate_limit = Float(None, config=True, allow_none=True, help="""(msg/sec)
```
minrk: Probably use `0` for the no-limit indicator. This is what we do elsewhere, so we should be consistent.

minrk:
I like this server-side limit better than mine in #506. I will remove the server-side limit there, but we should still use the client-side js changes in that PR.

notebook/services/kernels/handlers.py
```diff
@@ -244,6 +264,85 @@ def on_message(self, msg):
...
+                    self.log.warn("iopub messages resumed")
+                
+            # Increment the bytes and message count
+            self._iopub_msg_count += 1
+            self._iopub_byte_count += sum([len(x) for x in msg_list])
+            
+            # If the elapsed time is equal to or greater than the limiter
+            # interval, check the limits, set the limit flags, and reset the
+            # message and data counts.
+            if elapsed_time >= self.limit_poll_interval:
```
minrk: If poll interval is 1s, this means that I can exceed the rate limit, as long as I do it before then. There should probably be an additional condition checking whether the rate for the interval has already been exceeded:

```python
min_msg_rate = float(self._iopub_msg_count) / self.limit_poll_interval
min_data_rate = float(self._iopub_data_count) / self.limit_poll_interval

if (
    elapsed_time >= self.limit_poll_interval
    or min_data_rate >= self.iopub_data_rate
    or min_msg_rate >= self.iopub_msg_rate
):
```

In fact, maybe elapsed_time shouldn't be considered at all, just considering the total for a given window on every message.

minrk: In other words, unless we check the rates on every message, it will always be possible to exceed them with a single message.

jdfreder: Sure, I think that is a good idea.  What I ended up doing was changing the rate poll interval to a rate checking window.  I no longer reset the message or byte counts, instead I keep two queues for changing the counts sometime in the future.  Whenever I change the counts, I also add the opposite to the queue timed for now + window.  See 69faf0f

jdfreder:
Thanks for the review @minrk , see 7232a6a .  I'm going to test this to make sure it works in the current state, I'll ping back with the status.

jdfreder:
Everything seems to work **but** it does rate limit **all** iopub messages.  Do you think I should add logic to avoid limiting certain iopub messages?

blink1073:
Yes, please do not filter `status` messages, since they are used by the client.

blink1073:
Add `shell_reply` to that as well.

jdfreder:
> Yes, please do not filter status messages, since they are used by the client.

:heavy_check_mark: 

In 7aaccbd I don't filter status, kernel info, comm info, and exec request msgs.

notebook/services/kernels/handlers.py
```diff
@@ -244,6 +266,87 @@ def on_message(self, msg):
...
+            msg = self.session.msg("stream",
+                content={"text": error_message, "name": "stderr"},
+                parent=parent
+            )
+            msg['channel'] = 'iopub'
+            self.write_message(json.dumps(msg, default=date_default))
+            
+        channel = getattr(stream, 'channel', None)
+        msg_type = json.loads(fed_msg_list[1])['msg_type']
+        if channel == 'iopub' and msg_type not in ['status', 'comm_info_reply', 'kernel_info_reply', 'execute_input']:
```
minrk: use a set literal for `in` checks. And `kernel_info_reply` cannot come on the iopub channel.

jdfreder: Thanks, I checked and `comm_info_reply` isn't iopub either, but `comm_open` is, which probably shouldn't be filtered.

notebook/notebookapp.py
```diff
@@ -187,7 +187,12 @@ def init_settings(self, ipython_app, kernel_manager, contents_manager,
             },
             version_hash=version_hash,
             ignore_minified_js=ipython_app.ignore_minified_js,
-            
+
+            # rate limits
+            iopub_msg_rate_limit=ipython_app.iopub_msg_rate_limit,
+            iopub_data_rate_limit=ipython_app.iopub_data_rate_limit,
+            limit_window=ipython_app.limit_window,
```
minrk: sorry for being pedantic, but perhaps call this `rate_limit_window`, since in the config context it's not clear what would be limited.

jdfreder: No, don't be sorry, it's good!  

:heavy_check_mark:  9e2c95d

minrk:
I opened a PR against your branch to fix the failures on Python 3.5.

jdfreder:
Thanks!  Merged

Carreau:
What's the status ?

jdfreder:
Good to go, I merged @minrk 's commit

jdfreder:
Probably just fell bellow the radar...

PR #935, opened 2016-01-08
Remove deprecated usage of `log.warn` prefer `log.warning`.

Carreau:
Treating just this notebookapp for now.

-- 
I'd like to find a nice way to turn some DeprecationWarning (and other warings) into errors using nose, without using a wrapper script, but it seem annoying.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -123,7 +123,7 @@ class DeprecationHandler(IPythonHandler):
             console.warn('`/static/widgets/js` is deprecated.  Use `nbextensions/widgets/widgets/js` instead.');
             define(['%s'], function(x) { return x; });
         """ % url_path_join('nbextensions', 'widgets', 'widgets', url_path.rstrip('.js')))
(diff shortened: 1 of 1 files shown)


takluyver:
It looks like someone wrote a nose plugin for that:

https://github.com/Bernhard10/WarnAsError

I don't know how robust it is, but it's pretty simple.

Carreau:
Yeah I saw that, but it turns **all** into error, I want to actually filter. 

I have the plugin working, I need to figureouthow to make entry point with flit.

takluyver:
http://flit.readthedocs.org/en/latest/entrypoints.html :-)

Carreau:
> http://flit.readthedocs.org/en/latest/entrypoints.html :-)

Yeah i found it, but nose does not seem to find the plugin when I do that,

Carreau:
see https://github.com/Carreau/nose_warnings_filters/issues/1

Carreau:
Can we merge that and workout testing later ?

PR #937, opened 2016-01-08
fix dev notation and regex.

Carreau:
:-(


willingc:
@Carreau Since I think this is what is making Travis grumpy on tests choking on version, I'm going to merge this.

PR #938, opened 2016-01-08
Update doc for 4.1 announcement

willingc:
@Carreau Updating the docs for 4.1 official announcement. Congrats!

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,7 +6,7 @@ Jupyter notebook changelog
 A summary of changes in the Jupyter notebook.
 For more detailed information, see `GitHub <https://github.com/jupyter/notebook>`__.
 
-.. _release-4.1.x:
+.. _release-4.1.0:
 
 4.1.0
 -----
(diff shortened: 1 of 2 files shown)


docs/source/changelog.rst
```diff
@@ -6,7 +6,7 @@ Jupyter notebook changelog
 A summary of changes in the Jupyter notebook.
 For more detailed information, see `GitHub <https://github.com/jupyter/notebook>`__.
 
-.. _release-4.1.x:
```
takluyver: Do we want a label for every release, or just for every feature release? Should we leave a 4.1.x label in as well so we can always cross-link to that section?

willingc: @takluyver We certainly could leave 4.1.x. I'm cool to do whatever you and @Carreau think is best  since you have a better handle on frequency of releases and schedules.

willingc:
Going to merge this since the announcement is out and Travis keeps failing on one js test and its different.

docs/source/changelog.rst
Carreau: I think in the long run 4.1.x is enough, we can keep 4.1.0 also, and have 4.1.1 related news in under 4.1.0 link.

Issue #928, opened 2016-01-05
Update `log.warn` to `log.warning`


Carreau:
Identical to ipython/ipython#9091

As said in the documentation : [`warn` is a deprecated alias](https://docs.python.org/3/library/logging.html?highlight=logging#logging.Logger.warning), we should try not to use it. 

`log.warn` seem to be used in the following files:

```
   1 notebook/services/api/handlers.py
   1 notebook/services/security/handlers.py
   1 notebook/services/sessions/handlers.py
   2 notebook/services/contents/handlers.py
   3 notebook/base/handlers.py
   3 notebook/services/contents/filemanager.py
   3 notebook/services/contents/manager.py
   3 notebook/services/kernels/handlers.py
   3 notebook/services/kernels/kernelmanager.py
   3 setupbase.py
   7 notebook/base/zmqhandlers.py
  10 notebook/notebookapp.py
```

As in other issue bonus point to add a warnings filter that turn usage of `log.warn` in the package into an error in CI.

diego898:
Hello,

I'd like to try the bonus @Carreau mentioned in ipython/ipython#9091, though he said that it would likely be hard. Any other advice to get started?

Carreau:
See https://github.com/ipython/ipython/pull/9110 I tried but it does not work, I don't know why. You can try to figure it out !

takluyver:
Reading up on the [warnings module](https://docs.python.org/3/library/warnings.html) is a good place to start. We need to make quite a specific filter, so it doesn't fail because of other warnings.

I'm not quite sure how we'd do it for this repo - for ipython we have our own test runner (iptest), so it's easier to set up things like that before the tests start. For all the other repos, we use [nose](https://nose.readthedocs.org/en/latest/) - there's probably some way of setting up things like this before any code runs, but I'm not sure exactly what that is.

diego898:
ok Ill take a look at ipython/ipython#9110 first, as it seems more direct. Thanks!

Carreau:
> I'm not quite sure how we'd do it for this repo - for ipython we have our own test runner (iptest), so it's easier to set up things like that before the tests start. For all the other repos, we use nose - there's probably some way of setting up things like this before any code runs, but I'm not sure exactly what that is.

For nose, see this PR #940 , I wrote a nose extension to do that (I need to rebase the PR though)

JamiesHQ:
[...]

PR #940, opened 2016-01-09
show warnings during tests

Carreau:
There was no nose plugin that I can find to enable warnings selectively, so [I wrote one](https://github.com/Carreau/nose_warnings_filters).

This is to complement issues like https://github.com/jupyter/notebook/issues/928, where we get rid in our codebase of `log.warn`, and turn the warning into an error not to reintroduce it.

I guess `nose_warnings_filters` would need to be a bit more stable before we use it, but it's a bit a chicken and egg problem. It can stabilise if it's not used. So advice appreciated.

### DIFF
--- a/notebook/services/contents/tests/test_contents_api.py
+++ b/notebook/services/contents/tests/test_contents_api.py
@@ -17,7 +17,7 @@ from ..filecheckpoints import GenericFileCheckpoints
 from traitlets.config import Config
 from notebook.utils import url_path_join, url_escape, to_os_path
 from notebook.tests.launchnotebook import NotebookTestBase, assert_http_error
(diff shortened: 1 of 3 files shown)


notebook/files/handlers.py
```diff
@@ -49,16 +53,28 @@ def get(self, path, include_body=True):
                     self.set_header('Content-Type', 'application/octet-stream')
                 else:
                     self.set_header('Content-Type', 'text/plain')
+<<<<<<< 07b2d80453090d7427409b997d0566a136b8492e
```
jdfreder: whoops! :astonished:

Carreau: Finaly rebased :-)

Issue #471, opened 2015-09-21
Support for moving files


mrares:
This is a feature request, for usability reasons it would be beneficial if users could move files between Folders in the file view, allowing them to organise their notebooks after creation.

hosemn:
I ran into this lately, also, but discovered a work-around (non-ideal, but usable when SSH is not an option or the notebook server is not local): rename the notebook from the UI, but put the desired directory path in front of the name. The notebook server will relocate the notebook into the directory. I also think it would be nice to have a UI element for moving a notebook, but for now you should be able to work around it using this.

mrares:
Yes, that is one very creative solution, I will use it as a workaround for now.

NHDaly:
I'm working on a pull request now to add moving files. :)

SandboChang:
I have another solution, which may not be ideal in terms of data security but is easy to implement:
Simply link the root directory to Dropbox.

This is the current way I am running one server for my group in the university. Dropbox allows an easy way to manage multiple files using their website, however if your codes are very sensitive then you may not want them on Dropbox for unknown disclosure issue.

annieeby:
Renaming with directory path yields this error: 

"Invalid notebook name. Notebook names must have 1 or more characters and can contain any characters except :/\. Please enter a new notebook name."

my attempted file name (to move to folder Flatiron Introduction to Python):
./Flatiron Introduction to Python/Lists With Maps

Vinayk93:
**SOLUTION**:-
jupyter create ipynbcheckoint folder for the files previous state.
open file explorer look for that folder inside ipynbcheckpoint (hidden) most of the cases.
Create your new file there.

jmg:
I made some improvements over the move functionality by adding a folder selector. The PR is https://github.com/jupyter/notebook/pull/4604

MrAnalyticals:
A workaround solution I use is to download the notebook file to your local machine and then upload it to the required folder, not forgetting to delete the old version from the wrong location of course.

PR #941, opened 2016-01-10
Adds a Move button to the notebooklist interface next to Rename.

NHDaly:
With this button, you can now move files or directories to any path in
the notebook, including creating new directories (like 'mv -p'). To move
a file, you specify the new destination directory's path.

This achieves the same effect as renaming a file with a new relative
path, but is more clear to users unfamiliar with unix commands --
especially with regard to moving to a parent directory.

Can only move one file at a time currently.

Addresses Issue #471.

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -785,15 +785,16 @@ define([
     };
 
     NotebookList.prototype.move_selected = function() {
+        var that = this;
+
         // TODO(nhdaly): Support moving multiple items at once.
-        if (this.selected.length !== 1){
+        if (that.selected.length !== 1){
             return;
         }
 
-        var that = this;
-        var item_path = this.selected[0].path;
-        var item_name = this.selected[0].name;
-        var item_type = this.selected[0].type;
+        var item_path = that.selected[0].path;
+        var item_name = that.selected[0].name;
+        var item_type = that.selected[0].type;
 
         // Open a dialog to enter the new path, with current path as default.
         var input = $('<input/>').attr('type','text').attr('size','25').addClass('form-control')
@@ -808,6 +809,7 @@ define([
             title : "Move "+ item_type,
             body : dialog_body,
             buttons : {
+                Cancel : {},
                 OK : {
                     class: "btn-primary",
                     click: function() {
@@ -830,8 +832,7 @@ define([
                             console.warn('Error durring moving :', e);
                         });
                     }
-                },
-                Cancel : {}
+                }
             },
             open : function () {
                 // Upon ENTER, click the OK button.
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -835,6 +835,7 @@ define([
                     }
                 }
             },
+            // TODO: Consider adding fancier UI per Issue #941.
             open : function () {
                 // Upon ENTER, click the OK button.
                 input.keydown(function (event) {
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -775,6 +775,7 @@ define([
                     }
                 });
                 input.focus();
+                // Highlight the filename (up to the filetype suffix) in the input field.
                 if (input.val().indexOf(".") > 0) {
                     input[0].setSelectionRange(0,input.val().indexOf("."));
                 } else {
@@ -842,9 +843,8 @@ define([
                         return false;
                     }
                 });
+                // Put the cursor at the end of the input.
                 input.focus();
-                // Highlight the current path in the input box.
-                input.select();
             }
         });
     };
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -843,11 +843,8 @@ define([
                     }
                 });
                 input.focus();
-                if (input.val().indexOf(".") > 0) {
-                    input[0].setSelectionRange(0,input.val().indexOf("."));
-                } else {
-                    input.select();
-                }
+                // Highlight the current path in the input box.
+                input.select();
             }
         });
     };
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -129,6 +129,7 @@ define([
 
             // Bind events for action buttons.
             $('.rename-button').click($.proxy(this.rename_selected, this));
+            $('.move-button').click($.proxy(this.move_selected, this));
             $('.shutdown-button').click($.proxy(this.shutdown_selected, this));
             $('.duplicate-button').click($.proxy(this.duplicate_selected, this));
             $('.delete-button').click($.proxy(this.delete_selected, this));
@@ -550,6 +551,15 @@ define([
             $('.rename-button').css('display', 'none');
         }
 
+        // Move is only visible when one item is selected, and it is not a
+        // running notebook.
+        // TODO(nhdaly): Add support for moving multiple items at once.
+        if (selected.length === 1 && !has_running_notebook) {
+            $('.move-button').css('display', 'inline-block');
+        } else {
+            $('.move-button').css('display', 'none');
+        }
+
         // Shutdown is only visible when one or more notebooks running notebooks
         // are selected and no non-notebook items are selected.
         if (has_running_notebook && !(has_file || has_directory)) {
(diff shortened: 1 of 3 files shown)


NHDaly:
In this Pull Request the button only supports Moving 1 selected item at a time, but I've left a TODO in the code and also opened Issue #942 to address moving multiple.

Should I reference the issue number from the TODO? I'm still fairly new to open source codebases. :)

Thanks!

notebook/static/tree/js/notebooklist.js
```diff
@@ -697,6 +707,73 @@ define([
         });
     };
 
+    NotebookList.prototype.move_selected = function() {
+        // TODO(nhdaly): Support moving multiple items at once.
+        if (this.selected.length !== 1){
+            return;
+        }
+
+        var that = this;
```
captainsafia: I think it would make more sense to move this variable declaration to the top of the function and use `that` instead of `this` throughout.

captainsafia:
Good stuff, @NHDaly! I posted some notes inline.

notebook/static/tree/js/notebooklist.js
```diff
@@ -697,6 +707,73 @@ define([
...
+                                        .text(e.message || e)),
+                                buttons: {
+                                    OK: {'class': 'btn-primary'}
+                                }
+                            });
+                            console.warn('Error durring moving :', e);
+                        });
+                    }
+                },
+                Cancel : {}
```
captainsafia: We're looking to change our button orders so that the "Cancel" is always on the left and the action is on the right. So these should be switched to be compatible with future changes.

notebook/static/tree/js/notebooklist.js
NHDaly: Done! Thanks. (Extremely new to Javascript; why do we even create the `that` variable? I was just following the convention of the surrounding functions.)

notebook/static/tree/js/notebooklist.js
NHDaly: Done. That makes sense; thanks!

NHDaly:
Cool, thanks for looking at this! :) I've addressed your suggestions in ce24663! :)

Cheers!

notebook/static/tree/js/notebooklist.js
captainsafia: Good question, @NHDaly! I'm gonna give explaining it a shot and let me know if I am unclear. It all boils down to scoping. In JavaScript, the `this` keyword is used to refer to the object that the function you are currently in is tied to. In this case, `this` references `NotebookList`. You can confirm this by adding a `console.log(this)` statement to the beginning of the `move_selected` function. Now, let's say that inside your `move_selected` function, you create a new object with functions as you do with the dialog in this case. Your `this` now references a completely different object within those functions. You can test this out by adding a `console.log(this)` right before line 741. But what if in this new context you want to reference the `NotebookList` as you do on line 741? Well, you store the value of `this` in a variable called `that` when `this` references the object you later want to reference and then use `that` in functions where `this` does not reference that object.

So in reality, you don't particularly need it for the first few lines at the top since `this` references `NotebookList` in those cases but I like to declare at the top and use it consistently so that it is obvious that throughout the code anywhere `that` is used we are referencing a `NotebookList` object.

Let me know if any of that needs clarification! `this` is definitely a tricky topic in JavaScript.

notebook/static/tree/js/notebooklist.js
```diff
@@ -697,6 +707,71 @@ define([
...
+                // Upon ENTER, click the OK button.
+                input.keydown(function (event) {
+                    if (event.which === keyboard.keycodes.enter) {
+                        d.find('.btn-primary').first().click();
+                        return false;
+                    }
+                });
+                input.focus();
+                // Highlight the current path in the input box.
+                input.select();
```
NHDaly: Open Question: This is how the Rename box works, which makes sense. But should the Move button start with a pre-selected path? I feel like more often you're going to be modifying the path rather than overwriting it with an entirely new path, so maybe this should just move the cursor to the end instead of selecting the input.

Thoughts? :)

notebook/static/tree/js/notebooklist.js
NHDaly: Thanks Safia, nope that makes tons of sense now. Gotcha! Clever! Unnamed functions ... pretty crazy.

Thanks for taking the time to explain that to me! :)

notebook/static/tree/js/notebooklist.js
captainsafia: So I'm almost tempted to say that the Move functionality should prompt the user to select a directory to move a file into with a dropdown. I say this for two reasons.

1. To make the creating a directory and moving something into a directory two distinctly different tasks.
2. To conform with more tradition UI where the user doesn't enter a path to move files to manually.

I'm not sure how difficult creating a dropdown would be, but happy to help you work through it.

ellisonbg: The move UI in the GitHub webapp is really cool.

On Wed, Jan 20, 2016 at 7:55 AM, Safia Abdalla <notifications@github.com>
wrote:

> In notebook/static/tree/js/notebooklist.js
> <https://github.com/jupyter/notebook/pull/941#discussion_r50272758>:
>
> > +                            console.warn('Error durring moving :', e);
> > +                        });
> > +                    }
> > +                }
> > +            },
> > +            open : function () {
> > +                // Upon ENTER, click the OK button.
> > +                input.keydown(function (event) {
> > +                    if (event.which === keyboard.keycodes.enter) {
> > +                        d.find('.btn-primary').first().click();
> > +                        return false;
> > +                    }
> > +                });
> > +                input.focus();
> > +                // Highlight the current path in the input box.
> > +                input.select();
>
> So I'm almost tempted to say that the Move functionality should prompt the
> user to select a directory to move a file into with a dropdown. I say this
> for two reasons.
>
>    1. To make the creating a directory and moving something into a
>    directory two distinctly different tasks.
>    2. To conform with more tradition UI where the user doesn't enter a
>    path to move files to manually.
>
> I'm not sure how difficult creating a dropdown would be, but happy to help
> you work through it.
>
> —
> <https://github.com/jupyter/notebook/pull/941/files#r50272758>.
>



-- 
Brian E. Granger

captainsafia: Hm. Now that I think about it @ellisonbg's recommendation is a good one and you won't have to steer too far from the code that you have currently written to make it happen.

[Here](https://help.github.com/articles/moving-a-file-to-a-new-location/) is the help page for GitHub's move feature. So the way I imagine the user experience for this working is as follows.

1. User selects a checkbox on a file in the file tree.
2. User clicks the move button in the toolbar above.
3. Filename in the file tree gets converted into a file path relative to the directory the user is in.
4. User writes within the textbox the new path.
5. User hits enter and the file is moved to a new location if there is no error.

Thoughts on this, @NHDaly?

ellisonbg: Part of what is cool about the Github approach is that as you type paths
with / it detects that and moves those into the path part. Really slick.

On Wed, Jan 20, 2016 at 8:58 AM, Safia Abdalla <notifications@github.com>
wrote:

> In notebook/static/tree/js/notebooklist.js
> <https://github.com/jupyter/notebook/pull/941#discussion_r50282784>:
>
> > +                            console.warn('Error durring moving :', e);
> > +                        });
> > +                    }
> > +                }
> > +            },
> > +            open : function () {
> > +                // Upon ENTER, click the OK button.
> > +                input.keydown(function (event) {
> > +                    if (event.which === keyboard.keycodes.enter) {
> > +                        d.find('.btn-primary').first().click();
> > +                        return false;
> > +                    }
> > +                });
> > +                input.focus();
> > +                // Highlight the current path in the input box.
> > +                input.select();
>
> Hm. Now that I think about it @ellisonbg <https://github.com/ellisonbg>'s
> recommendation is a good one and you won't have to steer too far from the
> code that you have currently written to make it happen.
>
> Here <https://help.github.com/articles/moving-a-file-to-a-new-location/>
> is the help page for GitHub's move feature. So the way I imagine the user
> experience for this working is as follows.
>
>    1. User selects a checkbox on a file in the file tree.
>    2. User clicks the move button in the toolbar above.
>    3. Filename in the file tree gets converted into a file path relative
>    to the directory the user is in.
>    4. User writes within the textbox the new path.
>    5. User hits enter and the file is moved to a new location if there is
>    no error.
>
> Thoughts on this, @NHDaly <https://github.com/NHDaly>?
>
> —
> <https://github.com/jupyter/notebook/pull/941/files#r50282784>.
>



-- 
Brian E. Granger

NHDaly: @captainsafia: I believe the flow you have described is exactly how the `Rename` button already works. You can use `Rename` to move a file to a different directory by changing the relative path. E.g:
To move a file to a parent directory:
 1. Select checkbox on file.
 2. Press Rename. The textbox is pre-populated with just the filename + extension (e.g. "my_notebook.pynb")
 3. Prepend `../` to the name (e.g. changing it to "../my_notebook.pynb")
 4. Enter.

That's all done with relative paths. Is that what you were describing? However, I don't think this is intuitive for a user not familiar with Unix, who wouldn't think to rename a file to "../<filename>" to move it up a directory. That's why I wanted to add a Move button.


@ellisonbg + @captainsafia: Yeah, wow, that *is* a slick UI. I love it! That said, I think it would be good to separate that fancy UI into a separate PR? It works almost the same as this code, but builds a fancier UI on top of it. If we're happy with the semantics in this PR, I'd prefer starting a new one for the fancier interface. Does that sound okay?

NHDaly: @ellisonbg + @captainsafia: Also, I wonder about how well the github style interface would support moving multiple files? I have opened #942 to support moving multiple files from the Move dialogue.

I guess maybe you could do the same thing as the Github UI, but leave off the file name?

NHDaly:
Friendly ping? :)

minrk:
@NHDaly thanks!

NHDaly:
No problem! (Yay!! My first merged-in GitHub PR! haha :dancer:)

captainsafia:
giphy-26


Congrats! Looking forward to more great PRs from you, @NHDaly!

NHDaly:
hahaha thanks. Awesome gif.

willingc:
@NHDaly Nice job! Welcome to Open Source and Jupyter :surfer: @captainsafia sends great gifs.

annieeby:
sorry for dumb nube question but how do i actually use your cool move button to move files in Jupyter?

PR #947, opened 2016-01-12
Convert documentation notebooks to markdown instead of rst

willingc:
This PR takes the examples and tutorials notebooks (ipynb files) and converts them to markdown instead of reStructuredText. Sphinx then builds both the Markdown and rst source files. Current advantage of using Markdown is that the headings in the source notebooks convert more accurately to Markdown.


Carreau:
I'm ok with that if you like it. You are the one write most of the docs, so I would leave the decision to you.

willingc:
Holding off for the moment. The nbconverted notebooks to markdown or rst are not properly handling Font Awesome fonts.

willingc:
I'm going to close this PR since we can likely use nbsphinx to achieve a similar result.

PR #958, opened 2016-01-15
Use require.toUrl for help_links

vbraun:
The kernel doesn't know about the `base_url` of the notebook, so he
cannot generate local urls. This is important if you want to serve
documentation locally, for example see
http://trac.sagemath.org/ticket/19885

Require.toUrl rewrites non-relative paths according to the path
configuration. For example, a help_link defined like
~~~~
{
    'text': 'Sage Documentation',
    'url': 'kernelspecs/foo/bar.html',
},
~~~~
is then correctly transformed to `/base_url/kernelspecs/foo/bar.html`


Carreau:
to backport to 4.2

Carreau:
backported.

PR #960, opened 2016-01-15
Acknowledge `ignotre_minified_js` on auth, and tab-> space

Carreau:
As asked on ML.


Carreau:
Objection on having that only on 5.0 and not on 4.2 ?

minrk:
@Carreau 5.0 should be fine, since the option only makes sense on dev anyway.

PR #963, opened 2016-01-15
add ctrl-alt-[ shortcut for indentAuto

### LINKED ISSUE #951: Is there a "reindent" option?
I pasted some code and the code was not indented. Is there a "reindent" option like the one employed by SublieText?

If not, kindly consider it a a feature.

minrk:
aka `cmd-opt-[`

indentAuto is CodeMirror's best guess for the indentation.

This is the shortcut from TextMate, I didn't look up other editors if there's a better choice.

I also moved the common cell commands to Cell base config.

closes #951

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -114,9 +114,15 @@ define([
             readOnly: false,
             theme: "default",
             extraKeys: {
-                "Cmd-Right":"goLineRight",
-                "End":"goLineRight",
-                "Cmd-Left":"goLineLeft"
+                "Cmd-Right": "goLineRight",
+                "End": "goLineRight",
+                "Cmd-Left": "goLineLeft",
+                "Tab": "indentMore",
+                "Shift-Tab" : "indentLess",
+                "Cmd-Alt-[" : "indentAuto",
+                "Ctrl-Alt-[" : "indentAuto",
+                "Cmd-/" : "toggleComment",
+                "Ctrl-/" : "toggleComment",
             }
         }
     };
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -125,11 +125,7 @@ define([
     CodeCell.options_default = {
         cm_config : {
             extraKeys: {
-                "Tab" :  "indentMore",
-                "Shift-Tab" : "indentLess",
                 "Backspace" : "delSpaceToPrevTabStop",
-                "Cmd-/" : "toggleComment",
-                "Ctrl-/" : "toggleComment"
             },
             mode: 'text',
             theme: 'ipython',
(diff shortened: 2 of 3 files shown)


notebook/static/notebook/js/cell.js
```diff
@@ -114,9 +114,15 @@ define([
...
-                "Cmd-Left":"goLineLeft"
+                "Cmd-Right": "goLineRight",
+                "End": "goLineRight",
+                "Cmd-Left": "goLineLeft",
+                "Tab": "indentMore",
+                "Shift-Tab" : "indentLess",
+                "Cmd-Alt-[" : "indentAuto",
+                "Ctrl-Alt-[" : "indentAuto",
+                "Cmd-/" : "toggleComment",
+                "Ctrl-/" : "toggleComment",
```
jdfreder: I was going to suggest adding a block comment shortcut, but I checked CodeMirror.commands, and it's not available :sob:

jdfreder:
:+1: to this shortcut

elmerehbi:
Thanks for the shortcut though this is not what I was implying in the other issue. I made a [screencast ](https://youtu.be/KLxfEa-N1GU)to show this.

minrk:
@elmerehbi that looks an awful lot like the feature that this adds:

autoindent

what's different that you are looking for?

elmerehbi:
@minrk Exactly what I'm looking for. What is it?

minrk:
That's the shortcut added by this PR: ctrl-alt-[

elmerehbi:
Well I remember trying it but it didn't give the same result (not to this
code I grabbed online). i'll give it another try.

On Wed, Feb 17, 2016 at 5:02 PM Min RK <notifications@github.com> wrote:

> That's the shortcut added by this PR: ctrl-alt-[
>
> —
> <https://github.com/jupyter/notebook/pull/963#issuecomment-185272329>.
>
-- 
Sincerely Yours,
Ibrahim EL MEREHBI

elmerehbi.com <http://bit.ly/iem_com>

elmerehbi:
Well I remember trying it but it didn't give the same result (not to this code I grabbed online). i'll give it another try.

elmerehbi:
@minrk I have version 4.0.6 from Anaconda. The shortcut is not in the list. How can I get it?

takluyver:
The bug was closed because the shortcut was added to the development version. You'll have to wait for a new release (or run from the development version) to see the change.

It's currently targeted for 5.0 because it's a new feature - look at the right hand side of this issue page. But it's probably small enough that we could put it in a 4.x release. @minrk ?

elmerehbi:
@takluyver Great. Thanks for the clarification.

jdfreder:
> But it's probably small enough that we could put it in a 4.x release. @minrk ?

I think new features can get into minor releases.  It's only features that change existing behavior that cannot.  The key binding used here wasn't already in use.  I guess it's debatable though, because users may have it bound now (although I think user config would overwrite it regardless).

PR #969, opened 2016-01-18
multiple-undo for cell deletion

### LINKED ISSUE #712: Delete history for multiple undo
I created this issue [some time ago](https://github.com/ipython/ipython/issues/4924) before the split from IPython, and I assume it now belongs here. While it is easy to quickly delete multiple cells in a notebook, it is currently impossible to undo beyond the last cell deletion. This results in users having to, at best, dig through version control as a remedy, or at worst, suffer data loss. Being able to undo multiple cell deletions would address this.

minrk:
use a stack for undelete_backup instead of a singleton

avoids data loss on multiple undelete

Since this is only for undelete, it's possible for intermediate edits to prevent undeleted cells from going into the exact right place. But it's more important that the cells be recoverable than they be recovered to the exact right place every time.

closes #712

cc @fonnesbeck, who's been patiently waiting for this for ages.

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -1020,9 +1020,6 @@ define(function (require) {
             this.events.trigger('delete.Cell', {'cell': cell, 'index': indices[i]});
         }
 
-        // Flip the backup copy of cells back to first-to-last order
-        undelete_backup.cells.reverse();
-
         var new_ncells = this.ncells();
         // Always make sure we have at least one cell.
         if (new_ncells === 0) {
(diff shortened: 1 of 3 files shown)


dsblank:
+1 Excellent! Is this too extreme to backport into 4.x? The semester hasn't begun yet, so if it happened sooner than later, that would be a semester of happy students and instructors!

takluyver:
Nice. This doesn't look too invasive, I think it would be reasonable to put it into a 4.x release.

fonnesbeck:
Fantastic, thanks. I will test the heck out of this.

jdfreder:
Something's not right about the cell ordering, maybe one too many .reverse calls?  See the following action shot, where the header and text cells are swapped:

demo

willingc:
@minrk I see the same behavior as @jdfreder . It seems to happen when you skip around the notebook and delete cells at the bottom of notebook then top then bottom and then try to undo. It doesn't seem to happen when you delete in order down from 1 to 5. reverse issue seems a likely culprit.

willingc:
Hmm... I wonder if I ran into this:

>Since this is only for undelete, it's possible for intermediate edits to prevent undeleted cells from going into the exact right place.

I'm pretty sure that at least once I clicked in a cell though didn't change contents.

minrk:
@willingc I think it should be doable to track correctly as long as the only actions are delete/undelete. Once cells are added or moved, that's when I think we can expect it will show up a little funny.

@jdfreder I think I fixed the reversal, and added tests that should cover it.

jdfreder:
@minrk thanks, I've verified that your change fixes the problem for me.

jdfreder:
:+1: for merge

mytrjp:
I think the easiest way (until developers handle this issue)to retrieve your Ipython history, write them all into an empty file and done! You need to check by the date you created your last script. Obviously, it is going to be the last part of your Ipython history. Good luck!

To write your Ipython history into a file:

`%history -g -f anyfilename`

PR #974, opened 2016-01-19
Add gif to docs for notebook creation

willingc:
This helps to clarify the user interface when creating a new notebook. See https://github.com/jupyter/jupyter/issues/97


willingc:
@Carreau Please review when you have time. We can at least fix the docs until the UI label gets restyled for "New" notebooks.

Carreau:
Usually in general i feel like if we **have to** explain something with image, or animation something is wrong. 

And also gifs ave to be redone at every UI change, so in general I (personal opinion), would be in favor  of avoiding gifs/images when necessary.

willingc:
@Carreau I agree that having to explain the UI too much is a sign that improving an aspect of the UI. For now, we are being kind to the users by adding the visual. We can always remove after the UI change lands.

Your point on changes over time is well taken. The rule of thumb that I am working under is adding a GIF or image should directly impact the reader's understanding. In general, I love things like the Tour of the Notebook to help users become familiar with features.

Welcome back :D

PR #979, opened 2016-01-21
update swagger API doc from jupyter-js-services

minrk:


blink1073:
Is this an exact copy of what was in `jupyter-js-services`?

minrk:
@blink1073 yes, other than bumping the version to 5.

minrk:
@blink1073 actually, after checking it with swagger, there were some validation errors that I've now fixed.

blink1073:
LGTM

notebook/services/api/api.yaml
```diff
@@ -27,8 +27,315 @@ parameters:
...
+    post:
+      summary: Restore a file to a particular checkpointed state
+      parameters:
+        - $ref: "#/parameters/path"
+        - $ref: "#/parameters/checkpoint_id"
+      tags:
+        - contents
+      responses:
+        204:
+          description: Checkpoint created
```
jasongrout: This description should be something like "Checkpoint restored", not created.

jasongrout:
One small comment, but other than that, looks good to me.

minrk:
@jasongrout thanks! fixed.

jasongrout:
I'm trying to check the rendering, and for some reason it isn't rendering: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/minrk/notebook/swagger-doc/notebook/services/api/api.yaml

jasongrout:
It looks like a bug in the swagger renderer.  I'm looking into it.

jasongrout:
Yep, definitely looks like a bug in swagger: https://github.com/swagger-api/swagger-ui/issues/1903

jasongrout:
How do I kickstart Travis again?

blink1073:
Does hitting the restart button here not work for you?  https://travis-ci.org/jupyter/notebook/jobs/103882148

screen shot 2016-01-21 at 11 58 29 am

jasongrout:
I hit it several times, and finally all of them passed.

willingc:
@jasongrout, @blink1073, @minrk Do we want to link to the swagger API docs from the Developer section of the Sphinx docs or README.md?

blink1073:
I'd say so (to both).

willingc:
Ping me when the rendering bug for swagger is solved, send me a link to where it's rendered, and I'm happy to update both.

jasongrout:
Yeah, I'd say that we figure out how to statically render that page and include it in the notebook as a static file.  It's silly that we are depending on a hosted service to do the static rendering.

jasongrout:
and include it in the hosted documentation as well...

blink1073:
https://www.npmjs.com/package/bootprint-swagger

PR #987, opened 2016-01-22
save before copy if notebook is dirty

### LINKED ISSUE #986: "Make a copy" menu item loses unsaved changes
Sometimes I find myself making changes to a notebook, then realise that I should keep my original notebook for comparison reasons.

So I use the "Make a copy" option to save a copy of the notebook with my changes (and go back to undo them from my original file). However, this make a copy of the current *notebook state on disk, neglecting any unsaved changes*.

I appreciate that this may be subjective, but this behaviour seems quite counter-intuitive to me. My expectation of this option is that it should operate more like "Save As...", so that the current contents of the notebook gets saved to a new filename.

minrk:
in copy_notebook, rather than relying on async: false in event handlers, which is ignored.

closes #986

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2578,23 +2578,32 @@ define(function (require) {
 
     /**
      * Make a copy of the current notebook.
+     * If the notebook has unsaved changes, it is saved first.
      */
     Notebook.prototype.copy_notebook = function () {
         var that = this;
         var base_url = this.base_url;
         var w = window.open('', IPython._target);
         var parent = utils.url_path_split(this.notebook_path)[0];
-        this.contents.copy(this.notebook_path, parent).then(
-            function (data) {
-                w.location = utils.url_path_join(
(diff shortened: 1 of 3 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -2578,23 +2578,32 @@ define(function (require) {
...
-            }
-        );
+        var p;
+        if (this.dirty) {
+            p = this.save_notebook();
+        } else {
+            p = Promise.resolve();
+        }
+        p.then(function () {
+            that.contents.copy(that.notebook_path, parent).then(
```
jdfreder: If you add a `return` both here and the line above then `copy_notebook` notebook method will return an appropriate promise, which is useful for callers of the method.

Carreau:
relaunched tests.

notebook/static/notebook/js/notebook.js
minrk: Good call. Added.

Carreau:
Relaunched test. 
Candidate for backport.

PR #988, opened 2016-01-23
Fix ANSI code for "bold"

mgeier:
This is **untested**, since I couldn't build the notebook locally.

However, there's a similar error in [nbconvert](https://github.com/jupyter/nbconvert/blob/560a1fc8ead384e1953349f2300185ee7041882c/nbconvert/filters/ansi.py#L28), where changing `01` to `1` fixes the problem.

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -207,7 +207,7 @@ define([
 
     //Map from terminal commands to CSS classes
     var ansi_colormap = {
-        "01":"ansibold",
+        "1":"ansibold",
         
         "30":"ansiblack",
         "31":"ansired",


mgeier:
It may not be that simple since leading zeros seem to be allowed ...

Carreau:
> It may not be that simple since leading zeros seem to be allowed ...

likely true. 

you might want to change the [RegularExpression](https://github.com/mgeier/notebook/blob/patch-1/notebook/static/base/js/utils.js#L315): `/\033\[(0?[01]|22|39)?([;\d]+)?m/g` to have an extra pair of bracket, or `parseInt` [there](https://github.com/mgeier/notebook/blob/patch-1/notebook/static/base/js/utils.js#L233) and change the dict from strings as keys to ints.

mgeier:
I'm planning to implement this similarly to https://github.com/jupyter/nbconvert/pull/225 (basically porting it to JavaScript), but I don't know when I'll have the time.

If somebody else wants to tackle this, go ahead!

mgeier:
I've created #1230, which will resolve this issue.

PR #992, opened 2016-01-25
use _.isEqual to check for metadata changes

minrk:
JSON.stringifiy is unreliable because it is key-order sensitive

closes #964

PR #996, opened 2016-01-26
Use underscore's isEqual to compare objects for equality

wadevries:
This is more robust than JSON.stringifying both operands, because the order of keys is relevant for JSON.stringify, but not for object equality. This led to 'false positive' inequality reports, which led to notebooks being set to dirty immediately after opening, which caused an autosave directly after load, which is annoying when you want to navigate away quickly.


takluyver:
Beaten to the punch by #992, I'm afraid. ;-)

minrk:
Bonus: I think the patches might be character-identical. Clearly on the right track!

wadevries:
Heh, exactly my thoughts!

PR #998, opened 2016-01-26
use x-access for directory listing test in is_hidden

minrk:
instead of actually listing because listdir can be slow.

closes ipython/ipython#9165

### DIFF
--- a/notebook/utils.py
+++ b/notebook/utils.py
@@ -112,7 +112,7 @@ def is_hidden(abs_path, abs_root=''):
                 return True
         else:
             # use x-access, not actual listing, in case of slow/large listings
-            if not os.access(abs_path, os.X_OK):
+            if not os.access(abs_path, os.X_OK | os.R_OK):
                 return True
     
     # check UF_HIDDEN on any location up to root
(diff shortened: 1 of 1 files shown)


takluyver:
This makes sense to me, but we're relying on a Unix-y idiom. Have we ensured that whatever it does on Windows produces a sane result? Do we need another check for directory accessibility on Windows?

minrk:
I tested that it didn't fail on Windows, at least, but after a little more digging I discovered that it just always returns True on Windows, no matter what. Fixed to use stat on Windows, which fails with PermissionDenied on unreadable dirs.

notebook/utils.py
```diff
@@ -102,12 +102,18 @@ def is_hidden(abs_path, abs_root=''):
...
+        if sys.platform == 'win32':
+            # can't trust os.access on Windows because it seems to always return True
+            try:
+                os.stat(abs_path)
+            except OSError:
+                # stat may fail on Windows junctions or non-user-readable dirs
+                return True
+        else:
+            # use x-access, not actual listing, in case of slow/large listings
+            if not os.access(abs_path, os.X_OK):
```
takluyver: Should we use `os.X_OK | os.R_OK`? It's probably not important, but it's possible to have a directory that you can cd into and read files, but not list its contents.

minrk: Sure. Done.

PR #1014, opened 2016-01-27
Use conda on Readthedocs

### LINKED ISSUE #1012: Update docs requirements.txt
Changing the build on jupyter_client recently seems to have caused the docs to fail to build. The version of setuptools is not being set correctly for RTD (18.2 RTD vs needed 18.5). docs/requirements.txt needs update. 

cc/@takluyver

takluyver:
closes #1012 
closes #1013


willingc:
See note re: sphinx==1.3.1 on #1012

takluyver:
Thanks, I just added `sphinx>=1.3` to the requirements.

willingc:
@takluyver ReadTheDocs hasn't been able to build the docs without failing since we made this change. I spent a good chunk of my time going through options or workarounds. Here's where it is failing the build (on js):
screen shot 2016-01-29 at 9 53 53 am

Thoughts on a solution? Do we really need the js for building the docs?

takluyver:
I think we need to install the notebook for some of the API docs, and installing it tries to build the JS. But there is a conda package of bower, so maybe it's not too hard:

http://anaconda.org/javascript/bower

minrk:
We don't need bower on the system, just npm. So if you can get nodejs at all, we should be set.

I faced the same issue on JupyterHub a bit ago, and did a perhaps hacky workaround: don't install the notebook, just add the notebook dir to sys.path in conf.py. (#1025)

takluyver:
The latest doc build succeeded, after I merged #1025 . Thanks, Min!

PR #1020, opened 2016-01-28
only import ssl if it's used

### LINKED ISSUE #1019: Make ssl an optional import
Not everyone wants to (or can, for license reasons) use openssl.  It looks like we import ssl at https://github.com/jupyter/notebook/blob/a3e074c3a7e29eb3ec8b94627370698458f048ce/notebook/notebookapp.py#L22, but the *only* place the import is used is at https://github.com/jupyter/notebook/blob/a3e074c3a7e29eb3ec8b94627370698458f048ce/notebook/notebookapp.py#L873 and https://github.com/jupyter/notebook/blob/a3e074c3a7e29eb3ec8b94627370698458f048ce/notebook/notebookapp.py#L879, both just for the constants.

Thanks to @williamstein for making us aware of the issue.

minrk:
SSL stdlib module may be unavailable in rare cases.

closes #1019

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -19,11 +19,9 @@ import re
 import select
 import signal
 import socket
-import ssl
 import sys
 import threading
 import webbrowser
-
 
 from jinja2 import Environment, FileSystemLoader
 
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -870,13 +868,17 @@ def init_webapp(self):
             ssl_options['keyfile'] = self.keyfile
         if self.client_ca:
```
jasongrout: Is there a reason this entire if statement isn't moved down below in the else block?

jasongrout: never mind, figured the logic out.

jasongrout:
Travis errors are for an unrelated bug that appears to be fixed in master.  Thanks, Min!

Carreau:
Marking as 4.2 to think about backporting.

PR #1022, opened 2016-01-28
Update the notebook widget hack to reference the new jupyter-js-widgets-nbextension package

jdfreder:
Accomplice to https://github.com/ipython/ipywidgets/pull/336

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -246,7 +246,7 @@ class NotebookWebApplication(web.Application):
             try:
                 widgets = __import__('ipywidgets')
             except:
-                app_log.warning('jupyter-js-widgets-nbextension package not installed.  Widgets are unavailable.')
+                app_log.warning('Widgets are unavailable. Please install jupyter-js-widgets-nbextension or ipywidgets 4.0')
         if widgets is not None:
             handlers.append(
                 (r"/nbextensions/widgets/(.*)", FileFindHandler, {
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -239,16 +239,21 @@ class NotebookWebApplication(web.Application):
         handlers.extend(load_handlers('services.security.handlers'))
         
         # BEGIN HARDCODED WIDGETS HACK
+        widgets = None
         try:
             widgets = __import__('jupyter-js-widgets-nbextension')
+        except:
+            try:
+                widgets = __import__('ipywidgets')
+            except:
+                app_log.warning('jupyter-js-widgets-nbextension package not installed.  Widgets are unavailable.')
+        if widgets is not None:
             handlers.append(
                 (r"/nbextensions/widgets/(.*)", FileFindHandler, {
                     'path': widgets.find_static_assets(),
                     'no_cache_paths': ['/'], # don't cache anything in nbextensions
                 }),
             )
-        except:
-            app_log.warning('jupyter-js-widgets-nbextension package not installed.  Widgets are unavailable.')
         # END HARDCODED WIDGETS HACK
         
         handlers.append(
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -240,15 +240,15 @@ class NotebookWebApplication(web.Application):
         
         # BEGIN HARDCODED WIDGETS HACK
         try:
-            import ipywidgets
+            widgets = __import__('jupyter-js-widgets-nbextension')
             handlers.append(
                 (r"/nbextensions/widgets/(.*)", FileFindHandler, {
-                    'path': ipywidgets.find_static_assets(),
+                    'path': widgets.find_static_assets(),
                     'no_cache_paths': ['/'], # don't cache anything in nbextensions
                 }),
             )
         except:
-            app_log.warning('ipywidgets package not installed.  Widgets are unavailable.')
+            app_log.warning('jupyter-js-widgets-nbextension package not installed.  Widgets are unavailable.')
         # END HARDCODED WIDGETS HACK
         
         handlers.append(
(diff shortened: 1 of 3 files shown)


notebook/notebookapp.py
```diff
@@ -240,15 +240,15 @@ def init_handlers(self, settings):
         
         # BEGIN HARDCODED WIDGETS HACK
         try:
-            import ipywidgets
+            widgets = __import__('jupyter-js-widgets-nbextension')
```
Carreau: Well, that will make the notebook not work with current ipywidgets. Won't it ?

and `jupyter-js-widgets-nbextension` feel weird. It's likely **not* a valid python package name,. You shouldn't **have to** use import.

notebook/static/notebook/js/main.js
```diff
@@ -55,7 +55,7 @@ require([
     "use strict";
 
     // BEGIN HARDCODED WIDGETS HACK
-    utils.load_extension('widgets/notebook/js/extension').catch(function () {
+    utils.load_extension('widgets/extension').catch(function () {
```
Carreau: same coment here, that woudl break notebook with **current** ipywidgets.

notebook/notebookapp.py
jdfreder: It works, but I'd be really happy for a better name

notebook/static/notebook/js/main.js
jdfreder: Good point , I'll fall back to ipywidgets if the nbextension fails

notebook/notebookapp.py
Carreau: try `/-/_/` that automatically done by python.

notebook/static/notebook/js/main.js
jdfreder: :heavy_check_mark:

notebook/notebookapp.py
```diff
@@ -239,16 +239,21 @@ def init_handlers(self, settings):
         handlers.extend(load_handlers('services.security.handlers'))
         
         # BEGIN HARDCODED WIDGETS HACK
+        widgets = None
         try:
-            import ipywidgets
+            widgets = __import__('jupyter-js-widgets-nbextension')
```
Carreau: if you have `jupyter-js-widgets-nbextension` installed, does `import jupyter_js_widgets_nbextension as widets` works ?

notebook/notebookapp.py
```diff
@@ -239,16 +239,21 @@ def init_handlers(self, settings):
...
+        except:
+            try:
+                widgets = __import__('ipywidgets')
+            except:
+                app_log.warning('jupyter-js-widgets-nbextension package not installed.  Widgets are unavailable.')
```
Carreau: Message inaccurate : neither of  ipywidgets or ... is installed.

notebook/notebookapp.py
jdfreder: no

notebook/notebookapp.py
jdfreder: Well installing ipywidgets alone won't make it work, unless it's an old version of ipywidgets.  Basically I want this message to tell the user they should go install jupyter-js-widgets-nbextension.   How about "Widgets are unavailable.  Please install jupyter-js-widgets-nbextension"?

Carreau: Hum... i'm not sure, because people **should** be able to install only an old version of ipywidgets. 
I think the error message should reference both just in case.

jdfreder: The only problem is that installing a new version of ipywidgets will not work - because ipywidgets will only contain the Python code.

jdfreder: `Widgets are unavailable.  Please install jupyter-js-widgets-nbextension or ipywidgets 4.0`

jdfreder: `Widgets are unavailable. Please install jupyter-js-widgets-nbextension or ipywidgets 4.0 (5.0 and later will not work!)`

Carreau: > Widgets are unavailable. Please install jupyter-js-widgets-nbextension or ipywidgets 4.0

+1

jdfreder: :heavy_check_mark:

blink1073:
I can confirm this works with https://github.com/ipython/ipywidgets/pull/336, pending a small change there.

jdfreder:
Updated to point to the new, more pythonic name, of `widgetsnbextension`.

willingc:
@jdfreder Do any of these need to change: https://github.com/jupyter/notebook/search?utf8=%E2%9C%93&q=ipywidgets ?

Also, do you wish any changes to the docs?

jdfreder:
Ah thanks @willingc , I need to change the docker file, I missed that.  The other two are already changed in the PR

Carreau:
> widgetsnbextension

+1, thanks.

willingc:
@jdfreder No problem. I did a search on the notebook docs and didn't see any references to ipywidgets. If you see something or want to add something, just let me know.

+1 on the PR

minrk:
@jdfreder why is this marked for notebook 4.2? Is it needed to keep 4.x working (i.e. 4.1 is broken)?

jdfreder:
No, without it users won't be able to load the newer widgets- we can target that for 5 if you prefer

Carreau:
Yes, 4.2 is mostly bug fix, otherwise it's painful we have to backport all the things.
Retagged as 5.

lrq3000:
Note that ipywidgets 4.1.0 seems to already implements this PR (contrary to the tag?), so widgetsnbextension is required as early as ipywidgets >= 4.1.0 as [reported here](widgetsnbextension).

PR #1025, opened 2016-01-29
don't install on RTD

minrk:
add repo to sys.path instead

fixes RTD builds after #1014


takluyver:
I think that's a perfectly reasonable solution.

willingc:
Thanks @takluyver and @minrk.

PR #1036, opened 2016-02-02
disable MathJax renderer selection menu

### LINKED ISSUE #1035: [MathJax] unsupported user configurations
We had a bug filed upstream at https://github.com/mathjax/MathJax/issues/1347 which seems to come down to Jupyter Notebooks not shipping all MathJax components, in particular not all output options.

If our educated guess is correct, we'd suggest to disable the corresponding MathJax UI elements to avoid users from accidentally breaking things for themselves.

minrk:
since we don't ship all renderers, and MathJax can hang if a nonexistent renderer is selected.

closes #1035


Carreau:
Need backport.

Issue #1037, opened 2016-02-02
Multiple MathJax + Incomplete MathJax


milindsmart:
I am using Jupyter with IPython from Anaconda3 on Windows 7, notebook v4.1.0 which seems to contain MathJax 2.5.3 (I had to verify MD5 hashes manually, there is absolutely no mention of MathJax version anywhere).

I tried to switch the math renderer to NativeMML, but it could not find the required config.js. Adding the files directly from the MathJax release did not help (seems to be loading only those files listed in some config var, not the actual directories apparently).

The bigger issue is that I currently have 3 versions of MathJax on my system without having tried any repair of my own : 

1. C:\Users\MilindAnaconda3\pkgs\mathjax-2.2-0\Lib\site-packages\IPython\html\static\mathjax - v2.2.0
2. C:\Users\Milind\.ipython\nbextensions\mathjax - v2.4
3. C:\Users\Milind\Anaconda3\pkgs\notebook-4.1.0-py34_0\Lib\site-packages\notebook\static\components\MathJax - v2.5.3

It appears that 3. is being used currently.

At this point I cannot figure out what is Anaconda's problem, and what is Jupyter's and what is MathJax's, so wanted to document it here first. 

I'd be happy to help in fixing this issue (and others of this level), I love the concept of Jupyter. But I don't know what's going on.

minrk:
I'm not sure what the source of 1. is, perhaps conda used to separate MathJax as a package from IPython, but added it as a dependency. That would only affect the pre-Jupyter notebook (IPython < 4).

Number 2. is a remnant of IPython < 4, when we didn't ship mathjax, due to its size, and the default was to use a CDN. If you wanted local mathjax, you could install it as an nbextension. This no longer has any effect as of notebook 4.0, because we bundle MathJax as a plain old dependency.

The only relevant one if you are using current versions is 3, now that the notebook bundles a stripped MathJax as a normal dependency.

milindsmart:
That was informative and explains the situation somewhat, thanks. I should have mentioned that I upgraded Conda from v2 to v3, sorry about that. What remains is 2 things : 

1. Assuming that I'm adamant about MathML, how would I ensure that the required files are accessible to the frontend? Just putting the files in the correct location has no effect : the files are just not found.
2. Why a stripped copy of MathJax? I'm no authority, but I thought "bundling" meant a local copy with no relation to the direct release,  and "dependency" meant a separate package that the package manager recognizes as required.

[...]

PR #1039, opened 2016-02-02
Organize contributor/developer documentation sections

willingc:
Adding the code snippet for a development install to the documentation's Developer FAQ.  Inspired by @minrk's response in #1037.


takluyver:
We explain this in more detail in the development_js doc:
http://jupyter-notebook.readthedocs.org/en/latest/development_js.html

Is it worth making a link to that page, and/or improving that part of the documentation, rather than duplicating the information?

minrk:
My comment in #1037 was meant for added debugging. It shouldn't be necessary, and probably shouldn't be added here, since it gives the false impression that it is necessary. `pip install -e .` should always be sufficient, and if it ever isn't, it is because of a bug somewhere.

willingc:
@takluyver and @minrk: Good suggestions. 

I think it would be helpful to update the logical outline/flow of all the Developer (defined as those contributing code to the notebook source) documentation. 

Right now, the developer information is spread across the README, development_js.rst (which includes both js development as well as creating a release and should likely be split into two shorter documents), and development_faq.rst (which is intended for those contributing code to the notebook source).

If it's cool with you, I'm going to use this PR to reorganize the three docs mentioned above to be more contributor friendly.

minrk:
Sounds great!

willingc:
@takluyver @minrk Here's the first pass at reorganizing the existing docs for Developers. I have not yet moved the README contents into it. Feel free to toss in comments. My next step will be moving some things from the README to the docs and then referring via link from the README to docs (or the reverse). Screenshots below for your convenience.

### Navigation from main contents page and sidebar
screen shot 2016-02-02 at 11 41 18 am

### Developer Landing Page for contributors to the Notebook
screen shot 2016-02-02 at 11 41 46 am

### Installing Javascript Machinery page (post-edit)
screen shot 2016-02-02 at 11 42 17 am

### Making a release page (post-edit)
screen shot 2016-02-02 at 11 42 40 am

takluyver:
It seems a bit odd to have making a release taking up so much space under 'want to contribute' - that information isn't relevant to new contributors.

Also, more broadly, the heading 'Want to contribute?' feels a bit misleading at present - we have information on how to set up a development environment, but really no guidance to actually contributing.

I think it would be nice to have an overview similar to the CPython developer guide, that provides an overview of the contribution process: https://docs.python.org/devguide/

willingc:
@takluyver Thanks for the comments. 

I decreased the depth of the Table of Contents section 'Want to Contribute'. 

I also divided the section into project specific info and more general Jupyter contribution info which is at jupyter.readthedocs.org.

docs/source/development_js.rst
```diff
@@ -1,81 +1,52 @@
...
+     Select a pre-built installer
+     on the `Node.js`_ website. The installer will include Node.js and
+     Node's package manager, :program:`npm`.
 
-This will automatically fetch the remaining dependencies (bower, less) and
-install them in a subdirectory.
+   - Using system's package manager:
+     Install Node.js and :program:`npm` using the
+     system's package manager. For example, the command for Ubuntu or Debian
+     is:
```
damianavila: You can probably mention conda here as well... because it is trivial to get node with it:

`conda install -c javascript nodejs`

This will also include `npm`...

willingc:
Thanks @minrk. @damianavila good point re: conda. I will put it in when I correct a minor typo error.

PR #1041, opened 2016-02-03
Clarify docker image contents in README

parente:
* jupyter/notebook is HEAD of master and built on every merge
* Point to docker-stacks for stable releases / full environments


parente:
Doh. Just saw that all branches are built. Nevermind! I thought only master was going out.

PR #1050, opened 2016-02-05
Ensure user site packages is on PYTHONPATH for notebook test server

takluyver:
We modify `$HOME`, so it was failing to find my user site packages when trying to start kernels in the tests. Setting PYTHONPATH lets it use these packages.

### DIFF
--- a/notebook/tests/launchnotebook.py
+++ b/notebook/tests/launchnotebook.py
@@ -73,6 +73,7 @@ class NotebookTestBase(TestCase):
         data_dir = TemporaryDirectory()
         cls.env_patch = patch.dict('os.environ', {
             'HOME': cls.home_dir.name,
+            'PYTHONPATH': os.pathsep.join(sys.path),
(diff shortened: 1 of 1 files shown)


Carreau:
+1

minrk:
Do we let PYTHONPATH pass down to kernels? Maybe we shouldn't, since that's a bad plan for mixed Python versions.

takluyver:
That's possibly true, but in that case I'd want to try to run these tests without modifying $HOME, since that makes the kernel fail to start on my system.

willingc:
If running the test server when using conda, wouldn't setting PYTHONPATH potentially give unexpected results?

takluyver:
I don't think using conda would be a problem - it is setting PYTHONPATH only so that the same packages are importable by a subprocess after HOME is changed. If the import paths are correct in the notebook test runner, they should be correct for starting a kernel in the same environment.

Min's concern was about passing on PYTHONPATH when running kernels in a different environment or Python version. That could cause problems, but I don't think we run any such kernels in the notebook tests.

PR #1051, opened 2016-02-05
Run nosetests with verbose flag on Travis

takluyver:
Hopefully stops tests from hanging after completion - seems to work on Jenkins.


Carreau:
> Hopefully stops tests from hanging after completion - seems to work on Jenkins.

Is circle supposed to pass ? (restarted )

takluyver:
I have also seen the hanging failure on Circle, but I didn't do anything with that in this PR. I'll add that.

PR #1055, opened 2016-02-06
Allow all MathJax output formats

### LINKED ISSUE #1037: Multiple MathJax + Incomplete MathJax

minrk:
rather than white-listing HTML-CSS and scrubbing all others. This means the various other renderers (perhaps most importantly, MML) are available, not just HTML-CSS.

Fonts other than STIX-Web are still scrubbed, since much of the size is due to the many font choices.

This increases the size of a notebook install by ~3.5k/19MB or 18%.

A smaller change would be to *only* add MML, which would have a negligible affect on install size.

closes #1037

### DIFF
--- a/notebook/static/notebook/js/mathjaxutils.js
+++ b/notebook/static/notebook/js/mathjaxutils.js
@@ -29,9 +29,11 @@ define([
                     styles: {'.MathJax_Display': {"margin": 0}},
                     linebreaks: { automatic: true }
                 },
-                MathMenu: {
-                    showRenderer: false,
-                },
+            });
+            MathJax.Hub.Register.StartupHook("MathMenu Ready", function () {
+              var renderers = MathJax.Menu.menu.Find("Math Settings").submenu.Find("Math Renderer").submenu;
+              // disable SVG output, which we don't ship
+              renderers.Find("SVG").disabled = true;
             });
             MathJax.Hub.Configured();
         } else if (window.mathjax_url !== "") {
--- a/notebook/static/notebook/js/mathjaxutils.js
+++ b/notebook/static/notebook/js/mathjaxutils.js
@@ -30,11 +30,6 @@ define([
                     linebreaks: { automatic: true }
                 },
             });
-            MathJax.Hub.Register.StartupHook("MathMenu Ready", function () {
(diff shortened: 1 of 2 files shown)


Carreau:
5.0 of 4.2 ?

+1

minrk:
I'm not sure. I could go either way.

minrk:
@jasongrout you brought up some follow up that suggested maybe we should include SVG anyhow. Did you want to check in here?

minrk:
For reference: Adding SVG with a single font (like we do with HTML-CSS) would increase install size by 2.5 MB, or an additional ~13%.

rgbkrk:
Is install size a real worry here? This doesn't seem so bad, I'm ready to merge this.

takluyver:
I don't think an extra 2.5MB is a problem. IIRC, the full mathjax bundle was about 10x the size of our entire package, which seemed a bit much, but I think it's fine to selectively add bits back in if there's a use for them.

minrk:
Okay, this now includes all output formats, including SVG. SVG is scrubbed down to a single font, just like we do with HTML-CSS.

PR #1062, opened 2016-02-08
Revisions to js dev doc

### LINKED ISSUE #685: improve dev Install instruction on (Linux/ubuntu)
I've tried the dev-install instruction on ubuntu yesteday on someones laptop. 

Even after 30min of tinkering I was completely unable to install the notebook. 
In particular `node` was `nodejs`, I had to `apt-get install  npm`, bower was requiring `sudo`, and was calling `/usr/bin/env node` instead of `nodejs`, even though the `sudo python setup.py [css|js|...]` failed at various step, and reconstructing the command to run by hand was a pain. 

I'm not sure if it was this particular system, or a more general linux/ubuntu case, 
as in particular it seem to work flawlessly on Travis.

minrk:
- use `pip install -e .`
- clarify that `setup.py js css` is only needed for intermediate updates, not first-run

[This Dockerfile](https://gist.github.com/d5844104f6dad5b90a40) shows minimal steps to getting setup, starting from scratch Ubuntu.

closes #685

### DIFF
--- a/docs/source/development_js.rst
+++ b/docs/source/development_js.rst
@@ -29,22 +29,35 @@ following steps are used when making a Notebook release.
 
          sudo apt-get install nodejs-legacy npm
 
-#. Build the JavaScript and CSS by running:
+#. Install the notebook:
 
+   In the notebook repo, do a development install:
+   
+   .. code:: bash
+   
+       pip install -e .
+
+#. Rebuild JavaScript and CSS
+
+   There is a build step for the JavaScript and CSS in the notebook.
+   You will need to run this command whenever there are changes to JavaScript or LESS sources:
+  
    .. code:: bash
 
-       python setup.py css js
+       python setup.py js css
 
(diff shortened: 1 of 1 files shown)


docs/source/development_js.rst
```diff
@@ -29,22 +29,36 @@ following steps are used when making a Notebook release.
...
+   
+       pip install -e .
 
-       python setup.py css js
+#. Rebuild JavaScript and CSS
 
-   This command will automatically fetch the remaining dependencies (bower,
+   There is a build step for the JavaScript and CSS in the notebook.
+   If you are modifying the JavaScript or LESS sources, you may need to rebuild
+   their targets from time to time. This can be done with:
```
takluyver: 'from time to time' - doesn't it need to be any time you modify JS/CSS (assuming you want to test it)?

minrk: Yup, updated to actually describe the condition

PR #1064, opened 2016-02-08
scroll long tracebacks to the bottom

minrk:
since that's usually where the most important info is

related: jupyter/nbconvert#233 which will start producing big tracebacks on latex failures


Carreau:
+!

PR #1078, opened 2016-02-10
Add a make spelling option for doc builds

willingc:
Adds the ability to run ``make spelling`` for the Sphinx docs.

Any errors will be logged to the build directory after running ``make spelling``. I have corrected any spelling errors in all documents that haven't been autobuilt or converted.

I don't think this needs to run on every build. A cron job or running before each release could also be reasonable to add in the future. For now, the developer needs to run ``make spelling`` manually (which is the same case for ``make linkcheck``.


docs/source/conf.py
```diff
@@ -335,3 +336,6 @@
     'nbformat': ('http://nbformat.readthedocs.org/en/latest/', None),
     'jupyter': ('http://jupyter.readthedocs.org/en/latest/', None),
 }
+
+spelling_lang='en_US'
```
takluyver: en what? ;-)

minrk:
Neat!

willingc:
Thanks @minrk and @takluyver.

PR #1084, opened 2016-02-12
Refresh travis.yml and ~~coveralls~~ codecov to better handle stalled builds

willingc:


takluyver:
:+1: seems like a good idea

willingc:
@takluyver Thanks. Going to focus on travis before approaching tox since tox/conda combo complicates things.

willingc:
After much slogging through Travis, pip, and coveralls docs as well as Travis build reports, I believe this configuration is more stable.

* Removed ``group: edge`` since it was a temporary inclusion https://github.com/jupyter/notebook/commit/f05dfae0f3b2368bead6e374779ac21f7948e245 and is not needed since Python 3.5 is set to 3.5.1.

* Set python to version 3.5.1 since the default 3.5 on Travis does not yet include 3.5.1 and lower versions of 3.5 had issues.

* Before install: upgrade pip separately to most recent version to avoid any dependency nonsense. Then, upgrade the key packaging and testing libraries.

* Split coveralls out as a separate install. I can't explain why but it seems to work better this way.

* Also reactivated the coveralls.io repo since for some reason it was disabled

* Added fast_finish option to cause Travis to end on success or initial failure.

Carreau:
> Also reactivated the coveralls.io repo since for some reason it was disabled

Thats because coverall.io is annoying and keep asking for too much access to member of the org (access to all repository including private, read and write) so we started migrating to https://codecov.io which is: 
  - more responsive to support request.
  - more polite in support request
  - have an extension for github to show coverage on a per file basis in github UI.

So strong -1 on reenabling coverall. But fine if merge as is I can remove later.

> Added fast_finish option to cause Travis to end on success or initial failure.

-1 , we have random failures that will force us to rerun all instead of just the failing one(s) . And we have many groups, which test differents part, which mean that you cannot in one glance see all the failing test. 
I hate matplotlib that does that, and don't even test your code if it's not pep8 compliant.

+1 on the rest

willingc:
@Carreau I much prefer codecov.io so no objection to using that over coveralls. Will strip coveralls out and look any changes needed for codecov.io.

Fair enough on fast_finish, I debated either enable or disable (handy when debugging the travis.yml). Any thoughts on speeding up the tests. They take a looong time for Travis to execute.

I'll make the coveralls and fast_finish changes. Thanks for the review.

willingc:
@Carreau Changes made.

If someone with permissions can activate codecov.io, it should work (working on my local branch).

takluyver:
We have coverage: https://codecov.io/github/jupyter/notebook

Thanks @willingc

Carreau:
Awesome thanks !

minrk:
Nice!

Before you put too much time into tox, I would start with a bit of caution. If we aren't *using* tox on things like Travis, I wouldn't bother with it. A while ago ago, tox was added to IPython, but nobody used it so it regularly fell well out of date. In fact, when I started writing this I thought we had removed tox, but now that I look it's still there and wildly out of date. From the looks of it, it hasn't been in a working state for about a year, and nobody seems to have noticed. It ends up being another thing to maintain that doesn't actually get used.

willingc:
Thanks @minrk. Thanks for the heads up. When I actually implemented this and conda was working well, I realized tox didn't buy us anything. I forgot to change the issue title :)

PR #1086, opened 2016-02-13
Add download button to the tree template

VladMironov:
First of all, thank you for this amazing library.

As a Kaggler I found it *extremely* painful to open files before download them (usually it's some .csv predictions to send for and usually there're bunch of them).

So I decided to add Download button for single files inside the tree template.
Works like a charm for me — disappears when it needs to, downloads any single file except running ipython.


VladMironov:
According to this issue https://github.com/ipython/ipywidgets/issues/56 travis test should be ok, just need a restart.

takluyver:
Thanks, I restarted the failing Travis build, and it worked this time.

VladMironov:
It's possible to add multi-file/folder download via zip but I'm not sure about necessity of this move.

PR #1087, opened 2016-02-13
Updated Move Dialog's buttons per #744 for #742.

NHDaly:
Changes "OK" -> "Move" and sets "Cancel" as default button.

I think I added this button in between the commits for #744 and it being merged in, so it got missed.
This is part of #742.


takluyver:
Thanks!

PR #1100, opened 2016-02-17
Added better testing for file paths

### LINKED ISSUE #911: Cannot rename notebook if name change only involves capitalization
@numeratus opened ipython/ipython#9095

Say, I want to rename "Capitalization.ipynb" as "capitalization.ipynb", I get:

>File already exists: notebooks/capitalization.ipynb

Workaround is to change name in multiple steps. 

Jupyter notebook server is 4.0.6

`Python 2.7.11 (default, Dec  5 2015, 23:52:42) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]`

captainsafia:
Closes #911.

`os.path.exists` can be case-insensitive depending on the OS.

cc: @minrk

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -460,7 +460,9 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         old_os_path = self._get_os_path(old_path)
 
         # Should we proceed with the move?
-        if os.path.exists(new_os_path):
+        _, new_name = os.path.split(new_os_path)
+        old_dir, _ = os.path.split(old_os_path)
+        if new_name in os.listdir(old_dir):
             raise web.HTTPError(409, u'File already exists: %s' % new_path)
 
         # Move the file


minrk:
Thanks! Looks like the test is failing for real. Can you investigate that?

I believe this change isn't quite the right fix, because it enables clobbering any existing file if the case doesn't match (on a case-insensitive filesystem). In order to fix #911 more precisely, I think the check needs to be for whether the destination name in the case of a collision actually refers to the same file as the original. Something like:

```python
if exists(destination):
    if destination really is origin:
        proceed
    else:
        raise 409
```

The really-is-origin check can probably use `os.stat`:

```python
# assume we have already verified that a and b both exist
is_same_file = (os.stat(a).st_ino == os.stat(b).st_ino)
```

to verify that the two paths actually refer to the same file.

takluyver:
[os.path.samefile()](https://docs.python.org/3/library/os.path.html?highlight=samefile#os.path.samefile) ?

takluyver:
Gah, Python 2 doesn't have that function for Windows.

Drop Py 2 support, amirite @Carreau ? ;-)

minrk:
@takluyver good call. That's exactly what we want, though we will have to ship our own copy (even though it's exactly the same) for Windows + py2. My code above is not quite correct because it must also compare `st_dev`.

If we want to be lazy about it, we could only perform the check if `hasattr(os.path, 'samefile')`, so that Windows + py2 won't be able to recapitalize files, but every other case will.

captainsafia:
> Thanks! Looks like the test is failing for real. Can you investigate that?

Drat! I've been in JavaScript land for so long I ran the JavaScript tests and not the Python ones.

> That's exactly what we want though we will have to ship our own copy (even though it's exactly the same) for Windows + py2.

I'll cook something up.

Carreau:
> Drop Py 2 support, amirite @Carreau ? ;-)

Yessss

captainsafia:
Argh! I made an idiotic mistake and accidentally deleted my old fork while I was trying to get an updated version of upstream using the web UI. :cry: 

I'll close this for now but submit another PR with the appropriate fix tomorrow night.

PR #1104, opened 2016-02-17
Added initial issue template

captainsafia:
Now that GitHub allows you to create templates for issues, I figured it would be a good idea to make one based on the "Submitting a Bug Report" directions in the docs.

Happy to iterate on this.

cc: @willingc

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,7 @@
+### Problem
+
+### Reproducable Code Sample
+
+### Environment Configuration
+
+### Sample Test Case or Expected Behavior


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+### Problem
+
+### Reproducable Code Sample
```
takluyver: I don't think this is really the right structure for bugs on the notebook repo, because many (most?) issues are about UI, and don't necessarily involve a code sample. We've been happy enough for years without an issue template, so I don't think we need to rush to create one - it's more important to get something that helps more people than it obstructs.

Carreau:
+1 an @takluyver comment.

willingc:
@captainsafia Thanks for thinking about enhancements. I'm not very familiar with GitHub issue templates. Perhaps a good way to go would be try it on a repo that gets less activity than the notebook to see how it works in the wild.

takluyver:
> I'm not very familiar with GitHub issue templates.

I'd be amazed if you were, because Github only launched the feature yesterday ;-)

willingc:
@takluyver Oh good. I thought perhaps I was living under a rock. ;-)

Carreau:
I think this does have enough traction (for now).
a better way would be a link **in the notebook** that uses the following format:

```
https://github.com/jupyter/notebook/issues/new?title=my%20issue&body=the%20text
```

To pre-fill the issue with system information. Though we might get too many issues, and so not want to make this too obvious to find.

PR #1109, opened 2016-02-19
[WIP] Support Vim mode in notebook

parleur:
This pull request is functional but not usable right now.

It switches the keybindings of all codemirror instance (inspired by the editor). It obviously raises issue in UI interface I aim to discuss here.

- Issues (just here to be exhautive):
  - Double modal editor: when vim mode is activated in CM, it add a modal editor inside the Jupyter's modal management... Yes press Enter then press a to edit ... not usable
  - Escape (managed by Jupyter) unfocus CM even in Vim edit mode.

- Solutions proposed. Obviously vim mode should modify the Jupyter keymap also. I propose following changes:
     - Down (Jup managed ) --> Ctrl+Down : switch cell
     - Esc ( Jup managed ) --> Esc (Jup managed): not even ADD a modal state but track CM modal state and synchronize Jupyter modal and CM modal
     - 'a' and 'i' (CM vim mod) --> Same thing, escape from vim edit mode AND from Jupyter edit mode

That's it. Feel free to discuss, that's the main purpose of this PR


Carreau:
Thanks @parleur  I'll try to look at that soon !

Carreau:
High level comments: 

Could we have a tick marker on the menu for the currently use keymap ?

notebook/static/notebook/js/codecell.js
```diff
@@ -561,6 +564,30 @@ define([
...
+    CodeCell.prototype._set_codemirror_options = function (options) {
+        // update codemirror options from a dict
+        var codemirror = this.code_mirror;
+        $.map(options, function (value, opt) {
+            if (value === null) {
+                value = CodeMirror.defaults[opt];
+            }
+            codemirror.setOption(opt, value);
+        });
+        var that = this;
```
Carreau: forgotten var that=this ?

notebook/static/notebook/js/keyboardmanager.js
```diff
@@ -126,6 +136,21 @@ define([
             'q' : 'jupyter-notebook:close-pager',
         };
     };
+    
+    KeyboardManager.prototype.get_default_vim_command_shortcuts = function() {
+      var shortcut = this.get_default_command_shortcuts();
+      delete shortcut['down'];
+      shortcut['ctrl-down'] = 'jupyter-notebook:select-next-cell';
+      delete shortcut['up'];
+      shortcut['ctrl-up'] = 'jupyter-notebook:select-previous-cell';
```
Carreau: I dont' think we want to bind to `ctrl-something`, it will  screw up with windows manager on some linux/osx

Carreau:
OK, I think that we want to avoid vim shortcut for now, at it seem to be complex. Get the logic right for default/emacs/sublime, and deals with vim later.

Carreau:
Regardless of that, pretty sweet, and seem to mostly work for me.

notebook/static/notebook/js/notebook.js
```diff
@@ -3037,6 +3037,37 @@ define(function (require) {
...
+    */
+    Notebook.prototype.switch_keymap_to = function(keymap) {
+        var vimMode = true;
+        var cells = this.get_cells();
+        var cell = cells[1];
+        if (keymap === "vim"){
+            vimMode = true;
+            this.keyboard_manager.mode = 'vim-command';
+        } else if ( keymap === 'sublime' || keymap === 'default') {
+            vimMode = false;
```
Carreau: if you set vimMode to `null`, the update option will remove the key from the config file, which is cleaner I think.

notebook/static/notebook/js/notebook.js
```diff
@@ -3037,6 +3037,37 @@ define(function (require) {
...
+            this.keyboard_manager.mode = 'command';
+        } else {
+            console.log('No valid keymap specified. Valid keymaps are in {"vim", "sublime", "default"}');
```
Carreau: and `emacs` ?

minrk:
Awesome, people are going to be really happy about this.

jdfreder:
Sweet! Have you had a chance to try @ivanov 's [vimception](https://github.com/ivanov/ipython-vimception)?

hack-c:
This looks sweet. @parleur Can you provide some context on what changes were made and where review is most helpful? THanks!!

jdfreder:
Closed?  Did you find an alternative way to implement it?

Carreau:
> Closed? Did you find an alternative way to implement it?

I thin @parleur want a bit less attention, and yes, we discussed an alternative for vim keybinding, and had a long conversation on how key handling works in general in the notebook.

hack-c:
ah great, thanks @Carreau. I'm interested in helping out with "emacs mode" if/when there's an opening.

Carreau:
The emacs key bindings are not that hard, the issue is with Vim that already have a dual command/edit mode. 

We (@parleur and I) think it should be possible to "never" be in jupyter command mode, and have extra keybindings for codemirror that trigger actions at notebook level. 

So  you can do things like `<esc>:run all`, or `:<esc>:above` , or alike. 
It needs some refactor, like replacing 

```javascript
if (this.mode === 'edit') {
    return this.edit_shortcuts.call_handler(event);
} else if (this.mode === 'command') {
  return this.command_shortcuts.call_handler(event);
}
```

by something like:

```javascript
new_mode = this.mode[this.current_mode].call_handler(event)
this.current_mode = new_mode
```
To gain some flexibility, and some patching of markdown cell to switch the rendering on focus.

jdfreder:
^ @Carreau I like that idea

parleur:
Hi everyone,
This PR is more difficult than I thought, so I wanted some time to pursue in my side (means testing many thing an discard it). I reopen it because I need Travis working, and I because I am not lost anymore. Yet it still far from a serious working PR.

### What are the goal?

- enable vim-mode without crasy three mode to manage
- be able to dynamically switch edition mode
- make the previous stuff convenient ( and it's far to be a detail ...)

### What works

- mostly trap the user in notebook edition mode (except when doing things with mouse) so that working with the notebook could be done in keyboard only when proper shortcut for every notebook functionality will be finished
- Switch and switch back between vim mode and default mode without too much bug.

### Minimum work to make this usable ( a lot !!! )

- new cell should be configured in correct edition mode
- up/down keyboard aren't handled properly
- ex command in vim mode should access notebook feature
- visual interface for ex command is very poor and should be corrected

notebook/static/notebook/js/keyboardmanager.js
```diff
@@ -28,21 +28,36 @@ define([
...
         this.enabled = true;
         this.pager = options.pager;
         this.quick_help = undefined;
         this.notebook = undefined;
         this.last_mode = undefined;
+		this.current_mode = "command";
         this.bind_events();
         this.env = {pager:this.pager};
         this.actions = options.actions;
-        this.command_shortcuts = new keyboard.ShortcutManager(undefined, options.events, this.actions, this.env );
```
Carreau: For API compat reason we likely want to define a property names `command_shortcut` that return `this.mode['command']`, but we can do that later.

notebook/templates/edit.html
```diff
@@ -58,9 +58,9 @@
                 <li class="divider"></li>
                 <li class="dropdown-header">Key Map</li>
                 <li id="menu-keymap-default"><a href="#">Default<i class="fa"></i></a></li>
-                <li id="menu-keymap-sublime"><a href="#">Sublime Text<i class="fa"></i></a></li>
+                <li id="menu-keymap-sublime"><a href="#">Sublime Text <i class="fa"></i></a></li>
                 <li id="menu-keymap-vim"><a href="#">Vim<i class="fa"></i></a></li>
-                <li id="menu-keymap-emacs"><a href="#">emacs<i class="fa"></i></a></li>
+                <li id="menu-keymap-emacs"><a href="#">emacs<i class="fa"></i></a></li
```
Carreau: Oops, missing closing `>`

Carreau:
Thanks for the all recap !

notebook/static/notebook/js/keyboardmanager.js
parleur: Yes of course ... but I havn't yet precisely to track all occurence of old API through tests. I 'll do that at the end.

Carreau:
of interest. 
https://github.com/lambdalisue/jupyter-vim-binding

parleur:
Here a working cell configuration. Mainly, I modified Cell to pass anything from Cell.class_config to subclasses. Of course if subclasses ( namely CodeCell, MarkdownCell, ...) overwrite config if defined.

parleur:
Now initialise the notebook in a correct mode saved in config.

Carreau:
I think you might need to rebase.

parleur:
A stupid question: you suggest to rebase on top on current master:HEAD, I am used to merge new HEAD into my dev branch ( I ve read somewhere, it is dangerous to rebase any published commit ). I am wrong? Or is it equivalent ?

Carreau:
Usually we avoid merging with HEAD. 
It is not **dangerous** to rebase, it can just be a bit more complicated. And anyway locally your commits are still in the reflog, so you cannot **lose** work. 

I can show you how to rebase next  Tuesday on Skype if you like.

Carreau:
Also current commit is https://github.com/parleur/notebook/commit/2832c132643967e681907d5ed85d9d773880f89c

So now that it is linked here, whatever you do, we can get it back from GitHub.

parleur:
A little comment about dynamic Codemirror keymap switching.
CodeMirror have a bug/strange behavior with handlers so that when a handler have been registered before a keymap switch then it is treated __after__ the newly declared codemirror core handler. The solution is to unbind the jupyter handler then switch keymap and rebind it.

Carreau:
> A little comment about dynamic Codemirror keymap switching.
> CodeMirror have a bug/strange behavior with handlers so that when a handler have been registered before a keymap switch then it is treated after the newly declared codemirror core handler. The solution is to unbind the jupyter handler then switch keymap and rebind it.

Another solution would be forcing to reload page after switching keymap..

parleur:
Hmm it would solve the Codemirror bug ... but likely raise other issues. At minimum
- auto saving before reloading
- interrupt computation by keymap switching
Unbind and rebind is proper and simpler (it is needed only about the keydown hander). The cleanest way would be to patch codemirror but it is out of my scope.

Carreau:
We can likely assume people won't switch keymaps too many time. So I think reloading (with a warning) is acceptable if we want to go this route.

parleur:
The vim keymap works and going back to default mode.  It's quick and dirty but it is an example of pleasant notebook navigation using only the keyboard. 

Note that it work using a (https://github.com/codemirror/CodeMirror/issues/3881)[patched codemirror] (already merged) that can be backported onto the current codemirror version used by Jupyter.

Carreau:
THanks for fixing things upstream. In general CodeMirror is released monthly, so we never have to wait too much between releases.

Carreau:
related extension: https://github.com/lambdalisue/jupyter-vim-binding

parleur:
This PR in now close. After many time spent on in it appears that dealing a keymap layer need a proper keyevent handling, and it is not a refacto that can be done under the hood here.

Alok:
I wonder if Neovim could be embedded, like how [ActualVim](https://github.com/lunixbochs/actualvim) works, rather than just reimplementing vim-mode like
every other project with a vim emulation mode.

PR #1113, opened 2016-02-19
Fix long lines escaping from the cell container

### LINKED ISSUE #1003: Long input lines of code overwrite the input cell bounds on Chrome
Notebook v4.1, no custom css.

Firefox 43 / Ubuntu (expected behavior):

image

Chrome 48 / Ubuntu (cell too wide):

image

minrk:
by setting:

    min-width: 0

on div.inner_cell ([reference](https://drafts.csswg.org/css-flexbox/#min-size-auto)).

(*obviously* min-width needs to be defined for it to respect its maximum width. Yay, CSS!)

Fixes what [we thought](https://github.com/ipython/ipython/issues/7724) was a Firefox bug, but Chrome devs have [pointed out](https://bugs.chromium.org/p/chromium/issues/detail?id=582141#c4) that Firefox was doing it right all along, and all the other browsers were wrong.

A bugfix in Chrome 48 brought Chrome's behavior in sync with Firefox, which we thought was a bug.

closes #1003

We should backport at least to 4.2, and possibly to 4.1.1 if we're going to do one.

### DIFF
--- a/notebook/static/notebook/less/cell.less
+++ b/notebook/static/notebook/less/cell.less
@@ -97,16 +97,9 @@ div.cell {
 }
 
 div.inner_cell {
+    min-width: 0;
     .vbox();
     .box-flex1();
-}
-
-@-moz-document url-prefix() {
-    div.inner_cell {
-        // hack around FF bug causing cell to expand when lines are long
-        // instead of scrolling
-        overflow-x: hidden;
-    }
 }
 
 /* input_area and input_prompt must match in top border and margin for alignment */


notebook/static/notebook/less/cell.less
```diff
@@ -97,18 +97,11 @@ div.cell {
 }
 
 div.inner_cell {
+    .min-width: 0;
```
Carreau: I think you meant `min-width` without leading dot ?

minrk: Yup, fixed. Weirdly, it seems LESS was happy to accept the typo.

willingc:
Thanks!

xcosmos6:
Hi I am having this issue again with chrome Version 62.0.3202.89. Can anyone please help me? Thank you! My jupyter notebook was install via anaconda on centos.

PR #1115, opened 2016-02-19
refuse to run as root user

### LINKED ISSUE #1074: Refuse to start as root ?
Should we refuse to run the notebook if we are root ? (unless passed an --allow-root flag). 
Should we do the same for IPython itself ?

secant:
Hi! I'm a undergraduate at UC Berkeley working with Matthias (@carreau). This is to resolve #1074. Currently, running the notebook as root will cause the program to terminate, but the --allow-root flag isn't working/being recognized and I'm not sure why.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1110,11 +1110,13 @@ class NotebookApp(JupyterApp):
         This method takes no arguments so all configuration and initialization
         must be done prior to calling this method."""
         try:
-            if os.geteuid() == 0 and not self.allow_root:
-                self.log.critical("Running as root is forbidden. Use --allow-root to bypass.")
-                self.exit(1)
+            is_root = os.geteuid() == 0
         except AttributeError as e:
-            pass #need to add Windows
+            import ctypes
+            is_root = ctypes.windll.shell32.IsUserAnAdmin() == 1
+        if is_root and not self.allow_root:
+            self.log.critical("Running as root is forbidden. Use --allow-root to bypass.")
+            self.exit(1)
 
         super(NotebookApp, self).start()
 
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1113,8 +1113,8 @@ class NotebookApp(JupyterApp):
             if os.geteuid() == 0 and not self.allow_root:
                 self.log.critical("Running as root is forbidden. Use --allow-root to bypass.")
                 self.exit(1)
-        except OSError as e:
-            pass
+        except AttributeError as e:
+            pass #need to add Windows
 
         super(NotebookApp, self).start()
 
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -450,7 +450,7 @@ class NotebookApp(JupyterApp):
         help="Set the Access-Control-Allow-Credentials: true header"
     )
     
-    allow_root = Bool(False, config=False, 
+    allow_root = Bool(False, config=True, 
         help="Whether to allow the user to run the notebook as root."
     )
 
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1110,13 +1110,11 @@ class NotebookApp(JupyterApp):
         This method takes no arguments so all configuration and initialization
         must be done prior to calling this method."""
         try:
-            is_root = os.geteuid() == 0
+            if os.geteuid() == 0:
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -445,6 +450,10 @@ def _log_format_default(self):
         help="Set the Access-Control-Allow-Credentials: true header"
     )
     
+    allow_root = Bool(False, config=False,
```
Carreau: You just have `config=False` here, which make it not configurable. Just switch to `config=True` and it works.

Carreau:
Do you want to add a note id `docs/sources/changelog.rst` ?

willingc:
@secant Welcome! Matthias is a wonderful mentor. I'm part of the Jupyter team at Cal Poly.

@Carreau beat me to the help. I did restart Travis (our continuous integration and testing service). It is green.

Carreau:
> It is green.

Yeah, we can try to make a test. Not sure it is worth though. 
If we do that, we can monkey patch `os.geteuid()` to return always Zero. But I'm not sure it's worth testing that.

willingc:
@Carreau No need for tests on this. I was just explaining what Travis was :)

rgbkrk:
Hey this is cool to do.

Does Windows need to be cased off?

secant:
@Carreau 
Under what version in the changelog would I add this change? 4.0.10? Or a new one? :bowtie:

takluyver:
Hi @secant !

* In the changelog, start a new section called 'Development' for now - we'll rename it when it actually goes into a release.
* I think @rgbkrk is right, it needs a separate check for Windows. I guess that the `geteuid` function is simply not there in Windows, so it would throw AttributeError.

notebook/notebookapp.py
```diff
@@ -1100,6 +1109,13 @@ def start(self):
         
         This method takes no arguments so all configuration and initialization
         must be done prior to calling this method."""
+        try:
+            if os.geteuid() == 0 and not self.allow_root:
+                self.log.critical("Running as root is forbidden. Use --allow-root to bypass.")
+                self.exit(1)
+        except OSError as e:
+            pass
```
takluyver: Is there some condition that we're expecting to catch here? Why would `geteuid()` fail? If we're catching this, I think we should at least log the error so that there's an indication that something has gone wrong.

Carreau: No geteuid raise OSError on windows (according to the docs). So the try/catch is the dealing with windows case.

rgbkrk: :+1:

takluyver: @Carreau What docs are those? I rebooted into Windows to check, and it ain't there:

```
>>> os.geteuid()
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'geteuid'
```

secant: @takluyver Whoops, I read the [docs](https://docs.python.org/2/library/os.html) too fast 
> Note All functions in this module raise OSError in the case of invalid or inaccessible file names and paths, or other arguments that have the correct type, but are not accepted by the operating system.

Should be excepting AttributeError, my bad!

rgbkrk: No problem!

notebook/notebookapp.py
```diff
@@ -1100,6 +1109,15 @@ def start(self):
         
         This method takes no arguments so all configuration and initialization
         must be done prior to calling this method."""
+        try:
+            is_root = os.geteuid() == 0
+        except AttributeError as e:
+            import ctypes
+            is_root = ctypes.windll.shell32.IsUserAnAdmin() == 1
```
takluyver: I wouldn't do this - admin users in Windows aren't really equivalent to root on Unix, and I don't think it makes sense to block them from running the notebook. Admin users on Windows are more like the sudoers group on Linux.

notebook/notebookapp.py
```diff
@@ -1100,6 +1109,15 @@ def start(self):
...
+        if is_root and not self.allow_root:
+            self.log.critical("Running as root is forbidden. Use --allow-root to bypass.")
```
takluyver: I'd say something like "is not recommended" - it's not quite forbidden.

secant: Gotcha, will make the changes.

minrk:
Thanks, @secant, welcome to the project!

Carreau:
Thanks for taking care of this ! Sorry I was mostly offline this Week-end.

PR #1119, opened 2016-02-20
Clarify keyboard shortcut help for "split cell"

### LINKED ISSUE #1008: Shortcut help for 'split cell' shortcut - display broken
screenshot from 2016-01-27 11 05 50

I would never call that key 'subtract', either. I'd normally say 'dash', and I'd consider 'hyphen' or 'minus' valid alternatives ('minus' is the symbol, 'subtract' is the operation it indicates). The Unicode character it inserts is called HYPHEN-MINUS.

Also, I don't think the shortcut should be duplicated in help. We should pick one way to represent it.

bjlange:
### OSX
screen shot 2016-02-20 at 12 13 16 pm
### Everything else
image

### OSX
screen shot 2016-02-20 at 11 52 01 am
### Everything else
screen shot 2016-02-20 at 11 52 30 am

I went with "Minus" for places where "- -" gets confusing and kept the "-" on OSX where there aren't dashes to make things confusing, per brief discussion in #1008. Open to other alternatives.

Closes #1008

### DIFF
--- a/notebook/static/notebook/js/keyboardmanager.js
+++ b/notebook/static/notebook/js/keyboardmanager.js
@@ -79,7 +79,6 @@ define([
             'up'                  : 'jupyter-notebook:move-cursor-up',
             'down'                : 'jupyter-notebook:move-cursor-down',
             'ctrl-shift--'        : 'jupyter-notebook:split-cell-at-cursor',
-            'ctrl-shift-subtract' : 'jupyter-notebook:split-cell-at-cursor',
         };
     };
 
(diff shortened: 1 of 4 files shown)


takluyver:
:+1: thanks!

It looks like we're hitting rate limiting on Travis, so the tests haven't run yet. So long as they pass, this looks fine.

bjlange:
JS tests weren't passing, turns out they were using "ctrl-shift-subtract" to trigger splits, so removing that from keyboardmanager.js broke them. I changed them to use "ctrl-shift--", hope that's OK. Splitting with the shortcut still works in Chrome and Firefox for me (which have different keycodes for '-', may have been the origin of this to begin with).

minrk:
Thanks, @bjlange!

PR #1121, opened 2016-02-20
Updated Dockerfile for widgetsnbextension

### LINKED ISSUE #1114: The docker file should also install python2 widgetsnbextension
Please add this line:
pip2 install widgetsnbextension
after
pip3 install widgetsnbextension

So it won't generate 404 error:

404 GET /nbextensions/widgets/notebook/js/extension.js

brandles:
Closes #1114.

cc: @captainsafia

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,6 +79,7 @@ RUN BUILD_DEPS="nodejs-legacy npm" && \
     DEBIAN_FRONTEND=noninteractive apt-get install -yq $BUILD_DEPS && \
     \
     pip3 install --no-cache-dir /usr/src/jupyter-notebook && \
+    pip2 install widgetsnbextension && \
     pip3 install widgetsnbextension && \
     \
     npm cache clean && \


takluyver:
@jdfreder is this right? It seems like the nbextension stuff should only need to be installed once, so I'm not sure if #1114 correctly identified the cause of the issue.

Looking at the widgetsnbextension wheel on PyPI, it refers to a `static/` data directory which is not in fact anywhere in the package.

rgbkrk:
@takluyver They both need the kernel side python code though right? Not just the frontend assets?

minrk:
Yeah, and `widgetsnbextension` should only be notebook-side, right? That's the point of it?

jdfreder:
Hey, don't know why I didn't see this ping, sorry.  `widgetsnbextension` is the nbextension for widgets, which is language agnostic.  `ipywidgets` is the Python backend.

This fix should not be needed now - ipywidgets depends on widgetsnbextension.

PR #1126, opened 2016-02-22
[WIP] add trust to markdown cells

minrk:
This restores the pre-3.0 ability to have fully arbitrary HTML/CSS/JS in markdown cells when the notebook is trusted.

- cell.render checks for trust before rendering
- 'safe' (no sanitization to do) markdown is automatically trusted
- 'unsafe' and untrusted markdown is left unrendered until user requests render
- cell.execute (the shift-enter/run all action) sets explicit trust, just like code cells

See discussion in #1123

TODO:

- UI indicator beyond unrendered markdown on load
- support trust of markdown cells in nbformat (will require new nbformat release, or wrappers in this repo to handle the trusted flag server-side)
- changing what's considered for signatures needs to invalidate trust of all existing notebooks.
  Since current notebook signatures only consider code cell output for trust,
  we need to take care that previously sanitized markdown in trusted notebooks does not end up being trusted implicitly on upgrade.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,7 @@
     "codemirror": "~5.8",
     "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.2.0",
-    "google-caja": "5669",
+    "google-caja": "6005",
     "jquery": "components/jquery#~2.0",
     "jquery-ui": "components/jqueryui#~1.10",
     "marked": "~0.3",
--- a/notebook/static/base/js/security.js
+++ b/notebook/static/base/js/security.js
@@ -70,11 +70,13 @@ define([
         return h.html();
     };
     
-    var sanitize_html = function (html, allow_css) {
+    var sanitize_html = function (html, allow_css, hook) {
         /**
          * sanitize HTML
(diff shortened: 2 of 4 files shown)


jdfreder:
So if you handed me a notebook file and I was running this branch, I'd see everything as cells, not in  pretty, rendered format?

minrk:
@jdfreder no, this would be the relevant part:

>  'safe' (no sanitization to do) markdown is automatically trusted

Only the cells where sanitization would change the content would be left unrendered. Basically, all cells that work today would show up as rendered, just like they do now. But cells that include 'unsafe' content (as determined by caja) would be left unrendered until you decided to run them.

The only change with respect to current behavior is for these cells with untrusted content:

- today: the unsafe content is always stripped, but the safe remainder is still rendered.
- this pr: the unsafe content *may* be displayed, but won't be on load if the notebook is untrusted.

Carreau:
Needs rebase.

Carreau:
This is going stale, I'm going to mark it as "Closed PR" to find it easily, and close it to keep the PR queue short. 

Thanks all for working on that.

PR #1127, opened 2016-02-22
Don't force save before download if notebook isn't writable

### LINKED ISSUE #1112: cannot "download as IPython notebook" a read-only notebook
When I open a read-only ipython notebook I cannot download it as an ipynb file, even if it hasn't been modified.

Steps to reproduce:
  1. make some existing `*.ipynb` file readonly on your file system
  2. open it with jupyter notebook (the editor will flash "auto-save disabled" and show a small logo "this notebook is read-only")
  3. select   File > Download as > IPython notebook (*.ipynb) (this just opens a blank page, does not download anything, and shows no error on the console)

Context: I'm trying to setup a public Jupyter installation where users can test some of my tools.  In this installation I have some example notebooks that I have marked as read-only because I don't want them to be modified by mistake, but I'd still like them to be downloadable.

minrk:
closes #1112

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -79,7 +79,7 @@ define([
         ) + "?download=" + download.toString();
         
         var w = window.open('', IPython._target);
-        if (this.notebook.dirty) {
+        if (this.notebook.dirty && this.notebook.writable) {
             this.notebook.save_notebook().then(function() {
                 w.location = url;
             });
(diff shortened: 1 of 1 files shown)


takluyver:
It appears that Travis failed to see this, for some reason.

notebook/static/notebook/js/menubar.js
```diff
@@ -79,7 +79,7 @@ define([
         ) + "?download=" + download.toString();
         
         var w = window.open('', IPython._target);
-        if (this.notebook.dirty) {
+        if (this.notebook.dirty && that.notebook.writable) {
```
takluyver: I think it should be `this` instead of `that`.

Carreau: Yep.

PR #1130, opened 2016-02-22
Mark thread for test server as daemonic

takluyver:
I don't understand how, but I suspect that the threaded server is somehow involved in the hangs we see after the tests complete. This marks it as daemonic so that if it somehow escapes it won't keep the
test process alive.

Heisendebugging the slow way.

### DIFF
--- a/notebook/tests/launchnotebook.py
+++ b/notebook/tests/launchnotebook.py
@@ -115,7 +115,7 @@ class NotebookTestBase(TestCase):
                 # set the event, so failure to start doesn't cause a hang
                 started.set()
                 app.session_manager.close()
-        cls.notebook_thread = Thread(target=start_thread)
+        cls.notebook_thread = Thread(target=start_thread, daemon=True)
         cls.notebook_thread.start()
         started.wait()
         cls.wait_until_alive()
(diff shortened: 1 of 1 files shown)


takluyver:
Nope, [that's not it](https://travis-ci.org/jupyter/notebook/jobs/111017688).

@minrk I recall hanging-on-close issues with the zmq context before. Could that be happening here? How would we debug it?

minrk:
That's definitely a possibility. I'll see if I can figure out a way to diagnose it better.

willingc:
Complete shot in the dark. I wonder if linger needs to be set or set to a shorter interval???

https://kerpanic.wordpress.com/2015/07/22/zmq-hanging-at-context-termination/

takluyver:
`jupyter_client` sets the linger to 1000ms (in `ConnectionFileMixin._create_connected_socket`). But maybe a ZMQ socket is being created somewhere else and getting the default infinite linger?

minrk:
I doubt that linger would be the issue, but I can investigate to be more certain. If zmq is involved, it's more likely that a socket is somehow never getting closed. I'll see if I can get to a situation where I can reproduce the bug on a machine where I can do interactive things.

takluyver:
I can sometimes reproduce it by sshing into the Jenkins server, activating the environment used for the jack-of-none builds and running `nosetests notebook`. But even there, it seems less likely to occur than when it's run by Jenkins. :confused:

minrk:
The actual problem should be fixed by #1136. I still think making this thread a daemon makes sense, so merging anyway.

PR #1136, opened 2016-02-24
channel.closed is a method

### LINKED ISSUE #1043: Intermittent problem with tests hanging after completion
I'm not sure that this bug is actually in notebook, but so far I've only seen it from running the notebook test suite. Something goes wrong in the cleanup process after the tests, which results in the test process never exiting. I've seen this on Jenkins and now on Circle CI - it hangs until the CI system gives up on the build and fails it. IIRC, @Carreau has also seen this error running the tests locally.

Investigating now, but any suggestions are welcome.

```
............................................../usr/local/lib/python3.4/dist-packages/nbformat/validator.py:147: UserWarning: No schema for validating v2 notebooks
  warnings.warn("No schema for validating v%s notebooks" % version, UserWarning)
...............................................................S..........Failed to import the site module
  File "/usr/lib/python3.4/site.py", line 78, in <module>
    import os
  File "/usr/lib/python3.4/os.py", line 57, in <module>
    import posixpath as path
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2222, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 2164, in _find_spec
  File "<frozen importlib._bootstrap>", line 1940, in find_spec
  File "<frozen importlib._bootstrap>", line 1914, in _get_spec
  File "<frozen importlib._bootstrap>", line 2050, in find_spec
KeyboardInterrupt
........................................................................S
Ran 193 tests in 13.701s

OK (SKIP=2)
Error in atexit._run_exitfuncs:
  File "/usr/local/lib/python3.4/dist-packages/IPython/core/history.py", line 755, in writeout_cache
    self._writeout_input_cache(conn)
  File "/usr/local/lib/python3.4/dist-packages/IPython/core/history.py", line 739, in _writeout_input_cache
    (self.session_number,)+line)
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread.The object was created in thread id 139968591865600 and this is thread id 139968726718272
```

minrk:
since we weren't calling it, `channel.closed` is always True, so the restart channel was never closed, causing process teardown to hang *sometimes*, depending on garbage collection.

I also added explicit teardown of the zmq Context at the end of each test group,
which increases our chances of noticing when we aren't cleaning up our sockets  properly.

closes #1043

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -115,7 +115,7 @@ class MappingKernelManager(MultiKernelManager):
         
         def finish():
             """Common cleanup when restart finishes/fails for any reason."""
-            if not channel.closed:
+            if not channel.closed():
                 channel.close()
             loop.remove_timeout(timeout)
             kernel.remove_restart_callback(on_restart_failed, 'dead')
--- a/notebook/tests/launchnotebook.py
+++ b/notebook/tests/launchnotebook.py
@@ -18,6 +18,7 @@ except ImportError:
     from mock import patch #py2
 
 from tornado.ioloop import IOLoop
+import zmq
 
 import jupyter_core.paths
 from ..notebookapp import NotebookApp
@@ -131,6 +132,16 @@ class NotebookTestBase(TestCase):
         cls.notebook_dir.cleanup()
         cls.env_patch.stop()
         cls.path_patch.stop()
+        # cleanup global zmq Context, to ensure we aren't leaving dangling sockets
+        def cleanup_zmq():
+            zmq.Context.instance().term()
+        t = Thread(target=cleanup_zmq)
+        t.daemon = True
+        t.start()
+        t.join(5) # give it a few seconds to clean up (this should be immediate)
+        # if term never returned, there's zmq stuff still open somewhere, so shout about it.
+        if t.is_alive():
+            raise RuntimeError("Failed to teardown zmq Context, open sockets likely left lying around.")
 
     @classmethod
     def base_url(cls):


minrk:
Woo, Travis passed first try without needing a kick.

takluyver:
Nice find. Let's see what happens!

takluyver:
BTW, I found it confusing checking this, because while zmq streams have a `.closed()` method, sockets have a `.closed` property. I guess neither can really be changed now, but it is unfortunate.

minrk:
That's exactly what caused it. The ZMQStream API, and thus `.closed()` is inherited from tornado's IOStream.

Carreau:
Doh. 

Could we make `channel.closed` a property that return a callable with `__bool__` and `__zero__` that raise an exception ? It should prevent most misuses.

takluyver:
I don't think we as Jupyter should create a subclass just to do that. It might be reasonable to do in pyzmq because it's so easy to confuse these. It is rather special-casing, though.

willingc:
Good sleuthing. Any objection to adding a comment on ZMQStream API ``closed()`` inherited from tornado's IOStream to prevent future devs from changing to sockets ``.closed`` property?

Carreau:
> It might be reasonable to do in pyzmq because it's so easy to confuse these

Yes, that was my though. The discussion started here, so it seemed more relevant. 
And yes it is special casing, but I just want to consider the possibility. 

> Any objection to adding a comment on ZMQStream API closed() inherited from tornado's IOStream to prevent future devs from changing to sockets .closed property?

No objection. But I'm not sure it's useful. It's is inherently obvious from the doc wether it is a property or a method, the hard part is when you misuse it in actual code.

willingc:
@Carreau I completely agree for the Documentation. By comment, I meant a comment in the actual code ;-)

PR #1139, opened 2016-02-24
simplify js build steps

### LINKED ISSUE #1137: setup js does not reliably install webpack
Webpack was recently added as a development dependency, but when I went to run `setup js`, it did not get installed.

Looking at `setupbase.py`, it appears that it only considers running `npm install` if it's also going to run bower. I suspect this is wrong, but I don't know how these pieces fit together.

minrk:
- Run npm, bower install every time (except from sdist).
  Removes need to check npm, bower sources.
  These commands aren't slow enough that we need to save this time (~1s).
- Only check `built/index.js` as build target,
  since webpack is only a single step now.

closes #1137

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -470,17 +470,27 @@ class JavascriptVersion(Command):
     
     def run(self):
         nsfile = pjoin(repo_root, "notebook", "static", "base", "js", "namespace.js")
+        lines = []
+        found = False
         with open(nsfile) as f:
-            lines = f.readlines()
-        with open(nsfile, 'w') as f:
-            found = False
-            for line in lines:
+            for line in f.readlines():
                 if line.strip().startswith("Jupyter.version"):
-                    line = '    Jupyter.version = "{0}";\n'.format(version)
                     found = True
+                    new_line = '    Jupyter.version = "{0}";\n'.format(version)
+                    if new_line == line:
+                        # no change, don't rewrite file
+                        return
+                    lines.append(new_line)
+                else:
+                    lines.append(line)
+
+        if not found:
+            raise RuntimeError("Didn't find Jupyter.version line in %s" % nsfile)
+        
+        print("Writing version=%s to %s" % (version, nsfile))
+        with open(nsfile, 'w') as f:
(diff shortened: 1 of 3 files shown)


blink1073:
:+1:!

takluyver:
Travis failures:

```
      File "/tmp/pip-build-zy691q2m/notebook/setupbase.py", line 506, in run
        targets.extend(js.targets)
      File "/opt/python/3.5.1/lib/python3.5/distutils/cmd.py", line 103, in __getattr__
        raise AttributeError(attr)
    AttributeError: targets
```

setupbase.py
```diff
@@ -326,8 +326,8 @@ def run(cmd, *args, **kwargs):
     return check_call(cmd, *args, **kwargs)
 
 
-class Bower(Command):
-    description = "fetch static client-side components with bower"
+class JavascriptDependencies(Command):
+    description = "fetch static client-side components with npm and bower"
```
takluyver: It's also fetching JS build tools like less and webpack, right?

jdfreder: yup

jdfreder:
Thanks!  This is great!

setupbase.py
minrk: Yup, changed to be more generally accurate "Javascript dependencies"

setupbase.py
```diff
@@ -444,56 +420,39 @@ def initialize_options(self):
...
+        if self.force or not os.path.exists(self.target):
+            print("Missing %s" % self.target)
             return True
-        target_mtime = mtime(target)
-        for source in self.sources(name):
+        target_mtime = mtime(self.target)
+        for source in self.sources():
             if mtime(source) > target_mtime:
-                print(source, target)
+                print('%s > %s' % (source, self.target))
```
takluyver: Should this be using distutils' logging machinery, at a level that's suppressed by default? I could imagine this getting very verbose.

takluyver: Nvm, just spotted that it will only print once, because of the short circuit return.

PR #1142, opened 2016-02-25
Fix rename_file to be case insensitive

### LINKED ISSUE #911: Cannot rename notebook if name change only involves capitalization

captainsafia:
Closes #911.

Follow up on #1100.

cc: @minrk @takluyver

### DIFF
--- a/notebook/utils.py
+++ b/notebook/utils.py
@@ -141,6 +141,23 @@ def is_hidden(abs_path, abs_root=''):
 
     return False
 
+def same_file(path, other_path):
+    """
+    Check if path and other_path are hard links to the same file. This is a
+    utility implementation of Python's os.path.samefile which is not available
+    with Python 2.x and Windows.
+
+    Parameters
+    -----------
+    path:       String representing a path to a file
+    other_path: String representing a path to another file
+
+    Returns
+    -----------
+    same:   Boolean that is True if both path and other path are the same
+    """
+    return os.stat(path).st_ino == os.stat(other_path).st_ino
+
 def to_os_path(path, root=''):
     """Convert an API path to a filesystem path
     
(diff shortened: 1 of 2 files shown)


notebook/utils.py
```diff
@@ -141,6 +141,23 @@ def is_hidden(abs_path, abs_root=''):
...
+    Parameters
+    -----------
+    path:       String representing a path to a file
+    other_path: String representing a path to another file
+
+    Returns
+    -----------
+    same:   Boolean that is True if both path and other path are the same
+    """
+    return os.stat(path).st_ino == os.stat(other_path).st_ino
```
minrk: I think we mentioned in #1100 that this also needs to compare `st_dev`, to ensure that we're checking inodes on the same device. From [CPython](https://github.com/python/cpython/blob/master/Lib/genericpath.py#L81):

```python
def samestat(s1, s2):
    """Test whether two stat buffers reference the same file"""
    return (s1.st_ino == s2.st_ino and
            s1.st_dev == s2.st_dev)


# Are two filenames really pointing to the same file?
def samefile(f1, f2):
    """Test whether two pathnames reference the same actual file"""
    s1 = os.stat(f1)
    s2 = os.stat(f2)
    return samestat(s1, s2)
```

(we don't need it to be two functions)

captainsafia:
@minrk Thanks! Updated code.

minrk:
Thanks!

PR #1144, opened 2016-02-25
Update tests/README.md

julienr:
This updates tests/README.md to point towards using `notebook.jstest` instead of `iptest` (which does not seem to work anymore).


willingc:
Good changes. Thanks @julienr :surfer:

PR #1145, opened 2016-02-25
Freeze selection when performing actions on selected items

minrk:
Avoids async changes to selection causing events to misfire.

Fixes delete being a no-op due to select being cleared by shutdown_selected.

cc @epifanio, who reported that delete stopped working

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -801,7 +801,8 @@ define([
 
     NotebookList.prototype.move_selected = function() {
         var that = this;
-        var num_items = that.selected.length;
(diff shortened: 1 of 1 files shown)


jdfreder:
Thanks @minrk !  Was there an issue associate with this?

minrk:
Yup, I mentioned it above, but maybe it wasn't clear: delete had stopped working because of the change to clear selection after taking an action, since delete calls shutdown first, which clears `that.selection`. Then, by the time the delete action actually fired, `that.selection` was empty, so there were no items to delete.

jdfreder:
Ah, no I meant a GitHub issue #

minrk:
Oh, no. Reported on gitter.

PR #1149, opened 2016-02-26
[Dockerfile change]: Add 2 dependencies to python build layer

dsifford:
Hi there. I'll first start by saying, the regular contributors to this project are much more knowledgeable than I am, so if there is some inherent reason why my additions do not exist, feel free to reject this pull (but please tell me why so I can learn! :innocent: )

I noticed that, in it's current form, I cannot simply run this image, open a terminal, and install the scipy package using pip. After some research, I found that the failures occur because that particular package (and others) depends on `libatlas-base-dev` and `gfortran`.

I created a Dockerfile based off of the current image and added the two packages using apt and I can confirm that this change allows scipy to be installed using pip without any issues. (I also installed `pip-tools` on my image for other reasons, so I suppose my test wasn't exactly a 100% clean test - If this build fails, that'll be why)

Thanks so much for your stellar work on this! :beers:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 # Installs Jupyter Notebook and IPython kernel from the current branch
 # Another Docker container should inherit with `FROM jupyter/notebook`
 # to run actual services.
+#
+# For opinionated stacks of ready-to-run Jupyter applications in Docker, 
+# check out docker-stacks <https://github.com/jupyter/docker-stacks>
 
 FROM jupyter/ubuntu_14_04_locale_fix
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
The base jupyter notebook image is meant to be standalone with the core notebook features. For docker images that include fortran, scipy, etc. check out [docker-stacks](https://github.com/jupyter/docker-stacks).

dsifford:
@rgbkrk Doh! Sorry for overlooking. :blush:

rgbkrk:
That's ok. It makes me wonder if we're missing discoverability there. Would you mind adding a comment to the top of the Docker file referencing Docker stacks?

dsifford:
Done. 

Comment added to line 5 + removed both initial additions from the python layer.

Thanks again.

rgbkrk:
Awesome. Thank you!

PR #1165, opened 2016-03-03
Add about page

blink1073:
screen shot 2016-03-03 at 12 28 27 pm


minrk:
Awesome, thanks @blink1073!

willingc:
@blink1073 What's neat is that as the alpha moves forward this page can be used to highlight or point out added functionality :smile:

blink1073:
Yep, thanks for pointing out the blindspot!

Issue #411, opened 2015-09-10
the title has double backslash 


Jeffiy:
when I set the notebook_dir end with backslash.
![1](https://cloud.githubusercontent.com/assets/5069316/9778856/7e062948-57a8-11e5-95c8-4a4a2132c91d.png)
the title has double backslash
![2](https://cloud.githubusercontent.com/assets/5069316/9778859/852a79e0-57a8-11e5-99a2-f44e61f500de.png)
Is this a problem?

Carreau:
you should have any problem, but we could try fix it still.

Carreau:
**shouldn't**, typed too fast.

Jeffiy:
yeah, that is a minor problem. Thanks.

minrk:
rstrip `os.sep` on notebook_dir should do it.

captainsafia:
@Carreau — I'll take on this one for 4.1 Can you set me as assignee?

Carreau:
> I'll take on this one for 4.1 Can you set me as assignee?

We cannot assign people without commit rights. Just dropping a line that you are working on it is sufficient.

Carreau:
> We cannot assign people without commit rights. Just dropping a line that you are working on it is sufficient.

Actually I gave the "student" group, read access so that I can assign you.

captainsafia:
:smile:

PR #1168, opened 2016-03-04
allow notebook-dir to be root

minrk:
we were stripping `/`, which is fine except when that's the whole string,
which we were then interpreting as `''`, which gives the CWD.

This is a regression in 4.1 (#411)

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -805,6 +805,11 @@ class NotebookApp(JupyterApp):
 
     def _notebook_dir_validate(self, value, trait):
         # Strip any trailing slashes
+        # *except* if it's root
+        _, path = os.path.splitdrive(value)
+        if path == os.sep:
+            return value
+
         value = value.rstrip(os.sep)
 
         if not os.path.isabs(value):
--- a/notebook/tests/test_notebookapp.py
+++ b/notebook/tests/test_notebookapp.py
@@ -68,6 +68,11 @@ def test_nb_dir_with_slash():
         app = NotebookApp(notebook_dir=td)
         nt.assert_false(app.notebook_dir.endswith("/"))
 
+def test_nb_dir_root():
+    root = os.path.abspath(os.sep) # gets the right value on Windows, Posix
+    app = NotebookApp(notebook_dir=root)
+    nt.assert_equal(app.notebook_dir, root)
+
 def test_generate_config():
     with TemporaryDirectory() as td:
         app = NotebookApp(config_dir=td)


jmarrec:
On my Win10 x64 I still have the problem where it won't serve the root of either my C:\ or D:\ drives.

```
notebook                  4.2.1                    py35_0
jupyter                      1.0.0                    py35_3
jupyter_client            4.3.0                    py35_0
jupyter_console        4.1.1                    py35_0
jupyter_core             4.1.0                    py35_0
```

jmarrec:
I'm not familiar at all with the code base of this project, but I did a bit of digging since this issues has been bugging me for quite some time (see this [SO](http://stackoverflow.com/questions/33366624/jupyter-refuses-to-serve-hidden-directory-d-on-windows) post from October 2015). I hope these findings are helpful.

It throws an error here https://github.com/jupyter/notebook/blob/master/notebook/tree/handlers.py#L43

The end reason is on this line: https://github.com/jupyter/notebook/blob/master/notebook/utils.py#L135, which is called (I believe) here: https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L154

I think that the line

    path = path.strip('/')

might need to be replaced by:

    path = path.strip(os.sep)

so that it works on windows too. Indeed, is_hidden('C:') works fine (attrs is 22), whereas is_hidden('C:\\') fails (attr is 22)


Just above here: https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L871, it's may be needed to add the following two lines:

     # Start by normalizing the path, otherwise checking if equal to os.sep will fail
     value = os.path.normpath(value)

So far with this it still won't work, not sure why

takluyver:
> path = path.strip('/')

I think this is called on an 'API style' path. These are like URL paths in that they always use forward slashes, even on Windows.

>  value = os.path.normpath(value)

This doesn't appear to make any difference, though:

```
In [1]: import ntpath

In [2]: ntpath.normpath("C:\\")
Out[2]: 'C:\\'
```

jmarrec:
I actually tried passing `C:`, `C:\\`, `C:/` and therefore found normpath useful for the latter, otherwise the line `path == os.sep:` returns False all the time.

I'm not sure what's the solution, I just know it's still finding it to be hidden.

is_hidden('/'), is_hidden('C:/'), is_hidden('C:\\') all return True.

is_hidden('//'), is_hidden('C:') return false.

is_hidden ends up being called with "C:\\" as the abs_path, should that be API style or windows style (C:, not C:\)?

Aside from whether this makes sense to check for hidden or not (I'm sure you have good reasons to have implemented this), the is_hidden is maybe not complete either. It seems that it finds my D: drive to be hidden (even if I passed 'D:' to it), I think because I have a single hidden directory. I use Box Sync, and I have a hidden `.boxsync` directory as a result.

I think as a temporary workaround I'll force is_hidden to return False all the time.

Nomii5007:
@minrk hello sir.. I am facing same problem. couldn't run jupyter form C:\. and also couldn't understand the solution you provided here. I am using anaconda3 4.1.1 in windows 8.1 64 bit. can u help me?

ar-jan:
This is still a problem, and it looks like it's because of administrator / read-only / hidden folders present in the root of the drive as @jmarrec hinted. I created a symlink folder pointed at my e:\ drive, which does allow Jupyter Notebook to start, but the jupyter notebook command gives `E:\E-Jupyter\_SMSTaskSequence doesn't exist` in the command prompt window.

Perhaps Jupyter Notebook could not try to read/serve any administrator/read-only/hidden folders and files?

PR #1180, opened 2016-03-07
Clarify notebook heading in new menu

takluyver:
As discussed on jupyter/jupyter#97, it's not very clear that the 'Notebooks' entry in the new menu is supposed to be a heading instead of a disabled menu entry. This aims to make the distinction clearer.

Before:

nb new menu before

After:

notebook new menu after

### DIFF
--- a/notebook/static/tree/less/tree.less
+++ b/notebook/static/tree/less/tree.less
@@ -256,6 +256,13 @@ ul#new-menu {
     right: 0;
 }
 
+#new-menu .dropdown-header {
+    font-size: 10px;
+    border-bottom: 1px solid #e5e5e5;
+    padding: 0 0 3px;
+    margin: -3px 20px 0;
+}
+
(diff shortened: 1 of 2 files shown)


Carreau:
+1 for the `:`. 

The 10px is too small IMHO. 

Alternatively:
```
#new-menu .dropdown-header {
    border-bottom: 1px solid #e5e5e5;
    padding: 0 6px 3px;
    margin: -3px 0px 0;
}
```

screen shot 2016-03-07 at 12 23 21

I **think** the dedent convey even more that this is a section **header**.

takluyver:
I did a dedented mockup on jupyter/jupyter#97, but @cameronoelsen didn't include a dedent in any of his proposed designs, so I'm assuming there's a design-y reason not to do that. This implements my favourite of the designs he proposed.

Carreau:
Ok, merging, let's leave people to complain.

PR #1187, opened 2016-03-08
add missing url-encoding of base_url in terminal/edit templates

minrk:
This was fixed in the notebook and tree views, but overlooked in edit, terminal.

This completes the fix in #472 and should be backported to 4.2, and 4.1.1 if there is one.


yuvipanda:
+1. I'd also appreciate a 4.1.1 :D

yuvipanda:
Tested to work

minrk:
Thanks, @yuvipanda!

PR #1199, opened 2016-03-11
[WIP] [Image attachment] Add visual indicator that a drop is possible

### LINKED ISSUE #1186: Add indication that file drop in MD cell will work.
When you are dropping a file in an MD cell, the UI that the file will be attached (or not) is too subtil. 
Would it be possible to add a specific class to the cell when dragging is in the right place ? 

We can Bikeshed on the style later.

julienr:
This adds a visual indicator that a drop is possible when the user drags an image file over a markdown cell. This fixes #1186

This currently highlight the markdown cell with a red border as you can see in the animation below. Some work is needed on the design side, comments/ideas are welcome.

Note that this depends on this code mirror PR : https://github.com/codemirror/CodeMirror/pull/3882

TODO :
- [x] Improve design
- [x] Avoid triggering the effect when drag'n'droping text in edit mode
- [x] Wait for codemirror PR to be merged
- [x] Update bower dependency to the official codemirror

quickcast-11-03-2016-08-20-03

### DIFF
--- a/notebook/static/notebook/less/textcell.less
+++ b/notebook/static/notebook/less/textcell.less
@@ -48,8 +48,9 @@ h1,h2,h3,h4,h5,h6 {
     display:none;
 }
 
-.text_cell .droppable{
-  border: 1px solid #ff0000;
+.text_cell .dropzone .input_area {
+  border-color: #FCFF1E;
+  background-color: #FCFFE3;
 }
 
 .cm-header-1,
--- a/notebook/static/notebook/less/textcell.less
+++ b/notebook/static/notebook/less/textcell.less
@@ -48,6 +48,10 @@ h1,h2,h3,h4,h5,h6 {
     display:none;
 }
 
+.text_cell .droppable{
+  border: 1px solid #ff0000;
+}
+
 .cm-header-1,
 .cm-header-2,
 .cm-header-3,
--- a/notebook/static/notebook/less/textcell.less
+++ b/notebook/static/notebook/less/textcell.less
@@ -49,8 +49,8 @@ h1,h2,h3,h4,h5,h6 {
 }
 
 .text_cell .dropzone .input_area {
-  border-color: #FCFF1E;
-  background-color: #FCFFE3;
+  border: 2px dashed #bababa;
+  margin: -1px;
 }
 
 .cm-header-1,
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "backbone": "components/backbone#~1.2",
     "bootstrap": "components/bootstrap#~3.3",
     "bootstrap-tour": "0.9.0",
-    "codemirror": "julienr/CodeMirror#fix_dragleave",
+    "codemirror": "codemirror/CodeMirror#master",
     "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.2.0",
     "google-caja": "5669",
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "backbone": "components/backbone#~1.2",
     "bootstrap": "components/bootstrap#~3.3",
     "bootstrap-tour": "0.9.0",
-    "codemirror": "codemirror/CodeMirror#master",
+    "codemirror": "~5.14",
     "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.2.0",
     "google-caja": "5669",
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "backbone": "components/backbone#~1.2",
     "bootstrap": "components/bootstrap#~3.3",
     "bootstrap-tour": "0.9.0",
-    "codemirror": "~5.8",
+    "codemirror": "julienr/CodeMirror#fix_dragleave",
     "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.2.0",
     "google-caja": "5669",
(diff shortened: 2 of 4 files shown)


ellisonbg:
@cameronoelsen want to have a go at the visual design of this hover effect?

Carreau:
Any new on visual design ?

notebook/static/notebook/less/textcell.less
```diff
@@ -48,6 +48,10 @@ h1,h2,h3,h4,h5,h6 {
     display:none;
 }
 
+.text_cell .droppable{
```
Carreau: can you target the `.input_area` in particular, which already have a border, so this will avoid the change of size. And you can set only border-color.

Carreau:
What would append if we merge that with an old version of codemirror ? Can we just remove the border on cell render ? So that at least this is in, and the behavior will just "improve" on next release of CM.

cameronoelsen:
This problem actually brought up a design issue related to the field being edited that I have seen for a little while. Currently when you are editing a cell and its the one selected (like in the example gif attached), the side indicator changes from blue to this green:

screen shot 2016-04-08 at 2 46 17 pm

But, why do we need to change this color on the side when it is actually the cell being edited? To make it apparent to the user that it is the current cell being edited and that it is the one selected, couldn't we just make the cell itself turn from the gray background to white along with adding a blue glow that matches the blue on the left side showing it is the cell selected. This accomplished both tasks of showing which cell is being selected along with that it is in focus/editing mode:

editing

Using this method of cell selection enables us to then solely change the color of the hue of the glow around the cell to show that we are dropping an image into the cell/doing a different action than editing the text within the cell:

dropzone

I think doing so also fixes some collisions in colors that we will face if we add yet another color to our Notebook UI. Thoughts?

Carreau:
The 1px border and glow change of color is not sufficient enough for some mildly colorblind people. Heave the change of the border. 

For a cell with long output, changing the color of the focus area does not help. 

Example : 

screen shot 2016-04-08 at 16 16 41

You would not know whether you are editing or not.

julienr:
> What would append if we merge that with an old version of codemirror ? Can we just remove the border on cell render ? So that at least this is in, and the behavior will just "improve" on next release of CM.

Well, we could check for CodeMirror.version >= 5.13.5 before installing the listeners for this functionnality. But now that the code mirror PR has been merged, can't we just bump the requirement in bower.json to 5.13.5 ?

minrk:
> can't we just bump the requirement in bower.json to 5.13.5 ?

We certainly can and should, as soon as 5.13.5 is released.

julienr:
I've changed the effect to highlight the target dropzone in yellow. Suggestions for a better color choice are welcome :) 

The effect is removed on cell render, so it looks like that with codemirror < 5.13.5
codemirror_5_8

And like that with the codemirror master :
codemirror_5_13_5

A remaining issue is that the effect gets triggered if you use select - drag'n'drop to move text within the cell. I need to work on this.

I am also not 100% confident that the implementation works across all browsers. I've tested Firefox and Chrome seems ok, but the HTML5 drag'n'drop API [has a lot of quirks](http://stackoverflow.com/questions/7110353/html5-dragleave-fired-when-hovering-a-child-element).

Carreau:
> The effect is removed on cell render, so it looks like that with codemirror < 5.13.5

I think that a minor annoyance. I would be in favor of living with that on master until 5.13.5 is released. 

And I don't mind the yellow.

ellisonbg:
Solid yellow is too loud for this. Visual design should reflect the
relative importances on the page and we should always err towards the least
dramatic styling that gets the job done.

Some examples of this type of UI:

* Gmail uses a thick, inset dotted grey line with "Drop Files Here" text in
that same grey.
* GitHub uses a yellow shadow outside their usual grey border.

Based on this, I prefer:

* Grey rather than yellow
* Simple text
* Inset rather than outset
* Border/line rather than solid fills

@cameronoelsen

On Sat, Apr 16, 2016 at 3:14 AM, Matthias Bussonnier <

> The effect is removed on cell render, so it looks like that with
> codemirror < 5.13.5
>
> I think that a minor annoyance. I would be in favor of living with that on
> master until 5.13.5 is released.
>
> And I don't mind the yellow.
>
> —
> You are receiving this because you commented.
> <https://github.com/jupyter/notebook/pull/1199#issuecomment-210785188>
>



-- 
Brian E. Granger

julienr:
Following @ellisonbg remarks, here is the new look with a dashed grey inset border.

dropzone_dashed_gray

We could add a gmail-like "Drop your files here" text, but this require adding some DOM overlay elements (as opposed to just tagging the input_area with a class). Any thoughts ?

There is also an unrelated usability question : currently, if you drag'n'drop an image on something that's not an edit mode text area, this will cause the browser to navigate to the image and display it (leaving the notebook). We could prevent that by registering a ``$(document).on("dragover")`` that prevents this default behavior. In gmail for example, they do this only if you have a compose window open.

Carreau:
> There is also an unrelated usability question : currently, if you drag'n'drop an image on something that's not an edit mode text area, this will cause the browser to navigate to the image and display it (leaving the notebook). We could prevent that by registering a $(document).on("dragover") that prevents this default behavior. In gmail for example, they do this only if you have a compose window open.

We can do that in a subsequent PR. and refine later anyway.

New version of Codemirror has been released, so we can assume the drag event are now handeled.

If you just switch back the codemirror version to 5.14 or above I can merge.

julienr:
@Carreau I've bumped codemirror to 5.14.

Carreau:
Thanks !

julienr:
Thanks for the merge and for the feedbacks !

PR #1212, opened 2016-03-15
Fix notebook scrolling in JupyterLab

blink1073:
cf https://github.com/jupyter/jupyter-js-plugins/pull/105

Also adds more shortcuts and palette items:

screen shot 2016-03-15 at 12 40 24 pm

### DIFF
--- a/notebook/static/lab/package.json
+++ b/notebook/static/lab/package.json
@@ -6,7 +6,7 @@
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "dependencies": {
-    "jupyter-js-plugins": "^0.11.0",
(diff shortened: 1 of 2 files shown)


blink1073:
cc @SylvainCorlay

minrk:
Nice! Scrolling works, but I can no longer type the letter 'i' in a cell. Any idea what's up there?

blink1073:
Ack!  We'll need another PR to pick up the fixes to the interrupt/restart shortcuts that caused your problem.

blink1073:
Nevermind, I just released a patch version of plugins so anyone updating will get the latest, we can wait for the next round of changes.

minrk:
Awesome, thanks.

Issue #1214, opened 2016-03-15
Option to toggle header in edit mode.


Summersa:
#1215
I think there should be an option to toggle the header in edit mode.

This is currently what the view option have in the edit page.

![headheaderbefore](https://cloud.githubusercontent.com/assets/15676683/13789970/eb4655be-eaab-11e5-8b6e-b68b666ecd1e.png)

Carreau:
#1215 should have fixed that. 

Thanks !

PR #1215, opened 2016-03-15
Toggle Header in Edit page

Summersa:
#1214 

There is no function to hide the header in edit mode. Now I know my syntax is wrong. I couldn't find a way to do this from the edit page javascript file. If I could get some guidance and than I could write it properly. Right now I just put the javascript within the html edit file.

Before:
headheaderbefore
 #After:
headheaderafter

### DIFF
--- a/notebook/templates/edit.html
+++ b/notebook/templates/edit.html
@@ -65,7 +65,9 @@ data-file-path="{{file_path}}"
             </li>
             <li class="dropdown"><a href="#" class="dropdown-toggle" data-toggle="dropdown">View</a>
               <ul id="view-menu" class="dropdown-menu">
-                <li id="menu-line-numbers"><a href="#">Toggle Line Numbers</a></li>
+              <li id="toggle_header" title="Show/Hide the IPython Notebook logo and notebook title (above menu bar)">
+              <a href="#">Toggle Header</a></li>
+              <li id="menu-line-numbers"><a href="#">Toggle Line Numbers</a></li>
               </ul>
             </li>
             <li class="dropdown"><a href="#" class="dropdown-toggle" data-toggle="dropdown">Language</a>
--- a/notebook/static/edit/js/menubar.js
+++ b/notebook/static/edit/js/menubar.js
@@ -112,6 +112,11 @@ define([
         });
         
         // View
+
+        this.element.find('#toggle_header').click(function (){
+            $("#header-container").toggle();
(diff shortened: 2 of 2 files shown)


notebook/templates/edit.html
```diff
@@ -17,10 +19,16 @@
...
-    <span class="last_modified"></span>
+    <span id="notebook_name" class="filename"></span>
+    <span class="checkpoint_status"></span>
+    <span class="autosave_status"></span>
+</span>
+
+<span id="kernel_logo_widget">
+  {% block kernel_logo_widget %}
+  <img class="current_kernel_logo" src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7"/>
+  {% endblock %}
```
minrk: The changes here to the save widget shouldn't be needed, since they are all notebook-specific.

notebook/templates/edit.html
```diff
@@ -95,5 +109,12 @@
 
     {{super()}}
 
+<script>
+$(document).ready(function(){
+    $("#toggle_header").click(function(){
```
minrk: You can put this code in MenuBar.bind_events in notebook/static/edit/js/menubar.js.

To see updates to the js while you are working, it's a good idea to run:

    npm run build:watch

in the notebook repo. That way, your changes will get recompiled when you make them.

Summersa: Thank you for your experience. Just wasn't sure where you guys wanted the javascript. Yea I remember putting in stuff I didn't need thought I got rid of it. 

Also npm run build:watch does not work for me. I'm just using sublime save and refresh page. Probably a tad slower but that's alright for now.

I'll put the changes soon.

Summersa:
This will be good now.

minrk:
Great, thanks! It looks like you have a few extra commits that aren't part of this PR. Can you reset to the commit with the real changes here? There are a few ways to do this (rebase, reset).

Since it looks like there are a few commits at the end you don't need, you can reset to the real changes with:

    git reset d120344 --hard

And then re-push, adding `-f` to allow rewriting history:

    git push -f # may need to specify origin and/pr branch, depending on your git situation

captainsafia:
I think this is good to merge! :+1:

minrk:
Terrific, thanks @Summersa!

PR #1221, opened 2016-03-16
retry js tests on travis

minrk:
This is a lazy thing to do because intermittently failing tests are often actual bugs (usually in the tests themselves). But if we are just going to kick Travis to re-run when this happens, we might as well let Travis do it automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
     - pip install -f travis-wheels/wheelhouse file://$PWD#egg=notebook[test]
 
 script:
-    - 'if [[ $GROUP == js* ]]; then python -m notebook.jstest ${GROUP:3}; fi'
+    - 'if [[ $GROUP == js* ]]; then travis_retry python -m notebook.jstest ${GROUP:3}; fi'
     - 'if [[ $GROUP == python ]]; then nosetests -v --with-coverage --cover-package=notebook notebook; fi'
 
 matrix:


takluyver:
From to the docs, this will retry up to three times if the exit status is not zero. Given the frequency of intermittent failures, that's probably reasonable, though it will make things substantially slower when there is a real, replicable failure.

My bigger concern is that we might introduce a real bug which causes a test failure ~50% of the time. With the current test regime, we'd have a good chance of seeing that quickly, but if failing tests are automatically re-run, it can hide much better. Of course, if we always manually re-run failing tests anyway...

blink1073:
You could use the `true || echo "hello"` syntax to only ever retry once.

minrk:
@takluyver your point is certainly valid, but examining my own (bad) behavior I often don't even look at the failure before relaunching it once. The only time I do that is if I see that the same test group failed on multiple versions of Python, which suggests a real bug.

I dusted off my (possibly wrong) dice-rolling probabilities from ages ago to check out how likely a failure needs to be for this to hide it: https://nbviewer.jupyter.org/gist/minrk/a87b89aa46a24ea77b0f

If there's a failure in a single test group 50% of the time, the fact that we are running each group twice (py2, py3) means that these retries will reduce that from at-least-one failure in 75% of builds to at-least-one failure in 25% of builds, which is still noticeable. If it's affecting all the js test groups equally, it would still produce a single failure 65% of the time, as opposed to the current 99.6% of the time.

What this approach hides most is low-probability failures. For instance, a js failure affecting all js test groups 10% of the time will only show up 1% of the time after this change, rather than 60% of the time now.

takluyver:
:+1: for modelling it in a notebook :-). Those are roughly the numbers I was thinking of, though my point would have been better made by a low-probability failure.

However, I agree with you - even though I do generally scan over the output from a failure before restarting it, the non-determinate ones are most likely to fail with a timeout, and I see enough of those that I'll probably restart it without much thought. And we have our nightly runs on Jenkins to increase the visibility of real errors as well.

willingc:
I'm going to go ahead and merge this since we can change if needed down the road.

@takluyver Where are the Jenkins logs kept?

Has anyone tried https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image with the JS builds?

takluyver:
You can see Jenkins logs on https://jenkins.jupyter.org/ . If you're keen, you can also subscribe to the [ipython-build mailing list](http://lists.ipython.org/listinfo.cgi/ipython-build-ipython.org) to get messages when builds fail. Not all jobs are set up to send emails, though.

I think our main issue with Travis is that certain operations are much slower on Travis than in any realistic use case, so things time out even when we set the wait time really generously, and it hits what are technically race conditions that would never normally come up.

PR #1230, opened 2016-03-18
Re-factor ANSI color handling

mgeier:
This is based on a PR I made for `nbconvert`: https://github.com/jupyter/nbconvert/pull/225.

It also contains the "upgrade" to 16 ANSI colors from a future PR, see https://github.com/jupyter/nbconvert/issues/245 and https://github.com/jupyter/nbconvert/pull/259.

I've not yet updated the tests because I'd like to get some general feedback before.

WARNINGS:

* I'm new to JavaScript, this is my first non-trivial use of it. Please tell me anything that can be improved or made more idiomatic.
* I don't know if this has any performance implications
* The implementation uses some ES6 features, I hope there is some kind of transpiler involved to make this work on older browsers

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -288,7 +288,7 @@ define([
             }
             var chunk = str.substring(start, match.index);
             if (chunk) {
-                if (bold && 0 <= fg && fg < 8) {
+                if (bold && typeof fg === "number" && 0 <= fg && fg < 8) {
                     fg += 8;  // Bold text uses "intense" colors
                 }
                 var classes = [];
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -239,13 +239,13 @@ define([
             g = numbers.shift();
             b = numbers.shift();
             if ([r, g, b].some(function (c) { return c < 0 || 255 < c; })) {
-                throw new RangeError();
+                throw new RangeError("Invalid range for RGB colors");
             }
         } else if (n === 5 && numbers.length >= 1) {
             // 256 colors
             var idx = numbers.shift();
             if (idx < 0) {
-                throw new RangeError();
+                throw new RangeError("Color index must be >= 0");
             } else if (idx < 16) {
                 // 16 default terminal colors
                 return idx;
@@ -261,10 +261,10 @@ define([
                 // grayscale, see http://stackoverflow.com/a/27165165/500098
                 r = g = b = (idx - 232) * 10 + 8;
             } else {
-                throw new RangeError();
+                throw new RangeError("Color index must be < 256");
             }
         } else {
-            throw new RangeError();
+            throw new RangeError("Invalid extended color specification");
         }
         return [r, g, b];
     }
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -4,9 +4,10 @@
 define([
     'codemirror/lib/codemirror',
     'moment',
+    'underscore',
     // silently upgrades CodeMirror
     'codemirror/mode/meta',
-], function(CodeMirror, moment){
+], function(CodeMirror, moment, _){
     "use strict";
     
     // keep track of which extensions have been loaded already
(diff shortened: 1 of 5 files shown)


mgeier:
When I'm running the tests, I'm getting this (the same happens at Travis CI):

```
Page Error
  Message:   SyntaxError: Parse error
  Call stack:
Page Error
  Message:   SyntaxError: Parse error
  Call stack:
```

Does that mean that ES6 is not supported?
Or is it some other problem?

minrk:
@mgeier ES6 isn't used in the notebook js yet, so you have to use ES5.

mgeier:
@minrk OK, will do. What are the plans for ES6 support?

minrk:
We are adding it in new projects in various places (JupyterLab, ipywidgets, etc.). I'm not sure if we will add ES6 support to the old notebook js, as it is planned to be replaced by the totally rewritten frontend code in JupyterLab, which is mostly using TypeScript and ES6.

mgeier:
OK, thanks, that sounds reasonable.

I've re-based my initial commit (7bee6c5c2d6934989ee542d46af3d814de7c5bec).

Then I've removed the ES6 stuff (245287a83c7c633eccba6a9b530846dafad61416).
During that, I've also found out that conversion from a string to an integer number is insanely complicated in JavaScript and extended my code a bit. I hope now it's correct, but I've used another regex, so it might have become a bit slower.

Finally, I've updated the tests and added a new test case for overlapping formatting, which wasn't supported before (9d052588b574b218d830a62f8f8efc8484776caf).
During that I found a little bug, so adding this test case already payed off.

Now Travis CI doesn't complain anymore, hooray!

I would love to hear some feedback about the new code!

What about the colors? Does everybody like them? Any suggestions for improvements?

notebook/static/base/js/utils.js
```diff
@@ -192,165 +193,231 @@ define([
...
+                numbers.push(parseInt(item));
             } else {
-                attrs["class"] += " " + ansi_colormap[n];
-            }
-        } else if (n == "38" || n == "48") {
-            // VT100 256 color or 24 bit RGB
-            if (numbers.length < 2) {
-                console.log("Not enough fields for VT100 color", numbers);
+                // Ignored: Invalid color specification
+                numbers.length = 0;
```
minrk: No need to set length = 0 if we are returning here, plus length should be treated as read-only in general.

notebook/static/base/js/utils.js
```diff
@@ -192,165 +193,231 @@ define([
...
+                b = (idx - 16) % 6;
+                b = b > 0 ? 55 + b * 40 : 0;
+            } else if (idx < 256) {
+                // grayscale, see http://stackoverflow.com/a/27165165/500098
+                r = g = b = (idx - 232) * 10 + 8;
+            } else {
+                throw new RangeError();
             }
+        } else {
+            throw new RangeError();
```
minrk: It's a good idea to be informative on these RangeErrors, indicating what expected condition is not satisfied

notebook/static/notebook/js/outputarea.js
```diff
@@ -659,9 +659,9 @@ define([
     var append_text = function (data, md, element) {
         var type = 'text/plain';
         var toinsert = this.create_output_subarea(md, "output_text", type);
+        data = utils.fixCarriageReturn(data);
```
minrk: Is this reordering required? I'm not objecting to it, just wondering if there was a bug caused by the previous order.

notebook/static/base/js/utils.js
mgeier: I actually have to clear `numbers` here, because there might already be some entries from previous loop iterations.

Would you prefer if I used

```js
numbers = [];
```

Or something entirely different?

notebook/static/base/js/utils.js
mgeier: OK

notebook/static/notebook/js/outputarea.js
mgeier: I'm not sure.
It seems more logical to me to first fix the carriage returns and then convert to HTML.

And it's done like this at a different place already.
At least it should be consistent, right?

Either way, the handling of CRs w.r.t. the terminal state machine isn't quite correct.
But I guess in practice, this has little relevance.

mgeier:
@minrk Thanks for your comments, I've added a commit: 52fae535344e46f9b7951d13d14dc68d7b301088

notebook/static/base/js/utils.js
mgeier: I just saw that there is an error because I'm not returning an Array in this case!

Would it be adequate to do this:

```js
return [];
```

?

mgeier: ... or should I rather use `break;`?

mgeier: As a different option, I tried to inline this function: db651f79741d7bcd36c7bd51ae525262f7650d8c

Now the `numbers` Array is created only once, and never re-assigned.

What do you think about this?
If you don't like it, I can remove this commit again.

minrk:
Thanks!

mgeier:
Thanks for merging!

willingc:
Thanks @mgeier!

KristofferC:
Thank you for this, looking forward to the new release.

PR #1231, opened 2016-03-18
Add `cookie_options` to make cookie args configurable

minrk:
also make cookies httponly by default, since we do not need or want access to cookies in js

### DIFF
--- a/notebook/auth/login.py
+++ b/notebook/auth/login.py
@@ -40,6 +40,7 @@ class LoginHandler(IPythonHandler):
     def post(self):
         typed_password = self.get_argument('password', default=u'')
         cookie_options = self.settings.get('cookie_options', {})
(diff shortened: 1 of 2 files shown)


willingc:
@minrk This looks fine. I'm trying to reason through if a user can have a malicious entry in the cookie_options preconfigured before executing ``self.set_secure_cookie(self.cookie_name, str(uuid.uuid4()), **cookie_options)``. Nothing's jumping out at me after looking at the tornado source, but thought I would mention just in case you see anything. If not, merge :+1:

minrk:
@willingc there's loads the user can do when in control of the server config, including arbitrarily change the behavior of any part of the server, so I don't think we need to consider malice in server-side config.

Thanks for the review!

Issue #1204, opened 2016-03-14
Casperjs notebook/save.js skipping some asserts


julienr:
While working on my own js tests, I ran into something strange. It seems that calling ``open_dashboard`` in a JS test causes casperjs to silently skip all following steps.

This can be seen in travis builds. For example in the [latest one]( https://travis-ci.org/jupyter/notebook/jobs/115562243#L1294), the ``notebook/save.js`` tests gives the following output :

```
Test file: /home/travis/build/jupyter/notebook/notebook/tests/notebook/save.js  
PASS Save did not fail
PASS Save OK
PASS Save with complicated name
PASS path OK
PASS checkpoints OK
Test file: /home/travis/build/jupyter/notebook/notebook/tests/notebook/shutdown.js 
```

But if you look at the [test code](https://github.com/jupyter/notebook/blob/master/notebook/tests/notebook/save.js#L75), you see that there are two more asserts following the ``checkpoints OK`` one. The full test should print ``Escaped URL in notebook list`` and ``Notebook name is correct``. To confirm, if you add ``this.test.assert(false);`` in the last ``then``, you'll still get a passing test.

I can reproduce the same behavior on my OSX machine with phantomjs=2.1.1 and casper=1.1.0-beta5.

I tried debugging by adding some logging at the beginning of the ``notebook/save.js`` file :
```
    casper.on('remote.message', function(message) {
        this.echo('remote message caught: ' + message);
    });

    casper.options.onStepComplete = function(casper, result) {
        console.log('step complete ' + result);
    };
```

You can see the full output in [this gist](https://gist.github.com/julienr/1f7c3a953c5a418d7a95) : I get the following error printed :
```
remote message caught: Object {
  "readyState": 4,
  "responseText": "{\"reason\": null, \"message\": \"No such file or directory: has#hash and space and unic\\u00f8\\u2202e.ipynb\"}",
  "responseJSON":
    "reason": null,
    "message": "No such file or directory: has#hash and space and unicø∂e.ipynb"
  "status": 404,
  "statusText": "Not Found"
}
```

If I run with an external notebook server (using ``--url=http://localhost:8889``), I get some websocket errors with status code 426 as you can see in the second log of the gist. I am not sure if this is related.

I am quite new to phantomjs & casperjs, so I'm not really sure what's going on. Ideas welcome :-)

julienr:
[...]

PR #1232, opened 2016-03-18
Use casper.open() instead of casper.start() to open the dashboard.

julienr:
This is a fix for #1204 
Calling ``casper.start()`` multiple times seems to cause casper to skip subsequent then() call, skipping some asserts. Using ``casper.open()`` instead fixes this.

Example test run on save.js with ``python -m notebook.jstest notebook/save.js``

#### Before
```
Test file: notebook/tests/notebook/save.js
PASS Save did not fail
PASS Save OK
PASS Save with complicated name
PASS path OK
PASS checkpoints OK
PASS 5 tests executed in 4.354s, 5 passed, 0 failed, 0 dubious, 0 skipped.
```
#### After
```
Test file: notebook/tests/notebook/save.js
PASS Save did not fail
PASS Save OK
PASS Save with complicated name
PASS path OK
PASS checkpoints OK
PASS Escaped URL in notebook list
PASS Notebook name is correct
PASS 7 tests executed in 5.459s, 7 passed, 0 failed, 0 dubious, 0 skipped.
```

### DIFF
--- a/notebook/tests/util.js
+++ b/notebook/tests/util.js
@@ -694,17 +694,31 @@ casper.wait_for_dashboard = function () {
     casper.waitForSelector('.list_item');
 };
 
-casper.open_dashboard = function () {
+/**
+ * Open the dashboard page
+ * @param {bool} use_start - If true, will use casper.start(), otherwise
+ *      casper.open(). You should only set it to true if the dashboard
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #1234, opened 2016-03-20
Make keyboard shortcuts declarative

Carreau:
look in config for `keys.command.unbind`, `keys.edit.unbind` for which shortcuts to unbind, 
then `keys.command.bind`, `keys.edit.bind` for new shortcut to bind.

Done after feedback at JupyterDays.

Command Palette hover of Shortcut area show the action name as a tooltip so that users know what to bind.

Cleanup Js here and there, and replace `(command)` in command palette with `(command mode)` to remove confusion with Cmd keys on Mac.


notebook/static/base/js/keyboard.js
```diff
@@ -404,14 +409,15 @@ define([
...
-        {
+        var that = this;
+        try {
           this._remove_leaf(shortcut, this._shortcuts);
           if (!suppress_help_update) {
             // update the keyboard shortcuts notebook help
             this.events.trigger('rebuild.QuickHelp');
           }
         } catch (ex) {
+          console.warn('shortbut', shortcut, '...',this._shortcuts);
```
minrk: shortbut?

notebook/static/notebook/js/keyboardmanager.js
```diff
@@ -42,6 +42,26 @@ define([
...
         this.edit_shortcuts.add_shortcuts(this.get_default_common_shortcuts());
         this.edit_shortcuts.add_shortcuts(this.get_default_edit_shortcuts());
+
+
+        this.config = options.config;
+        var that = this;
+
+        this.config.loaded.then(function(){ 
+            
+            (((that.config.data.keys||{}).edit||{})
```
minrk: This is extremely dense, and I can't really follow where the different parentheses start and stop. Maybe expand it a little bit?

minrk:
A few minor comments, but :+1:

notebook/static/notebook/js/keyboardmanager.js
Carreau: expanded, but that look a bit ugly for just avoiding typeerror if one of them is undefined: 

```
 +            try {
 +                edit_unbind = that.config.data.keys.edit.unbind;
 +            } catch (e) {
 +                if (e instanceof TypeError) {
 +                    edit_unbind = [];
 +                } else {
 +                    throw e;
 +                }
 +            }
```

ellisonbg:
Is this API compatible?

Carreau:
> Is this API compatible?

With what ?

Carreau:
@willingc added a bit of documentation. Can I get your feedback on it ?

Carreau:
> Is this API compatible?
> With what ?

More especially: it is purely an addition of a config section. It maps what user were doing in their `custom.js` with a corresponding config entry (we can bikeshed on the name if you like). 

There is 0 API deletion, 0 API addition, 0 API modification. Just the **provisional** (added in the docs) addition of a config section in the docs. 

So I'm not sure what your question of API compatibility means.

docs/source/extending/keymaps.rst
```diff
@@ -0,0 +1,88 @@
+Customize keymaps
+=================
+
+.. note::
+
+    DEclarative Custom Keymaps is a provisional feature with unstable API which is not
+    guarantied to be keep in future versions odf the notebook, and can be
+    removed or changed without warnings.
```
SamLau95: spelling nits:

```
DEclarative -> Declarative
guarantied -> guaranteed
be keep -> be kept
odf -> of
```

notebook/static/base/js/keyboard.js
```diff
@@ -404,8 +408,8 @@ define([
          *  The shortcut error should be explicit here, because it will be
          *  seen by users.
          */
-        try
-        {
+        var that = this;
```
SamLau95: I might be missing something but it doesn't look like `that` is being used anywhere in this function?

notebook/static/notebook/js/keyboardmanager.js
```diff
@@ -42,6 +42,46 @@ define([
...
+                edit_unbind = that.config.data.keys.edit.unbind;
+            } catch (e) {
+                if (e instanceof TypeError) {
+                    edit_unbind = [];
+                } else {
+                    throw e;
+                }
+            }
+
+            edit_unbind.forEach(function(u){that.edit_shortcuts.remove_shortcut(u);});
```
SamLau95: Can this be shorted to:

```
edit_unbind.forEach(that.edit_shortcuts.remove_shortcut);
```
?

notebook/static/base/js/keyboard.js
Carreau: Oh, yeah, just had an experiment here that I removed.

notebook/static/notebook/js/keyboardmanager.js
Carreau: No, I actually tried, and it was calling the prototype in the wrong context, telling me that `undefined has no attribute i_dont-remember_what` my guess is that forEach will actually `.bind(..)` the callback to something else.

SamLau95: Ah, you're right. I think it's because a function has the scope of the object it was defined on, so by default `remove_shortcut` has the scope of the `edit_shortcuts` object.

ES6 gives a very concise way to get around most `this` problems with arrow functions, but alas, it's [not supported by most browsers yet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility).

Carreau: Yeah, I know. We could think of using es6 transpiler as now we use webpack.... we might be able to use:

`edit_unbind.forEach(that.edit_shortcuts.remove_shortcut, that.edit_shortcuts);` but I'm not sure if:
1) it works, 
2) it's understandable by everyone. 

So I'll be in favor of something obvious.

Carreau:
All updated.

minrk:
Thanks!

willingc:
@carreau Thanks! I'll give a look at the content now that it's merged. Spent yesterday cleaning up the other notebook stuff.

PR #1235, opened 2016-03-21
Allow kernel id to take precedence over name

blink1073:
This allows both kernel name and kernel id to be provided in the request, making the request implementation simpler.

cc @minrk


blink1073:
Motivation: `NotebookSession.changeKernel(name: string, id?: string): Promise<IKernel>;`.  If the id is provided, that should take precedence over name.

minrk:
:+1:

PR #1237, opened 2016-03-21
Allow a session to connect to an existing kernel

blink1073:


blink1073:
This is ready for review.  I allowed it to use the kernel name if the kernel id is not active, for compatibility with the existing notebook.

minrk:
Nice! And I think we do want to keep the new-kernel if one doesn't exist, since that should still be the majority use case.

PR #1249, opened 2016-03-22
JupyterLab Updates

### LINKED ISSUE #1248: Impossible to close/move `notebook tutorial` in lab
Observed on 7174d9d

To reproduce:
1. Select "notebook tutorial" from the command palette.
2. Verify that the tab with the tutorial can only be resized but not closed or moved around.

Desired outcome:
It should be possible to close the notebook tutorial.

blink1073:
- Fixes usability issues addressed in other jupyter-js-* repos, and fixes #1248.

- Allows JupyterLab to update to the latest patch version of `jupyter-js-plugins` when building.

- Needed to move the static `build` files to the `notebook/lab` directory so they could be served.


blink1073:
@minrk, this is good to go and has a lot of utility.

notebook/lab/handlers.py
```diff
@@ -18,12 +19,12 @@ def get(self):
...
 
-
 #-----------------------------------------------------------------------------
 # URL to handler mappings
 #-----------------------------------------------------------------------------
 
-
 default_handlers = [
     (r"/lab", LabHandler),
+    (r"/lab/(.*)", AuthenticatedFileHandler,
```
minrk: I suspect this wants to be a regular StaticFileHandler (or more precisely FileFindHandler, to get identical caching behavior to the existing static files). We don't normally authenticate access to the basic js/css.

blink1073: Thanks, updated.

blink1073:
:party:

PR #1251, opened 2016-03-22
Remove index.rst.

Carreau:
The actually used file is `Examples and Tutorials Index` in teh same
folder.


willingc:
Double check that this works without this file. It didn't a few months ago.

Carreau:
Actually it works if you remove all files from `rstversions/` why are here anyway ? Are they supposed to be in the repository ? I thought we decided **not** to keep the rst versions of notebooks.

minrk:
We decided not to commit them, but then building them on RTD stopped working, so they were restored while we figured out a more reliable way to build notebooks into sphinx. I think if we adopt nbsphinx here like we did in nbconvert, we can safely drop the rstversions.

willingc:
@minrk Yes, that's the approach that I'm working on. There are a few build errors within the notebooks (nbconvert errors) that need to be resolved even if using nbsphinx. Working on fixing those today.

Issue #1198, opened 2016-03-11
Subatomic writes with docker spawned notebooks


ryanlovett:
We're using JupyterHub with SwarmSpawner and have assigned users a memory limit of 2GB. When user kernels get killed off, they are sometimes left with zero-length .ipynb files and a .~filename.ipynb backup of it.

I don't know how far the the notebook should go to ensure atomic writes, i.e. if kernel killing is pathological. If so can/should dockerspawner cope with it?

takluyver:
Are you killing the kernel or the notebook server and all kernels? The
kernels don't write the notebook files, so killing them should not cause
that. Is it possible to do a softer shutdown, I.e. sigterm, instead of hard
killing it?

ryanlovett:
Sorry, it is the notebook server that is getting killed, so s/kernel/notebook server/g. Docker is [OOM killing](https://docs.docker.com/engine/reference/run/#user-memory-constraints) our [swarmspawner](https://github.com/data-8/jupyterhub-deploy/blob/master/roles/jupyterhub/files/swarmspawner.py) spawned containers when they want to use more than `extra_host_config['mem_limit'] = '2g'`.

I think we can try to reduce the frequency of these events by setting a 2GB memory reservation and increasing the limit. I don't know how far we can take that given hardware constraints. If we disabled OOM for the container, it would just stall since the notebook server is the only thing running in the container and there's nothing else to free up.

But all that being said, can the notebook prevent zeroing out the .ipynb file under such conditions?

takluyver:
We had more atomic writing before, but it caused problems with things like network filesystems and folders synced by services like Dropbox, so we got to what we have now.

Does the `.~filename.ipynb` backup contain a valid notebook with the contents you want? Our atomic-ish compromise should ensure there's always a usable copy of the notebook, even if it's in the backup file. However, trying to do anything reliably with filesystems is a big pain, so it might not be working as intended.

ryanlovett:
Yes, the backup does contain a valid notebook. There are also older versions available as checkpoints.

I'm sure it can be very difficult in trying to cope with various filesystems and filesystem like things. :)

Does it make sense to checkpoint during this write rather than use the .~filename.ipynb mechanism? That way the user has a way to easily restore it if necessary.

takluyver:
Checkpoints are kind of different - we use that so that if a user deletes some cells just before an autosave, they still have their manual save as the checkpoint to go back to.

[...]

PR #1253, opened 2016-03-22
Restore atomic intermediate if notebook is invalid.

ryanlovett:
See previous discussion at https://github.com/jupyter/notebook/issues/1198.

### DIFF
--- a/notebook/services/contents/fileio.py
+++ b/notebook/services/contents/fileio.py
@@ -42,6 +42,17 @@ def copy2_safe(src, dst, log=None):
         if log:
             log.debug("copystat on %s failed", dst, exc_info=True)
 
+def path_to_intermediate(path):
+    '''Name of the intermediate file used in atomic writes.
+
+    The .~ prefix will make Dropbox ignore the temporary file.'''
+    dirname, basename = os.path.split(path)
+    return os.path.join(dirname, '.~'+basename)
+
+def path_to_invalid(path):
+	'''Name of invalid file after a failed atomic write and subsequent read.'''
+    dirname, basename = os.path.split(path)
+    return os.path.join(dirname, '.invalid-'+basename)
 
 @contextmanager
 def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
@@ -73,9 +84,8 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
     if os.path.islink(path):
         path = os.path.join(os.path.dirname(path), os.readlink(path))
 
-    dirname, basename = os.path.split(path)
-    # The .~ prefix will make Dropbox ignore the temporary file.
-    tmp_path = os.path.join(dirname, '.~'+basename)
+    tmp_path = path_to_intermediate(path)
+
     if os.path.isfile(path):
         copy2_safe(path, tmp_path, log=log)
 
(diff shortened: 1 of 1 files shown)


notebook/services/contents/fileio.py
```diff
@@ -42,6 +42,17 @@ def copy2_safe(src, dst, log=None):
...
+    '''Name of the intermediate file used in atomic writes.
+
+    The .~ prefix will make Dropbox ignore the temporary file.'''
+    dirname, basename = os.path.split(path)
+    return os.path.join(dirname, '.~'+basename)
+
+def path_to_invalid(path):
+	'''Name of invalid file after a failed atomic write and subsequent read.'''
+    dirname, basename = os.path.split(path)
+    return os.path.join(dirname, '.invalid-'+basename)
```
takluyver: Personally I'd use a non-hidden filename for this.

notebook/services/contents/fileio.py
```diff
@@ -42,6 +42,17 @@ def copy2_safe(src, dst, log=None):
...
+def path_to_intermediate(path):
+    '''Name of the intermediate file used in atomic writes.
+
+    The .~ prefix will make Dropbox ignore the temporary file.'''
+    dirname, basename = os.path.split(path)
+    return os.path.join(dirname, '.~'+basename)
+
+def path_to_invalid(path):
+	'''Name of invalid file after a failed atomic write and subsequent read.'''
+    dirname, basename = os.path.split(path)
```
takluyver: IndentationError

notebook/services/contents/fileio.py
```diff
@@ -249,11 +259,24 @@ def _read_notebook(self, os_path, as_version=4):
...
+
+            if not self.use_atomic_writing or not os.path.exists(tmp_path):
                 raise HTTPError(
                     400,
-                    u"Unreadable Notebook: %s %r" % (os_path, e),
+                    u"Unreadable Notebook: %s %r" % (os_path, e_orig),
                 )
 
+            # Move the bad file aside, restore the intermediate, and try again.
+            copy2_safe(os_path, path_to_invalid(os_path))
```
takluyver: I think this should be a rename rather than a copy. Copy has to open both files, read the data and write it back again, and it can fail if the disk is full. Rename changes the name of the file in the directory without reading it, and I don't think it can fail because of a full disk.

However, `os.rename()` fails on Windows if the destination file already exists. On Python 3, there's [os.replace](https://docs.python.org/3/library/os.html#os.replace) which does what we need, but we'll need to remove the destination file ourselves on Windows + Python 2.

notebook/services/contents/fileio.py
ryanlovett: Will fix the comment on line 53, thanks!

notebook/services/contents/fileio.py
ryanlovett: How about os.path.join(dirname, basename+'.invalid') ?

notebook/services/contents/fileio.py
ryanlovett: Would the strategy employed by atomic_writing() be sufficient? It removes if 'nt' and the destination exists, then does the rename.

takluyver: Yeah, we decided that's 'atomic enough'.

Our `atomic_writing` context manager is a bit of a misnomer now. We initially had a more truly atomic writing mechanism, but it caused problems with things like Dropbox and network folders, so we switched to something simpler but not as robust.

Carreau:
Ok, let's try that !

Thanks !

ryanlovett:
Thanks!

PR #1259, opened 2016-03-24
Small fixes to notebook docs

willingc:
- Add links in README
- Update file name for environment.yml
- Clarify dropping --user in a virtualenv

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -61,8 +61,8 @@ steps::
     pip install -e . --user
 
 If you want the development environment to be available for all users of your
-system (assuming you have the necessary rights), just drop the ``--user``
-option.
+system (assuming you have the necessary rights) or if you are installing in a
+virtual environment, just drop the ``--user`` option.
 
 
 Rebuilding JavaScript and CSS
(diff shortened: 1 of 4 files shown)


captainsafia:
:+1: :tada:

willingc:
@captainsafia thanks!

@mgeier Thanks for the review. I'm not sure why ``pip install --user`` worked on your system in a virtual environment. See [pip docs](https://pip.pypa.io/en/stable/user_guide/#user-installs) and screenshot of a virtual environment and Python 3.5.1 reflect desired usage:
screen shot 2016-03-25 at 12 17 15 am

mgeier:
@willingc Sorry, I should have been clearer: I was talking about `venv` which is part of Python since version 3.3. It is typically also available from the commandline as `pyvenv`.
The error you are showing happens with an old-style `virtualenv` and apparently also with `virtualenvwrapper`.

I think this should be the preferred way to create a virtual environment:

    python3 -m venv myenvname

This fits nicely with `python3 -m pip install ...` and `python3 setup.py ...`.

I think its not worth cluttering the documentation with legacy solutions like `virtualenv`.
Especially if this misleads new users into using obsolete tools.

willingc:
>I think its not worth cluttering the documentation with legacy solutions like virtualenv.
Especially if this misleads new users into using obsolete tools.

``virtualenv`` is an actively maintained project of pypa and one of the two recommended ways of creating a virtual environment https://packaging.python.org/en/latest/installing/#creating-virtual-environments 

Let's try to keep this productive and focused on the contributors (some of which are going to use conda, some venv, and some virtualenv).

mgeier:
OK, fair enough.
I'm just trying to simplify things ...

By not adding this comment now, this paragraph is already future-proof. By adding it now, somebody will have to remove it in the future or it will just stay there, distracting readers.

willingc:
:+1:

PR #1266, opened 2016-03-26
Minor nbsphinx-related tuning

mgeier:
I wanted to introduce `nbsphinx` to the Notebook docs but then I saw that @willingc beat me to it: #1255

That's great, thanks!

Here I have just a few minor adjustments (hopefully "improvements") ...

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -1,0 +1,9 @@
+/* 24px margin from readthedocs theme */
+div.nblast {
+    margin-bottom: 19px !important;  /* padding has already 5px */
+}
+
+/* ... except between notebook cells! */
+div.nblast + div.nbinput {
+    margin-top: -19px !important;
+}
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -1,9 +1,10 @@
 /* 24px margin from readthedocs theme */
-div.nblast {
-    margin-bottom: 19px !important;  /* padding has already 5px */
+div.nbinput.nblast,
+div.nboutput.nblast {
+    margin-bottom: 19px;  /* padding has already 5px */
 }
 
-/* ... except between notebook cells! */
+/* ... except between code cells! */
 div.nblast + div.nbinput {
-    margin-top: -19px !important;
+    margin-top: -19px;
 }
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -118,7 +118,7 @@ language = None
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['_build']
+exclude_patterns = ['examples/Notebook/.ipynb_checkpoints']
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -186,6 +186,10 @@ html_theme = 'sphinx_rtd_theme'
 #       since it is needed to properly generate _static in the build directory
 html_static_path = ['_static']
 
+html_context = {
+    'css_files': ['_static/custom.css'],
+}
+
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
 # directly to the root of the documentation.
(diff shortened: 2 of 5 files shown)


docs/source/_static/custom.css
```diff
@@ -0,0 +1,9 @@
+/* 24px margin from readthedocs theme */
+div.nblast {
+    margin-bottom: 19px !important;  /* padding has already 5px */
```
minrk: Is there a way to do this without `!important`? that's always something we should avoid if we can.

mgeier: I don't know, I couldn't find one.

But I think in this case it doesn't matter, because there will be no further CSS "derived" from this.

Anyway, if somebody knows a way, I'm happy to change this!

willingc: @jdfreder Thoughts?

Carreau: where is `.nblast` created ? Can't you use somehting like `:last-child` ? Or an example page that use it ?

jdfreder: > where is .nblast created ?

Ditto, I just checked out the PR, cleaned and built the docs, and opened one of the nbsphinx (example notebook) docs and I didn't see it on the page. `document.querySelector('.nblast')`

mgeier: `.nblast` is used wherever there are code cells, e.g. in https://jupyter-notebook.readthedocs.org/en/latest/examples/Notebook/Running%20Code.html

I didn't want to create a separate reST directive as container for a "code cell" (i.e. the combination of an `nbinput` directive and 0..N `nboutput` directives). Currently, the `nbsphinx` directives for input/output cells can be used on the top level, see http://nbsphinx.readthedocs.org/en/latest/rst.html#sphinx-directives-for-jupyter-notebook-cells.
Since I don't have a reST container, I also don't know how to get the parts of a "code cell" into a CSS container like the `code_cell` class used by the Notebook/`nbviewer`. If you know a (reasonably simple) way to do that, please tell me!
As a work-around, I introduced the [:no-output:](https://github.com/spatialaudio/nbsphinx/blob/2181601ed469f309d649732719adcb354f3240ad/nbsphinx.py#L74) and [:more-to-come:](https://github.com/spatialaudio/nbsphinx/blob/2181601ed469f309d649732719adcb354f3240ad/nbsphinx.py#L102) flags for the `nbinput`/`nboutput` directives, which translate to the `nblast` CSS class. This class is used in the last group of a "code cell", i.e. in an input cell if no output is available or in an output cell if it's the last of a group of outputs.

Since (currently) there is no CSS container class for a "code cell" in `nbsphinx`, I cannot use `:last-child`. I also wouldn't have to if there were, because then I could simply use the `margin` settings of the container.

But probably I can do this with some other CSS magic?

mgeier: Another option would be to fix the RTD theme itself, but I can't do that since I have no clue about SASS.

Carreau: so 
```
div.nbinput.nblast {
    margin-bottom: 19px;  /* padding has already 5px */
}
```

See to work for me, and the second `!important` seem to have no effect to me.

mgeier: Thanks @Carreau! I added your suggestion in 9c56f4577db0b9c823d4f34c3350f982ef835368.

docs/source/conf.py
```diff
@@ -118,7 +118,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['_build']
```
Carreau: do we want to keep `_build` in the list of ignored patterns ?

Carreau:
1 comment, +1 otherwise

docs/source/conf.py
mgeier: When building the docs with `make html` (as described in [the docs](https://jupyter-notebook.readthedocs.org/en/latest/contributing.html#building-the-documentation)), the built files end up in `docs/build/` and not in `docs/source/_build`, so I don't think it's necessary to have latter in the `exclude_patterns`.

But if you want it there anyway, just say the word, and I'll add it back!

Carreau: No problem, thanks for the precision.

Carreau:
Ok, merged, let's see if people complain.

PR #1275, opened 2016-03-29
fix validation warnings when validating

minrk:
map is lazily evaluated, so doesn't show messages

### DIFF
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -511,9 +511,11 @@ def validate_nbextension(require, logger=None):
     
     if logger:
         if warnings:
-            logger.warn(u"      - Validating: problems found:")
-            map(logger.warn, warnings)
-            map(logger.info, infos)
+            logger.warning(u"      - Validating: problems found:")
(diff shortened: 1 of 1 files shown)


notebook/nbextensions.py
```diff
@@ -511,9 +511,9 @@ def validate_nbextension(require, logger=None):
     
     if logger:
         if warnings:
-            logger.warn(u"      - Validating: problems found:")
-            map(logger.warn, warnings)
-            map(logger.info, infos)
+            logger.warning(u"      - Validating: problems found:")
+            [logger.warning(warning) for warning in warnings]
+            [logger.info(info) for info in infos]
```
takluyver: I don't like using a listcomp when we don't care about the result. It's not actually any saving over doing:

```python
for warning in warnings: logger.warning(warning)
```

minrk: I agree, I just what's used below. Will fix.

jdfreder:
Thanks! Also, good to know.  :snake:

PR #1277, opened 2016-03-29
Update link to swagger docs

blink1073:
Note that the swagger docs still do not render due to a bug in swagger.


captainsafia:
:tada: :dancer:

PR #1278, opened 2016-03-29
Downgrade hardcoded ipywidgets hack

jdfreder:


jdfreder:
Whitespace problem, fixing now...

PR #1288, opened 2016-03-31
Correctly render markdown code blocks with unknown language

### LINKED ISSUE #1209: Ungraceful failure for code blocks from unknown languages
In a markdown cell, a code block with an unknown language renders as "Type Markdown and LaTeX: α^2" instead of (as one would expect) just printing the code with no highlighting.

BerndSchwarzenbacher:
Closes #1209 

Are the tests in the correct file?

### DIFF
--- a/notebook/jstest.py
+++ b/notebook/jstest.py
@@ -230,8 +230,6 @@ class JSController(TestController):
         includes = '--includes=' + os.path.join(js_test_dir,'util.js')
         test_cases = os.path.join(js_test_dir, self.section)
         self.cmd = ['casperjs', 'test', includes, test_cases, '--engine=%s' % self.engine]
-        # self.cmd = ['casperjs', 'test', includes, test_cases, '--engine=%s' % self.engine,
-        #             '--remote-debugger-port=9000']
 
     def setup(self):
         self.ipydir = TemporaryDirectory()
--- a/notebook/jstest.py
+++ b/notebook/jstest.py
@@ -230,6 +230,8 @@ class JSController(TestController):
         includes = '--includes=' + os.path.join(js_test_dir,'util.js')
         test_cases = os.path.join(js_test_dir, self.section)
         self.cmd = ['casperjs', 'test', includes, test_cases, '--engine=%s' % self.engine]
+        # self.cmd = ['casperjs', 'test', includes, test_cases, '--engine=%s' % self.engine,
+        #             '--remote-debugger-port=9000']
 
     def setup(self):
         self.ipydir = TemporaryDirectory()
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -113,7 +113,7 @@ define(function (require) {
                         }
                     }, function (err) {
                         console.log("No CodeMirror mode: " + lang);
-                        callback(err, code);
+                        callback(null, code);
                     });
                 }
             });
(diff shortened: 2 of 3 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -113,7 +113,7 @@ define(function (require) {
                         }
                     }, function (err) {
                         console.log("No CodeMirror mode: " + lang);
-                        callback(err, code);
+                        callback(null, code);
```
BerndSchwarzenbacher: I'm unsure if it's the desired behavior to suppress the error here.
I think [marked](https://github.com/chjj/marked) doesn't produce a html output when `err != null` in this callback.

BerndSchwarzenbacher:
With this fix the notebook is rendered correctly.
My test doesn't pass though.

I'm unsure if the test is coded correctly.
After the kernel is killed, the correct html would be rendered.
So maybe I misunderstood something with asynchrony when calling `this.evaluate()`.

I also tried to debug like it's described here:
https://drupalize.me/blog/201410/using-remote-debugger-casperjs-and-phantomjs
But I can't set breakpoints in functions which are passed to `this.evaluate()`.

Any help is appreciated.

willingc:
@BerndSchwarzenbacher I've restarted the Travis tests since the Travis may have timed out on the tests and reported a false failure. I will leave it to @Carreau to review the PR contents.

notebook/static/notebook/js/notebook.js
minrk: It might be worth logging the error, if not raising it, just in case the error is caused by something else. From the looks of the code, it seems like the first condition above (`if (!mode)`) is meant to be the one handling a missing mode, but it is now triggering the require error path. I'm not quite sure what changed, but I'm guessing one or the other is never being called.

minrk:
@BerndSchwarzenbacher I think your code all right, but your test is running afoul of async rendering of markdown. I'll write up how to reliably wait for the rendered result.

minrk:
@BerndSchwarzenbacher I made https://github.com/BerndSchwarzenbacher/notebook/pull/1 which adds the necessary async waits to your tests. The casper async testing stuff is far from obvious.

BerndSchwarzenbacher:
@minrk Thanks a lot for your help!
I added logging of the error and I had a look at the source of `CodeMirror.getMode`. It seems to me, that it actually never returns an empty object. So we might delete the extra `if (!mode)` case.
What do you think?

notebook/jstest.py
```diff
@@ -230,6 +230,8 @@ def __init__(self, section, xunit=True, engine='phantomjs', url=None):
         includes = '--includes=' + os.path.join(js_test_dir,'util.js')
         test_cases = os.path.join(js_test_dir, self.section)
         self.cmd = ['casperjs', 'test', includes, test_cases, '--engine=%s' % self.engine]
+        # self.cmd = ['casperjs', 'test', includes, test_cases, '--engine=%s' % self.engine,
+        #             '--remote-debugger-port=9000']
```
minrk: I think we're good to go if we drop out this debug comment.

BerndSchwarzenbacher: Ok, I will get rid of that. Just to check if I understood correctly: we keep the `if (!mode)` case?

minrk: Yes, for now. We can investigate later if it's unreachable and should be removed, but we don't need to do that right now.

Issue #1146, opened 2016-02-25
Untrusted notebook user experience


Phyks:
Hi,

I got some issues with `%matplotlib notebook` which was discarding my figures when reloading the notebook, and replacing them with `<IPython.core.display.Javascript object>`. I could not figure out why, and I discovered it was due to the fact that my notebook was not trusted.

Maybe it could be a good idea to add a big warning bar on top of the notebook if running in untrusted mode, stating something like "this notebook runs in untrusted mode, some features will be disabled.", and it could improve user experience a lot.

Thanks

Phyks:
I am having a quick look at it and think I can make a PR to handle it, if it can be useful.

I am wondering what is the best implementation:
* Either to check for the `trusted` status of the notebook, and display a banner on top of the page, as described initially in this issue.
* Or handle it at the `OutputArea` level (https://github.com/jupyter/notebook/blob/932f85c7593b2eada7a2095bd1f30d3f6ea48115/notebook/static/notebook/js/outputarea.js#L558) and append a banner in sanitized output areas.

What do you think?

minrk:
If you would like to mock up an example of a better UI for untrusted output, that would be great! My first inclination would be the single global indicator, not per-output.

cc @ellisonbg @cameronoelsen

juhasch:
I added a trusted/not trusted indicator (checkmark or cross) for notebooks in the [python-markdown](https://github.com/ipython-contrib/IPython-notebook-extensions/tree/master/nbextensions/usability/python-markdown) extension:
![](https://raw.githubusercontent.com/ipython-contrib/IPython-notebook-extensions/master/nbextensions/usability/python-markdown/trusted.png)

Not saying it should be done like this, but it might give an idea how it could looke like.

Phyks:
:+1: This looks like the perfect thing to do for me. @juhasch Maybe you want to handle a PR yourself? or want me to extract it from your extension?

juhasch:
You are welcome to strip out the indicator an make a PR for the notebook.

Phyks:
UI is far from perfect, but I started a PR at https://github.com/jupyter/notebook/pull/1293.

I think we can discuss there of UI improvements. Thanks!

gnestor:
Merged https://github.com/jupyter/notebook/pull/1658#issuecomment-237954619

Thanks @Phyks for your contribution.

PR #1293, opened 2016-04-01
[WIP] Add a trusted notebook indicator

Phyks:
This pull request adds a "trusted notebook" indicator, as discussed in https://github.com/jupyter/notebook/issues/1146.

Currently, the implementation is to put a "Notebook is trusted" / "Notebook is untrusted" sentence in the menubar on top of the screen. 

Still, UI is not really good at the moment and can be discussed further (hence the "WIP" tag). For instance, displaying a notification explaining what is an untrusted notebook upon opening an untrusted notebook could be a good idea IMO.

Trusted notebook UI:

Untrusted notebook UI:

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -287,6 +287,9 @@ define([
             that.update_nbconvert_script(langinfo);
             that.add_kernel_help_links(data.kernel.info_reply.help_links || []);
         });
+        this.events.on('trust_changed.Notebook', function (event, trusted) {
+            that.set_trusted_indicator(trusted);
+        });
     };
     
     MenuBar.prototype._add_celltoolbar_list = function () {
@@ -415,6 +418,18 @@ define([
         });
         
     };
+
+    MenuBar.prototype.set_trusted_indicator = function (trusted) {
+        // TODO
+        var el = this.element.find('span.trusted-indicator');
+        if (trusted) {
+            el.text('Notebook is trusted.').removeClass("untrusted").addClass("trusted");
+        }
+        else {
+            el.text('Notebook is untrusted.').removeClass("trusted").addClass("untrusted");
+        }
+    };
+
 
     return {'MenuBar': MenuBar};
 });
--- a/notebook/static/notebook/less/menubar.less
+++ b/notebook/static/notebook/less/menubar.less
@@ -17,6 +17,20 @@
   .navbar-collapse {
     clear: left;
   }
+  .trusted-indicator {
+      position: relative;
+      display: block;
+      padding: 10px 15px;
+      padding-top: 6px;
+      padding-bottom: 6px;
+      font-weight: bold;
+  }
+  .trusted-indicator.trusted {
+      color: green;
+  }
+  .trusted-indicator.untrusted {
+      color: red;
+  }
   
 }
 
--- a/notebook/templates/notebook.html
+++ b/notebook/templates/notebook.html
@@ -40,6 +40,7 @@ data-notebook-path="{{notebook_path | urlencode}}"
     <span id="notebook_name" class="filename"></span>
     <span class="checkpoint_status"></span>
     <span class="autosave_status"></span>
+    <span class="trusted_indicator"></span>
 </span>
 
 <span id="kernel_logo_widget">
@@ -301,6 +302,9 @@ data-notebook-path="{{notebook_path | urlencode}}"
                         <li title="About Jupyter Notebook"><a id="notebook_about" href="#">About</a></li>
                         {% endblock %}
                     </ul>
+                </li>
+                <li>
+                    <span class="trusted-indicator"></span>
                 </li>
               </ul>
             </div>


Carreau:
Thanks for doing that !

I think we want something a more subtle, likely  an icon (from https://fortawesome.github.io/Font-Awesome/icons/ ?) that change depending on the state, and likely with less strength in color.

My guess is that @ellisonbg and @cameronoelsen will want to make some design mocks !

Thanks !

An tip: you might be able to set the text with css using `content:` that would allow to change that with themes !

Phyks:
Thanks for the `content:` type in the CSS!

If anyone has some design mocks, I'll gladly update the PR accordingly. Else, I can try to come up with a better UI, but am not very good at it :)

ellisonbg:
I will work with @cameronoelsen on the design side

On Fri, Apr 1, 2016 at 4:37 PM, Lucas Verney <notifications@github.com>
wrote:

> Thanks for the content: type in the CSS!
>
> If anyone has some design mocks, I'll gladly update the PR accordingly.
> Else, I can try to come up with a better UI, but am not very good at it :)
>
> —
> <https://github.com/jupyter/notebook/pull/1293#issuecomment-204608241>
>



-- 
Brian E. Granger

Phyks:
Ok. Thanks!

Carreau:
@ellisonbg @cameronoelsen News on design ?

cameronoelsen:
We were thinking that 1.) the design should be something that will work in both the current notebook as well as JupyterLab and 2.) it should be subtle and somehow connect to the actual document being run (hence why the alert is at the very top of the notebook document before the user starts running the cells). It seemed that when we put the untrusted symbol and text in the toolbar, it gets confused with either the kernel that is running or whether the document was autosaved or not.

Seeing that "trusted" means that the user has run all of the cells in the document, we could either have this "untrusted notification" be only dismissable (no buttons, only clickable to dismiss) or let the user decide if they either want to dismiss it or trust it on the spot (action buttons that are similar to the dialogs that are being built in JupyterLab).

## Untrust notification with action buttons (dismiss and trust)
buttons

## Untrust notification with only dismiss on click
nobuttons

cameronoelsen:
I personally am leaning towards the notification only being dismissable without the buttons (since it is smaller and doesn't require user input)

minrk:
all-gray might be a little too subtle for untrusted

takluyver:
I definitely think it shouldn't have the trust button - it makes it too easy to trust a notebook without really understanding what that means. It looks like a dialog, and we know that users seeing a dialog typically just click the first thing they see that looks like it will make it go away.

Carreau:
> I definitely think it shouldn't have the trust button - it makes it too easy to trust a notebook without really understanding what that means. It looks like a dialog, and we know that users seeing a dialog typically just click the first thing they see that looks like it will make it go away.

Agreed.

cameronoelsen:
@Carreau @takluyver Do you guys like the placement of the dialog box in the top left of the document instead of the toolbar?

Carreau:
> all-gray might be a little too subtle for untrusted

Not sure about that. All grey seem fine to me. 

Though:
I would likely put notification on right, as many user will have a title or text in first cell, and thus the notification will be hiding content.

cameronoelsen:
@Carreau Here is the notification on the righthand side, good point!
nobuttonsright

cameronoelsen:
Or if we have it on the lefthand side, is there a way to have the content of the notebook fall under the dialog? I like it on the left only because it aligning with most of the content in the toolbar

takluyver:
> Do you guys like the placement of the dialog box in the top left of the document instead of the toolbar?

I'm not sure. I'm a bit concerned that it will feel like unnecessary clutter. When you scroll the notebook, does it scroll too, or float wherever you are?

Also, I think it should have an `(i)` for more information about what that means.

cameronoelsen:
@takluyver It will definitely stay in place at the top and not float as you scroll, I think when a user loads up the document and sees it, it doesn't have to get more in the way while scrolling. Trying out adding an information icon to see how that looks, should have it posted soon

cameronoelsen:
## Dialog with information icon

nobuttonsi

minrk:
Last one looks nice to me

ellisonbg:
I think this is converging. I like the last one, but my only comment is
that the "!" and "i" look similar visually but are not aligned vertically.
I agree we need someway for a user to get more information. Maybe some
things to try:

* Just an "i" in a circle?
* Vertically align the "!" and "i"?
* Use text "This notebook is untrusted (info)"?



On Fri, Apr 8, 2016 at 4:07 PM, Min RK <notifications@github.com> wrote:

> Last one looks nice to me
>
> —
> You are receiving this because you were assigned.
> <https://github.com/jupyter/notebook/pull/1293#issuecomment-207642422>
>



-- 
Brian E. Granger

Phyks:
I like the last one. As @minrk I think light grey might be too subtle?

Concerning @cameronoelsen remark that on the left it would be aligned, is there a reason for the toolbar to be slightly shifted to the right with respect to the notebook content? (or is it an illusion)

cameronoelsen:
@Phyks I took a screenshot of the notebook, I think you are right with it being aligned to the right and not aligning on the left-hand side:

group

I just opened a new issue related to it here @Phyks #1345. Nice find!

cameronoelsen:
@ellisonbg I think this one looks more natural. Adding the information icon inside the circle and the spacing and alignment between everything helped a lot with putting pertinent information closer to each other. What do you think?

newest

Phyks:
@cameronoelsen Thanks for the double check!

Phyks:
@cameronoelsen Concerning your latest screenshot, I think it is a bit hard to distinguish between the exclamation point (not clickable) and the information button (which is clickable). Isn't it a bit misleading?

And, if the alignment issue can be easily fixed, what about putting the notification on the left hand side? (not sure it is way better though)

cameronoelsen:
@Phyks Does this change help with that confusion of which one is actionable vs. not?

newer

cameronoelsen:
Also, @Phyks I like the notification at the top left of the document more so too, but @Carreau made a good point that most user notebook titles are aligned left at the top of the document and having the notification on the left hand side will most likely cover a good amount of notebooks with titles there.

leftnew

But, if there is a way to make the content of the document to start underneath of the notification or if this issues doesn't affect as many users as we are thinking, I definitely think the left is where to place it since it aligns with the toolbar content.

cameronoelsen:
Here is the notification with the information symbol enlarged to be the height of the font:

righttool

ellisonbg:
I think this last version works pretty well. I know the grey isn't too
strong, but I think that reflects the relative importance of the
notification well.

On Tue, Apr 12, 2016 at 4:03 PM, Cameron Oelsen <notifications@github.com>
wrote:

> Here is the notification with the information symbol enlarged to be the
> height of the font:
>
> [image: righttool]
> <https://cloud.githubusercontent.com/assets/6437976/14478183/267f91f4-00c8-11e6-9eb3-39da25286047.png>
>
> —
> <https://github.com/jupyter/notebook/pull/1293#issuecomment-209141191>
>



-- 
Brian E. Granger

Carreau:
> Concerning your latest screenshot, I think it is a bit hard to distinguish between the exclamation point (not clickable) and the information button (which is clickable). Isn't it a bit misleading?

We don't need to have only the icon clickable, the all notification can, so I would just remove the `i` in the circle, but we can try that later. 

Also, dumb question, we already have a notification area, which is meant for notification, notification that can be sticky. Why not use it ?

Phyks:
@cameronoelsen :+1: for your last design!

cameronoelsen:
@Carreau What does the notification area / individual notifications look like in the notebook?

Carreau:
Look like that:

screen shot 2016-04-14 at 13 30 21

Carreau:
Trying to revive this one. 

@ellisonbg there is already most of the code ready, could this be iterated on by your students, to get something in ?  it would be nice to have @Phyks code integrated. Thanks.

gnestor:
I'd be willing to implement whatever design spec we've agreed upon. It looks like consensus is currently https://github.com/jupyter/notebook/pull/1293#issuecomment-209141191. 

Requirements: 

* Display the notification if the notebook is untrusted
* Clicking on the notification will dismiss it?

@Carreau Do we currently have a notification/alert component that I could use to implement this? We have Bootstrap's [alert](http://getbootstrap.com/components/#alerts) which offers the same functionality but different presentation.

Carreau:
> @Carreau Do we currently have a notification/alert component that I could use to implement this? We have Bootstrap's alert which offers the same functionality but different presentation.

No we don't have anything specific, just the [notification widgets/notification are](https://github.com/jupyter/notebook/blob/master/notebook/static/base/js/notificationarea.js), but they are in the toolbar.

I don't think we should dismiss the notification on click,we might just want to pop-up a longer dialog that explain what untrusted means.

juhasch:
From the experience with my own little trusted notebook indicator [here](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/python-markdown), I would like to add, that it is useful to have a persistent indication if the notebook is trusted.

gnestor:
@Carreau I just saw [your comment](https://github.com/jupyter/notebook/pull/1293#issuecomment-209889622) about the notifiction area. This seems like a good solution given that's how other notifications (such as kernel error) are presented. I will share a quick implementation of this...

gnestor:
See https://github.com/jupyter/notebook/pull/1658 for an implementation using the notebook's notification widgets.

gnestor:
Merged https://github.com/jupyter/notebook/pull/1658#issuecomment-237954619

Carreau:
Thanks a lot all, to @gnestor to get the ball rolling on that and to @Phyks to starting that originally.

PR #1304, opened 2016-04-05
Missing font-awesome css file.

lduchesne:
I'm not sure if this is the right fix, but the font-awesome*.css files were not copied over to the build/ directory when doing "python setup.py build". Running the notebook from that build/ directory resulted in missing icons everywhere.


willingc:
@lduchesne Thanks for the PR :fish_cake:

Issue #1316, opened 2016-04-07
Copy Past issues


Carreau:
With the recent introduction of copy-past by @takluyver, there are a few issues: 

1) cannot copy past call output. 

(screen cast 1 whwre I press Ctrl-C on keyboard after selecting)

![cv](https://cloud.githubusercontent.com/assets/335567/14360304/90c96296-fcaa-11e5-8125-8ab87352f519.gif)

2) cannot past in debugger prompt field:

![debug](https://cloud.githubusercontent.com/assets/335567/14360344/c948c8c8-fcaa-11e5-9b15-0108d647b7f9.gif)

Carreau:
Just as a side note, I am happy that we merged @takluyver quick as it allowed us to **discover** bugs that we would have missed otherwise.

takluyver:
Fix coming up

Carreau:
Reopening 2, not fixed.

PR #1319, opened 2016-04-07
Don't paste cells in chrome when keyboard manager is disabled

takluyver:
Closes gh-1316 (the second part of it, which I hadn't noticed before ;-)

If a widget, an input box or a dialog has focus, we should allow paste to paste text by default. We already take care to disable the keyboard manager in these cases, so I check this to decide to handle the paste event.

In Firefox this isn't needed, because Ctrl+V is registered as the shortcut to bring up the paste dialog, so it only works when the keyboard manager is enabled.

### DIFF
--- a/notebook/static/notebook/js/clipboard.js
+++ b/notebook/static/notebook/js/clipboard.js
@@ -68,6 +68,19 @@ function paste(event) {
   event.preventDefault();
 }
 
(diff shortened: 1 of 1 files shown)


Carreau:
Works ! Thanks !

PR #1323, opened 2016-04-08
Build source map for lab

blink1073:
cf https://github.com/jupyter/jupyter-js-notebook/issues/160#issuecomment-207215889


Carreau:
Cool ! Thanks !

PR #1328, opened 2016-04-10
The path for widgets has changed

SylvainCorlay:
cc @minrk @jdfreder


jdfreder:
Thanks @SylvainCorlay

PR #1333, opened 2016-04-11
remove what's new from index

minrk:
use |release| variable to ensure correct value.

move 4.1 release announcement link to 4.1 changelog, away from index.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,12 @@ Jupyter notebook changelog
 A summary of changes in the Jupyter notebook.
 For more detailed information, see `GitHub <https://github.com/jupyter/notebook>`__.
 
+.. tip::
+
+     Use ``pip install notebook --upgrade`` or ``conda upgrade notebook`` to
+     upgrade to the latest release.
+
+
 .. _release-4.2.0:
 
 4.2.0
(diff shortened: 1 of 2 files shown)


minrk:
I think we may need to change this more. I'm not sure about having a duplicate of the most recent changelog on the front page, especially since I didn't realize until now (after I've cut the 4.2 beta) that I had two places to add the same info for every release. I also don't really know how to have something like a link to a release announcement in docs when the release announcement doesn't exist yet.

@willingc what do you think we should do about the what's new copy on the index for 4.2?

willingc:
@minrk I think what prompted the "What's new" on the index page was the upgrade questions when 4.1 was released. I think it is probably better going forward to remove the "What's new" from the index page. Instead, we could:
1. add release upgrade commands to the release notes document
2. add a link to the blog post (announcement) in the README. It would then be captured in version control but wouldn't need to happen before the release since it's not part of the formal docs.

@minrk What do you think?

minrk:
That sounds like a good plan. Do you want to make those changes? I can close this one.

willingc:
@minrk I made the changes as a PR against your branch.

blink1073:
@wiilingc, isn't it 4am there??

PR #1338, opened 2016-04-12
avoid double-base-url when launching browser

### LINKED ISSUE #1336: 4.2.0b1 - Overriding base_url duplicates that path on startup
Congrats on the beta! But I found a little bug:

Starting up with a different base url opens a browser with the new path duplicated before appending the tree path. For example, if I use the command line tag `--NotebookApp.base_url=new_path`, it opens the browser pointing at `http://localhost:port/new_path/new_path/tree`. The server's at the right place (pasted a startup console log below), it just sends the browser to the wrong place.

Tested this with both the command-line configuration and a config file.  
(Mac OSX 10.11.3, installed both with pip and from source)

```
(jupyter)wjriehl:[jupyter-4.2.0b1]> jupyter notebook --NotebookApp.base_url=new_path
[I 17:00:05.669 NotebookApp] Serving notebooks from local directory: /Users/wjriehl/Projects/jupyter-4.2.0b1
[I 17:00:05.670 NotebookApp] 0 active kernels 
[I 17:00:05.670 NotebookApp] The Jupyter Notebook is running at: http://localhost:8888/new_path/
[I 17:00:05.670 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[W 17:00:05.982 NotebookApp] 404 GET /new_path/new_path/tree (::1) 18.75ms referer=None
```

minrk:
default_url contains base_url, and so does connection_url

closes #1336

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1211,7 +1211,8 @@ class NotebookApp(JupyterApp):
                 relpath = os.path.relpath(self.file_to_run, self.notebook_dir)
                 uri = url_escape(url_path_join('notebooks', *relpath.split(os.sep)))
             else:
-                uri = self.default_url
+                # default_url contains base_url, but so does connection_url
+                uri = self.default_url[len(self.base_url):]
             if browser:
                 b = lambda : browser.open(url_path_join(self.connection_url, uri),
                                           new=2)


Carreau:
+1

PR #1265, opened 2016-03-25
[POC] Shortcut editor (for command mode)

Carreau:
Proof of concept.

1) Ask Webpack to compile es6, because now that we have webpack, why not ? 
2) install react/react-dom.
3) Learn React, and write a module that allows you to edit keyboard shortcuts. 
4) enable eslint for es6.

Todo:
  - Actually persist the new shortcuts
  - actually persist the shortcut we unbind. 
  - design.
  - figure out how to destroy the react components. 
  - do not fail silently if : 
    - shortcut already attached to another action. 
    - Shortcut invalid.
  - decide on JSX.
  - name things correctly.

PIng @rgbkrk and @jdfreder because I have no clue what I am doing. 

---

Personal note, es6 is much less WTF than es5 , but still WTF. 
JSX: replace `<` by `?>` and `/>` by `<?` and you get PHP.

PR #1347, opened 2016-04-13
Create a shortcut editor for the notebook.

Carreau:
Replace #1265 

Also introduce es6, and write the new keyboard shortcut editor as pure es6.
Does not use react.

--- 

[edit] 
Update description after lots of refactor.

Create shortcut editor for the notebook

1) finish the step allowing the use of es6

  - this include some tweak to web pack configuration to speed up
      recompile in watch mode (in particular cache sourcemaps).
  - enable eslint (error only), on obvious mistakes.
  - setup babel to compile to es5 as a target.

2) Make the test pass under Casper that does not always have
`Function.prototype.bind` defined, which we cannot patch only in the
tests.

3) Write an actual shortcut editor that list and allow to modify most of
the command mode shortcut.

The logic to persist the shortcuts is a bit tricky as there are default
keyboard shortcuts, and so when you "unbind" them you need to re-unbind
them at next startup. This does not work for a few shortcut for
technical reasons: `<Esc>`, `<Shift>`, as well as `<Ctrl-Shift-P>` and `<F>`
which register asynchronously, so are not detected as "default"
shortcuts.


sccolbert:
Should preact be a runtime dependency instead of a dev dependency?

Carreau:
> Should preact be a runtime dependency instead of a dev dependency?

Likely yes, reflect of `--save-dev`.

Carreau:
reflex*,

and it will be bundled in the min.js that will generated by webpack and be shipped, so not actually not sure.

package.json
```diff
@@ -27,5 +30,7 @@
   },
   "dependencies": {
     "moment": "^2.8.4"
```
sccolbert: this is a syntax error, you need a comma

package.json
```diff
@@ -27,5 +30,7 @@
...
+    "preact": "^4.5.1",
+    "preact-compat": "^1.7.0",
```
sccolbert: No trailing comma.

Carreau:
ping @jdfreder , you're the one that know the most things about notebook test. Test seem to not run at all with this PR, but I can't seem to figure out the reason why, as the same step works flawlessly locally.

Any ideas ?

Carreau:
BTW, this now works, and can now use es6.

sccolbert:
What do you mean by "can use es6"?

Carreau:
> What do you mean by "can use es6"?

Sorry, was unclear, you can use es6 feature in all the js files (typically `notebook/static/*/js/*.js`), webpack use the a babel loader when minifying, so will understand `let`, `const`, `import {} from ...` etc. There is no extra steps.

sccolbert:
Ah - I didn't realize we were using babel. Are we still targeting ES5 runtimes?

Carreau:
> Are we still targeting ES5 runtimes?

Yes. We can use something else than babel, I guess we can use a typescript loader as well, I have no issues with that. 

One thing I did not include is the ability to use `.jsx`, but that should be pretty easy to add too if we decide to do so.

sccolbert:
I don't think we need to use the TS loader at this point. We've been doing ago just shipping the built `.js` files for our code which is written in TS.

sccolbert:
Also, I think (but could be wrong) that preact has an `h()` function, which can reduce the boilerplate needed by `createElement`.

Carreau:
> Also, I think (but could be wrong) that preact has an h() function, which can reduce the boilerplate needed by createElement.

Yes it has a `h()` function, though they don't really explain how/why to use it, expect using `/** @jsx h */` in `jsx` files, and it looks mostly as `s/createElement/h/g`, but I might be missing something.

Carreau:
Ok, merging to get people to complain.

PR #1348, opened 2016-04-13
pin casper to 1.1.0-beta5

minrk:
to see if this fixes the shutdown test failure

If this works, we can get tests back to passing on master while we figure out what's up with casper in another PR.


minrk:
It works! Merging now, so Travis can be green again.

PR #1349, opened 2016-04-13
Dockerfile: Added --no-browser argument to CMD

FrederikNS:
The docker images doesn't contain any browsers, therefore there is no need for trying to launch a browser within the docker container.


Carreau:
Thanks !

PR #1351, opened 2016-04-14
ES6 and prerequisite to shortcut editor

Carreau:
For whatever reason casper seem to be unhappy when I actually introduce es6 things in modules, 
so this is a subset of #1347 where tests are (or at least should be) passing on travis.

### DIFF
--- a/notebook/static/notebook/js/keyboardmanager.js
+++ b/notebook/static/notebook/js/keyboardmanager.js
@@ -36,12 +36,12 @@ define([
         this.bind_events();
         this.env = {pager:this.pager};
         this.actions = options.actions;
-        this.command_shortcuts = new keyboard.ShortcutManager(undefined, options.events, this.actions, this.env );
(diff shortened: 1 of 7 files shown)


notebook/static/notebook/js/keyboardmanager.js
```diff
@@ -10,7 +10,7 @@
 
 define([
     'base/js/utils',
-    'base/js/keyboard',
+    'base/js/keyboard'
```
minrk: Stripping trailing commas makes me a little sad. Just curious: what brought this on?

Carreau: Sorry, tried enabling `eslint` and my editor complains. I can deactivate and revert that.

Carreau:
Fixed.

Carreau:
Yeah ! Thanks !

PR #1352, opened 2016-04-14
Make JupyterLab more self contained

blink1073:
Moves the ES6 promise polyfill, and font-awesome from the template to the webpack bundle.

We are still using MathJax and requirejs from bower.  It makes sense to only have one copy of MathJax in the repository since it is so big, and we'd might as well keep the known path to requirejs.

### DIFF
--- a/notebook/lab/package.json
+++ b/notebook/lab/package.json
@@ -6,12 +6,12 @@
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "dependencies": {
+    "es6-promise": "^3.1.2",
     "jupyter-js-plugins": "^0.11.8",
     "phosphide": "^0.9.0"
   },
   "devDependencies": {
     "css-loader": "^0.23.1",
-    "es6-promise": "^3.1.2",
     "file-loader": "^0.8.5",
     "json-loader": "^0.5.4",
     "rimraf": "^2.5.0",
--- a/notebook/lab/package.json
+++ b/notebook/lab/package.json
@@ -7,6 +7,7 @@
   "typings": "lib/index.d.ts",
   "dependencies": {
     "es6-promise": "^3.1.2",
+    "font-awesome": "^4.6.1",
     "jupyter-js-plugins": "^0.11.8",
     "phosphide": "^0.9.0"
   },
--- a/notebook/lab/package.json
+++ b/notebook/lab/package.json
@@ -11,7 +11,7 @@
   },
   "devDependencies": {
     "css-loader": "^0.23.1",
-    "es6-promise": "^3.0.2",
+    "es6-promise": "^3.1.2",
     "file-loader": "^0.8.5",
     "json-loader": "^0.5.4",
     "rimraf": "^2.5.0",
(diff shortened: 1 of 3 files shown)


blink1073:
cc @SylvainCorlay

notebook/lab/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "devDependencies": {
     "css-loader": "^0.23.1",
-    "es6-promise": "^3.0.2",
+    "es6-promise": "^3.1.2",
```
sccolbert: should this be a runtime dep?

SylvainCorlay: So we discussed this on gitter and my reasoning for `jupyter-js-widgets` was that only the file that was intended to be webpacked explicitly required `es6-promise`, and webpacking is a build-time thing.

blink1073: But it doesn't *need* to be Webpacked.

blink1073: In general

SylvainCorlay: Well, that is the thing (in jupyter-js-widgets), we have this `embed-webpack.js` which is not required by `index.js` and only intended to be webpacked.

blink1073:
I'm going to mark this WIP, since currently the `lab.html` template already provides a polyfill from bower among other things.  I'm going to clean this up to be self contained.

sccolbert:
So, I shouldn't hit the green button yet?

blink1073:
Updated, see original comment.

notebook/templates/lab.html
```diff
@@ -6,8 +6,6 @@
 
     <title>{{page_title}}</title>
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-    <link rel="stylesheet" type="text/css" href="{{static_url("components/font-awesome/css/font-awesome.css")}}"></link>
```
SylvainCorlay: FYI, for requirejs, instead of using bower, you can load it with scriptjs.

blink1073: No, because the whole thing needs to be self contained.  (I used to work in a lab with no internet access).

blink1073:
@SylvainCorlay, I finally got requirejs to play nicely with WebPack here:  88f9332.

notebook/templates/lab.html
SylvainCorlay: Yes, you can load it locally with scripjs.

blink1073: Ah, I see.

notebook/templates/lab.html
```diff
@@ -94,7 +92,6 @@
   "wsUrl": "{{ws_url| urlencode}}",
   "notebookPath": "{{notebook_path | urlencode}}"
 }</script>
-<script src="lab/bundle.js" type="text/javascript" charset="utf-8"></script>
```
minrk: Is it okay and/or necessary for lab to load before the require config?

blink1073: Argh!  Good point.

blink1073:
@minrk, updated.

minrk:
Looks okay to me. I'll let @SylvainCorlay merge in case he sees anything else.

SylvainCorlay:
👍

PR #1353, opened 2016-04-15
Escape file names in attachment headers

siyuwj:
When there are non-ascii characters in notebook file name, i.e., "文件名.ipynb", notebooks cannot be properly downloaded via configurable-http-proxy in [JupyterHub](https://github.com/jupyter/jupyterhub) and an HPE_INVALID_HEADER_TOKEN is reported. After I inspect the headers returned by the notebook, I find that the file name in Content-Disposition header is not escaped, which cause the error in [request module](https://github.com/request/request) of node.js. After escaping the file names, the bug is resolved.

### DIFF
--- a/notebook/nbconvert/handlers.py
+++ b/notebook/nbconvert/handlers.py
@@ -7,7 +7,7 @@ import io
 import os
 import zipfile
 
-from tornado import web
+from tornado import web, escape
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #1359, opened 2016-04-15
Update the session api in preparation for file and console sessions

### LINKED ISSUE #1358: Update Session to take an arbitrary name?
We talked today at the JupyterLab meeting about how the Session would map to a source code file or a console.  We could just set the `notebook:path` to be a file path or a console session name, but it would be more generic to have a session id take this interface:

```typescript
interface ISessionId {
  id: string;
  path: string;  
  type: string;   // one of { notebook, file, console }
  kernel: {
    id: string;
    name: string;
  }
}
```

For text files and notebooks, the path is the full path to the file.  For a console, the path is the cwd directory of the console (at startup).

blink1073:
Fixes #1358.

### DIFF
--- a/notebook/services/api/api.yaml
+++ b/notebook/services/api/api.yaml
@@ -639,12 +639,11 @@ definitions:
       id:
         type: string
         format: uuid
-      name:
+      path:
         type: string
-        description: the name of file, notebook, or console
-      directory:
-        type: string
-        description: the directory of the session
+        description: "The path to the session.  For console sessions, the path 
+         contains the cwd of the kernel and the name of the console session, 
+         which may be a uuid"
       type:
         type: string;
         description: session type
--- a/notebook/services/api/api.yaml
+++ b/notebook/services/api/api.yaml
@@ -357,15 +357,7 @@ paths:
           in: body
           required: true
           schema:
-            type: object
-            properties:
-              notebook:
-                type: object
-                properties:
-                  path:
-                    type: string
-                    format: path
-                    description: new path for notebook
+            $ref: '#definitions/Session'
       responses:
         200:
           description: Session
@@ -641,16 +633,13 @@ definitions:
         format: uuid
       path:
         type: string
-        description: "The path to the session.  For console sessions, the path 
-         contains the cwd of the kernel and the name of the console session, 
-         which may be a uuid"
+        description: path to the session
+      name:
+        type: string
+        description: name of the session
       type:
         type: string;
         description: session type
-        enum:
-          - notebook
-          - file
-          - console
       kernel:
         $ref: '#/definitions/Kernel'
   Contents:
--- a/notebook/services/api/api.yaml
+++ b/notebook/services/api/api.yaml
@@ -349,7 +349,7 @@ paths:
           schema:
             $ref: '#/definitions/Session'
     patch:
-      summary: "This can be used to rename the notebook, or move it to a new directory."
+      summary: "This can be used to rename the session."
       tags:
         - sessions
       parameters:
@@ -395,29 +395,14 @@ paths:
             items:
               $ref: '#/definitions/Session'
     post:
-      summary: "Create a new session, or return an existing session if a session for the notebook path already exists"
+      summary: "Create a new session, or return an existing session if a session of the same name already exists"
       tags:
         - sessions
       parameters:
         - name: session
           in: body
           schema:
-            type: object
-            properties:
-              notebook:
-                type: object
-                required:
-                  - path
-                properties:
-                  path:
-                    type: string
-                    description: path to notebook file
-              kernel:
-                type: object
-                properties:
-                  name:
-                    type: string
-                    description: "Kernel spec name, defaults to default kernel spec"
+            $ref: '#definitions/Session'
       responses:
         201:
           description: Session created or returned
@@ -429,7 +414,7 @@ paths:
               type: string
               format: url
         501:
-          description: Kernel not available
+          description: Session not available
           schema:
             type: object
             description: error message
@@ -654,12 +639,19 @@ definitions:
       id:
         type: string
         format: uuid
-      notebook:
-        type: object
-        properties:
-          path:
-            type: string
-            description: path to notebook
+      name:
+        type: string
+        description: the name of file, notebook, or console
+      directory:
+        type: string
+        description: the directory of the session
+      type:
+        type: string;
+        description: session type
+        enum:
+          - notebook
+          - file
+          - console
       kernel:
         $ref: '#/definitions/Kernel'
   Contents:
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -478,6 +478,8 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
 
     def get_kernel_path(self, path, model=None):
         """Return the initial API path of  a kernel associated with a given notebook"""
+        if self.isdir(path):
+            return path
         if '/' in path:
             parent_dir = path.rsplit('/', 1)[0]
         else:
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -478,7 +478,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
 
     def get_kernel_path(self, path, model=None):
         """Return the initial API path of  a kernel associated with a given notebook"""
-        if self.isdir(path):
+        if self.dir_exists(path):
             return path
         if '/' in path:
             parent_dir = path.rsplit('/', 1)[0]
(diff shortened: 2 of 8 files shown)


blink1073:
Ran locally and the current session still works, still need to update tests.

blink1073:
@minrk, this is ready to go.

blink1073:
@parente, this affects kernel gateway and would require a minor change.  TLDR: instead of `{ notebook: path }` we have `{ path, type: 'notebook' }`.

blink1073:
Optionally, we could allow both APIs until 5.0 is released.

blink1073:
Although, master *is* 5.0.

blink1073:
Updated to allow the previous API to still work, since there is no harm in doing so.

notebook/services/api/api.yaml
```diff
@@ -654,12 +639,18 @@ definitions:
...
+        description: "The path to the session.  For console sessions, the path 
+         contains the cwd of the kernel and the name of the console session, 
+         which may be a uuid"
+      type:
+        type: string;
+        description: session type
+        enum:
+          - notebook
+          - file
+          - console
```
parente: A slight tangent:

I think the [EclairJS project](https://github.com/EclairJS/eclairjs-node) is using sessions in the kernel gateway to assign kernels a name and later look them up by that name. They do this to avoid having to track kernel IDs in their client and to rely on the already stateful kernel gateway to do it for them. 

They're currently using the `notebook` field in the current version of the API as the place to put that name. Path + type makes that feel a little less hacky, though path is still a bit of a misnomer (even the description above says it might contain a UUID) and none of the types named above really apply to the EclairJS use case. 

Is there a way to make these new field names more generically applicable and less specific to notebook / jupyter lab features? I realize this is opened against the notebook repo, but I'm thinking aloud about how it might cater to both these applications plus others in the more open ended use cases of the kernel gateway.

blink1073: Okay, so perhaps `path` and `name` as separate fields makes more sense, and allow `type` to take any value?

minrk: alternately, `name` could be an opaque key, and only interpreted as a path when `type == notebook`.

parente: Opaque `name` would work for notebooks as @minrk says, consoles which according to the swagger doc uses a UUID, EclairJS which uses some sort of a UUID/name as well, and any user use case that requires a string identifier.

parente: /cc @bpburns

blink1073: My concern is that we still need a path, to tell the kernel which directory to take as its cwd.

https://github.com/jupyter/notebook/blob/master/notebook/services/sessions/sessionmanager.py#L90

minrk: Good point. In that case, maybe leave path, and add name adjacent to it. Notebook might ignore name and just use path.

blink1073: We could avoid all ambiguity by using `cwd` and `name`.

parente: `cwd` and `name` are clearer to me.

parente: I just noticed the `type` is declared as an enum in the spec. I think it's good that there's a declared list of types known by the notebook server, but I think it has to document that other types are allowed as well. I'm thinking of cases where people write JLab plug-ins that want to manage other types of sessions, and of apps like Eclair talking to KG, declaring their own session types.

blink1073: Yep, I hence my earlier comment that `type` should be allowed to take any value ;)

blink1073: I'm waiting for consensus before updating the PR.

blink1073: @minrk, do you agree with `cwd` and `name`?

minrk: I have mixed feelings about CWD, since it bakes filesystem knowledge into the API. When interpreting cwd, is the CM path used, or the real filesystem? We want a kernel to start in a particular dir when the notebook it starts with is on the same filesystem, but it's unclear what should happen when the Contents paths don't map onto the kernel's filesystem, or how/when that should be communicated.

blink1073: The filesystem knowledge is already in the API, since we are starting a kernel in a specified directory.  The fact that it is relative shouldn't matter.

minrk: There is no filesystem knowledge in the current REST API. The Contents API path is exposed, but that need not map to any filesystem at all. The SessionManager asks the ContentsManager "What should the CWD of the kernel be for this file/notebook". The *client* of the session API does not have any way to influence this, and it shouldn't for notebooks. For notebooks, the information should continue to be just two pieces of information: the notebook path according to the ContentsManager and the kernel name, according to the KernelManager.

blink1073: Right, in this context `cwd` is the path according to the ContentsManager.  We still need the same information to appropriately start any type of session in the correct directory.  Take, for example, a "Start Console Session in this Directory" option in the file browser.

minrk: Then maybe I'd leave it as `path`, and not `cwd`. For instance, with a notebook you ask for a kernel associated with a given notebook. There's no directory that's part of the request, nor should there be. Same goes for a file, I would think (we still don't have anything associating kernels with non-notebook files, but I'm guessing we might want it to be the same). But a console might be associated with a directory, as you point out.

blink1073: So then we have `path` which may be a path to a file or a directory, `name` with the file name or some other id, and `type` which can be anything?  That would mirror the contents API.

minrk: I think so, which means it's really back to just adding `name` as something frontend applications can use to categorize sessions if `path` is not sufficient. Which means that notebooks probably also won't use `name`, and can just rely on type and path.

minrk: Do we require name to be unique?

blink1073: From what I can tell in the current code, the only uniqueness is the tuple of `(session_id, path, kernel_id)` by virtue of the database.  So one could have multiple sessions for the same path, which I think is fine in a collaborative scenario.  The contents manager is expecting a full path to a file [here](https://github.com/jupyter/notebook/blob/master/notebook/services/sessions/sessionmanager.py#L89), so we'd have to make up a file name for a console, for example.
Given that these are *web* paths and not necessarily file paths, and don't have to reflect a file stored on disk, it may still make more sense then to just use `path` and let the "file name" be whatever one wants, such as a uuid.

minrk: I like requiring `path` to be an API-path, rather than opaque, so if there is a need for setting the 'name' of a session to something unrelated to paths, I would let name be a distinct field.

blink1073: Do we then update `ContentsManager.get_kernel_path` to check for a directory vs. a file, or add a fake file before passing it?

blink1073: Actually, there isn't a clean way to tell the difference because they aren't actual file paths, so we have to go with the convention that a directory ends with `/`, which the `ContentsManager` handles just fine.

blink1073:
Updated.

notebook/services/api/api.yaml
minrk: The ContentsManager should be able to know if it's a file or directory, because it contains the mapping of path to actual location on disk (if there's a disk). `ContentsManager.dir_exists(path)` should be sufficient.

blink1073: Thanks, updated.

blink1073:
FWIW, I intend to support both the old and new API in `jupyter-js-services` as a shim.

blink1073:
(And this PR preserves the old API on the server as well).

gnestor:
@blink1073 Is this ready to merge?

blink1073:
I'm not sure if we had consensus on this.  cf https://github.com/jupyter/jupyterlab/issues/260

blink1073:
@minrk, thoughts on moving forward with this.  It looks like we really do want a way to distinguish session types, and not use a "fake" file extension for consoles.

minrk:
I think this is solid, and makes sense, and we should do it. Merging so we can see how it goes.

blink1073:
This will not help JupyterLab unless we backport it to 4.2, or provide a custom `SessionManager` in JupyterLab.

minrk:
It would be 4.3 since 4.2 is already out. I'd rather it get some tested use before it goes in a notebook release. How do we go about doing that?

blink1073:
I can make a PR in services that explicitly uses this API, and make sure it passes against notebook master.  Then, once 4.3 is released, we bump the minimum version of the notebook for services and JLab.

Issue #1350, opened 2016-04-14
Should we hardcode CSS mimetype?


takluyver:
Users on Windows have a recurring issue with CSS being served with the wrong mimetype, which makes the browser refuse to load it, showing the ugly unstyled page. For example:

* ipython/ipython#6910
* ipython/ipython#7024
* ipython/ipython#9364
* http://stackoverflow.com/q/36521428/434217

Tornado's StaticFileHandler [calls](https://github.com/tornadoweb/tornado/blob/b6998247702a960f0c73004a6b5f8ae775607c03/tornado/web.py#L2556) `mimetype.guess_type()` to get the `Content-Type` header. On Windows, this queries the registry, and certain applications insert incorrect mimetype entries in there. We know the mimetype CSS files always need to be served with (`text/css`).

So far we've told people that their system is misconfigured and they need to fix that. I think this has come up often enough that we should work around it in our own code. This would involve serving files with a SFH subclass that overrides `get_content_type`.

We could also ask about hardcoding a check in tornado itself - the mimetype for CSS files shouldn't be a 'guess' at all. But I guess tornado sees less need to deal with misconfigured Windows systems than we do.

minrk:
Might be worth pinging tornado about it, too.

takluyver:
Pinged tornado (tornadoweb/tornado#1690): they don't think a fix belongs in tornado, but they did suggest a simple workaround:

```python
mimetypes.add_type('text/css', '.css')
```

As they said, changing global state isn't great, but if global state is just plain wrong...

PR #1360, opened 2016-04-15
Override mimetype for .css files

takluyver:
This should avoid the issue several people have had with unstyled pages on misconfigured Windows systems. I think practicality beats purity here.

Closes gh-1350


takluyver:
@minrk do we want this for 4.2.1, or is it a bit too significant?

minrk:
Let's do it!

PR #1370, opened 2016-04-19
use os.replace instead of os.rename if available

### LINKED ISSUE #1270: atomic_writing use os.replace instead of os.rename
When it make sense, which is python 3.

varmas:
Closes #1270 

@captainsafia 

Yay!

### DIFF
--- a/notebook/services/contents/fileio.py
+++ b/notebook/services/contents/fileio.py
@@ -30,6 +30,16 @@ except ImportError: #PY2
     from base64 import encodestring as encodebytes, decodestring as decodebytes
 
 
+def replace_file(src, dst):
+    """ replace dst with src
+
+    switches between os.replace or os.rename based on python 2.7 or python 3
+    """
+    try:
+        os.replace(src, dst)
+    except:
+        os.rename(src, dst)
+
 def copy2_safe(src, dst, log=None):
     """copy src to dst
 
@@ -104,7 +114,7 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
         if os.name == 'nt' and os.path.exists(path):
             # Rename over existing file doesn't work on Windows
             os.remove(path)
-        os.rename(tmp_path, path)
+        replace_file(tmp_path, path)
         raise
 
     # Flush to disk
@@ -277,8 +287,8 @@ class FileManagerMixin(Configurable):
             # Rename over existing file doesn't work on Windows
             if os.name == 'nt' and os.path.exists(invalid_file):
                 os.remove(invalid_file)
-            os.rename(os_path, invalid_file)
-            os.rename(tmp_path, os_path)
+            replace_file(os_path, invalid_file)
+            replace_file(tmp_path, os_path)
             return self._read_notebook(os_path, as_version)
 
     def _save_notebook(self, os_path, nb):


captainsafia:
Hey @Carreau! We tried to address the issue that you opened. I think this is what you wanted it but if not let us now.

We are doing this PR at PyData Chicago hack nights.

Carreau:
Hey ! Thanks @varmas hope you are enjoying PyData Chicago hack nights !

Sorry for the delay I'm in europe, I'll make a few inline comments !

notebook/services/contents/fileio.py
```diff
@@ -30,6 +30,16 @@
...
 
 
+def replace_file(src, dst):
+    """ replace dst with src
+
+    switches between os.replace or os.rename based on python 2.7 or python 3
+    """
+    try:
+        os.replace(src, dst)
+    except:
```
Carreau: So here don't leave a catch all `except:` clause or you will catch unwanted errors like permission errors, disk full ...and so on, and so forth, so when you can, it is better to be explicit about exception you catch when possible. 

Here it's likely not possible, so you you will have to check with an `if/else ` and `sys.version_info >= (3,3)`

minrk: A check for AttributeError should be sufficient.

takluyver: In general I think it's best to avoid catching `NameError` and `AttributeError`, because they can both hide typos and other easy mistakes. It's probably not a big deal here, because there's only one attribute in use and it's calling a standard library function (which we can probably assume doesn't have trivial bugs). But it's best to cultivate good habits. ;-)

notebook/services/contents/fileio.py
```diff
@@ -30,6 +30,16 @@
     from base64 import encodestring as encodebytes, decodestring as decodebytes
 
 
+def replace_file(src, dst):
+    """ replace dst with src
+
+    switches between os.replace or os.rename based on python 2.7 or python 3
```
takluyver: This isn't super helpful ;-). When writing docstrings or comments, think about why someone might be reading it. Mentioning Python 3 and 2.7 is a good start, it suggests that one is better but not available on the older version, but it doesn't say which, nor why it's better.

notebook/services/contents/fileio.py
```diff
@@ -104,7 +114,7 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
         if os.name == 'nt' and os.path.exists(path):
             # Rename over existing file doesn't work on Windows
             os.remove(path)
```
takluyver: This bit should be moved into `replace_file` - the idea is that `os.replace()` can atomically replace `path` if it already exists, whereas `os.rename()` can't on Windows. So we only need this extra check when `os.replace()` isn't available.

notebook/services/contents/fileio.py
```diff
@@ -277,8 +287,8 @@ def _read_notebook(self, os_path, as_version=4):
             # Rename over existing file doesn't work on Windows
             if os.name == 'nt' and os.path.exists(invalid_file):
                 os.remove(invalid_file)
```
takluyver: Same here.

captainsafia:
Thanks for the comments, guys!

@varmas: If you want to review the statements made about the `except` clause and docstrings and make the changes yourself, feel free to do so. If you'd like me to work through the review comments with you, let me know and we can try it at the next meeting. Also, the team can help you out virtually if you are OK with that. 🎉 💃

varmas:
woah, what time do you guys start your day!?!

@captainsafia I'll go for it myself, I was a bit careless yesterday. Will of course save it for the next meetup if I can't get this PR successful

minrk:
@varmas we are scattered across the globe, so you get reviews 24/7  🌎🌍🌏

minrk:
@varmas will you have a chance to finish this one? If not, that's totally fine. We can do the last touches if you are too busy.

minrk:
Merging here, will finish up addressing the review in a follow-up PR. Thanks, @varmas!

Carreau:
Thanks @varmas  and @minrk

PR #1374, opened 2016-04-19
Enforce npm dedupe

### LINKED ISSUE #1373: Node De-Duplication
Npm v3+ installs are maximally flat by default, but versions before that require `npm dedupe` to ensure that two versions of the same package with the same version are not installed.  This is a problem when a module stores module-private data structures that are meant to be shared and loaded only once.

I propose that the version of npm be checked on install, and if it is 3+, install with no modification.  Otherwise, attempt to run `npm dedupe` after `npm install`.  If that fails (as has been seen due to file permissions [issues](https://github.com/jupyter/atom-notebook/issues/32#issuecomment-184922403)), prompt the user to install npm 3+.  

I can make the necessary changes to `setupbase.py`.

blink1073:
Fixes #1373.

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -336,13 +336,16 @@ def run(cmd, *args, **kwargs):
 
 def npm_install(cwd):
     """Run npm install in a directory and dedupe if necessary"""
+
     try:
         run(['npm', 'install', '--progress=false'], cwd=cwd)
     except OSError as e:
         print("Failed to run `npm install`: %s" % e, file=sys.stderr)
         print("npm is required to build a development version of the notebook.", file=sys.stderr)
         raise
-    version = check_output('npm --version', shell=True).decode('utf-8')
+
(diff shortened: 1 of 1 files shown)


setupbase.py
```diff
@@ -333,6 +334,24 @@ def run(cmd, *args, **kwargs):
...
 
+def npm_install(cwd):
+    """Run npm install in a directory and dedupe if necessary"""
+    try:
+        run(['npm', 'install', '--progress=false'], cwd=cwd)
+    except OSError as e:
+        print("Failed to run `npm install`: %s" % e, file=sys.stderr)
+        print("npm is required to build a development version of the notebook.", file=sys.stderr)
+        raise
+    version = check_output('npm --version', shell=True).decode('utf-8')
```
minrk: shell=True here, False above and below?

blink1073: Thanks, fixed.

PR #1378, opened 2016-04-20
Protection againt bad custom.js

### LINKED ISSUE #1363: Protect against bad custom.js
if custom.js has a bad require, (eg require something only available on 5.x , like `require('notebook')`) all hell break loose and the page do not load. 

That should not be the case.

Marking as 4.3 as it is a bit annoying to debug each time you revert to an older version.

parleur:
Just embed it in a try/catch

closes #1363

### DIFF
--- a/notebook/static/edit/js/main.js
+++ b/notebook/static/edit/js/main.js
@@ -25,7 +25,13 @@ require([
     notificationarea
     ){
     "use strict";
-    requirejs(['custom/custom'], function() {});
+
+    try {
+        requirejs(['custom/custom'], function() {});
+    } catch(err) {
+        console.log("Error loading custom.js from edition service. Continuing and logging");
+        console.warn(err);
+    }
     
     page = new page.Page();
 
--- a/notebook/static/notebook/js/main.js
+++ b/notebook/static/notebook/js/main.js
@@ -52,8 +52,13 @@ require([
 
     // Pull typeahead from the global jquery object
     var typeahead = $.typeahead;
-
-    requirejs(['custom/custom'], function() {});
+    
+    try{
+        requirejs(['custom/custom'], function() {});
+    } catch(err) {
+        console.warn("Error processing custom.js file. Logging the error and continuing")
+        console.log(err);
+    }
 
     // compat with old IPython, remove for IPython > 3.0
     window.CodeMirror = CodeMirror;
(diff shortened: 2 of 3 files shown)


Carreau:
Closes https://github.com/jupyter/notebook/issues/1363 I guess ?

takluyver:
I think we should call @parleur 'Sprint', because he's so good at fixing issues tagged 'Sprint friendly' :-)

Carreau:
@parleur aka 'Sprint', make also puns in his commits.

minrk:
custom.js is also loaded on the `tree` and `edit` pages.

parleur:
I am not really sure about the purpose of custom.js, but maybe it would be meaningful to have a single entry point for js injection? 
I mean, knowing there is three (and it's not obvious), I would first check in custom.js from where the script has been called.

So what about split custom.js in `notebook_custom.js`, `edit_custom.js`, and `tree_custom.js` ?

Carreau:
> So what about split custom.js in notebook_custom.js, edit_custom.js, and tree_custom.js ?

These kind of already exists, and they are nbextensions. If you drop a file in `.jupyter/nbextension/` and modify `~/.jupyter/nbconfig/tree.json`, `~/.jupyter/nbconfig/notebook.json`, `~/.jupyter/nbconfig/edit.json` like so:

```
cat ~/.jupyter/nbconfig/tree.json
{
  "load_extensions": {
    "jupyter_cms/dashboard/main": true,
    "ipyparallel/main": true,
    "assignment_list/main": true
  }
}
```
(here I have 3 extensions enabled)

You can tell which extension to enable on which page, independently. And these files and layout is programmatically modifiable, which is better to install automatically multiple extensions. `custom.js` is just an artifact of the past (really we should just get rid of it ), but people are still using it, and it's damned convenient.

minrk:
Custom.js is an early escape hatch for a lack of configurability in the frontend. As we add that configurability, the need for custom.js diminishes, but it isn't gone yet. It is my hope that JupyterLab, for instance, will have sufficient configurability and not load custom.js.

takluyver:
Do we want this in 4.2.1?

Carreau:
Manually backporting this one, because of the  webpack thing it's unhappy.
The backporting tool will mistakenly says this has not been backported.

PR #1389, opened 2016-04-23
Enable older version of ipywidgets on notebook 4.2

SylvainCorlay:
Fixes #1386.

@minrk  This makes it work but I can't believe this is the best way to do this.

PR #1399, opened 2016-04-27
Enable widgets 4.1

SylvainCorlay:
@minrk this is a backport to 4.x of #1389.

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -79,6 +79,22 @@
               }
           }
       });
+
+      define("bootstrap", function () {
+          return window.$;
+      });
+
+      define("jquery", function () {
+          return window.$;
+      });
+
+      define("jqueryui", function () {
+          return window.$;
+      });
+
+      define("jquery-ui", function () {
+          return window.$;
+      });
     </script>
 
     {% block meta %}


jcb91:
Hi, this appears to have broken [a template](https://github.com/jcb91/IPython-notebook-extensions/blob/721b2f7f9939182f58146924ebd957a6e4fa2a19/src/themysto/nbextensions_configurator/templates/nbextensions_configurator.html) based on `page.html`, which now throws an error from requirjs with a stack beginning `TypeError: $ is not a function`. It seems that in this template, `window.$` is undefined. I'm not really sure how/where `window.$` is supposed to get defined, and it seems a bit strange to set a jquery path in [notebook/templates/page.html#L33](https://github.com/jupyter/notebook/blob/ee44d1947f0ad9e3ca309ac6d9d1aa1778e9d0ee/notebook/templates/page.html#L33), only to re-define it further down the page in [notebook/templates/page.html#L88](https://github.com/jupyter/notebook/blob/ee44d1947f0ad9e3ca309ac6d9d1aa1778e9d0ee/notebook/templates/page.html#L88).

So, yeah, any info on what the thinking behind this is, and what am I doing wrong appreciated :)

jcb91:
relevant stack traces:

```
TypeError: $ is not a function
Stack trace:
@http://localhost:12341/a%40b/static/base/js/events.js:23:12
...
newContext/context.onScriptLoad@http://localhost:12341/a%40b/static/components/requirejs/require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1711:21
 require.js:900
```
```
TypeError: $ is undefined
Stack trace:
@http://localhost:12341/a%40b/static/base/js/keyboard.js:63:9
...
newContext/context.onScriptLoad@http://localhost:12341/a%40b/static/components/requirejs/require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1711:21
 require.js:900
```
```
TypeError: $ is not a function
Stack trace:
get_body_data@http://localhost:12341/a%40b/static/base/js/utils.js:521:19
...
newContext/context.onScriptLoad@http://localhost:12341/a%40b/static/components/requirejs/require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1711:21
 require.js:900
```

minrk:
@SylvainCorlay any ideas?

jcb91:
so, if I insert the main notebook minified javascript below:

```html
<script src="/static/notebook/js/main.min.js" type="text/javascript" charset="utf-8"></script>
```

 the require errors go away. I assume that this is because `window.$` has already been successfully included by the jquery included by the minified script before the `define` call, since the same is not true when using the non-minified tag

```html
<script src="/static/notebook/js/main.min.js" type="text/javascript" charset="utf-8"></script>
```

jcb91:
indeed, if the `NotebookApp.ignore_minified_js` config value is set `True`, the notebook page also fails to load, with a string of javascript console errors about undefined `$`

jcb91:
I've so far got around this by wrapping my require call in another one to handle the potentially-undefined jquery:

```
require(['jquery'], function (jq) {
	// hack to fix notebook 4.2.1
	if (jq === undefined) {
		require.undef('jquery');
		require.undef('jquery-ui');
		require.undef('jqueryui');
		require.undef('bootstrap');
	}
	require(['nbextensions/nbextensions_configurator/main'], function (nbext_config_module) {
		nbext_config_module.build_page();
	});
});
```

jasongrout:
I'm trying to duplicate the original issue in 4.2.0:

```
conda create -n widgets python=3 notebook=4.2.0 ipywidgets=4.1.1 nb_conda_kernels
source activate widgets
```
Running the code in the original issue had no problems:
```
from ipywidgets import widgets
from IPython.display import display
s = widgets.IntSlider()
display(s)
```

jasongrout:
Digging deeper, we are using a jqueryui that does not support AMD loading (we have 1.10, AMD loading came in 1.11). No wonder things are messed up.

jasongrout:
Digging even deeper, one problem I had was that loading main.min.js was stomping all over other things (like the jquery module definition), so we'd end up with multiple copies of jquery overwriting each other.

jcb91:
:astonished:

PR #1409, opened 2016-04-30
Configurable MathJax config

poke1024:
MathJax supports different combined configuration modes, like TeX-MML-AM_HTMLorMML or TeX-MML-AM_SVG (http://docs.mathjax.org/en/latest/config-files.html). Currently, Jupyter hard codes "TeX-AMS_HTML-full" in its notebook.html template. This makes it difficult for experimental Jupyter kernels which need or are designed around other modes (like MML, i.e. MathML).

In the long run, it would be great if Jupyter had a kernel specific MathJax setting in this regard. This, however, is beyond my capabilities and I guess it would be a rather big change, since MathJax config need to be stored statically on the server side.

What I'm suggesting in this pull request therefore is to make the MathJax config mode configurable via the Jupyter config file, so developers can easily work with their experimental kernels in the needed MathJax mode without having to patch the Jupyter source code. They still need to provide their custom MathJax config files in notebook/static/components/MathJax/config, but this is easier, and so the change would still be of great utility.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -238,6 +238,7 @@ class NotebookWebApplication(web.Application):
             nbextensions_path=ipython_app.nbextensions_path,
             websocket_url=ipython_app.websocket_url,
             mathjax_url=ipython_app.mathjax_url,
+            mathjax_config=ipython_app.mathjax_config,
             config=ipython_app.config,
             config_dir=ipython_app.config_dir,
             jinja2_env=env,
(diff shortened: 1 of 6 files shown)


notebook/notebookapp.py
```diff
@@ -728,6 +729,15 @@ def _mathjax_url_changed(self, name, old, new):
         else:
             self.log.info("Using MathJax: %s", new)
 
+    mathjax_config = Unicode("", config=True,
+        help="""The MathJax.js configuration file that is to be used."""
+    )
+    def _mathjax_config_default(self):
```
minrk: You don't need a dynamic default here, it can be in the constructor:

```python
mathjax_config = Unicode("TeX-AMS_HTML-full,Safe", config=True...
```

Issue #1408, opened 2016-04-29
.sh file won't open in text editor?


jessehoff:
Trying to open a text file with .sh extension. Other text files of various extension open fine.
When I try open the .sh file it tries to download
Working on chrome in a mac with a server side notebook.

takluyver:
On my system, shell files have mimetype `application/x-sh` - our rule for when to try to edit things is to check for the `text/` media type. We think the most general solution is to have *edit* and *view/download* buttons for all file types (possibly with a blacklist of definitely-not-text filetypes like image/png that we know we can't edit).

captainsafia:
There was discussion related to this with respect to #1143. I'm working on a more permanent solution for this as we speak!

minrk:
This was fixed by #1413

kxiluri:
Hello everyone, i have the problem you describe here, but i do not know how to fix it.
Could anyone please help.
When i click on a .txt file it renders fine in the text editor, but when i click on .sh it does nothing.
Kind Regards
kiriaki

frankr6591:
Same here... I am on mac and when I click on .sh ... it starts Xcode... want to edit inside jupyter.

kxiluri:
i had figured it out it i will try to remember
stay tuned
k

Sent from my iPad

On Aug 26, 2017, at 8:30 AM, frankr6591 <notifications@github.com<mailto:notifications@github.com>> wrote:


Same here... I am on mac and when I click on .sh ... it starts Xcode... want to edit inside jupyter.

takluyver:
There were some further changes to this for 5.1, which we're just getting ready to release. If you want to help us test the release candidate, you can upgrade to it:

```
pip install --upgrade --pre notebook
```

PR #1413, opened 2016-05-02
Add PERMITTED_MIMETYPES to mimetype check

captainsafia:
This is a follow up to #1143 and #1408.

I think the easiest and most scalable solution would be to set up a PERMITTED_MIMETYPES array that we can add strange but editable mime types to. We can build the download and edit functionality on top of this.

cc: @takluyver @minrk

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -52,6 +52,10 @@ define([
             'sort-name': 0
         };
         this._max_upload_size_mb = 25;
+        this.PERMITTED_MIMETYPES = [
+          'application/javascipt',
+          'application/x-sh',
+        ];
     };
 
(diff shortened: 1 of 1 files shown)


minrk:
Makes sense. I would probably do `EDIT_MIMETYPES`, since it's specifically for those that should go to the editor.

captainsafia:
Updated.

takluyver:
I suspect that it may be better to offer an 'edit' button for each file, separate from the view/download action, and then introduce a blacklist for files that we know we can't edit, rather than this whitelist approach. There are files like HTML and SVG that we can open in a text editor, but the user might also want to view them in the browser.

minrk:
Edit button + blacklist sounds great in the long run. Merging this as an incremental improvement, which we can remove when the edit button is added.

PR #1418, opened 2016-05-03
log exceptions in nbconvert handlers

minrk:
when nbconvert fails, log the traceback

cc @janschulz who reported this [on conda-forge](https://github.com/conda-forge/nbconvert-feedstock/issues/3#issuecomment-216256276)

### DIFF
--- a/notebook/nbconvert/handlers.py
+++ b/notebook/nbconvert/handlers.py
@@ -8,6 +8,7 @@ import os
 import zipfile
 
 from tornado import web, escape
+from tornado.log import app_log
 
 from ..base.handlers import (
     IPythonHandler, FilesRedirectHandler,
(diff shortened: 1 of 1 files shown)


notebook/nbconvert/handlers.py
```diff
@@ -70,6 +71,7 @@ def get_exporter(format, **kwargs):
     try:
         return Exporter(**kwargs)
     except Exception as e:
+        app_log.exception("Could not construct Exporter: %s", Exporter)
```
takluyver: Should this also include the exception object?

minrk: `log.exception` does include the current exception and traceback.

notebook/nbconvert/handlers.py
```diff
@@ -103,6 +105,7 @@ def get(self, format, path):
                 }
             )
         except Exception as e:
+            self.log.exception("nbconvert failed: %s", e)
```
takluyver: In that case surely this *doesn't* need the exception object?

minrk: It's not strictly necessary, but I like having the single-line summary.

takluyver: OK. I find it a bit odd that you have done this in two different ways within one PR, but that's not important.

PR #1420, opened 2016-05-03
Make Importing Notebooks guide work again.

yuvallanger:
Please review the code

### DIFF
--- a/docs/source/examples/Notebook/Importing Notebooks.ipynb
+++ b/docs/source/examples/Notebook/Importing Notebooks.ipynb
@@ -40,7 +40,7 @@
    "outputs": [],
    "source": [
     "from IPython import get_ipython\n",
-    "from IPython.nbformat import current\n",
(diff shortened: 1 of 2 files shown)


docs/source/examples/Notebook/Importing Notebooks.ipynb
```diff
@@ -130,7 +130,7 @@
     "                                       \n",
     "        # load the notebook object\n",
     "        with io.open(path, 'r', encoding='utf-8') as f:\n",
-    "            nb = current.read(f, 'json')\n",
+    "            nb = read(f, current_nbformat)\n",
```
takluyver: This should be specified as `4` explicitly - the idea of requiring it as a parameter is that code like this won't break each time we update the format, because it will be translated back to the format the code understands. Using `current_nbformat` means it will break.

yuvallanger:
Is this what you had in mind?

takluyver:
Yep, thanks! :cake:

PR #1429, opened 2016-05-04
Add codecov.yml

Carreau:
And lower coverage threshold to 60%
this should prevent coverage to fail on simple patch like
https://github.com/jupyter/notebook/pull/1418

--- 

Ping @minrk. 
Cf documentation for [codecov.yml](https://codecov.io/gh/jupyter/notebook/settings/yaml)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,8 +3,8 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 0.1
+        threshold: 10
     patch:
       default:
         target: auto
-        threshold: 0.1
+        threshold: 10
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0.1
+    patch:
+      default:
+        target: auto
+        threshold: 0.1


takluyver:
Is it possible for it to check change in coverage, rather than the total number? i.e. it might be handy to fail if coverage dropped by, say, >= 1%.

Carreau:
> Is it possible for it to check change in coverage, rather than the total number? i.e. it might be handy to fail if coverage dropped by, say, >= 1%.

I looked for this setting but havent found it:

```
Coverage Target

number; 0 <= N <= 100 | auto

auto: (default) Codecov will compare coverage against the parent commit to determin if coverage increased or decreased.

number: Codecov will set the status based on if coveage was greater then or equal to the target.
```

```
Threshold

numeric; 0 <= N <= 100

To submit this notification the coverage must change beyond the threshold.

threshold: 1.25
```

But never seem like a delta.

takluyver:
"the coverage must change beyond the threshold." sounds like what I'm thinking of. So I guess:

```
number: auto
threshold: 0.1
```

(It sounds like it only shows up at all if it changes by more than threshold one way or the other, so I've set it quite low so we can get a +1 when coverage increases).

Carreau:
> (It sounds like it only shows up at all if it changes by more than threshold one way or the other, so I've set it quite low so we can get a +1 when coverage increases).

that's not how I understand it, I understand it as coverage bot making a comment. 

But that mean we'll get a coverage not passing for -0.1

Carreau:
updated.

takluyver:
> But that mean we'll get a coverage not passing for -0.1

We can always ignore it and merge anyway, and it's probably a prod in the right direction.

Carreau:
sure, let see and update if necessary. I just dont want to get failure fordrom of 0.01 in coverage.

minrk:
I tried to disable these commit statuses entirely at the team config level when codecov decided to throw out all user preferences a bit ago. I don't like coverage being part of the test-passing status, so I would rather just turn it off than try to pick some heuristic for what should count as "passing"

takluyver:
The automated comments annoy me, but I do like seeing coverage changes in the PR status.

Carreau:
I tend to agree with Min on the general idea, but I think that a drop of 10% in coverage, is likely a mistake where you accidently skip or drop some of the tests (which happened to me more than once).

minrk:
I agree with @takluyver that having a little info in the status is useful. I just don't think coverage should be considered for failure. So if it were up to me, I would probably tell it to set a coverage of 1% as the threshold for success. @Carreau if you want to set the drop threshold to something large like a 10-20% drop in total coverage, that's probably enough that it won't fail unless something really is wrong.

Carreau:
Threshold set to 10.

PR #1432, opened 2016-05-06
Document `kernel.js` extensions.

Carreau:
--- 
Ping @willingc


willingc:
@Carreau I just created a PR against your branch :smile:

willingc:
Thanks @Carreau.

PR #1450, opened 2016-05-13
[WIP] Unify JupyterLab with the main npm build

blink1073:
This currently works (and results in one less npm install), but needs some cleanup.


minrk:
Cool, thanks @blink1073! What cleaning up do you need to do?

blink1073:
Just general housekeeping on the changes, making sure I didn't add/remove too much.

blink1073:
@minrk, I though about this more over the weekend, and I think it would be prudent to hold off on this until we switch over to `phosphor-ui`, since the existing phosphor repos have a `dedupe` post install script and we have previously established in this repo that we would only `dedupe` once and only if using an older `npm`.

minrk:
Sure, that makes sense.

blink1073:
Closing in favor of #1506.

PR #1453, opened 2016-05-13
added terminado_settings

robnagler:
Specifically, to support overriding terminado's shell_command with something that allows flags. We are using JupyterHub, and it's important for our users to get a clean shell (bash --login).

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -608,6 +608,9 @@ class NotebookApp(JupyterApp):
             help="Supply overrides for the tornado.web.Application that the "
                  "Jupyter notebook uses.")
     
+    terminado_settings = Dict(config=True,
+            help='Supply overrides for terminado. Currently only supports "shell_command".')
+
     cookie_options = Dict(config=True,
         help="Extra keyword arguments to pass to `set_secure_cookie`."
              " See tornado's set_secure_cookie docs for details."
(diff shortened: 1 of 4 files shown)


notebook/notebookapp.py
```diff
@@ -608,6 +608,9 @@ def _webapp_settings_changed(self, name, old, new):
             help="Supply overrides for the tornado.web.Application that the "
                  "Jupyter notebook uses.")
     
+    terminado_settings = Dict(config=True,
+            help='Supply overrides for terminado. Currently only supports "shell_command".')
```
takluyver: Are there other things you envisage putting in here? Would it be simpler to have `terminals_shell_command` as a configurable thing by itself?

robnagler: I didn't want to add anything I didn't need right now. I could see someone wanting to add environment variables or , for instance. It also seems to me that it is also consistent with `tornado_settings`.

takluyver: It's a reasonable enough approach. I think tornado_settings is a bit of a special case because there are so many things that might go in there, but I'd be happy either way. Let's give others a chance to look at this.

minrk:
Since you made this PR from your master branch, it looks like it's pulling in work from other PRs. Do you want to get this back to just the proposed change?

robnagler:
I'll resubmit. Not sure how to remove commits from an existing PR.

PR #1458, opened 2016-05-16
typeahead.min.js (not typeahead.js)

robnagler:
page.html was referring to typeahead.js, which does not exist in the dist build.


minrk:
Thanks!

PR #1468, opened 2016-05-19
Allow root user to run tests.

### LINKED ISSUE #1056: re-enable CircleCI
Circle CI don't respect `circle.yml` test function anymore, which make all PRs fail. 
I've diable the webhook for now.

We can investigate reenabling later.

jtyberg:
There was a change to disallow root user to run the notebook server, which is all good, but it broke the Docker image build.  These changes enable the Docker image to build again.

FWIW, I'm not convinced that these changes should go in, or if the Dockerfile should get more love and run as a different user, like all the [docker-stacks](http://github.com/jupyter/docker-stacks) do.

Fixes #1056

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # Another Docker container should inherit with `FROM jupyter/notebook`
 # to run actual services.
 #
-# For opinionated stacks of ready-to-run Jupyter applications in Docker, 
+# For opinionated stacks of ready-to-run Jupyter applications in Docker,
 # check out docker-stacks <https://github.com/jupyter/docker-stacks>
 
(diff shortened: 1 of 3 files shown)


minrk:
I'm not convinced the dockerfile on this repo should exist at all, and we should probably just point to docker-stacks from now on.

jtyberg:
It's hard to disagree.  Clearly, no one is using it.  I reached for it because I thought it was the way to do a little dev without polluting my system, but volume mounting the source directory from my host is hampering that effort.

Carreau:
[Backporting this one manually](https://github.com/jupyter/notebook/commit/e1ff71d3cfcca9d6495c45505a13fd866a13110e), the backporting tool will mistakenly say it has not been backported.

PR #1476, opened 2016-05-22
Don't require patch coverage for CI

minrk:
Keep it running so the number is visible, but don't use it as a CI criterion.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,5 +6,4 @@ coverage:
         threshold: 10
     patch:
       default:
-        target: auto
-        threshold: 10
+        target: 0%


minrk:
BTW, our coverage numbers are not especially informative because we only record what's tested by the Python tests, not any of the what's exercised by the js tests. As a result, I'm not sure how valuable coverage-as-ci is on this repo.

Carreau:
> BTW, our coverage numbers are not especially informative because we only record what's tested by the Python tests, not any of the what's exercised by the js tests. As a result, I'm not sure how valuable coverage-as-ci is on this repo.

I think there is way to also submit coverage with JS, but that would be easier if tested in node.
We'll figure that out at some point in the future.

minrk:
> I think there is way to also submit coverage with JS

To be clear, I mean that our *Python* coverage is significantly under-reported, because it doesn't include Python code only exercised by our js tests (websocket handlers, etc.).

PR #1482, opened 2016-05-24
Respect 'editable' cell metadata

### LINKED ISSUE #1402: True read-only cells
nbgrader currently has pseudo "read-only" cells where they are "read-only" in the sense that their contents get saved to the nbgrader database, and then overwritten during grading if they have been changed by the student. I have reports that this is confusing, and requests for true read-only cells where the cell actually cannot be edited at all in the live notebook.

I think this would be really useful to have for nbgrader. We could keep it as an undocumented feature for the moment, just like "undeletable" cells are, until we can figure out an appropriate UI for it.

yuvipanda:
- Readonly cells can not be split or merged or deleted
- Readonly cells can be executed
- End users can still modify this by editing the metadata
  themsevles directly

Fixes #1402

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -498,48 +498,46 @@ define([
 
 
     /**
-     * can the cell be split into two cells (false if not deletable
-     * or if readonly)
+     * can the cell be split into two cells (false if not deletable)
      *
      * @method is_splittable
      **/
     Cell.prototype.is_splittable = function () {
-        return this.is_deletable() && !this.is_readonly();
+        return this.is_deletable();
     };
 
 
     /**
-     * can the cell be merged with other cells (false if not deletable
-     * or if readonly)
+     * can the cell be merged with other cells (false if not deletable)
      * @method is_mergeable
      **/
     Cell.prototype.is_mergeable = function () {
-        return this.is_deletable() && !this.is_readonly();
+        return this.is_deletable();
     };
 
     /**
-     * is the cell readonly? only true (readonly) if
-     * metadata.readonly is explicitly true -- everything else
-     * counts as false
+     * is the cell edtitable? only false (readonly) if
+     * metadata.editable is explicitly false -- everything else
+     * counts as true
      *
-     * @method is_readonly
+     * @method is_editable
      **/
-    Cell.prototype.is_readonly = function () {
-        if (this.metadata.readonly === true) {
-            return true;
+    Cell.prototype.is_editable = function () {
+        if (this.metadata.editable === false) {
+            return false;
         }
-        return false;
+        return true;
     };
 
     /**
      * is the cell deletable? only false (undeletable) if
-     * metadata.deletable is explicitly false -- everything else
-     * counts as true
+     * metadata.deletable is explicitly false or if the cell is not
+     * editable -- everything else counts as true
      *
      * @method is_deletable
      **/
     Cell.prototype.is_deletable = function () {
-        if (this.metadata.deletable === false) {
+        if (this.metadata.deletable === false || !this.is_editable()) {
             return false;
         }
         return true;
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -498,20 +498,37 @@ define([
 
 
     /**
-     * can the cell be split into two cells (false if not deletable)
+     * can the cell be split into two cells (false if not deletable
+     * or if readonly)
+     *
      * @method is_splittable
      **/
     Cell.prototype.is_splittable = function () {
-        return this.is_deletable();
+        return this.is_deletable() && !this.is_readonly();
     };
 
 
     /**
-     * can the cell be merged with other cells (false if not deletable)
+     * can the cell be merged with other cells (false if not deletable
+     * or if readonly)
      * @method is_mergeable
      **/
     Cell.prototype.is_mergeable = function () {
-        return this.is_deletable();
+        return this.is_deletable() && !this.is_readonly();
+    };
+
+    /**
+     * is the cell readonly? only true (readonly) if
+     * metadata.readonly is explicitly true -- everything else
+     * counts as false
+     *
+     * @method is_readonly
+     **/
+    Cell.prototype.is_readonly = function () {
+        if (this.metadata.readonly === true) {
+            return true;
+        }
+        return false;
     };
 
     /**
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -171,6 +171,8 @@ define([
             if (that.keyboard_manager) {
                 that.keyboard_manager.enable();
             }
+
+            that.code_mirror.setOption('readOnly', that.is_readonly());
         });
         this.code_mirror.on('keydown', $.proxy(this.handle_keyevent,this));
         $(this.code_mirror.getInputField()).attr("spellcheck", "false");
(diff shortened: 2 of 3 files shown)


yuvipanda:
Should readonly cells be deletable? I'm unsure! Perhaps it's ok to keep those two orthogonal, but might be confusing?

Should it provide some other visual indicator of being readonly? Probably, although I'm not sure what form that should take.

TODO:
1. Add tests
2. Figure out if setting codemirror option on every focus has performance penalties, or if there is a better location to put that in.

I haven't really written any client side JS in a few years, so be gentle!

takluyver:
From a UI point of view, I'd expect readonly cells to also be undeletable; it's hard to imagine a scenario where you want a non-editable but deletable cell.

notebook/static/notebook/js/codecell.js
```diff
@@ -171,6 +171,8 @@ define([
             if (that.keyboard_manager) {
                 that.keyboard_manager.enable();
             }
+
+            that.code_mirror.setOption('readOnly', that.is_readonly());
```
takluyver: This is on code cells - do we want non-editable Markdown & raw cells as well?

takluyver:
Will it be awkward that the two protection options work opposite ways round - `deletable: false` but `readonly: true`? Do we want `editable: false` instead?

yuvipanda:
In general I prefer defaults to false than defaults to true, but since it is already 'deletable: false' rather than 'undeletable: true', I guess it makes sense to keep it consistent.

notebook/static/notebook/js/codecell.js
yuvipanda: Yeah, I've duplicated it in textcell.js, which seems to be the only other cell type defined? I can't really set this in the parent cell.js since no codemirror stuff happens there.

takluyver: There's MarkdownCell and RawCell, both in `textcell.js`, but TextCell is the parent class, so if setting it there works, that's good.

yuvipanda: Yup, just did :)

Carreau:
I'm fine with changing the default, be warned that @jhamrick nbgrader might need to be updated as, IIRC, it is the main usecase.

yuvipanda:
Yeah, I think this just needs tests. I'll try to pull in some time next
week to finish this up.

On Fri, Jun 3, 2016 at 5:17 PM, Matthias Bussonnier <

> I'm fine with changing the default, be warned that @jhamrick
> <https://github.com/jhamrick> nbgrader might need to be updated as, IIRC,
> it is the main usecase.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/jupyter/notebook/pull/1482#issuecomment-223723193>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB23qN_GylSQWIh-WGm5rjLOJwi2ICLks5qIMQBgaJpZM4IlKSo>
> .
>



-- 
Yuvi Panda T
http://yuvi.in/blog

jhamrick:
I think not editable should imply not deletable as well.

I'm fine with changing the default but if we do that (go from `deletable` to `undeletable`) could we build in something that would automatically update the notebook metadata if the old `deletable` metadata is found?

Also thanks @yuvipanda for implementing this!!

minrk:
Yup, any change like that should include automatic upgrading of the old format on load.

gnestor:
@yuvallanger Anything I can do to help get this merged?

gnestor:
@minrk I can work on this. To make sure I understand: if the `deletable` property is not available for a cell in a notebook, then we should provide a value for it depending on the cell's `editable` property (e.g. if `metadata.editable = false` then `metadata.deletable = false`, otherwise  `metadata.deletable = true`).

minrk:
@gnestor I believe that's correct, yes.

notebook/static/notebook/js/cell.js
```diff
@@ -515,14 +516,28 @@ define([
...
+    /**
+     * is the cell deletable? only false (undeletable) if
+     * metadata.deletable is explicitly false or if the cell is not
+     * editable -- everything else counts as true
+     *
      * @method is_deletable
      **/
     Cell.prototype.is_deletable = function () {
-        if (this.metadata.deletable === false) {
+        if (this.metadata.deletable === false || !this.is_editable()) {
```
gnestor: > I'm fine with changing the default but if we do that (go from deletable to undeletable) could we build in something that would automatically update the notebook metadata if the old deletable metadata is found?

@jhamrick This block will handle cases where `metadata.deletable` is undefined (`is_deletable()` will return `true`). Do you think that we need to write that value back to the metadata? It seems like this would only be necessary if a notebook were to be used in another client that doesn't define an `is_deletable()` equivalent.

jhamrick: Yes, I think it should update the old metadata. As @minrk said:

> Yup, any change like that should include automatic upgrading of the old format on load.

So even before it gets to the point of checking if the cell is deleteable, ideally the old metadata format should be updated.

gnestor:
@yuvipanda Can you enable "Allow edits from maintainers" for this PR? It's a [new Github feature](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests) and should be available in the right side bar for this PR.

yuvipanda:
Done! Sorry I haven't had time to finish this up.

gnestor:
@yuvipanda Sorry but I accidentally deleted your branch and I don't think I can restore it on your behalf. I think a `git reflog` should do the trick. If you can't figure it out either, then I can create a new PR...

yuvipanda:
I've restored it :)

gnestor:
@yuvipanda I think the "Allow edits from maintainers" permissions were reset after I deleted your branch. I submitted a PR to your fork that will upgrade a notebook's metadata on load to include `editable` and `deletable` properties. Can you merge pleeez?

yuvipanda:
Merged and re-enabled!

gnestor:
@minrk @jhamrick Can you test this? I have tested and verified that cell metadata will include `editable` and `deletable` properties. 

To test: 

1. Create a new notebook or open an existing notebook
2. Inspect notebook in a text editor, there should be no `editable` or `deletable` properties for cells
3. Save and Checkpoint
4. Inspect notebook in a text editor, there should be `editable` or `deletable` properties for cells

Carreau:
LGTM ! Thanks all !

jhamrick:
Would it be possible to include this for the next release (4.4) rather than 5.0? This is a *really* important feature for nbgrader to have that I thought would have ended up in the 4.3 release, but apparently it didn't. It would be unfortunate if all the instructors waiting on this feature have to wait until 5.0 for it.

Carreau:
let's see how hard it is to backport. 

@meeseeksdev backport to 4.x

Carreau:
See #1482.

Carreau:
I mean #2039

jhamrick:
Thanks! 😄

Carreau:
> Thanks! 😄

No problem. Also I whitelisted you so you can mention the bot to do backport or other stuff.

gnestor:
Can we release this in 4.3.2? It's currently in 4.x branch and we're about to do a 4.3.2 release...

minrk:
I think it' fine. I relabeled it as 4.3.2, since it's already been backported.

Issue #1439, opened 2016-05-10
Non-python syntax highlighting disabled


fonnesbeck:
In the current `notebook` master running on OS X 10.11.4 (Python 3.5.1), non-Python kernels do not have syntax highlighting enabled. Examples for the R and Julia kernels are depicted below:
## R

![screenshot 2016-05-10 09 06 01](https://cloud.githubusercontent.com/assets/81476/15151885/bba8c738-1698-11e6-8519-204c485c9e77.png)
## Julia

![julia](http://fonnesbeck-dropshare.s3.amazonaws.com/Screen-Shot-2016-05-10-10-26-28.png)

System information:

Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:24:55) 
[GCC 4.2.1 (Apple Inc. build 5577)]

Kernel: R version 3.2.5 (2016-04-14) via IRKernel 0.5

willingc:
@Carreau @takluyver Thoughts? I'm seeing the same behavior as Chris (thanks for filing the issue) using the IJulia 0.4.5 kernel with similar OS config using Chrome or Firefox.

takluyver:
I also see it, but I don't know what's causing it. There's no error in the JS console.

Carreau:
It's the same that webpack that f**-up #1431 too, `require` just does not
work anymore, or at least not correctly, and there are basically 2
different CodeMirror library loaded on the page. So the mode is indeed
loaded correctly, but attached to the wrong copy of the module, so when the
cell highlighting changes, codemirror will just not find the mode, and give
up.

Carreau:
And that unfortunately **is** #1047, that I just bisected, so the introduction of webpack **did** break all the kernels and many custom js and extensions.

Ping @ellisonbg , @fperez , @jdfreder as this is pretty important.

ellisonbg:
Ouch, thanks for tracking this down. I am traveling more today. @jdfreder
is going to work on Friday for us and may be able to help us figure this
out...

On Tue, May 10, 2016 at 10:52 AM, Matthias Bussonnier <
notifications@github.com> wrote:

> And that unfortunately *is* #1047
> <https://github.com/jupyter/notebook/pull/1047>, that I just bisected, so
> the introduction of webpack *did* break all the kernels and many custom
> js and extensions.
>
> Ping @ellisonbg <https://github.com/ellisonbg> , @fperez
[...]

PR #1490, opened 2016-05-25
Unify the codemirror imports

blink1073:
Fixes https://github.com/jupyter/notebook/issues/1439

screen shot 2016-05-25 at 8 31 56 am

screen shot 2016-05-25 at 8 20 20 pm

### DIFF
--- a/notebook/static/notebook/js/codemirror-ipython.js
+++ b/notebook/static/notebook/js/codemirror-ipython.js
@@ -5,34 +5,35 @@
 
 (function(mod) {
   if (typeof exports == "object" && typeof module == "object"){ // CommonJS
-    mod(require("codemirror/lib/codemirror"),
-        require("codemirror/mode/python/python")
+    mod(require("codemirror/lib/codemirror")
         );
   } else if (typeof define == "function" && define.amd){ // AMD
-    define(["codemirror/lib/codemirror",
-            "codemirror/mode/python/python"], mod);
+    define(["codemirror/lib/codemirror"], mod);
   } else {// Plain browser env
     mod(CodeMirror);
   }
 })(function(CodeMirror) {
     "use strict";
 
-    CodeMirror.defineMode("ipython", function(conf, parserConf) {
-        var pythonConf = {};
-        for (var prop in parserConf) {
-            if (parserConf.hasOwnProperty(prop)) {
-                pythonConf[prop] = parserConf[prop];
+    requirejs(['codemirror/mode/python/python'], function () {
+        CodeMirror.defineMode("ipython", function(conf, parserConf) {
+            var pythonConf = {};
+            for (var prop in parserConf) {
+                if (parserConf.hasOwnProperty(prop)) {
+                    pythonConf[prop] = parserConf[prop];
+                }
             }
-        }
-        pythonConf.name = 'python';
-        pythonConf.singleOperators = new RegExp("^[\\+\\-\\*/%&|\\^~<>!\\?]");
-        if (pythonConf.version === 3) {
-            pythonConf.identifiers = new RegExp("^[_A-Za-z\u00A1-\uFFFF][_A-Za-z0-9\u00A1-\uFFFF]*");
-        } else if (pythonConf.version === 2) {
-            pythonConf.identifiers = new RegExp("^[_A-Za-z][_A-Za-z0-9]*");
-        }
-        return CodeMirror.getMode(conf, pythonConf);
-    }, 'python');
+            pythonConf.name = 'python';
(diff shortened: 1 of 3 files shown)


blink1073:
Note: I tried using the RegExp version of `externals`, but I could not get it to work.

https://webpack.github.io/docs/configuration.html#externals

blink1073:
Using the following:

```javascript
    /codemirror/,
        function(context, request, callback) {
            if (/codemirror/.test(request)) {
                return callback(null, 'commonjs ' + request);
            }
            callback();
        }
```

Resulting in this output:

```javascript
/***/ function(module, exports) {

	module.exports = codemirror/lib/codemirror;

/***/ },
```

blink1073:
I was able to get a configuration that used `require()` for all CodeMirror imports, but I was getting the error `"Module name has not been loaded yet for context: _. Use require([])"`

blink1073:
I can't figure out how to get the python mode loaded, which is causing the error in the build.

blink1073:
The last commit fixes the python mode, but I'm not sure how to verify whether anything else from the addons might be broken.

Carreau:
Thanks for trying to tackle that. 

I guess the only way to really have tested is to push it in the wild. 
I'll try to give it a try tomorrow, and merge if it works.

webpack.config.js
```diff
@@ -38,7 +38,9 @@ var commonConfig = {
       bootstrap: '$',
       bootstraptour: 'Tour',
       'jquery-ui': '$',
-      typeahead: '$.typeahead'
+      typeahead: '$.typeahead',
+      'codemirror/lib/codemirror': 'commonjs codemirror/lib/codemirror',
+      'codemirror/mode/meta': 'commonjs codemirror/mode/meta',
```
Carreau: Great ! Thanks can you post a link of where you found this info ? 
In a separate PR I guess it will be useful to do the same for `addon`, `keymap`, and `theme`.

notebook/static/notebook/js/codemirror-ipython.js
```diff
@@ -5,34 +5,35 @@
...
   }
 })(function(CodeMirror) {
     "use strict";
 
-    CodeMirror.defineMode("ipython", function(conf, parserConf) {
-        var pythonConf = {};
-        for (var prop in parserConf) {
-            if (parserConf.hasOwnProperty(prop)) {
-                pythonConf[prop] = parserConf[prop];
+    requirejs(['codemirror/mode/python/python'], function () {
```
Carreau: One thing I'm worried about is that we **can** do that for the IPython mode, but some codemirror modes do have this pattern:

```
django/django.js:        require("../../addon/mode/overlay"));
dockerfile/dockerfile.js:    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"));
factor/factor.js:    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"));
gfm/gfm.js:    mod(require("../../lib/codemirror"), require("../markdown/markdown"), require("../../addon/mode/overlay"));
handlebars/handlebars.js:    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"), require("../../addon/mode/multiplex"));
htmlembedded/htmlembedded.js:        require("../../addon/mode/multiplex"));
modelica/modelica.js:    mod(require("../../lib/codemirror"));
nsis/nsis.js:    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"));
rst/rst.js:    mod(require("../../lib/codemirror"), require("../python/python"), require("../stex/stex"), require("../../addon/mode/overlay"));
rust/rust.js:    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"));
tornado/tornado.js:        require("../../addon/mode/overlay"));
twig/twig.js:    mod(require("../../lib/codemirror"),  require("../../addon/mode/multiplex"));
vue/vue.js:        require("../../addon/mode/overlay"),
```

So they will likely be broken...

Carreau:
Regardless of my comments, this is a strict improvement, so +1 in getting this in, and figuring out what to do next. 

Kudos to @blink1073 for figuring things out and having taken the time to do a patch !

minrk:
Backported to 4.x, and confirmed to be working with R, at least. Thanks! I'll give 4.2.1 a few more days to turn up issues, and then release 4.2.2 with this by the end of this week.

blink1073:
@Carreau, this was guess-and-check based on https://webpack.github.io/docs/configuration.html#externals.  We need to have a broader conversation about how to rectify Webpack and Requirejs for shared resources like CodeMirror.

minrk:
Sorry about inappropriately backporting - I reverted the backport, since this bug does not exist on 4.x. Properly updated everything to 5.0

PR #1501, opened 2016-05-31
Big upload

nsknojj:
Add jquery-file-upload module, can upload big files.


notebook/templates/tree (复件).html
```diff
@@ -0,0 +1,199 @@
+{% extends "page.html" %}
```
minrk: I assume you didn't actuall mean to commit this file

notebook/notebookapp.py
```diff
@@ -182,7 +182,7 @@ def init_settings(self, ipython_app, kernel_manager, contents_manager,
         jenv_opt.update(jinja_env_options if jinja_env_options else {})
 
         env = Environment(loader=FileSystemLoader(template_path), **jenv_opt)
-
```
minrk: Please try to avoid committing big whitespace changes like these.

notebook/services/bigupload/handlers.py
```diff
@@ -0,0 +1,103 @@
...
+            raise web.HTTPError(403)
+        self.write(self.generate_response({'files': files}))
+
+    def generate_response(self, content):
+        self.set_header("Content-Type", "text/plain")
+        return content
+
+    def handle_file_upload(self, path, name, body, content_range):
+        file = dict()
+        file_path = self.get_upload_path(path, name)
```
minrk: File uploads must go to the contents manager

notebook/static/bigupload/js/cors/jquery.postmessage-transport.js
```diff
@@ -0,0 +1,120 @@
+/*
```
minrk: We do not include javascript dependencies in our repo. We use bower to express those dependencies.

minrk:
You may also want to look at #536, which is a partial implementation chunked-file upload, which talks to the contents service. We will need to store files via the contents service and not force it to be on the local filesystem.

nsknojj:
Hi, @minrk . Sorry for replying you so late. Before I change the structure of the code, I want to know your idea about the big file upload function. 
I add an additional upload button, and the only function is to upload files, not related with saving files. Current design is a little different to the default upload button. 
When you choose a set of files and click "open", these files will start uploading immediately, there is no "start" button (though it can be added). 
Moreover, when an upload is done, the item bar will not destroy. Instead, I added a "Hide" button for user.
1. Do you think if I need to follow the default upload's logic?
2. Should I just leave one button for upload?
updemo

takluyver:
From a UX point of view, if we do this, we'll definitely want to have just one upload button - users shouldn't have to decide whether an upload is 'big'. I'll leave it to Min and others whether this and/or #536 make sense.

daniel1124:
We have solved the problem of loading large files by chunking. 
I would be happy to contribute the code. Please let me know what's the process. Thank you!

PR #1509, opened 2016-06-03
Upload

nsknojj:
Add the big file upload function
Old pull request: [1501](https://github.com/jupyter/notebook/pull/1501)


minrk:
@nsknojj you can update pull requests by continuing to push commits to your branch, you do not need to open new pull requests with every change.

Following discussion from #1501, For 'big file' support, it should definitely be hooked up to the contents manager and use the same upload button that already exists. Uploading 'big' files shouldn't be different from smaller ones.

notebook/static/custom/custom.css
```diff
@@ -4,4 +4,66 @@ Placeholder for custom user CSS
 mainly to be overridden in profile/static/custom/custom.css
 
 This will always be an empty file in IPython
```
minrk: This file needs to stay empty. If there is more CSS to include in this page, it should be added to  `static/tree/less/tree.less`.

nsknojj:
I've modify the UI to be the same as the original upload.

notebook/static/tree/js/fileupload.js
```diff
@@ -0,0 +1,1462 @@
+/*
```
minrk: External dependencies should not be included in this repo. If it is needed, it should be added as a dependency in `bower.json` and loaded with `require`.

nsknojj:
I don't know how to pass the travis-ci build test while I add a new dependency "require(['jquery.fileupload'] )" in main.js

gnestor:
@nsknojj Anything I can do to help?

nsknojj:
@gnestor 
I tried to add big file upload feature using thie existing plugin "blueimp-file-upload" https://github.com/blueimp/jQuery-File-Upload
In this [commit](https://github.com/jupyter/notebook/pull/1509/commits/2d795ad90b1d277e22604ddbc072df3165d13a33), the travis ci build did not raise error, because I add the blueimp-file-upload javascript as an internal file instead of using bower to maintain the dependency.
Frankly, I'm not familiar with web, JS, CSS, BOWER and such things and I'm busy recently. If you want to continue completing this feature, you can make changes on my version, start your owns or whatever you can do. Thank you for attention and help!

Sorry for poor English.

gnestor:
> Following discussion from #1501, For 'big file' support, it should definitely be hooked up to the contents manager and use the same upload button that already exists. Uploading 'big' files shouldn't be different from smaller ones.

@nsknojj I suggest that we close this PR and if the need for this feature still exists, follow @minrk's suggestion and use the contents manager.

PR #1510, opened 2016-06-03
Fix widget warning in case of ipywidgets 5.x

### LINKED ISSUE #1377: Widgets 5.0 warns that to upgrade notebook to 4.2 even with 4.2 installed
I.e. I get
```
/usr/local/lib/python3.5/site-packages/widgetsnbextension/__init__.py:30: UserWarning: To use the jupyter-js-widgets nbextension, you'll need to update
    the Jupyter notebook to version 4.2 or later.
  the Jupyter notebook to version 4.2 or later.""")
```
yet:
```
/private/tmp pip3 list | egrep 'notebook|widgets'
ipywidgets (5.0.0)
notebook (4.2.0)
widgetsnbextension (1.0.0)
```

This seems to be because the code hack in notebookapp.py still calls `widgets.find_static_assets` even when widgets 5.0 is installed. 
Comparing https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L276
to https://github.com/jupyter/notebook/blob/4.x/notebook/notebookapp.py#L236 It seems like the HACK on the 4.x branch has not been fully backported from master?

SylvainCorlay:
Fixes #1377

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -236,15 +236,18 @@ class NotebookWebApplication(web.Application):
         # BEGIN HARDCODED WIDGETS HACK
         # TODO: Remove on notebook 5.0
         try:
-            import ipywidgets as widgets
-            handlers.append(
-                (r"/nbextensions/widgets/(.*)", FileFindHandler, {
-                    'path': widgets.find_static_assets(),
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! I think we can get 4.2.1 out, now.

PR #1511, opened 2016-06-03
changes for 4.2.1

minrk:
With this, I can release 4.2.1 this evening.

cc @SylvainCorlay, @willingc, @JamiesHQ

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -12,6 +12,21 @@ For more detailed information, see `GitHub <https://github.com/jupyter/notebook>
      upgrade to the latest release.
 
 
+.. _release-4.2.1:
+
+4.2.1
+-----
+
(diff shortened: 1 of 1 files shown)


takluyver:
:+1:

SylvainCorlay:
:+1:

willingc:
Looks good. Perhaps in a future release, we can start moving toward adding pr/issue numbers to the bullets.

Carreau:
things like #1490 (if the right fix)  should really go into notebook soon, as all the extensions since 4.2 are broken.

Carreau:
> things like #1490 (if the right fix) should really go into notebook soon, as all the extensions since 4.2 are broken.

And all non-python kernels have no highlighting.

minrk:
@Carreau good call. I'll cut 4.2.1 now, and we can do 4.2.2 with #1490 next week.

PR #1518, opened 2016-06-07
Allow requests to POST in OPTIONS requests

chultquist:
Commit 65eb248209b5fdc9d1d430c0403fc6bfee1057f4 introduced a bug by not allowing the POST method in OPTIONS requests (which is wrong since several parts of the API accept POST requests, for example creating a new kernel).


minrk:
Indeed, thanks!

chultquist:
No problem :) Thanks for the quick merge!

PR #1519, opened 2016-06-07
Add title attribute to image tag

vanillajonathan:


minrk:
Thanks!

vanillajonathan:
Perhaps it should be an `alt` attribute or both an `alt` and a `title` attribute.

Issue #1508, opened 2016-06-02
Unify behavior of server/nb extensions enabling, user vs sys.prefix


bollwyvl:
Users are encountering exciting conditions with the extension enabling behavior.

We encountered that [here](https://github.com/conda/conda/issues/2548) where we have basically had to say `rm -rf ~/.jupyter`.

The behavior of both the nbextensions and the serverextensions are surprising. 
### nbextensions: `user` and `sys-prefix` merge, but `sys-prefix` wins

```
$ jupyter nbextension enable thisthing --py --sys-prefix
$ jupyter nbextension enable thisotherthing --py --sys-prefix
$ jupyter nbextension disable thisthing --py
$ curl /api/config/notebook
{"this-thing/main": true, "this-other-thing/main": true}
```

```
$ jupyter nbextension enable this-thing --py
$ jupyter nbextension disable this-thing --py --sys-prefix
$ curl /api/config/notebook
{}
```
#### Good
- merging works (of enabling)
#### Bad
- user loses
#### potential fixes

``` python
# traitlets/config/manager.py
def recursive_update(target, new):
   ...
   if not target[k] is None: # was not updating with 'false' before
   ...

# notebook/services/config/manager.py
class ConfigManager(LoggingConfigurable):
    def get(self, section_name):
        for p in self.read_config_path[::-1]: # previously reading in reverse order
```
### serverextensions: `~/.jupyter` conquers all

```
$ jupyter serverextension enable this-thing --py --sys-prefix
$ jupyter serverextension enable this-other-thing --py --sys-prefix
$ jupyter serverextension disable this-thing --py
$ jupyter notebook
Collisions detected in jupyter_notebook_config.py and jupyter_notebook_config.json config files. jupyter_notebook_config.json has higher priority: {
      "NotebookApp": {
        "nbserver_extensions": "{'this-thing': True, 'this-other-thing': True} ignored, using {'this-thing': False}"
      }
    }
```
### Good
- user wins!
### Bad
- no merge
#### potential fixes

``` python
# notebook/services/config/manager.py
def _ensure_subconfig(self):
   ...
   if isinstance(obj, dict) \ # _is_section_key(key)
                    and not isinstance(obj, Config):
```

ellisonbg:
Do you propose that both follow:

* Merge
* User wins

minrk:
[...]

PR #1526, opened 2016-06-10
reverse nbconfig load order

minrk:
list is in descending priority, so load should iterate back to front, to ensure user config wins

the easy part of #1508

### DIFF
--- a/notebook/services/config/manager.py
+++ b/notebook/services/config/manager.py
@@ -19,11 +19,12 @@ class ConfigManager(LoggingConfigurable):
     def get(self, section_name):
         """Get the config from all config sections."""
         config = {}
-        for p in self.read_config_path:
+        # step through back to front, to ensure front of the list is top priority
+        for p in self.read_config_path[::-1]:
             cm = BaseJSONConfigManager(config_dir=p)
             recursive_update(config, cm.get(section_name))
         return config
-            
+
     def set(self, section_name, data):
         """Set the config only to the user's config."""
         return self.write_config_manager.set(section_name, data)


lgfa29:
@minrk I have a `~/.jupyter/custom/custom.js` file that listens for the `app_initialized.NotebookApp` event and do some stuffs, like so:

```js
define(['base/js/events'], function(events) {
    events.on('app_initialized.NotebookApp', function() {
        ...
    });
});
```

After upgrading `notebook` to `4.2.2` the event handler is not being called anymore and I am not sure why. My thinking is that by reversing the order the nbconfig is loaded, the `app_initialized.NotebookApp` event is being fired before my `custom.js` code is loaded.

Does this make sense? Is there a better place to listen for `app_initialized.NotebookApp`?

minrk:
nbconfig shouldn't affect the custom.js load, it's a separate mechanism. Plus, the load order only affects when there are conflicts, which there shouldn't be on custom.js. It's possible that the app_initialized event has already fired when your module loads. Could you wait for `notebook_loaded.Notebook` instead?

lgfa29:
Sorry, I forgot to mention that I did try to listen for `notebook_loaded.Notebook`, but it still didn't work. 

I saw [this post](https://groups.google.com/forum/#!searchin/jupyter/app_initialized.NotebookApp%7Csort:relevance/jupyter/5F2gVEyvOVQ/JtQIRsJbDAAJ) which seems to be the same problem, so something on `4.2.2` changed how these events work on `custom.js`.

I will open an issue to discuss it further. Thanks for the help!

PR #1527, opened 2016-06-10
Update traitlets API

SylvainCorlay:
@Carreau @minrk would you be ok with me adding an intermediate base class called JupyterServerApp between NotebookApp and JupyterApp.

NotebookApp would only have what is specific to the current notebook, and all the rest would be moved up.

The goal is to be able to inherit from this JupyterServerApp in JupyterLabApp eventually.


SylvainCorlay:
@Carreau @minrk  would you be ok with me adding an intermediate base class called `JupyterServerApp` between `NotebookApp`  and `JupyterApp`.

`NotebookApp` would only have what is specific to the current notebook, and all the rest would be moved up.

The goal is to be able to inherit from this `JupyterServerApp` in `JupyterLabApp` eventually.

minrk:
Something like that, though we also want to be able to run an application that only provides the kernel service and not notebooks for things like the kernel gateway. We've packaged up the various endpoints into subdirs with this in mind, so maybe while you are putting together JupyterServer, make the list of services to load something that can be changed.

SylvainCorlay:
Yeah, It looks like most of it would go up. Certain things like

 - static directories
 - enabling pylab
 - deprecated APIs

could remain in `NotebookApp`...

minrk:
Yup

PR #1547, opened 2016-06-15
Load codemirror as a script tag

### LINKED ISSUE #1528: Notebook terminal broken
In recent builds of the notebook, the making a new terminal window fails.
Nothing is shown in the browser window, and the javascript console gives the error 

``require.js:140 Uncaught Error: Module name "codemirror/lib/codemirror" has not been loaded yet for context: _. Use require([])``

01ffff7 does not work, a5ad979 is working.

blink1073:
Fixes #1528

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -21,6 +21,8 @@
     <script src="{{static_url("components/bootstrap/js/bootstrap.min.js") }}" type="text/javascript" charset="utf-8"></script> <!-- extends window.$ -->
     <script src="{{static_url("components/bootstrap-tour/build/js/bootstrap-tour.min.js") }}" type="text/javascript" charset="utf-8"></script> <!-- window.Tour -->
     <script src="{{static_url("components/jquery-typeahead/dist/jquery.typeahead.min.js") }}" type="text/javascript" charset="utf-8"></script> <!-- extends window.$ -->
+    <script src="{{static_url("components/codemirror/lib/codemirror.js") }}" type="text/javascript" charset="utf-8"></script>
+    <script src="{{static_url("components/codemirror/mode/meta.js") }}" type="text/javascript" charset="utf-8"></script>
     <script src="{{static_url("components/requirejs/require.js") }}" type="text/javascript" charset="utf-8"></script>
     <script>
       window['staticURL'] = "{{static_url("", include_version=False)}}";
@@ -178,6 +180,19 @@
       define("jquery-ui", function () {
           return window.$;
       });
+
+      define("codemirror/lib/codemirror", function () {
+          return window.CodeMirror;
+      });
+
+      define("codemirror/mode/meta", function () {
+          return window.CodeMirror;
+      });
+
+      define("CodeMirror", function () {
+          return window.CodeMirror;
+      });
+
     </script>
 
     {% block meta %}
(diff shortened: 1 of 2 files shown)


blink1073:
This is not the final solution to the CodeMirror problem, but we will need this as part of it, and it allows the terminal to work.

blink1073:
I verified that syntax highlighting works for python and javascript, and that terminals work.  CodeMirror modes are loaded using requirejs explicitly.  Anything in `addon/` that is loaded by both Webpack and RequireJS will still affect the global `CodeMirror` instance, but would re-evaluate that code and state would be lost.

blink1073:
cc @SylvainCorlay @jasongrout

blink1073:
The following addons are loaded by WebPack currently:

```
edit/matchbrackets
edit/closebrackets
comment/comment
dialog/dialog
search/searchcursor
runmode/runmode
```

blink1073:
Based on further experimentation, the `addons` and `keymap` need to be explicitly loaded as script tags, and could be moved from the source files into the appropriate html templates.

minrk:
Why do you say that they need to be loaded as script tags? Is this because of webpack? They worked just fine with requirejs while we didn't use webpack.

blink1073:
I mean that right now the find/replace does not work in the editor page, but if I add the `addon` script tags, it does work.  I have not yet nailed down what that is.

blink1073:
The `addon/search` code is in the Webpack bundle, but it is not affecting the global `CodeMirror` instance.

blink1073:
Ah, there is *another* copy of `codemirror.js` in the Webpack bundle as well, picked up by relative paths in the `addon` files I imagine.

blink1073:
Yep, `../../lib/codemirror` calls that bundled version.

blink1073:
Fixed!  Success!!!!

blink1073:
screen shot 2016-06-16 at 6 53 41 am

blink1073:
screen shot 2016-06-16 at 6 54 45 am

blink1073:
And...

screen shot 2016-06-16 at 6 58 44 am

minrk:
Gotcha, thanks. So if people `require('codemirror')` now, whether in our webpacked files or at runtime, everything is AOK and there's just one CodeMirror module on the page?

minrk:
This seems related to the issues with Cell, where if one uses require to load the `cell` module and set global configuration, it's actually loading a different copy.

Is it impossible to bundle things with webpack such that `require` at runtime finds the webpacked module? If so, that seems like a severe problem, to the point that maybe we shouldn't be using webpack in the 'legacy' notebook.

blink1073:
Right, I verified that there is only one bundle.  I just added the ability to `require('codemirror')` no path qualifier.

blink1073:
The two are fundamentally at odds, yes.  This is a wider discussion that we need to have as a group.

blink1073:
I'd vote to use all `RequireJS` in the existing notebook and all `Webpack` in JupterLab.

minrk:
Thanks, that's my inclination, as well. I think with the clearer separation we have now with jupyterlab, it will be easier to 'stick with the old way' here, and adopt new tools over there.

minrk:
Merging here, as it fixes a severe issue, and we can figure out the bigger issue later on.

PR #1549, opened 2016-06-15
Allow completion picking by clicking option.

### LINKED ISSUE #1512: Can't click to select tab completion option
Does the tab completion feature of Jupyter use CodeMirror's Autocomplete? If so, why doesn't clicking on an option in the Jupyter tab completion dropdown work as it does in the [CodeMirror demo](https://codemirror.net/demo/xmlcomplete.html)?

natalieperna:
Resolve #1512.

The `focus` function was shadowing the `dblclick` function (which, probably at one time, allowed picking completion options by double clicking). I don't think the `focus` function was necessary, as clicking anywhere outside the selection box (i.e. in the editor), still changes focus just fine (at least for me on Chrome).

I also changed `dblclick` to `click`, since I think this is the expected behaviour for select widgets, opinions?

click-select

### DIFF
--- a/notebook/static/notebook/js/completer.js
+++ b/notebook/static/notebook/js/completer.js
@@ -255,6 +255,7 @@ define([
             var that = this;
             this.sel.click(function () {
                 that.pick();
+                that.editor.focus();
             });
             this._handle_keydown = function (cm, event) {
                 that.keydown(event);
--- a/notebook/static/notebook/js/completer.js
+++ b/notebook/static/notebook/js/completer.js
@@ -256,9 +256,6 @@ define([
             this.sel.dblclick(function () {
                 that.pick();
             });
-            this.sel.focus(function () {
-                that.editor.focus();
-            });
             this._handle_keydown = function (cm, event) {
                 that.keydown(event);
             };
--- a/notebook/static/notebook/js/completer.js
+++ b/notebook/static/notebook/js/completer.js
@@ -253,7 +253,7 @@ define([
 
             //build the container
             var that = this;
-            this.sel.dblclick(function () {
+            this.sel.click(function () {
                 that.pick();
             });
             this._handle_keydown = function (cm, event) {


notebook/static/notebook/js/completer.js
```diff
@@ -253,12 +253,9 @@ define([
 
             //build the container
             var that = this;
-            this.sel.dblclick(function () {
+            this.sel.click(function () {
                 that.pick();
             });
-            this.sel.focus(function () {
-                that.editor.focus();
-            });
```
Carreau: With all the focus problem we had I would be **extremely** worry of removing something like. 
a few questions: 

On linux/ windows you might need to click on the menu to be able to mouse scroll – does that still work, or can you, as in mac of, scroll even without focusing ? 

With your patch after clicking a completion the notebook is in an inconsistent state. Keybinding are not working correctly, until a reach the mouse and click in a cell, and even though menu can get stuck.


cantclose

That tend to hint toward the fact that `that.editor.focus()` is here on purpose.

Carreau:
Thanks a lot for investigating that ! It is a super brittle piece of code though, which is years old.  we have to be careful with what we are doing. 

Codemirror now [ships with an autocompleter](http://codemirror.net/demo/complete.html) , it might be worth investigating using that, instead of trying to fix the behavior of our clucky taped togeter completer.

natalieperna:
> On linux/ windows you might need to click on the menu to be able to mouse scroll – does that still work, or can you, as in mac of, scroll even without focusing ?

I think it's actually the other way around. Without this change, you cannot click the scroll wheel (at least in Chrome on Linux). Focus will immediately switch to the editor and the completion box will disappear. With this patch, you can click-and-drag the scroll wheel.
 
> With your patch after clicking a completion the notebook is in an inconsistent state. Keybinding are not working correctly, until a reach the mouse and click in a cell, and even though menu can get stuck.

I'm not sure how to check for an inconsistent state, but I've added `that.editor.focus();` after `that.pick();` and it appears to fix the other issues.

I understand if this is too much duct tape on duct tape to feel safe to include though.

minrk:
Thanks. From the code, I'm guessing the original intention was click to focus, so one could select without applying the completion, then double-click to apply it. I think this PR is an improvement for me, though, so I'm inclined to merge as-is.

In general, I agree that if someone wants to put in the time to send our completions to CodeMirror's own completion widget, that should be a big improvement.

natalieperna:
In the meantime, if we want this patch to our Jupyter version, is it possible to modify this through our `kernel.js` (if so, could you give me some suggestion how), or would we need to maintain our own patched version? There are also some other changes we want to make to the Completer that probably wouldn't be of interest to Jupyter core, so I'm curious whether this is possible.

> In general, I agree that if someone wants to put in the time to send our completions to CodeMirror's own completion widget, that should be a big improvement.

Is it possible to replace the Jupyter custom completion code with CodeMirror's completion widget without upgrading all of the CodeMirror code to the current version? The former sounds like something I might be able to do, but the latter, probably not.

takluyver:
> Is it possible to replace the Jupyter custom completion code with CodeMirror's completion widget without upgrading all of the CodeMirror code to the current version?

I think the version of Codemirror we already include has the completion widget, we're just not using it. We have (in master) 5.14, which is only a couple of months old.

https://github.com/jupyter/notebook/blob/c9a7e9df13eb4bbc888450e6d773815f97a52ad1/bower.json#L8

Carreau:
There is a small inconsistency possible if you click on the scrollbar, and then on one menu. But it appears like it can't persist. 

inc

I'll be with min on this one, it seem like an improvement, so let's try it, and if there is ever a critical think we did not think about, we can always revert. 

For your other question, yes you can probably ship things with your `kernel.js` but you will likely need to use private undocumented API and that will be something brittle. 

Merging ! 

Thanks ! Welcome to the party !

PR #1552, opened 2016-06-16
avoid clobbering ssl_options.ssl_version

### LINKED ISSUE #1028: Overriding ssl_options protocol and forcing TLSv1
Current code in [notebook/notebookapp.py#init_webapp](https://github.com/jupyter/notebook/blob/807f4501a2c507ee99076e9c724852519ca74412/notebook/notebookapp.py#L878) overrides user setting of ssl protocol version "ssl_version" and forces it to be TLSv1, so whatever the user configures in the jupyter_notebook_config file for this field will be ignored, and also some browsers like Chrome will say the connection uses an "obsolete cipher" when looking into the connection details.

A better approach for the whole ssl_context handling could be taking advantage of  an SSLContext object created through [ssl.create_default_context(*)](https://docs.python.org/3/library/ssl.html#ssl.create_default_context).
This function already provides safer defaults, which allows for protocol negotiation according to client/server capabilities while also declining SSLv2 and SSLv3. 

If it's not possible to use, due to SSLContext being on Python3.2+ and tornado that only allows a frozenset of keywords in the ssl_options dict, then it could be manually set to the user given value, as long as it is greater than [PROTOCOL_SSLV23](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_SSLv23), which has a value of 2, so the more modern protocols will always have a greater value, like TLSv1_2 which is 5.

This last suggestion would be a simple and extensible fix. 
I'd be happy to submit a PR with such, after hearing your feedback on the issue.

minrk:
use setdefault instead of setting directly, so that the user can override these values.

closes #1028

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1020,10 +1020,10 @@ class NotebookApp(JupyterApp):
         else:
             # SSL may be missing, so only import it if it's to be used
             import ssl
-            # Disable SSLv3, since its use is discouraged.
-            ssl_options['ssl_version'] = ssl.PROTOCOL_TLSv1
+            # Disable SSLv3 by default, since its use is discouraged.
+            ssl_options.setdefault('ssl_version', ssl.PROTOCOL_TLSv1)
             if ssl_options.get('ca_certs', False):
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks !

PR #1553, opened 2016-06-17
Remove the dockerfile.

Carreau:
We have too many issues related to the dockerimage.

Closes Many issues.


parente:
I take it this means we can also axe the https://hub.docker.com/r/jupyter/ubuntu_14_04_locale_fix/ base image (from https://github.com/jupyter/docker-stacks/tree/master/internal/ubuntu_14_04_locale_fix) which only existed (I think) to support the jupyter/notebook image. Or is / was it being used for anything else?

jakirkham:
Must have missed this one. In any event 👍. I sort of got the feeling we were going this way anyways. docker-stacks is just a better curated version. Couple notes though.

* Should we add a note in the Readme?
* Also there seems to be some other Docker stuff we missed (e.g. `.dockerignore`)

Issue #1546, opened 2016-06-14
Many tests leak temporary resources if a setup/setupclass fails.


ssanderson:
I'm working on adding encryption/decryption routines to [pgcontents](https://github.com/quantopian/pgcontents), which means I'm breaking all my tests pretty regularly.

One of the things I've noticed in the process is that many of the contents API tests (which I subclass to ensure that I have a conforming implementation) leak temporary state (in particular, temporary directories) if a `setUp` test method fails.  

This happens because `unittest` doesn't call a test suite's `tearDown` method if its `setUp` method fails, so if we get halfway through a `setUp` and fail, we don't clean up those resources in `tearDown`.  For per-method resources, the right way to fix this is generally to use [addCleanup](https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.addCleanup) to register a destructor for each allocated resource (functions registered with `addCleanup` are guaranteed to be called even if `setUp` fails).

Unfortunately, there's no good equivalent to `addCleanup` for resources created in an `setupClass`.  Correct handling of class-level resources generally involves manually maintaining a stack of destructor callbacks and ensuring that they're fired properly even if setupClass fails.  (We have a pretty intricate [example of this](https://github.com/quantopian/zipline/blob/master/zipline/testing/fixtures.py#L44-L191) over in Zipline.)

I'd like to at least update some of the `setUp` methods to use `addCleanup`.  Whether or not the project has the test-refactoring stomach to do fix usages of `setUpClass` is a more interesting question.

At any rate, I figured I should open an issue before I go through and refactor a whole bunch of tests.

minrk:
Sounds great! Please do make a PR when you are ready.

ssanderson:
PR for all the instance-level cleanups is here: https://github.com/jupyter/notebook/pull/1557.

takluyver:
#1557 was merged. Do we want to try to tackle the class-level cleanups as well? Or shall we close this?

PR #1557, opened 2016-06-18
Use addCleanup instead of tearDown in several test suites.

ssanderson:
xref https://github.com/jupyter/notebook/issues/1546


Carreau:
Thanks !

BTW, I was sad you could not make it in time for your talk at PyCon :-P

ssanderson:
@Carreau thanks!  We're a make-up at PyGotham in a few weeks, but I have a feeling we'll have some scheduling issues at the last minute ;).

Carreau:
> We're doing a make-up at PyGotham in a few weeks, but I have a feeling we'll have some scheduling issues at the last minute ;).

Well, at least you won't have to improvise your talk in front of a blank notebook !

PR #1561, opened 2016-06-22
[Post save hook] Raise web.HTTPError on fail

Fitz3012:
Warn the user of a failure on save.
Allow post_save_hook function to send exceptions to front


HenriJ:
We use the post_save_hook function to commit the notebook to a central notebook repository. Silent failure as it works today could cause data loss...
This PR would help us at least display to the user that the save failed

minrk:
Merging since I think this is a good idea, but I am a little concerned about the error indicating that the save failed, which is not true–a *side effect* of the save has failed.

PR #1565, opened 2016-06-24
MathJax MML support

poke1024:
Allows kernels to provide math output as MathML (MML) in addition to TeX by changing the default MathJax configuration from `TeX-AMS_HTML-full` to `TeX-AMS-MML_HTMLorMML-full`.

This allows the MathML-bound https://github.com/mathics/IMathics kernel to render its math output properly.

Details of the new and old MathJax configurations are available at http://docs.mathjax.org/en/latest/config-files.html.

This is a follow-up of #1456.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -161,7 +161,7 @@ class IPythonHandler(AuthenticatedHandler):
     
     @property
     def mathjax_config(self):
-        return self.settings.get('mathjax_config', 'TeX-AMS_HTML-full,Safe')
+        return self.settings.get('mathjax_config', 'TeX-AMS-MML_HTMLorMML-full,Safe')
 
     @property
     def base_url(self):
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -762,7 +762,7 @@ class NotebookApp(JupyterApp):
         else:
             self.log.info("Using MathJax: %s", new)
 
-    mathjax_config = Unicode("TeX-AMS_HTML-full,Safe", config=True,
+    mathjax_config = Unicode("TeX-AMS-MML_HTMLorMML-full,Safe", config=True,
         help="""The MathJax.js configuration file that is to be used."""
     )
 
--- a/setupbase.py
+++ b/setupbase.py
@@ -167,7 +167,7 @@ def find_package_data():
     mj = lambda *path: pjoin(components, 'MathJax', *path)
     static_data.extend([
         mj('MathJax.js'),
-        mj('config', 'TeX-AMS_HTML-full.js'),
+        mj('config', 'TeX-AMS-MML_HTMLorMML-full.js'),
         mj('config', 'Safe.js'),
     ])
     


takluyver:
For that matter, would it be useful to add a mathml mime type so kernels can send MathML directly, rather than going through Mathjax to create it?

rgbkrk:
> would it be useful to add a mathml mime type so kernels can send MathML directly, rather than going through Mathjax to create it?

It's been an ISO standard since last year though only supported fully on Firefox. I'm on board with pushing for this standard.

What JS libraries are available to render MathML? The only one I've found so far is MathJax.

minrk:
> What JS libraries are available to render MathML? The only one I've found so far is MathJax.

I think the answer to this is either MathJax, or let the browser do it (Firefox only, I think).

gnestor:
@takluyver Does this MathJax config depend on a MathML mime type or can we merge this and create a new issue regarding MathML mime type?

takluyver:
I'm not exactly sure about this, but I think it's orthogonal to my suggestion - I think this is about Mathjax rendering latex-y maths to MathML, whereas I'm talking about the kernel publishing MathML directly.

poke1024:
Can second that this is about MathJax configuration, namely MathJax being able to render MathML in addition to the currently configured TeX support. In Jupyter's current MathJax configuration, MathJax does not support MathML input at all.

As @rgbkrk noted, kernels wanting to use MathML directly can only do this on Firefox. Most browsers (including Chrome, which had MathML support but killed it some time ago) do not support MathML directly (see the MathML column under https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Web_technology_support) and thus need MathJax to convert MathML to something they can render (e.g. CSS and HTML). Even on Firefox, MathJax may add certain MathML features that Firefox's native MathML renderer does not support (see https://www.mathjax.org/mathjax-2-0-and-the-default-rendering-in-firefox/).

So, in short, Jupyter kernels wanting to output MathML need the change proposed in this PR.

takluyver:
OK, so this is about MathML input to MJ? I had misunderstood, then, I thought it was about MathML output.

Presumably the idea is that kernels would produce MathML embedded in HTML output? Do you think it would be helpful to add a MathML output type (akin to SVG output, which we already allow)?

poke1024:
Exactly, the kernel outputs MathML output as `<math>...</math>` embedded in HTML (see for example https://developer.mozilla.org/en/docs/Web/MathML/Element/math). I'm not sure how Jupyter output types work, but as MathML is an HTML element like `<svg>` it sounds quite logical to do the same thing as with SVG. Are there some docs I could look at?

takluyver:
We output a dictionary of alternative formats, keyed by mime types (see [display_data](http://jupyter-client.readthedocs.io/en/latest/messaging.html#display-data) messages). As far as I know, we don't explicitly prevent sending any format in there, but the frontend only handles formats it recognises, and we don't currently recognise MathML. IPython also provides convenient mechanisms to display the formats Jupyter recognises.

So what you're presumably doing at the moment is sending `text/html` output with embedded MathML. That's fine, but it would be neat if some time in the future we handled a native MathML output format.

takluyver:
I've created #1670 for taht.

gnestor:
Ok, to summarize: this PR simply updates MathJax config to accept MathML input in addition to TeX. Unless there are any objections, I'd like to merge this. @takluyver created https://github.com/jupyter/notebook/issues/1670 to discuss adding a MathML mime type and handler so that MathJax can not only accept `text/html` mime type but also `application/mathml+xml`.

rgbkrk:
I'm certainly in favor. This doesn't have to change any mimetypes at this time and we can handle it later.

gnestor:
Thanks @poke1024 for your contribution!

poke1024:
Changing `Math` to `HTML` makes it work for me using Jupyter 5/master:

```
from IPython.core.display import HTML
HTML(r"""<math mode="display" xmlns="http://www.w3.org/1998/Math/MathML">
 <semantics>
  <mrow>
    <mi>x</mi>
    <mo>=</mo>
    <mfrac>
      <mrow>
        <mo form="prefix">&#x2212;<!-- − --></mo>
        <mi>b</mi>
        <mo>&#x00B1;<!-- &PlusMinus; --></mo>
        <msqrt>
          <msup>
            <mi>b</mi>
            <mn>2</mn>
          </msup>
          <mo>&#x2212;<!-- − --></mo>
          <mn>4</mn>
          <mo>&#x2062;<!-- &InvisibleTimes; --></mo>
          <mi>a</mi>
          <mo>&#x2062;<!-- &InvisibleTimes; --></mo>
          <mi>c</mi>
        </msqrt>
      </mrow>
      <mrow>
        <mn>2</mn>
        <mo>&#x2062;<!-- &InvisibleTimes; --></mo>
        <mi>a</mi>
      </mrow>
    </mfrac>
  </mrow>
  <annotation encoding="TeX">
     x=\frac{-b\pm\sqrt{b^2-4ac}}{2a}
  </annotation>
  <annotation encoding="StarMath 5.0">
     x={-b plusminus sqrt {b^2 - 4 ac}} over {2 a}
  </annotation>
 </semantics>
</math>""")
```

I'm not sure what `IPython.display.Math` is supposed to do, but it doesn't seem to pass on the XML code MathJax then operates on.

The other example above (using `<apply>`) does not work as `<apply>` and the other elements are content markup. MathML differentiates between presentation markup (https://www.w3.org/TR/MathML3/chapter3.html) and content markup (https://www.w3.org/TR/MathML3/chapter4.html) and MathJax, as far as I know, currently focuses on supporting presentation markup. A short description on the two markup styles can be found here: https://developer.mozilla.org/de/docs/Web/MathML/Element/semantics. That's why `<semantics>` is optional and you can also pass in something like:

```
<math mode="display" xmlns="http://www.w3.org/1998/Math/MathML">
  <mrow>
        <msqrt>
          <msup>
            <mi>b</mi>
            <mn>2</mn>
          </msup>
        </msqrt>
  </mrow>
</math>
```

This is `<semantics>` content and understood as such by MathJax, but not tagged as such.

takluyver:
The `Math` display object expects Latex code passed into it, so things like `<` will be treated as characters, not as HTML syntax. We may want to add an option to that to treat the input as MathML rather than Latex (like the `Image` class can handle different image formats).

gnestor:
@poke1024 Thanks, that worked for me. Regarding Content MathML, I added support for it and AsciiMath to MathJax in https://github.com/jupyter/notebook/pull/1677.

@takluyver Next I will add formatters and displayers for MathML and AsciiMathto to ipython. It think we need to consider whether the `IPython.display.Math` will accept LaTeX, MathML, and AsciiMath or whether should just require users to use separate `MathML` and `AsciiMath` methods. I will create a new issue for this on ipython and link to it here...

Issue #1572, opened 2016-06-28
Text output processing needs to handle backspace


takluyver:
Originally reported as takluyver/bash_kernel#50, but it affects other kernels - e.g. run `!man ls` in an IPython notebook.

Certain programs (like `man`) appear to use an overwriting system for simple formatting, so `A<BS>A` makes a bold A and `_<BS>A` makes an underlined A. Where programs don't understand this, the second character replaces the first and you see unformatted text. However, we don't handle backspaces at all, so we show `AA` and `_A`.

This should go along with our handling of ANSI escapes and carriage returns.

PR #1573, opened 2016-06-28
Process backspace characters in output

takluyver:
These are like carriage return, but only affect a character before them instead of the whole line. I've checked that this makes the output from `man ls` look OK.

Closes gh-1572


minrk:
Cool! Is a test easy enough to add?

takluyver:
Voila, tests.

minrk:
🚀

PR #1577, opened 2016-06-29
Fix #1532

### LINKED ISSUE #1532: Directory listing should not show hidden or system files
Hello,

if a file is hidden (or is system file) by operating system in OS's file browser, it should be hidden also in Jupyter's file / directory listing.

This happens at least on my Windows 8 system, when I launch notebook and go to browser, it shows hidden and system files in the listing.

mkoske:
See discussion on #1532.


takluyver:
Thanks, this looks good to me.

PR #1579, opened 2016-06-30
Bundler extension API

parente:
For #1469

* [x] port existing code from jupyter_cms
* [x] switching to using content manager instead of absolute file paths
* [x] make API resources consistent with /nbconvert
* [x] make menu item ordering consistent 
* [x] add cli option for listing bundles
* [x] add zip bundler as example
* [x] port bundler tests
* [x] add bundler extension management tests
* [x] port documentation
* [x] decide if all bundler tools should remain here (I think yes, as examples, along with zip_bundler.py showing how to use them)


notebook/bundlerextensions.py
```diff
@@ -0,0 +1,259 @@
+# Copyright (c) Jupyter Development Team.
```
minrk: It might be nice to put this file inside notebook/bundler so it's all self contained.

scripts/jupyter-bundler
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from notebook.bundlerextensions import main
```
minrk: It would probably also be good for this entrypoint to be accessible in `notebook.bundler.__init__`, so that one can also launch this with `python -m notebook.bundler`

setup.py
```diff
@@ -180,6 +180,7 @@
             'jupyter-notebook = notebook.notebookapp:main',
             'jupyter-nbextension = notebook.nbextensions:main',
             'jupyter-serverextension = notebook.serverextensions:main',
+            'jupyter-bundler = notebook.bundlerextensions:main',
```
minrk: General question: `jupyter bundler` or `jupyter notebook bundler`?

parente: I was staying consistent with `jupyter nbextension` and `jupyter serverextension` which both affect the notebook alone but are top level commands. I don't mind making it a subcommand of notebook but might need a pointer on how that would work. It's not simply renaming the script `jupyter-notebook-bundler`, right?

minrk: Subcommands of notebook go through the earlier traitlets.config subcommand system. See the `subcommands` attribute on NotebookApp, where there should already be at least one example.

I'm not sure that it should change, but `bundler` seemed a little generic to be top-level, so I thought I'd ask.

parente: I think I prefer `jupyter <splat>` for consistency with the others. But we can definitely change the name from `bundler` to `bundlerextension` or something more descriptive.

parente:
@minrk Do you think this is reasonable to target for 4.3 instead of 5.0? There's only new API, no backward incompatibility. It would help us move the dashboard bundlers away from their dependency on another extension (jupyter_cms) sooner rather than later.

minrk:
It could be for 4.3, if we do that. We may be at the point of only doing bugfix backports for 4.x at this point, though.

parente:
For the docs, I'm thinking of adding a new `docs/source/extending/bundlerextensions.rst` and expanding the content in `docs/source/examples/Notebook/Distributing Jupyter Extensions as Python Packages.ipynb` to also mention what needs to be done to distribute bundler extensions.

Does this sound reasonable?

minrk:
Sounds good to me

parente:
I've checked off everything I had planned on my list here. There's examples, tests, doc, and the implementation itself.

willingc:
@parente Looks good to me. Thanks!

parente:
I think this is ready for a final review / merge.

Adding @Carreau and @ellisonbg who commented on the original associated issue and cc'ed @fperez @jasongrout.

torsstei:
It would be desirable to move this pull request towards merge soon. We have an exploitation of the cms (https://www.youtube.com/watch?v=xSRZgxAf9BQ) and a regular notebook core API would be what we would prefer.

gnestor:
@parente I'll review this. A couple questions:

* How can I get the `bundlerextension` command working? I pulled this PR branch and re-installed using `pip install -e . --user`...
* Once the `bundlerextension` extension is working, can I test the zip bundler using `jupyter bundlerextension enable --py notebook.bundler.zip_bundler --sys-prefix`?
* Anything else I should test?

parente:
@gnestor Thanks for having a look!

> How can I get the bundlerextension command working? I pulled this PR branch and re-installed using pip install -e . --user...

`jupyter bundlerextension` should just work. Are you saying it doesn't? I dev'ed locally by pip installing into a conda environment. Is `--user` breaking it?

> Once the bundlerextension extension is working, can I test the zip bundler using jupyter bundlerextension enable --py notebook.bundler.zip_bundler --sys-prefix?

This is definitely a good place to start. There's a tarball bundler in there too that you can try also.

> Anything else I should test?

https://github.com/jupyter-incubator/dashboards_bundlers/pull/46 makes the dashboards bundlers extension compatible with this new API. If you're feeling brave, you might try testing that.It'll take a bit of doing, namely getting the dashboard extension installed, the bundler extension built with the PR installed, and the dashboard server running to see if the bundler properly packs and sends one of the example notebook-dashboards to the dashboard server properly. If that's too much overhead (and it probably is), don't sweat it. I did test it once already when creating that PR. If this merges in, I'll test it against notebook HEAD.

gnestor:
> jupyter bundlerextension should just work. Are you saying it doesn't? I dev'ed locally by pip installing into a conda environment. Is --user breaking it?

Yep, the `--user` flag was the culprit.

zip and tarball bundlers worked for me! It took about 20-30 seconds for the first zip to fulfill. Each subsequent one was instant, along with the tarball.

gnestor:
@minrk How do you feel about merging this?

minrk:
@gnestor 👍

parente:
@minrk  @gnestor Not sure why `--user` didn't work. I can look into that as follow-on. Thanks for testing and merging!

gnestor:
@parente Ya please do because it could be my environment...

parente:
@gnestor I pulled master just now and ran:

```
cd notebook
source activate notebook-dev
pip install -e .
jupyter bundlerextension enable --py notebook.bundler.zip_bundler --user
jupyter notebook
```

Then I opened a notebook and triggered the bundler menu item. No issues.

I next tried the following outside a conda env:

```
cd notebook
pip install -e . --user
~/.local/bin/jupyter bundlerextension enable --py notebook.bundler.zip_bundler --user
~/.local/bin/jupyter notebook
```

No issues again.

gnestor:
@parente Ok good. Great work, this is an exciting addition! Any *bundle as Gist* bundlers yet? If not, that could be my first bundler contribution...

Issue #104, opened 2015-05-21
Paste into Jupyter terminal?


sjobeek:
Is there any way to paste text into the Jupyter terminal?   

On chrome I'm able to copy from the terminal no problem, but can't figure out how to get text in.

The terminal is quite handy, but being unable to paste in snippets of text makes it difficult to execute larger commands.  I may just not have been able to figure out how to do it.

rgbkrk:
I've noticed that I have to right click inside the little term js widget to get copy paste to work. That definitely needs a fix. Thanks for reporting it!

takluyver:
See also chjj/term.js#37 and chjj/term.js#51.

cboettig:
Looks like this has been resolved, yet I'm still not able to paste into the Jupyter terminal (in Ubuntu/Firefox).  I'm running jupyter via your current jupyter/datascience-notebook docker image, seems like it should be recent enough?  Any hints on debugging this?

yuvipanda:
@cboettig I don't think this is in the current released version, is set to milestone 4.1 which hasn't been released yet.

yuvipanda:
I can still repro this with the 4.1rc1 :(

yuvipanda:
Can someone re-open this? This is still happening...

minrk:
Some further clarification: The bug preventing paste in *all browsers* is fixed in notebook 4.1 (term.js 0.0.7). The bug preventing Firefox from pasting is apparently separate (chjj/term.js#51), and fixed in term.js master, but hasn't seen a release. There's nothing for us to do but wait for term.js 0.0.8.

yuvipanda:
Actually, the fix hasn't been merged into master, and term.js development seems to have stopped :(

aggFTW:
I'm not sure if this is a dumb question or not, but how are you getting pasting to work? `Ctrl+V` or `right click` don't work.

My Jupyter installation is:
```
The version of the notebook server is 4.2.0 and is running on:
Python 2.7.11 |Anaconda 2.3.0 (64-bit)| (default, Dec  6 2015, 18:08:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
```

takluyver:
@aggFTW Are you using Firefox? I think the fix for that is still not released.

aggFTW:
I'm using Chrome.

smokedsalmonbagel:
[...]

PR #1591, opened 2016-07-05
Switch to xterm for the terminal emulator

blink1073:
Xterm is actively maintained and is being used in Jupyterlab.

cf https://github.com/jupyter/notebook/issues/104#issuecomment-230443710

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   "dependencies": {
     "moment": "^2.8.4",
     "preact": "^4.5.1",
-    "preact-compat": "^1.7.0"
+    "preact-compat": "^1.7.0",
+    "xterm": "^0.33.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -27,9 +27,12 @@
     "babel-preset-es2015": "^6.6.0",
     "bower": "*",
     "concurrently": "^1.0.0",
+    "css-loader": "^0.23.1",
     "eslint": "^2.8.0",
+    "json-loader": "^0.5.4",
     "less": "~2",
     "requirejs": "^2.1.17",
+    "style-loader": "^0.13.1",
     "underscore": "^1.8.3",
     "webpack": "^1.12.13"
   },
--- a/setupbase.py
+++ b/setupbase.py
@@ -153,7 +153,6 @@ def find_package_data():
         pjoin(components, "underscore", "underscore-min.js"),
         pjoin(components, "moment", "moment.js"),
         pjoin(components, "moment", "min", "moment.min.js"),
-        pjoin(components, "term.js", "src", "term.js"),
         pjoin(components, "text-encoding", "lib", "encoding.js"),
     ])
 
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,6 @@
     "MathJax": "components/MathJax#~2.6",
     "moment": "~2.8.4",
     "requirejs": "~2.1",
-    "term.js": "chjj/term.js#~0.0.7",
     "text-encoding": "~0.1",
     "underscore": "components/underscore#~1.5",
     "jquery-typeahead": "~2.0.0"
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "moment": "^2.8.4",
     "preact": "^4.5.1",
-    "preact-compat": "^1.7.0"
+    "preact-compat": "^1.7.0",
+    "xterm": "^0.33.0"
```
takluyver: Just for my own understanding, why are there some dependencies in `package.json` and some in `bower.json`? Are these in package.json installed by npm? Are we trying to move away from bower?

minrk:
Woo!

We should look through the various open Issues pertaining to terminal and see which ones are closed by this. I know the terminal paste issue has been open for a *long* time.

And we should also make sure this works without webpack if we really are going to rollback the webpack change for the legacy pages (I looked into this for a while last week, and it looked like a lot of work).

takluyver:
Thanks @blink1073 :-)

blink1073:
It could be that the classic notebook will need to similarly expose a public API for v5+.

takluyver:
webpack is currently failing with ` ModuleNotFoundError: Module not found: Error: Cannot resolve module 'style-loader' in /tmp/pip-build-uaps7hmg/notebook/notebook/static/terminal/js`

blink1073:
@takluyver, awesome!  That means adding `bail: true` did what it intended. ;)

minrk:
@blink1073 have you figured out how to get runtime `require` to return the actual exact module instance (forgive my Python terminology) when it was bundled up with webpack? This is the backward-compatibility bit we need to figure out if we are going to keep webpack for the legacy endpoints.

blink1073:
@minrk, each WebPack bundle will *be* an AMD module in the notebook, exporting the desired API.

minrk:
That's very exciting. Looking forward to the PR. Thanks for working on it.

takluyver:
It's now failing in some Python code that checks that the expected JS files are in place - this probably just needs to be updated.

```
      File "/tmp/pip-build-ow9jpojc/notebook/setupbase.py", line 220, in check_package_data
        assert os.path.exists(path), "Missing package data: %s" % path
    AssertionError: Missing package data: notebook/static/components/term.js/src/term.js
```

setupbase.py
```diff
@@ -153,7 +153,6 @@ def find_package_data():
         pjoin(components, "underscore", "underscore-min.js"),
         pjoin(components, "moment", "moment.js"),
         pjoin(components, "moment", "min", "moment.min.js"),
-        pjoin(components, "term.js", "src", "term.js"),
```
takluyver: Does it make sense to replace it with a check for xterm.js? Or does the bundling mean there isn't a separate file for it?

blink1073: These are explicit checks for bower modules, there is no need IMO to check for node_modules.

takluyver:
Thanks, let's see how we get on with this.

yuvipanda:
\o/ Thank you all! The bot users of Wikimedia love you :)

On Tue, Jul 5, 2016 at 11:07 PM, Thomas Kluyver <notifications@github.com>
wrote:

> Merged #1591 <https://github.com/jupyter/notebook/pull/1591>.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/jupyter/notebook/pull/1591#event-713582050>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAB23iVHeBoMGnuBNyChw3ttyKyQ-lciks5qSsd2gaJpZM4JFKEt>
> .
>



-- 
Yuvi Panda T
http://yuvi.in/blog

mogthesprog:
Thanks for this guys! Seriously.... so many analysts will stop hating me because of this fix.

yuvipanda:
I've kinda of ported this to run on top of 4.2.1 with https://github.com/jupyter/notebook/pull/1613. Note that I haven't touched js in a long time - this works for me, but might not work for everyone!

PR #1599, opened 2016-07-06
fix notebook validation

minrk:
validate was imported from traitlets in #1527, clobbering validate from nbformat

### DIFF
--- a/notebook/services/contents/manager.py
+++ b/notebook/services/contents/manager.py
@@ -301,7 +301,7 @@ class ContentsManager(LoggingConfigurable):
         try:
             validate_nb(model['content'])
         except ValidationError as e:
-            model['message'] = u'Notebook Validation failed: {}:\n{}'.format(
+            model['message'] = u'Notebook validation failed: {}:\n{}'.format(
                 e.message, json.dumps(e.instance, indent=1, default=lambda obj: '<UNKNOWN>'),
             )
         return model
--- a/notebook/services/contents/manager.py
+++ b/notebook/services/contents/manager.py
@@ -13,7 +13,7 @@ from tornado.web import HTTPError
 
 from .checkpoints import Checkpoints
 from traitlets.config.configurable import LoggingConfigurable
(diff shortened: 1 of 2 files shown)


takluyver:
I'm a bit surprised we didn't have a test which caught this. Is there an easy way to add one that would exercise this code?

minrk:
Me, too, as a matter of fact. I found it because I couldn't open any notebooks. Not sure how that doesn't get any exercise, as anything that opens a notebook should hit this.

minrk:
Oh, now I know. It's because I'm on a branch of traitlets where this invalid call to `validate` raises instead of instantiating an object that gets immediately thrown away.

We don't seem to have a test here for notebooks that fail validation.

takluyver:
> We don't seem to have a test here for notebooks that fail validation.

Is it worth adding one? If you don't fancy doing it here, feel free to open an issue for it, and then I'll merge this.

Carreau:
We should start adding linters when possible, this would have been caught as one import clober another.

minrk:
I'll add a test.

Linters are nice, but can be trouble in CI. I like pyflakes since it mostly checks correctness, but since it gives no mechanism to tell it "No, really, it's fine" I don't think it's okay to use in CI. I've used flake8 + turning off most of its annoying checks to get basically pyflakes with some control.

minrk:
Test added.

takluyver:
Test failed ;-)

It's a Legacy Python thing:

```
  File "/home/travis/build/jupyter/notebook/notebook/services/contents/tests/test_contents_api.py", line 305, in test_get_nb_invalid
    self.make_txt(path, json.dumps(nb))
  File "/home/travis/build/jupyter/notebook/notebook/services/contents/tests/test_contents_api.py", line 169, in make_txt
    f.write(txt)
TypeError: must be unicode, not str
```

minrk:
Argle bargle. Should be okay this time.

takluyver:
Good job, thanks.

PR #1633, opened 2016-07-26
fix travis builds on 4.x branch

minrk:
- pin casperjs to 1.1.1 (1.1.3 doesn't appear to work)
- remove coveralls calls


minrk:
Merging since this is only a travis change, to get 4.x builds back again.

PR #1638, opened 2016-07-27
Remove `circle.yml`

jakirkham:
This was used as CI for the docker build. However, we got rid of the Docker image. So we don't need this either.


takluyver:
Thanks!

PR #1642, opened 2016-07-28
include cross-origin check when allowing login URL redirects

minrk:
allows cross-origin setups to use the login page and redirect back

### DIFF
--- a/notebook/auth/login.py
+++ b/notebook/auth/login.py
@@ -3,6 +3,10 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+try:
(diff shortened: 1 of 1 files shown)


blink1073:
👍

Carreau:
`s/redirect/redirect_unsafe/g`, `s/_redirect_safe/redirect/g` from a best practice ? Or, if it's to not break API, prepare for a switch in next major version ?

takluyver:
`redirect` is a method defined by tornado, so let's not change what it means.

Carreau:
> redirect is a method defined by tornado, so let's not change what it means.
>  Show all checks

Ok.

PR #1650, opened 2016-08-02
Load extension in predictable order.

Carreau:
Not doing this can lead to subtle incomprehensible issues when 2
extensions (eg, jupyterlab and jupyterlab_extensions) both register the
same handlers, that will randomly at each start get a different
extension take precedence.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -902,7 +902,8 @@ class NotebookApp(JupyterApp):
     nbserver_extensions = Dict({}, config=True,
         help=("Dict of Python modules to load as notebook server extensions."
               "Entry values can be used to enable and disable the loading of"
-              "the extensions.")
+              "the extensions. The extensions will be loaded in alphabetical "
+              "order.")
     )
 
     reraise_server_extension_failures = Bool(
@@ -1156,7 +1157,7 @@ class NotebookApp(JupyterApp):
             if not modulename in self.nbserver_extensions:
                 self.nbserver_extensions[modulename] = True
         
-        for modulename in self.nbserver_extensions:
+        for modulename in sorted(self.nbserver_extensions):
             if self.nbserver_extensions[modulename]:
                 try:
                     mod = importlib.import_module(modulename)
--- a/notebook/tests/test_serverextensions.py
+++ b/notebook/tests/test_serverextensions.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from unittest import TestCase
 try:
     from unittest.mock import patch
@@ -13,7 +14,12 @@ from traitlets.tests.utils import check_help_all_output
 
 from notebook.serverextensions import toggle_serverextension_python
 from notebook import nbextensions
+from notebook.notebookapp import NotebookApp
 from notebook.nbextensions import _get_config_dir
+
+
+from types import SimpleNamespace
+from collections import OrderedDict
 
 
 def test_help_output():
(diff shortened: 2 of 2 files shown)


notebook/notebookapp.py
```diff
@@ -902,7 +902,8 @@ def _update_server_extensions(self, change):
     nbserver_extensions = Dict({}, config=True,
         help=("Dict of Python modules to load as notebook server extensions."
               "Entry values can be used to enable and disable the loading of"
-              "the extensions.")
+              "the extensions. The extensiosn will be loaded in alphabetical "
```
fperez: typo and wording: should be "the extensions will be ..."

notebook/notebookapp.py
```diff
@@ -1156,7 +1157,7 @@ def init_server_extensions(self):
             if not modulename in self.nbserver_extensions:
                 self.nbserver_extensions[modulename] = True
         
-        for modulename in self.nbserver_extensions:
+        for modulename in sorted(self.nbserver_extensions.items(), key=lambda k,v:k):
```
fperez: Why do you need the full `.items()` call? Since you're only iterating on keys, as far as I can tell, a simple `sorted(self.nbserver_extensions)` should suffice...

Carreau: I read too fast :  `for modulename, value in ... ` you are right.

fperez:
I don't see any downside to this, given that we're just turning something that was effectively random into something at least predictable.

Before I merge, my only question is: in addition to the help string, is there any location in the docs where it's worth documenting the load order?

Carreau:
> don't see any downside to this, given that we're just turning something that was effectively random into something at least predictable.
> 
> Before I merge, my only question is: in addition to the help string, is there any location in the docs where it's worth documenting the load order?

Yes, and write a test also.

fperez:
+1, I'll wait then.​

gnestor:
@Carreau It looks like you updated the docstring. Do you want to write a quick test for `init_server_extensions` so we can merge?

Carreau:
> @Carreau It looks like you updated the docstring. Do you want to write a quick test for init_server_extensions so we can merge?

I'll try to do that soon, I have a few other things that I want to finish first while they are fresh in my head; apologies if I am not super responsive.

gnestor:
@Carreau No prob. I'll take a quick stab at it...

gnestor:
@Carreau Ya I'm not sure how to write a test for this. Assigning to you...

Carreau:
Ok, finally came up with a way to test that, which could be cleaner but should work. 

one question though, I'm wondering is enforcing ordering on keys is the right thing as it is apparently system dependant (leading zero might not count) and users might start to rely on this specific fact and publish extensions with specific names to force this ordering. 

I'm considering just saying that the order is stable on a specific configuration (without consideration of which order), and for example sort using a hash of the key.

gnestor:
@Carreau So you're proposing that we don't allow users to be able to specify loading order of extensions by sorting extensions by a hash of their name, for example? What if the loading order was based on the order in `jupyter_notebook_config.json`? This would allow users to specify the loading order of extensions there (as long as jupyter respects this order when writing to the file and doesn't arbitrarily rearrange the order like I've observed in `nbconfig/notebook.json`).

Carreau:
Ok, so github since to have missed it, but there is now tests.

notebook/notebookapp.py
```diff
@@ -1156,7 +1157,7 @@ def init_server_extensions(self):
             if not modulename in self.nbserver_extensions:
                 self.nbserver_extensions[modulename] = True
         
-        for modulename in self.nbserver_extensions:
```
takluyver: Should the deprecated `server_extensions` above be loaded as well?

takluyver: s/loaded/sorted/

takluyver:
@gnestor the extensions to load are stored in a JSON object rather than an array, so it's conceptually unordered. If the loading order of extensions matters, I'd say that either the extensions or our API are doing something wrong.

notebook/notebookapp.py
takluyver: Nevermind, I see that `server_extensions` is a list, so it's already a stable order.

Issue #1649, opened 2016-08-02
Use of undefined variable.


Carreau:
in `notebookapp.py`. `new` is not defined. 
```
@observe('pylab')
    def _update_pylab(self, change):
        """when --pylab is specified, display a warning and exit"""
        if new != 'warn':
            backend = ' %s' % change['new']
        else:
            backend = ''
        self.log.error("Support for specifying --pylab on the command line has been removed.")
        self.log.error(
            "Please use `%pylab{0}` or `%matplotlib{0}` in the notebook itself.".format(backend)
        )
        self.exit(1)
```

takluyver:
Looks like it just got overlooked in the traitlets API update. #1652 should fix it.

PR #1652, opened 2016-08-03
Fix missed in traitlets API overhaul

takluyver:
Closes gh-1649


Carreau:
I was in the middle  of another patch/debug so couldn't fix it on the fly.  Thanks for taking care of it.

PR #1658, opened 2016-08-04
Display "trusted" and "untrusted" notifications

gnestor:
demo

### DIFF
--- a/notebook/static/notebook/js/notificationarea.js
+++ b/notebook/static/notebook/js/notificationarea.js
@@ -351,9 +351,9 @@ define([
         // Notebook trust events
         this.events.on('trust_changed.Notebook', function (event, trusted) {
             if (trusted) {
-                tnw.info("Trusted");
+                tnw.set_message("Trusted");
             } else {
-                tnw.danger("Not Trusted", undefined, function() {
+                tnw.set_message("Not Trusted", undefined, function() {
                   that.notebook.trust_notebook();
                   return false;
                 });
--- a/notebook/static/notebook/js/notificationarea.js
+++ b/notebook/static/notebook/js/notificationarea.js
@@ -24,6 +24,7 @@ define([
     NotebookNotificationArea.prototype.init_notification_widgets = function () {
         this.init_kernel_notification_widget();
         this.init_notebook_notification_widget();
+        this.init_trusted_notebook_notification_widget();
     };
 
     /**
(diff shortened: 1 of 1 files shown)


gnestor:
Related #1293

gnestor:
@ellisonbg Can you UX test?

minrk:
Nice!

Carreau:
I would just not use the `danger` alternative as it's ok to view an untrusted notebook. 
I think the "classic" color for both trusted and untrusted might be sufficient. 

One of the reasons I think it would be better, is that I don't want this action to become something which is click-through like the UAC on windows. 

Otherwise +1

takluyver:
Yeah, we want to avoid creating something like the MS Office 'enable macros' dialog. Because when the computer says "This might be incredibly dangerous!", users say:

gnestor:
Ok, I just switched to classic style. 

One other question: Should the "Trusted" notification hide itself after a second or so? Currently both notifications are persisted. If the "Untrusted" notification is clicked, the user is prompted to trust the notebook. If the "Trusted" notification is clicked, the notification hides.

Carreau:
> Ok, I just switched to classic style.
> 
> One other question: Should the "Trusted" notification hide itself after a second or so? Currently both notifications are persisted. If the "Untrusted" notification is clicked, the user is prompted to trust the notebook. If the "Trusted" notification is clicked, the notification hides.

I think that's a good compromise. And it's easy enough to change afterward if needed. 
It will keep the UI clean for most users. 

As @Phyks was the first initiator of that, we can ask him his thoughts.

takluyver:
I think I'd be inclined to hide both notifications after 10 seconds or so, to reduce the likelihood that the user will click to make it go away. Even in blue, it still stands out quite a bit.

Carreau:
>  Even in blue, it still stands out quite a bit.

Both are gray now. IIUC.

Phyks:
Sorry, I did not have much time to follow it and eventually work on my PR lately. 

Thanks @gnestor ! Your design implementation looks really cool!

I would not be in favor of hiding the untrusted notification, as the user may not notice it before it goes away and the main purpose of this feature was to have an indication that a notebook is untrusted. The trusted notification can go away though I think, as it results from either a user interaction, or a notebook being safe from the loading.

gnestor:
Ok, @minrk should we switch this to 5.0 before merging or leave it at 4.3 and backport?

Carreau:
> Ok, @minrk should we switch this to 5.0 before merging or leave it at 4.3 and backport?
>  Show all checks

You can leave at 4.3 for the backport script to find the issue. 
Or you can decide this is only 5.0 and is a "new feature" and leave switch the milestone to 5.0; 
Up to you. I'm fine considering that as "new feature" and having one less thing to backport.

gnestor:
> I'm fine considering that as "new feature" and having one less thing to backport.

That's what I was thinking...

Carreau:
Yeah !

Issue #1454, opened 2016-05-13
Cannot change kernel if current kernel broken.


Carreau:
`Notebook Dashboard` > `New Notebook` > `broken/missing kernel` (like I missclick on a missing conda-env kernel).

I now cannot change kernel anymore, well it **says** it changed, but does not. 
Even after save/reload.

gnestor:
@Carreau How can I reproduce the broken/missing kernel part?

Carreau:
hey, I would duplicate and existing kernelspec and just replace the executable by gibrish
```
$ jupyter kernelspec list
Available kernels:
  ...
  python3      /usr/local/share/jupyter/kernels/python3
```

```
$ cd /usr/local/share/jupyter/kernels
$ cp -r python3/ brokenkernel
$ sed -i bkp s/ython/broken/g brokenkernel/*.json  # missing p for Python on purpose as kernel name start with uppercase P. 
```

Does that make sens ?

PR #1669, opened 2016-08-10
If kernel is broken, start a new session

gnestor:
Fixes https://github.com/jupyter/notebook/issues/1454

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -254,10 +254,8 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
             if (!existing_spec || ! _.isEqual(existing_spec, that.metadata.kernelspec)) {
                 that.set_dirty(true);
             }
-            // start session if the current session isn't already correct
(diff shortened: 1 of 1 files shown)


gnestor:
@Carreau Can you test?

Carreau:
> @Carreau Can you test?

Starting to test review that. Sorry for the delay !

Carreau:
Hum... does not seem to fix it for me... Let me make sure I have your branch...

Carreau:
Ha... it work 1/2 time...

i can do `Kernel A` -> `Broken Kernel` -> `Kernel B` successfully, but not `Kernel A` -> `Broken Kernel` -> `Kernel A`.

Carreau:
Ok, that's because this logic is faulty as well: 

https://github.com/jupyter/notebook/blob/fd03c87f431372fc5de459ecbba00fa3f5c15765/notebook/static/notebook/js/notebook.js#L258-L260

if the kernel is broken `that.session.kernel` still point to the **old** kernel, so it does nothing and does not start a new session...

gnestor:
@Carreau Good catch! Can you pull and test again?

Carreau:
Whooo ! Thanks !

Issue #1670, opened 2016-08-11
Handle MathML output


takluyver:
Follow on from #1565.

The mimetype for MathML appears to be `application/mathml+xml` ([source](https://www.w3.org/TR/MathML2/chapter7.html)).

gnestor:
@takluyver I'd be happy to look into this. A couple questions:

* Where are mime types defined in this repo?
* Where are handlers for mime types defined?

takluyver:
This file should be the most relevant one: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/outputarea.js

gnestor:
Thanks @takluyver! Here is an initial implementation https://github.com/jupyter/notebook/pull/1672.

gnestor:
Merged with https://github.com/jupyter/notebook/pull/1672

PR #1672, opened 2016-08-11
Add support for MathML mime type

gnestor:
Initial implementation of https://github.com/jupyter/notebook/issues/1670

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -245,6 +245,7 @@ define([
         'text/html',
         'text/markdown',
         'text/latex',
+        'application/mathml+xml',
         'image/svg+xml',
         'image/png',
         'image/jpeg',
@@ -445,7 +446,8 @@ define([
         }
         this._safe_append(toinsert);
         // If we just output latex, typeset it.
-        if ((json.data['text/latex'] !== undefined) ||
+        if ((json.data['application/mathml+xml'] !== undefined) ||
+            (json.data['text/latex'] !== undefined) ||
             (json.data['text/html'] !== undefined) ||
             (json.data['text/markdown'] !== undefined)) {
             this.typeset();
@@ -558,7 +560,8 @@ define([
         if (this.append_mime_type(json, toinsert, handle_inserted)) {
             this._safe_append(toinsert);
             // If we just output latex, typeset it.
-            if ((json.data['text/latex'] !== undefined) ||
+            if ((json.data['application/mathml+xml'] !== undefined) ||
+                (json.data['text/latex'] !== undefined) ||
                 (json.data['text/html'] !== undefined) ||
                 (json.data['text/markdown'] !== undefined)) {
                 this.typeset();
@@ -570,6 +573,7 @@ define([
     OutputArea.safe_outputs = {
         'text/plain' : true,
         'text/latex' : true,
+        'application/mathml+xml' : true,
         'image/png' : true,
         'image/jpeg' : true
     };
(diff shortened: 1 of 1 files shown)


gnestor:
@takluyver This adds support for "application/mathml+xml" and provides a `append_mathml` function that mirrors `append_latex` functionality. 

@poke1024 @rgbkrk @minrk Do you have any idea how to test this? Are there any kernels that output this mime type?

rgbkrk:
This is the chicken and egg problem for the project, which is that kernels won't implement/experiment with mimetypes that the notebook doesn't implement (let alone other frontends), while demand for new mimetypes in the notebook has to come from library authors and kernel developers knowing they "need" those mimetypes. Great to see this initial implementation!

notebook/static/notebook/js/outputarea.js
```diff
@@ -783,6 +787,18 @@ define([
...
     };
+    
+    var append_mathml = function (mathml, md, element) {
+        /**
+         * This method cannot do the typesetting because the mathml first has to
+         * be on the page.
+         */
+        var type = 'application/mathml+xml';
+        var toinsert = this.create_output_subarea(md, "output_mathml", type);
+        toinsert.text(mathml);
```
Carreau: wouldn't `.text()` escape the tags ?  should we use html, and make sure to sanitize ?

gnestor: Ok it looks like `append_html` or equivalents use `toinsert.append()`, so it would look like:

```js
var append_mathml = function (mathml, md, element) {
        var type = 'application/mathml+xml';
        var toinsert = this.create_output_subarea(md, "output_mathml", type);
        toinsert.append(mathml);
        element.append(toinsert);
        return toinsert;
};
```

How do we go about sanitizing the HTML?

gnestor:
> This is the chicken and egg problem for the project, which is that kernels won't implement/experiment with mimetypes that the notebook doesn't implement (let alone other frontends), while demand for new mimetypes in the notebook has to come from library authors and kernel developers knowing they "need" those mimetypes. Great to see this initial implementation!

@rgbkrk Theoretically, could I add a MathML [formatter](https://github.com/ipython/ipython/blob/bf3b45fbc38e224703b18a53d18ff03e4279d296/IPython/core/formatters.py) to ipython to test? The question then is: how does the kernel map outputs to mime types (e.g. how will ipython know that the output is "application/mathml+xml" type)? Is that what [display.py](https://github.com/ipython/ipython/blob/8185987f6af796024c3eac3cc36c5a5162c889fa/IPython/core/display.py) is for?

notebook/static/notebook/js/outputarea.js
takluyver: I think the key thing is to use the trust mechanism, as we do for HTML and JS output already. So outputs sent from a running kernel or loaded from a trusted notebook will be displayed without sanitisation, and outputs loaded from an *untrusted* notebook are either not displayed at all, or sanitised so they can't do anything dangerous.

IIRC the sanitisation library we use is Google Caja.

takluyver:
Yes, we should add a corresponding `MathMLFormatter` in IPython, which would recognise a special method like `_repr_mathml_`. The formatters themselves store the mimetypes of the output they produce, as a `format_type` attribute.

notebook/static/notebook/js/outputarea.js
gnestor: @takluyver I think this should handle sanitation: https://github.com/jupyter/notebook/pull/1672/commits/a711ddb45e9540b12ade13cbe48f91a558ffd0e3

takluyver:
That looks sensible to me, but I hope that someone more familiar with our JS code will have a look at it too.

rgbkrk:
Looks good to me relative to how other mime types are implemented.

rgbkrk:
Something good to follow up here with: it would be good to use constants (locally scoped shared variables in JS) for "application/mathml+xml".

gnestor:
@rgbkrk Are you suggesting that we use constants for all references to mime types in `outputarea.js`? I think this is a good idea...

rgbkrk:
@gnestor yup!

gnestor:
> Something good to follow up here with: it would be good to use constants (locally scoped shared variables in JS) for "application/mathml+xml".

@rgbkrk See https://github.com/jupyter/notebook/pull/1678

PR #1675, opened 2016-08-12
Update webpack and eslint config

gnestor:
Improve source maps and decrease eslint warnings

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,7 +7,7 @@
         "semi": 1,
         "no-cond-assign": 2,
         "no-debugger": 2,
-        "comma-dangle": 1,
+        "comma-dangle": 0,
         "no-unreachable" : 2
     }
 }
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -55,7 +55,8 @@ function buildConfig(appName) {
         entry: './notebook/static/' + appName + '/js/main.js',
         output: {
             filename: 'main.min.js',
-            path: './notebook/static/' + appName + '/js/built'
+            path: './notebook/static/' + appName + '/js/built',
+            publicPath: '/'
         },
         devtool: sourcemaps,
     });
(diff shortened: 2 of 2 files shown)


Carreau:
+1 to do that !

gnestor:
It looks like all the checks failed! 😮 @Carreau Can you tell why?

Carreau:
I'll have a look. 

Sorry 'm usually more responsive, but I had a lot of things that coulnd't that took priority.

It seem to be 3 of [these lines](https://github.com/jupyter/notebook/blob/master/setupbase.py#L124-L130) that try to ensure at install time that the `.maps` files are built.

Carreau:
https://github.com/gnestor/notebook/pull/1 shoudl fix the build.

gnestor:
@Carreau To make sure I understand: The checks were failing because I switched from `source-map` to `inline-source-map`?

Carreau:
Yes, building the python package was failing to include the map files that
where explicitely included.

It's not obvious from the log, as there is this weird combination of python
calling npm.

The error messages are probably due to some extra-cleverness from
setuptools of trying to figure out what went wrong in previous command.

gnestor:
"Checks have passed!"

Carreau:
> "Checks have passed!"

Aka: Travis is Happy.

@minrk , this inline the maps if we release it,  and make the files bigger, do you know of anyone that would be strongly affected on the hub side by these changes and would complain ?

gnestor:
@Carreau @minrk If inline maps prove to be an optimization issue in production, we can conditionally set the `sourcemap` config based on the `NODE_ENV` (e.g. `inline-source-map` for `development` and `source-map` for `production`). However, this may over-complicate things for `setupbase.py` and Travis.

rgbkrk:
:shipit:

minrk:
@Carreau I wouldn't expect any complaints about the size due to including maps.

PR #1678, opened 2016-08-13
Replace mime type strings with constants

gnestor:


notebook/static/notebook/js/outputarea.js
```diff
@@ -239,18 +239,30 @@ define([
...
+    const MIME_JAVASCRIPT = 'application/javascript';
+    const MIME_HTML = 'text/html';
+    const MIME_MARKDOWN = 'text/markdown';
+    const MIME_LATEX = 'text/latex';
+    const MIME_MATHML = 'application/mathml+xml';
+    const MIME_SVG = 'image/svg+xml';
+    const MIME_PNG = 'image/png';
+    const MIME_JPEG = 'image/jpeg';
+    const MIME_PDF = 'application/pdf';
+    const MIME_TEXT = 'text/plain';
```
rgbkrk: Glorious, thank you!

takluyver:
Out of interest, what is the advantage of this?

gnestor:
Based on my understanding, using constants this way a) centralizes information (in this case, the mime type strings) and b) can be used to catch errors if you're using a static analysis tool like Flow or Typescript. It's easy to imagine one of those `"application/mathml+xml"` accidentally ending up like `"application/matml+xml"` which would probably pass tests and make it into production while `MIME_MATML` would not if we using Flow. 

@rgbkrk What are your thoughts?

Carreau:
> Based on my understanding, using constants this way a) centralizes information (in this case, the mime type strings) and b) can be used to catch errors if you're using a static analysis tool like Flow or Typescript. It's easy to imagine one of those "application/mathml+xml" accidentally ending up like "application/matml+xml" which would probably pass tests and make it into production while MIME_MATML would not if we using Flow.

I think that's a good point, and even if not using flow, i guess with `"use strict"` you would get an error.

Thanks !

rgbkrk:
> Out of interest, what is the advantage of this?

The notebook is a fairly complex web app. Even if this was a compiled application, I'd prefer single constants for repeated strings to both give me name hinting (within Atom and Vim) and to catch my typos (`eslint` will catch this after all). What @gnestor says is exactly right.

takluyver:
Thanks both for taking the time to explain it to me.

PR #1680, opened 2016-08-14
fixes problem with disappearing MathJax mglyphs in output areas

poke1024:
This change allows MathML `mglyphs` with embedded data to be correctly rendered. This is an important feature for some applications using images embedded in math formulas.

To test this PR's change in a Python 3 notebook, use the attached notebook and evaluate all three cells.

Without this PR, the third step's `HTML` will output nothing. With the CSS change in this PR, an arrow svg will be visible as expected.

The second step is necessary to reconfigure MathJax to accept `mglyph' tags with embedded data, which is by default disabled due to Jupyter's safe MathJax mode. I will address this in a separate PR.

[mglyph.ipynb.zip](https://github.com/jupyter/notebook/files/416997/mglyph.ipynb.zip)

### DIFF
--- a/notebook/static/notebook/less/outputarea.less
+++ b/notebook/static/notebook/less/outputarea.less
@@ -74,6 +74,10 @@ div.output_area {
             max-width: none;
         }
     }
+
+    .mglyph>img {
+        max-width: none;
+    }
 }
 
 /* This is needed to protect the pre formating from global settings such


gnestor:
@poke1024 I can confirm that master currently will not render the SVG and that this PR will render it. In testing, I noticed that the less-watch script only recompiles LESS on a file change and not on `npm run build:watch` so I will submit a PR to fix that...

@Carreau @minrk Feel free to test as well but I think this is ready to merge.

PR #1695, opened 2016-08-20
Include bundler test files in package_data

takluyver:
This was causing a test failure on Jenkins. It only shows up if you install the package (not a dev install) and test against the installed version.

Ping @parente

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -198,6 +198,7 @@ def find_package_data():
     package_data = {
         'notebook' : ['templates/*'] + static_data,
         'notebook.tests' : js_tests,
+        'notebook.bundler.tests': ['resources/*', 'resources/*/*', 'resources/*/*/.*'],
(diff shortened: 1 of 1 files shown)


parente:
Thanks for catching that. Is there a Jenkins running somewhere that tests the installed package that I should check in the future?

takluyver:
No problem. https://jenkins.jupyter.org/ is our server. The jobs which test this repo are:

- Notebook
- notebook-windows
- jack-of-none (which installs and tests master from a range of our projects)

takluyver:
Merging this to try to fix the test failures.

parente:
Thanks. (I don't have merge perms here else I would have done it.)

takluyver:
Annoyingly, this does not seem to have solved things on Windows. I'm not sure why at the moment.

PR #1701, opened 2016-08-23
Update xterm.js to 1.1.2

yuvipanda:
Major feature upgrade is that IMEs now work, so people
can type non-latin characters into them!

Also has:
   - Better mouse support
   - Signed releases
   - Scrollbar

https://github.com/sourcelair/xterm.js/releases/tag/1.1.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,6 @@
     "moment": "^2.8.4",
     "preact": "^4.5.1",
     "preact-compat": "^1.7.0",
-    "xterm": "^1.0"
+    "xterm": "^1.1.2"
   }
 }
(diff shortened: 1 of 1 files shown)


minrk:
1.1 apparently hasn't been published to npm yet. 1.0 is still the [latest release](https://www.npmjs.com/package/xterm).

yuvipanda:
Yup, filed https://github.com/sourcelair/xterm.js/issues/244

yuvipanda:
They fixed the issue with their release script and pushed 1.1.3, so I've updated to that. Waiting for travis now.

yuvipanda:
Are these just flaky tests? I don't see how they are related...

takluyver:
Yup, looks like flaky tests. I just re-ran the failing one, and it passed. Let's see if this helps with #1724.

PR #1709, opened 2016-08-24
Fixed a bug in 'ConfigWithDefaults.get_sync'

superbobry:
The method returned the default value if the result of `_class_data` lookup was falsy (in the JS sense).

This bug also affected `ConfigWithDefaults.get`.


superbobry:
Also note that the documentation of `ConfigWithDefaults.get_sync` is a bit misleading, because the default is also returned in case `key` is not present in the loaded config.

takluyver:
This looks OK to me, but I'll let someone more familiar with JS double check.

PR #1715, opened 2016-08-26
Remove jupyter lab link.

SylvainCorlay:
Since jupyterlab is now a good citizen and is installed as a notebook extension. Besides, that link is dead when the extension is missing.


minrk:
Yup. If anything, this link should be provided by the lab extension.

PR #1725, opened 2016-08-31
Home url

Amerterasu:
Changes the functionality of the fa fa-home button on the page. Allows changes to the home button url which can be configured from the jupyterhub config file. Requires the jupyterhub github repository to accept changes to the spawner to allow this functionality.


minrk:
How would you navigate to the root directory listing if the home link is overridden? This link is part of the path breadcrumbs, and I think should always take you to the root tree listing. Note that the default_url, which is where clicking the Jupyter *logo* takes you is already configurable. That is most analogous to other web application "home URLs", not the little house in the breadcrumbs.

Amerterasu:
@anderbubble My plan was to replicate the feature that the logo provides for links such as the user's home directory. So if the user has full system traversal that if they decide they'd rather not always go back to root, but their home.

Carreau:
> @anderbubble My plan was to replicate the feature that the logo provides for links such as the user's home directory. So if the user has full system traversal that if they decide they'd rather not always go back to root, but their home. I was hoping this would provide more customization and maybe easier for some users to understand the meaning of that home button. I imagine that moving to root could be done by going one parent up at a time and that this wouldn't happen often if it is customized to a directory they regularly return to.

The ability to get back to the hub make sens, but I doubt  changing the meaning of the home icon is the right way to do it, though we really appreciate the thought, and will consider that. Another similar request we had was how to access multiple drives on windows whic would require modification of the UI as well. 


Side note and comment about the commit themselves when you make a pull request:

Can you avoid to add files like the sqlite one and other file not relevant to the change on the PR ?
If you are new to git feel free to ask questions. 

We also prefer not to trim whitespace in non modified lines of files, we would appreciate if you could configure your editor to not do that when contributing. The reasoning has mostly 3 points:
  - It create conflicts between PRs. 
  - it make review harder, 
  - It screws up looking at the history if we need to git blame. 


Feel free to ask if you have any questions.

anderbubble:
> I doubt changing the meaning of the home icon is the right way to do it

I'd argue that notebook is already breaking the meaning of the home icon in this sense. In every UI I've used, a home icon goes to the home directory. We're just trying to maintain that UI concept while allowing the user to navigate outside of their home directory as well.

> Another similar request we had was how to access multiple drives on windows whic would require modification of the UI as well.

Makes sense enough, and I don't want to undervalue the impact of changes to the UI. Still, allowing the behavior of the home button to be configured shouldn't break that for anyone who doesn't wish to change it.

minrk:
If the icon is the source of confusion, I would propose that we change it from a house to something else that indicates the "root" of the served directory. What we have is a breadcrumb for the root of the served directory (which usually does not include the user's home, so the home directory cannot be navigated to), not a link to 'home'. The "default URL" for a given notebook server is already defined and configurable, and currently linked from the Jupyter logo. This is what people are already using to default to users' home directories in JupyterHub instances with access to the whole filesystem. I don't think we should have two separate definitions for a default URL, though I'm happy to consider adding new links to the default URL if the upper-left logo convention is insufficient.

cc @ellisonbg for UI/icon/link questions

gnestor:
To avoid ambiguity, I suggest that the home icon be replaced with a [folder icon](http://fontawesome.io/icon/folder/) or the name of the current working directory (served directory).

anderbubble:
@gnestor I think a different icon is a good idea. A folder icon could work, in which case I'd advocate for hovertext that displays the served path. The name of the path itself might be ok; but I'm afraid of losing the visual abstraction that the current icon provides (e.g., people don't *have* to care what the served directory is).

I'd also put forward an icon of [http://fontawesome.io/icon/tree/ a tree] as a potential alternative (though I must admit I don't actually like *that* tree icon).

anderbubble:
I'm happy with where this discussion is going, btw. It seems like our desired functionality exists, it's just a UI concern. To that end, I want "clicking on the logo to get to the default directory" to be more discoverable. Until @Amerterasu discovered it as part of this effort, I had no idea it did that.

gnestor:
Folder icon:

image

Directory name:

image

anderbubble:
I take my suggestion of tree back, along with any reservations about using the folder icon. That looks great.

gnestor:
For simplicity's sake, I think we should just open a new PR with only the icon change and close this one given the extra files (e.g. sqllite) and trimmed whitespace. The only line that needs to be changed is https://github.com/jupyter/notebook/blob/master/notebook/templates/tree.html#L99.

anderbubble:
Can we get some indication of the function of clicking on the logo in that PR, too? Or should that be a yet separate PR?

gnestor:
@anderbubble Ok, good! I will create a new PR. Regarding the href of the Jupyter logo...

> The "default URL" for a given notebook server is already defined and configurable, and currently linked from the Jupyter logo.

@minrk Can you elaborate on how to configure the href of the Jupyter logo?

anderbubble:
@gnestor see `NotebookApp.default_url`

gnestor:
@anderbubble Ok, so you're familiar with how to configure the default URL. What more would you like to see changed?

minrk:
@gnestor I think what's missing is a visual indication that the logo in the upper-left corner is a clickable link that takes you somewhere, or where it takes you. While this is a convention on many websites (this one included), it's not necessarily clear that the logo is a link (though it does hover as a link, and show the not-super-clear hover text 'dashboard', I believe). Perhaps expanding the hover text to be clearer is sufficient?

anderbubble:
@gnestor is correct. I don't think expanding the hover text is particularly sufficient, though. I understand that "click on the site's logo" is a convention on websites for "return to the home/front page" I definitely didn't even think to try it in this context. The "click-on-logo-for-home" idiom implies a association between the top-level branding and the top-level-content: I'd almost expect clicking on that logo to take me to http://jupyter.org before I'd expect it to take me to whatever the default directory happens to be in the notebook file browser.

gnestor:
@anderbubble and @Amerterasu Ok, feel free to submit a new PR (or clean up this one as per @minrk's suggestion) proposing something more intuitive or user-friendly.

gnestor:
Let's close this PR for now. Feel free to open a new one to provide some visual indication that the Jupyter logo in the upper-left of the dashboard is a link and where it links to.

PR #1728, opened 2016-09-01
[WIP] add appveyor.yml to start running tests on Windows

minrk:
Doesn't pass yet, but it at least starts running the tests.

- use conda-forge scripts to bootstrap conda
- I don't plan to run the js tests, just the Python ones

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,4 +23,4 @@ install:
   - cmd: pip install .[test]
 
 test_script:
-  - nosetests -v notebook
+  - nosetests --exclude-dir notebook\terminal -v notebook
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,26 @@
+# miniconda bootstrap from conda-forge recipe
+matrix:
+  fast_finish: true
+
+environment:
+  CONDA_INSTALL_LOCN: "C:\\conda"
+  matrix:
+    - CONDA_PY: 35
+    - CONDA_PY: 27
+
+platform:
+  - x64
+
+build: off
+
+install:
+  - appveyor DownloadFile "https://raw.githubusercontent.com/pelson/Obvious-CI/master/bootstrap-obvious-ci-and-miniconda.py"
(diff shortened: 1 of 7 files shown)


takluyver:
:+1:

willingc:
Tests are good 👍

Carreau:
I enabled appveyor, it should ignore any branch without an `appveyor.yml` but I hope it will start to build this PR.

Thanks @minrk for starting this !

takluyver:
Closing and reopening to see if that triggers Appveyor.

takluyver:
Several test failures on Py2 appear to relate to it not getting errors when expected.

Carreau:
> Closing and reopening to see if that triggers Appveyor.

Oh, good point. 

> Several test failures on Py2 appear to relate to it not getting errors when expected.

Easy path forward if skip-if windows for now.

minrk:
I only got Python 3 passing yesterday, but moved on before dealing with the Python 2 errors. I'll spend a little more time figuring out exactly why they fail before punting to skip-on-Windows. It's a bit odd that they only fail on Python 2.

minrk:
Most of the py2 failures were because of a real bug. The samefile backport in #1142 didn't work on Python 2 because the stat fields used (st_ino, st_dev) always return 0 on Windows + Python 2, causing the function to return True for any pair of files. It works everywhere else, though. I updated the method to do case-insensitive file comparison plus full stat comparison (times included), and use the stdlib `os.path.samefile` where it's available (everywhere but win+py2).

An alternative would be to make the fill-in same file just do `path == other_path`, which would never give a false-positive, but would disable renaming with only changed case on Windows + Python 2. Since it works correctly on Python 3 due to samefile, I don't have a huge preference either way.

takluyver:
I also have no strong preference, but since you have this working now, let's go with this unless/until we want to do something different. :-)

Carreau:
Yeahh ! Thanks @minrk  !

PR #1748, opened 2016-09-06
Fix metadata causing error with docs build

### LINKED ISSUE #1746: Errors on Read the Docs build after recent PR
After #1742 Read the Docs is now erroring:

https://readthedocs.org/projects/jupyter-notebook/builds/4374465/

willingc:
Closes #1746 

Removes "anaconda-cloud" and second Python version from the metadata.


michaelpacer:
Lgtm, though I think the anaconda cloud thing is going to come up anytime anyone runs, saves and commits a notebook from a conda env. We might want to look into why that is breaking things.

willingc:
@michaelpacer I believe you are correct with the anaconda cloud thing. Let's keep an eye on the next notebook change in a doc to see why the python version was appended to the metadata instead of replacing the python version in the original PR.

PR #1757, opened 2016-09-12
Fix markdown mode loading

### LINKED ISSUE #1743: Markdown cells no longer soft wrap lines in master
After updating notebook with the current master this evening, I noticed that markdown cells no longer soft wrap long lines. This makes it very difficult to edit markdown cells. This occurs both with Safari and Chrome.

screen shot 2016-09-05 at 8 25 51 pm

Running 5.0.0.dev on OS X 10.12 and Python 3.5.

minrk:
and options_default in general, which has been ignored entirely for most cell types since #1676.

This attempts to resolve the separation of config and Class.options_default by using options_default as the defaults for the cell's config objects, eliminating a separate notion of 'options' and 'config'.

For some reason, I also had to add an explicit 'reload ipythongfm' action after that mode is defined, since we are no longer guaranteed that the IPython markdown mode is defined when markdown cells are getting created. I'm not sure exactly when that stopped working, but I imagine it's due to webpack and require fighting with each other (cc @blink1073)

I also added to the config section the ability to get the whole class-config object merged with defaults, since that's the data structure we want here.

Closes #1743

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -49,10 +49,9 @@ define([
         options = options || {};
         this.keyboard_manager = options.keyboard_manager;
         this.events = options.events;
-        var config = utils.mergeopt(Cell, options.config);
+        var config = options.config;
         // superclass default overwrite our default
         
-        this.placeholder = config.placeholder || '';
         this.selected = false;
         this.anchor = false;
         this.rendered = false;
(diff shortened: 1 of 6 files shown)


fonnesbeck:
This fixes the issue for me, thanks.

gnestor:
I have tested and it fixes Markdown syntax highlighting and line soft wrapping for me. Merging...

Issue #1661, opened 2016-08-06
Permanently disable header/toolbar


rornor:
Hi, please provide option for user preference to permanently hide header/toolbar.

From my perspective header is useless to me - it contains Jupyter logo and Notebook title, which I already have in my browser tab. Although it's height is reduced compared to early versions, why waste valuable vertical space?

I made JS start script to auto disable header/toolbar, but with latest version this HTML element is probably changed, and it is not the first time it is changed. So instead digging inside HTML DOM and writing JS, I thought to write to you.

Thanks

hainm:
You can also try other themes 

https://github.com/dunovank/jupyter-themes/blob/master/README.md

gnestor:
+1 for using other themes.

See https://github.com/ipython/ipython/issues/5623#issuecomment-40517625 for an example of how to modify custom.js to hide the header/menu.

rornor:
@gnestor Have you tried the example you have linked?

FYI, it does not work in current version of Jupyter, and my point was to avoid rewriting JS for such obvious thing like removing useless elements from workspace which should be included by default or at least allow user to set the preference in some config file.

OTOH just take a look at various pycon videos that use Notebook - by default almost all use half screen as presentation workspace, because of these elements that are enabled by default.

As for the skins, thanks to @hainm for suggested the workaround, but I don't want to use skins as I like the way Notebook looks.

gnestor:
@rornor The code I linked to is no longer compatible with the current version of the notebook. I've tested the following code and it works (if you're running 4.2.2, note https://github.com/jupyter/notebook/issues/1730 and put the code outside of `'app_initialized.NotebookApp'` callback):

```js
events.on('app_initialized.NotebookApp', function () {
    console.log('custom.js: hiding header and toolbar');
    $('#header-container').toggle();
    $('.header-bar').toggle();
    $('div#maintoolbar').toggle();
    events.trigger('resize-header.Page');
});
```

Screenshot:

![screenshot](http://g.recordit.co/DOj9to5V91.gif)

So, regarding your your issue:

* You can obviously toggle the header and toolbar using the menu but this config is not persisted, so that the next time you open the same notebook or any other notebook, you will need to re-toggle.
  * However, toggle status can be persisted using `nbconfig` (this is used to persist line number toggle status on master). I will look into doing the same for header/toolbar toggle status...
* You can customize the notebook UI using `custom.js` which offers users a lot of flexibility.

rornor:
[...]

PR #1423, opened 2016-05-03
[WIP] Toggle all line numbers, &  persist config 

Carreau:
Add 3 actions `show all`,`hide all`, `toggle`, expose toggle in the menu. 

Persist state to nbconfig. Also improve the config so that config option set on `Cell` apply to MD/Code/Raw unless overwritten. By a more specific config. 

Thus the line-numbering state is stored as:

```
{
  "Cell": { 
    "cm_config": {
      "lineNumbers": false
    }
  }
}
```

PR #1769, opened 2016-09-16
Persist header and toolbar toggle status in nbconfig

gnestor:
A first stab at https://github.com/jupyter/notebook/issues/1661

### DIFF
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -462,18 +462,6 @@ define(function(require){
                 env.notebook.show_command_palette();
             }
         },
-        'show-all-line-numbers': {
-            help : 'show line numbers in all cells, and persist the setting',
-            handler: function(env) {
-                env.notebook.line_numbers = true;
-            }
-        },
-        'hide-all-line-numbers': {
-            help : 'hide line numbers in all cells, and persist the setting',
-            handler: function(env) {
-                env.notebook.line_numbers = false;
-            }
-        },
         'toggle-all-line-numbers': {
             help : 'toggles line numbers in all cells, and persist the setting',
             icon: 'fa-list-ol',
@@ -481,21 +469,16 @@ define(function(require){
                 env.notebook.line_numbers = !env.notebook.line_numbers;
             }
         },
-        'toggle-toolbar':{
-            help: 'hide/show the toolbar',
-            handler : function(env){
-                env.notebook.toolbar = !env.notebook.toolbar;
-                $('div#maintoolbar').toggle();
-                events.trigger('resize-header.Page');
-            }
-        },
         'toggle-header':{
             help: 'hide/show the header',
             handler : function(env){
                 env.notebook.header = !env.notebook.header;
-                $('#header-container').toggle();
-                $('.header-bar').toggle();
-                events.trigger('resize-header.Page');
+            }
+        },
+        'toggle-toolbar':{
+            help: 'hide/show the toolbar',
+            handler : function(env){
+                env.notebook.toolbar = !env.notebook.toolbar;
             }
         },
         'close-pager': {
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -466,19 +466,91 @@ define(function(require){
             help : 'toggles line numbers in all cells, and persist the setting',
             icon: 'fa-list-ol',
             handler: function(env) {
-                env.notebook.line_numbers = !env.notebook.line_numbers;
+                var value = !env.notebook.line_numbers;
+                env.notebook.get_cells().map(function(c) {
+                    c.code_mirror.setOption('lineNumbers', value);
+                });
+                env.notebook.line_numbers = value;
+            }
+        },
+        'show-all-line-numbers': {
+            help : 'show line numbers in all cells, and persist the setting',
+            handler: function(env) {
+                env.notebook.get_cells().map(function(c) {
+                    c.code_mirror.setOption('lineNumbers', true);
+                });
+                env.notebook.line_numbers = true;
+            }
+        },
+        'hide-all-line-numbers': {
+            help : 'hide line numbers in all cells, and persist the setting',
+            handler: function(env) {
+                env.notebook.get_cells().map(function(c) {
+                    c.code_mirror.setOption('lineNumbers', false);
+                });
+                env.notebook.line_numbers = false;
             }
         },
         'toggle-header':{
             help: 'hide/show the header',
-            handler : function(env){
-                env.notebook.header = !env.notebook.header;
+            handler : function(env) {
+                var value = !env.notebook.header;
+                if (value === true) {
+                    $('#header-container').show();
+                    $('.header-bar').show();
+                } else if (value === false) {
+                    $('#header-container').hide();
+                    $('.header-bar').hide();
+                }
+                events.trigger('resize-header.Page');
+                env.notebook.header = value;
+            }
+        },
+        'show-header':{
+            help: 'show the header',
+            handler : function(env) {
+                $('#header-container').show();
+                $('.header-bar').show();
+                events.trigger('resize-header.Page');
+                env.notebook.header = true;
+            }
+        },
+        'hide-header':{
+            help: 'hide the header',
+            handler : function(env) {
+                $('#header-container').hide();
+                $('.header-bar').hide();
+                events.trigger('resize-header.Page');
+                env.notebook.header = false;
             }
         },
         'toggle-toolbar':{
             help: 'hide/show the toolbar',
-            handler : function(env){
-                env.notebook.toolbar = !env.notebook.toolbar;
+            handler : function(env) {
+                var value = !env.notebook.toolbar;
+                if (value === true) {
+                    $('div#maintoolbar').show();
+                } else if (value === false) {
+                    $('div#maintoolbar').hide();
+                }
+                events.trigger('resize-header.Page');
+                env.notebook.toolbar = value;
+            }
+        },
+        'show-toolbar':{
+            help: 'show the toolbar',
+            handler : function(env) {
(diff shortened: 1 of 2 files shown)


gnestor:
@Carreau I am following your lead from https://github.com/jupyter/notebook/pull/1423/commits/c04046430d2b10923ae3d08d367b629351283f1d. However, neither the `header` or `toolbar` properties are being persisted in `notebook.config.data` after refresh. I don't fully understand how nbconfig works so I would appreciate your input here (I've enabled "Allow edits from maintainers" so you should be able to push to this PR).

Carreau:
Yeah, nbconfig thing is complex. let me have a look. Thanks for that.

Carreau:
So it seem you get the config right, the problem is the asynchronicity and the config return defaults before being completly loaded. 

if you `cat ~/.jupyter/nbconfig/notebook.json` you can see the values get changed. 

You might want to add a `this.config.loaded.then(function(){stuff(that.config.data)}`

The `resolve`  part of the promise does not get passed the new config apparently, but that should be fixable in `notebook/services/config.js` around line 34:

```
 ConfigSection.prototype._load_done = function() {
        if (!this._one_load_finished) {
            this._one_load_finished = true;
            this._finish_firstload();  // might want to pass self.data here. 
        }
    };
```

Does that make sens ?

takluyver:
The idea with the config is that you can create a `ConfigWithDefaults` object, and then you have a choice of:

```javascript
// Get a value immediately, using the defaults if real values aren't loaded yet
visible = c.get_sync('toolbar_visible');

// Wait for the value if necessary
c.get('toolbar_visible').then(function(visible) {...});
```

gnestor:
@Carreau Your `this.config.loaded.then()` suggestion worked great...

Two questions:

1. Is there a place to define default config values so that I don't have to wonder if `this.config.data.Header` is defined or not?
2. Is there a better place to check the config values and hide/show the header/toolbar than in the initialization of the `Notebook` class (e.g. `Notebook.load_notebook`)?

takluyver:
> Is there a place to define default config values so that I don't have to wonder if this.config.data.Header is defined or not?

There's a `ConfigWithDefaults` class, used like this:

```javascript
configwd = ConfigWithDefaults(this.config, {
    header: true,
    toolbar: true,
});
configwd.get('header').then(function(header) {
    if (!header) {
        $('#header-container').hide();
        ...
    }
});
```

gnestor:
Thanks @takluyver! Ok I switched from `ConfigSection` to `ConfigWithDefaults` to take advantage of defaults. I also moved header/toolbar toggle logic from `actions.js` to `notebook.js` to be consistent with the line numbers toggle implementation. 

I think this is ready for review...

notebook/static/notebook/js/notebook.js
```diff
@@ -164,29 +164,79 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
...
+                    $('div#maintoolbar').show();
+                } else if (value === false) {
+                    $('div#maintoolbar').hide();
+                }
+                that.events.trigger('resize-header.Page');
+                that.class_config.set('Toolbar', value);
+            }
+        });
+        
+        this.class_config.get('Header').then(function(header) {
```
gnestor: To revisit my other question: Is there a better place to put this (like `Notebook.load_notebook`)?

notebook/static/notebook/js/notebook.js
```diff
@@ -164,29 +164,79 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
...
+                            'lineNumbers':value
+                        }
+                    }
+                });
+            }
+        });
+        
+        Object.defineProperty(this, 'header', {
+            get: function() {
+                return that.class_config.get_sync('Header');
```
gnestor: I switched from ConfigSection to ConfigWithDefaults to take advantage of defaults.

notebook/static/notebook/js/actions.js
```diff
@@ -462,38 +462,23 @@ define(function(require){
...
             handler: function(env) {
                 env.notebook.line_numbers = !env.notebook.line_numbers;
             }
         },
-        'toggle-toolbar':{
-            help: 'hide/show the toolbar',
+        'toggle-header':{
+            help: 'hide/show the header',
             handler : function(env){
-                $('div#maintoolbar').toggle();
```
gnestor: I also moved header/toolbar toggle logic from actions.js to notebook.js to be consistent with the line numbers toggle implementation.

notebook/static/notebook/js/notebook.js
```diff
@@ -164,29 +164,79 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
...
-          },
-          set: function(value){
-            this.get_cells().map(function(c) {
-              c.code_mirror.setOption('lineNumbers', value);
-            })
-            that.config.update({'Cell':{'cm_config':{'lineNumbers':value}}})
-          }
-          
-        })
+            get: function() {
```
gnestor: I tried to update the line numbers toggle logic to use `ConfigWithDefaults` (via `that.class_config.get_sync` and `that.class_config.set`) to be consistent with header and toolbar but it looks like the line numbers toggle status is being read from `notebook.config.data` regardless. `notebook.json` looks like:

```json
{
  "Notebook": {
    "Toolbar": false,
    "Cell": {
      "cm_config": {
        "lineNumbers": false
      }
    },
    "Header": false
  },
  "Cell": {
    "cm_config": {
      "lineNumbers": true
    }
  }
}
```

The top-level `Cell` is set using `notebook.config.update` while `Notebook.Cell` is set using `notebook.class_config.set`. Changing the value of `Notebook.Cell.cm_config.lineNumbers` has no effect even after changing the line numbers toggle logic, so the cells/Codemirror appear to be accessing `Cell.cm_config.lineNumbers` directly vs. accessing `notebook.lineNumbers`. I know this is probably confusing and that's because it is! It doesn't really matter because everything is working but it would be nice if these 3 properties shared the same implementation.

notebook/static/notebook/js/actions.js
```diff
@@ -462,38 +462,23 @@ define(function(require){
                 env.notebook.show_command_palette();
             }
         },
-        'show-all-line-numbers': {
```
takluyver: Actions are quite public API - they're used to define keyboard shortcuts - so I'd rather not remove them without a good reason.

notebook/static/notebook/js/notebook.js
```diff
@@ -164,29 +164,79 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
...
+                    $('div#maintoolbar').hide();
+                }
+                that.events.trigger('resize-header.Page');
+                that.class_config.set('Toolbar', value);
+            }
+        });
+        
+        this.class_config.get('Header').then(function(header) {
+            if (header === false) {
+                that.header = false;
```
gnestor: This is a good point, thanks for the feedback. I just pushed a new commit that moves the toggle logic to the actions and then calls the actions to toggle the header/toolbar on notebook load.

notebook/static/notebook/js/actions.js
gnestor: Good point. I added back the show/hide line numbers as well as adding show/hide actions for header and toolbar.

gnestor:
@Carreau Would you care to review? It's working for me and I think it's quite a bit cleaner given @takluyver's feedback. Two questions for you:

1. I would like `notebook.line_numbers` to use `ConfigWithDefaults` as well to keep these properties consistent. I tried it and the issue is this: `ConfigWithDefaults` (or `notebook.class_config.set`) saves the line numbers toggle state at `Notebook.Cell.cm_config.lineNumbers` and `ConfigSection` (or `notebook.config.update`) saves it at `Cell.cm_config.lineNumbers`. After switching to `ConfigWithDefaults`, the cells no longer reflected the state in `Notebook.Cell.cm_config.lineNumbers` after refresh. However, if I manually changed the value of `Cell.cm_config.lineNumbers` in `notebook.json`, then they would reflect that. **Sooo, the cells are not accessing the line numbers toggle state from `notebook.line_numbers` but appear to accessing it via `notebook.config.data.Cell.cm_config.lineNumbers`. My question for you: where does Codemirror look to determine whether a cell should display line numbers or not (specifically on load)?**
2. Is there a better place to put the "check nbconfig for header/toolbar toggle state and hide/show them accordingly" behavior than in the initialization of the `notebook` class (like `notebook.load_notebook` for example)?

Carreau:
I'm giving a seminar today and will talk to a lot of people before and
after so won't be online a lot. I'll have a look ASAP.

gnestor:
@Carreau Ok no problem!

Carreau:
> Notebook.Cell.cm_config.lineNumbers

this is due to the fact that line number could be toggle per-cell. 
I think this is/was a mistake, and I'm fine not storing this config there anymore.

Currently here is how it seem to work...:

Codemirror instances are created on the `create_element` methods of `codecell.js`, and `textcell.js` (` this.code_mirror = new CodeMirror(input_area.get(0), this._options.cm_config);` lines.

`this._options.cm_config` contain likely the `lineNumber` field which is set on the parent class `Cell` (from `cell.js`) generated on the line `this._options = utils.mergeopt({}, cell_config_data, class_config_data);`, and `class_config_data` get the config synchronously `class_config_data = this.class_config.get_sync();`.

Does that make (some) sens ?

gnestor:
Yep, it makes perfect sense. The culprit was https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/notebook.js#L1185. I ran into a little error but it's almost there...

gnestor:
@Carreau Let's merge this and I will work off of another branch to try to get `notebook.line_numbers` to use `ConfigWithDefaults`.

Carreau:
You are right, let's try ! Thanks !

PR #1732, opened 2016-09-02
yet another error-catching custom.js loader

minrk:
Previous fix (#1378) ended up delaying custom.js loading time until after main.js had completed executing due to async loads. This switches it up a bit to add a 'custom' shim that tries to load 'custom/custom', which should ensure the right loading order.

closes #1730.

This should go into 4.2.3, since this is a regression in 4.2.2

PR #1770, opened 2016-09-18
forward-port custom.js loading

minrk:
from 4.x branch

port of #1732 onto master


Carreau:
Failure seem legitimate.

minrk:
Yeah, this doesn't seem to work due to webpack differences. Since it now looks like 5.0 will remove webpack, I'll wait for that to re-issue.

PR #1781, opened 2016-09-22
Updated README.md

himani1:
Fixed minor mistakes in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,11 @@ computing.
 ![Jupyter notebook example](docs/resources/running_code_med.png "Jupyter notebook example")
 
 ### Jupyter notebook, the language-agnostic evolution of IPython notebook
-Jupyter notebook is language-agnostic HTML notebook application for the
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks for looking at this. You're right that the language there is a bit hard to read, but unfortunately I don't think your changes improve on what's currently there.

himani1:
@takluyver : I understand. There was a bit too much use of 'the' in that. I thought with the changes I have proposed, it'd be clearer to read.

takluyver:
Some of the 'the'-s are necessary for the text to make sense, though, and others are there to match the way we tend to talk about them - e.g. we say "the IPython kernel" rather than using "IPython kernel" as a name. However, we do use "Project Jupyter" as a name, so it doesn't need a 'the'.

himani1:
@takluyver : Ok, I get it now. Thanks for correcting me.

takluyver:
Thanks :-)

Issue #1778, opened 2016-09-21
Shared dependencies between npm and bower


gnestor:
The following dependencies are declared by both npm and bower:
- es6-promise
- moment

And the following are declared as npm `devDependencies`:
- requirejs
- underscore

@blink1073 @takluyver Does it make sense to clean this up? I tried moving them from bower to npm in https://github.com/jupyter/notebook/issues/1765 and I ran into multiple issues with Travis and Appveyor builds specifically around https://github.com/jupyter/notebook/blob/master/setupbase.py#L133.

blink1073:
Sounds good to me.   I'd say you can just remove those explicit checks since we're not shipping any code from node_modules.

takluyver:
This is an educated guess, but I think it may be that npm dependencies are used for build & development tooling, and bower dependencies are what we package into the notebook. If that's the case, it's probably OK for there to be some packages in both.

blink1073:
My impression is that bower is used for things that must be used as script tags, and npm for things that can be bundled.  The number of things that must be used as script tags is shrinking, and may be reduced to MathJax and RequireJS.  Even those can be npm installed and then moved from node_modules to a known location for serving.

blink1073:
I would say wait until we have implemented something like https://github.com/jupyter/notebook/issues/1554#issuecomment-245144140 to start exploring this.

minrk:
Until webpack was adopted, client-side js exclusively came from bower, and build utilities (e.g. bower itself, less) exclusively came from node_modules. The use of webpack has mixed this up quite a bit, as now *some* client-side js comes from node_modules (eventually, after building), while *most* comes from bower.

gnestor:
Right, webpack allows us to bundle npm modules along with our source in such a way that it can be loaded in the browser. If we decide to stick with webpack vs. abandoning it and returning to strictly RequireJS, then we could potentially ditch bower all together seeing as though even [mathjax](https://www.npmjs.com/package/mathjax) is available as an npm module.

gnestor:
I've created a PR to start trying things out because my attempts in https://github.com/jupyter/notebook/pull/1765 were breaking Travis and Appveyor left and right.

minrk:
Frontend deps are no longer provided by npm.

PR #1783, opened 2016-09-22
Remove duplicate dependencies in bower

gnestor:
An initial stab at https://github.com/jupyter/notebook/issues/1778. 

I've enabled "Allow edits from maintainers" so feel free to push to this PR...

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -6,17 +6,15 @@
     "bootstrap": "components/bootstrap#~3.3",
     "bootstrap-tour": "0.9.0",
     "codemirror": "~5.18",
-    "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.2.0",
     "google-caja": "5669",
     "jquery": "components/jquery#~2.0",
+    "jquery-typeahead": "~2.0.0",
     "jquery-ui": "components/jqueryui#~1.10",
     "marked": "~0.3",
     "MathJax": "components/MathJax#~2.6",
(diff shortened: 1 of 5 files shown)


blink1073:
I think you might also need to remove some script tags in `page.html` or the other templates.

gnestor:
@blink1073 Good call! Now what about instances where `moment` is defined using RequireJS (https://github.com/jupyter/notebook/blob/53ab302a98ca24b2a104a4ec7bb049ff4dcc1a09/notebook/static/base/js/utils.js#L6)? Will that look for `moment` in node_modules? If not, then we need to replace these with `import moment from 'moment'`...

blink1073:
It looks like `moment` is in the compiled WebPack bundle, so the code looks good as is.  Eventually, it may make sense to switch to all-CommonJS code in `static/`.

gnestor:
@blink1073 @minrk It looks like the tests are failing because ["ReferenceError: Can't find variable: Promise" ](https://travis-ci.org/jupyter/notebook/jobs/162261858#L922). I removed `es6-promise` from bower so now the tests will rely on webpack for `Promise`. Is the source not being built by webpack before tests are run?

minrk:
I believe the source is built by webpack in the tests, otherwise no js would be loaded because the main.js would be undefined.

blink1073:
Is the webpack even running?  It looks like the build is failing because you need to `require('webpack')` in order to use the `ProvidePlugin`.

PR #1793, opened 2016-09-28
[WIP] Export modules for AMD consumption

blink1073:
cf https://github.com/jupyter/notebook/issues/1554#issuecomment-245144140

- Break the pages into `vendor` and `main` using the WebPack [DllPlugin](https://github.com/webpack/docs/wiki/list-of-plugins#dllplugin).
- Use a script to generate a RequireJS bundle from the DLL manifest.
- The script will walk through the manifest contents and create `define` functions of the form:
     
```javascript
define('base/js/auth', function() { return vendor(100); });
```

- We get the path by replacing `./notebook/static/` with `''` and `./node_modules/` with `''` and stripping out the `.js` (along with making sure it has a `.js` extension first).
- The HTML template will include the RequireJS bundle after including the vendor bundle and RequireJS itself.
- After that, all modules on the page are available using the previous AMD require semantics.

Here is what the manifest looks like for the `terminal`:

```json
{
  "name": "vendor",
  "content": {
    "./notebook/static/terminal/js/main.js": 1,
    "./node_modules/xterm/src/xterm.css": 2,
    "./node_modules/css-loader/index.js!./node_modules/xterm/src/xterm.css": 3,
    "./node_modules/css-loader/lib/css-base.js": 4,
    "./node_modules/style-loader/addStyles.js": 5
  }
}
```

And the resulting shim is:

```javascript
define("terminal/js/main", function () { return vendor(1); });
define("css-loader/lib/css-base", function () { return vendor(4); });
define("style-loader/addStyles", function () { return vendor(5); });
```

## Status
The WebPack bundles and manifests and shims are created and the `tree` page is functional.


blink1073:
Drat.  In order for this to work all of the modules will have to be converted be fully CommonJS.  Otherwise the bulk of the modules end up in `1.vendor.js` and are not in the DLL manifest.

blink1073:
@gnestor, @minrk.  I am 100% convinced this will work if I implement the above, but it is going to be time consuming and invasive.

minrk:
I like it in principle, but I don't like how much work we are making you do on this. @blink1073 what do you think about the tradeoffs in terms of effort of getting webpack to not break requirejs here vs going back to requirejs with no webpack?

blink1073:
Looking at the `package.json`, all of the `dependencies` are available on `bower`.  I'd say go back to being `bower` + `requirejs`-only is the way to go.  It is the least brittle solution.

minrk:
> I'd say going back to being bower + requirejs-only is the way to go

That makes sense to me. I took a stab at exactly that a while back, but too many webpack changes had been made for it to be really easy, and I ran out of time.

blink1073:
I'll close this one then.  :(

PR #1798, opened 2016-09-29
Add base aliases for nbextensions apps

minrk:
Includes `--log-level` alias that was missing from install/enable applications


Carreau:
Kicked travis. +1.

Carreau:
And travis is happy.

PR #1802, opened 2016-10-01
Set ws-url data attribute when accessing a notebook terminal

ashishdahiya:
Notebook is not able to launch a terminal when notebook tries to connect to a web socket url specified via [NotebookApp.websocket_url option](http://jupyter-notebook.readthedocs.io/en/latest/config.html#options). 

The problem occurs when notebook/templates/terminal.html does not set 'data-ws-url' attribute in the body. Two files need to change to fix this:

[1] A change needs to be made to notebook/templates/terminal.html [similar to notebook.html](https://github.com/jupyter/notebook/blob/23ca9602c2fdd0cd705ec97c2feee909e7865e67/notebook/templates/notebook.html#L30) to support connecting to a websocket url specified via NotebookApp.websocket_url option when launching a terminal.

[2] terminal/js/main.js needs to be updated to consume wsUrl attribute set by terminal.html. If wsUrl is not set, fallback to location.host.

### DIFF
--- a/notebook/static/terminal/js/main.js
+++ b/notebook/static/terminal/js/main.js
@@ -38,8 +38,12 @@ require([
 
     var base_url = utils.get_body_data('baseUrl');
     var ws_path = utils.get_body_data('wsPath');
-    var ws_url = location.protocol.replace('http', 'ws') + "//" + location.host
-                                    + base_url + ws_path;
+    var ws_url = utils.get_body_data('wsUrl');
+    if (!ws_url) {
+        // trailing 's' in https will become wss for secure web sockets
+        ws_url = location.protocol.replace('http', 'ws') + "//" + location.host;
+    }
+    ws_url = ws_url + base_url + ws_path;
     
     var header = $("#header")[0];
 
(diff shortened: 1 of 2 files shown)


minrk:
Thanks! What are you setting ws_url for?  I had actually thought we removed that option because it was never used.

ashishdahiya:
Awesome! Thanks for the merge.

ws_url is a very useful option for us at [Instabase](https://www.instabase.com). 

We ran into a problem with GCloud load balancers. There are two types of them: Network and HTTP. (1) Network LB supports websockets but doesn't forward client ip to upstream, (2) HTTP LB forwards client ip to upstream but doesn't support websockets. 

We partitioned our traffic wherein websocket traffic went to a network lb and the rest went to a http lb. ws_url option saved the day for us and allowed us to keep our infrastructure stack fairly simple.

ashishdahiya:
Is there a timeline to push 4.3 milestone to GA? We use conda to install notebook and would love to have these changes in a conda notebook package.

Carreau:
Backported.

PR #1807, opened 2016-10-04
set dirty flag when output arrives

minrk:
and when the prompt changes

we were previously only setting dirty when execution finishes, leaving the notebook incorrectly clean during many changes before the output is complete.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -349,9 +349,11 @@ define([
             },
             iopub : {
                 output : function() { 
+                    that.events.trigger('set_dirty.Notebook', {value: true});
                     that.output_area.handle_output.apply(that.output_area, arguments);
                 }, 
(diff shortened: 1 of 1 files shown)


takluyver:
Makes sense. The Travis failures look like they're not just random, though.

minrk:
Indeed, improper use of this and that. Should be okay, now.

Carreau:
Backported.

garrett-pnnl:
I modified my local codecell.js with the changes listed here, but the autosave does not save when jupyter is executing a code. Are there any other modifications that needs to be done to enable this feature?

takluyver:
`codecell.js` and the other js files alongside it are bundled into a single big javascript file for quicker loading. The [contributing guide](http://jupyter-notebook.readthedocs.io/en/latest/contributing.html) has details of how to get the development version and build the css & js.

PR #1809, opened 2016-10-04
Container resize button

jsjaspreet:


Carreau:
Hi there ! 
Thanks for your pull request ! 

I am not sure what you are trying to achieve here, from what I can read you are trying to add API to use a versioning system as a content manager right ? And need APIs to pass the commit message around. 

First this need to have some conflicting in a couple of files:

```
notebook/static/notebook/js/menubar.js
notebook/templates/page.html
```

Second as a advice I would suggest you configure your editor to not autostrip trialing whitespace, it increases the chances of conflict, and screw up git history (in particular blam which is useful). 


If you have any questions feel free to ask !

ivanov:
Hey there, I'm going through old issues and it seems to me that it makes sense to close this one.

There are too many irrelevant commits here that just whitespace related, and the PR has been sitting idle for 3 and a half months without any activity.


Thanks everyone and happy hacking! :bowtie:

PR #1826, opened 2016-10-11
don’t clear callbacks for cell output until next execute

minrk:
adds callbacks.clear_on_done bool flag for kernel callbacks.

Default: true (preserve pre-existing behavior)

Setting to false means that the caller is taking responsibility for clearing callbacks.
Use this in cells to only clear callbacks on the next execution, for better handling of async output.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -340,6 +340,7 @@ define([
     CodeCell.prototype.get_callbacks = function () {
         var that = this;
         return {
+            clear_on_done: false,
             shell : {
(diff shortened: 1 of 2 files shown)


minrk:
Related to ipython/ipython#9969

minrk:
cc @jasongrout @blink1073 for JupyterLab. Following #9969 and other async-output discussions, I'm increasingly convinced that this is the right approach: Leave it to the cell to dispose of callbacks, which should happen on the next execution, and otherwise leave callbacks for a given execution registered as long as the output area associated with it is on the page. This eliminates the issue of trying to guess whether output is done.

PR #1828, opened 2016-10-11
Fix typo

piec:
Hi, I think some define(...) should be require(...)

### DIFF
--- a/docs/source/extending/frontend_extensions.rst
+++ b/docs/source/extending/frontend_extensions.rst
@@ -61,7 +61,7 @@ instance:
 
     // file my_extension/main.js
 
-    define([
+    require([
         'base/js/namespace'
     ], function(
         Jupyter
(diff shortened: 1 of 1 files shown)


Carreau:
I'm unsure about that.  Define is just defining a module while `require()` will actually  trigger its evaluation (when a module have a main function). Here we do want to _define_ them in particular they all rely on the `base/js/namespace` module to be available, and evaluation will be triggered by the value returned in
```
return {
             load_ipython_extension: load_ipython_extension
         };
```

Did you actually have things that don't work with `define` and do work with `require` ?

piec:
OK! I was ill informed about define, it's used to define and require at the same time
Thanks

Carreau:
> OK! I was ill informed about define, I didn't realize it's used to define and require at the same time
> Thanks

yeah the naming is quite... unfortunate. You can think of `define()` as the lazy of the two. It tells you that something does exist and will evaluate it when needed. This is generally useful when you have circular dependencies. 

It took me a while to understand it as well. It's kinda weird !


Thanks for try though ! We really appreciate !

PR #1833, opened 2016-10-13
Fix: Carriage return should behave like in console

lgeiger:
This PR makes carriage return symbols behave like in `jupyter console`. It uses [this function](https://github.com/lgeiger/escape-carriage/blob/master/index.js) to escape the carriage return.

Here's a example:
```python
print('this sentence\rthat\nwill make you pause')
```

`jupyter notebook` master prints:
```
that
will make you pause
```

`jupyter console` prints:

 ```
that sentence
will make you pause
```

\cc @rgbkrk

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -9,7 +9,7 @@ define([
     'codemirror/mode/meta',
 ], function(CodeMirror, moment, _){
     "use strict";
-    
+
     // keep track of which extensions have been loaded already
     var extensions_loaded = [];
 
@@ -236,7 +236,7 @@ define([
         "ansi-cyan-intense",
         "ansi-white-intense",
     ];
-    
+
     function _getExtendedColors(numbers) {
         var r, g, b;
         var n = numbers.shift();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: notebook/static/base/js/utils.js [hunk 1]
```diff
@@ -441,17 +441,7 @@ define([
...
-    function fixCarriageReturn(txt) {
-        var tmp = txt;
-        do {
-            txt = tmp;
-            tmp = txt.replace(/\r+\n/gm, '\n'); // \r followed by \n --> newline
-            tmp = tmp.replace(/^.*\r+/gm, '');  // Other \r --> clear line
-        } while (tmp.length < txt.length);
-        return txt;
-    }
+    var fixCarriageReturn = require('escape-carriage');
```
rgbkrk: It's been a while since I knew what the state of asset requiring was in the notebook, so I'll let someone else chime in here on how we should do a `require`.

gnestor:
Thanks @lgeiger. Since escape-carriage is such a tiny library, I suggest that you just copy and paste the function body vs. adding it as an NPM dependency. I created a [branch](https://github.com/gnestor/notebook/commit/d2a6221c2368e73da454cace4ccf8907807ba96f) and tested and it works fine. 

image

lgeiger:
@gnestor Thanks for testing!
We can just use your branch if you like keep the dependency list small. 👍

gnestor:
Alrighty, I'll just submit a new PR 👍

PR #1836, opened 2016-10-14
Fix: Carriage symbol should behave like in console

gnestor:
Continuation of https://github.com/jupyter/notebook/pull/1833

This PR makes carriage return symbols behave like in `jupyter console`. It uses [this function](https://github.com/lgeiger/escape-carriage/blob/master/index.js) to escape the carriage return.

Here's a example:
```python
print('this sentence\rthat\nwill make you pause')
```

`jupyter notebook` master prints:
```
that
will make you pause
```

`jupyter console` prints:

 ```
that sentence
will make you pause
```

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -444,12 +444,13 @@ define([
     // Remove chunks that should be overridden by the effect of
     // carriage return characters
     function fixCarriageReturn(txt) {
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks ! Appveyor failure is unrelated. 
Merrging.

Carreau:
Backported.

Issue #1820, opened 2016-10-07
Change favicon when busy


alangrafu:
It would be neat to provide some visual notification when a notebook is running, like changing the color of the favicon.

takluyver:
I like this idea. What about putting three dots through the middle of the logo while the kernel is busy?

PR #1837, opened 2016-10-15
Change favicon when busy

gnestor:
Closes https://github.com/jupyter/notebook/issues/1820

I’m open to _busy_ icon suggestions…

screenshot


takluyver:
:+1:

jasongrout:
I really like this change. Can we also not change the page title anymore, now that we have this much less distracting indicator in a tab? This may be for a different PR, but if you could do it in this PR, that would be awesome.

takluyver:
I'd also favour not changing the title - my tabs are often quite short, and "(Busy)" will take up ~all the visible space.

gnestor:
@jasongrout @takluyver Updated this PR with your suggestion...

notebook/static/notebook/js/notificationarea.js
```diff
@@ -242,27 +242,42 @@ define([
...
 
         this.events.on('kernel_ready.Kernel', function () {
             that.save_widget.update_document_title();
             $kernel_ind_icon.attr('class','kernel_idle_icon').attr('title','Kernel Idle');
             knw.info("Kernel ready", 500);
+            change_favicon('/static/base/images/favicon.ico');
         });
 
         this.events.on('kernel_idle.Kernel', function () {
             that.save_widget.update_document_title();
```
takluyver: We probably don't need this on each kernel idle event now, as we're no longer changing the title on kernel busy.

gnestor: I'm not sure I follow... It current changes the favicon to the busy icon on `'kernel_busy.Kernel'` and back to normal on `'kernel_idle.Kernel'` same for `'kernel_starting.Kernel kernel_created.Session'` and `'kernel_ready.Kernel'`. If we remove this from kernel idle, then the favicon will indicate that the kernel is busy even after it's idle/done.

takluyver: I meant specifically this line that calls `update_document_title()`.

gnestor: 👌

Carreau:
I kicked Appveyor.

takluyver:
Appveyor is having constant problems with Python 2 right now, failing with the error "The input line is too long." I don't know what to do about that - I'm ignoring it for now...

minrk:
Nice!

oessaid:
Is this supposed to work on Safari ? I'm using Version 12.1 (14607.1.40.1.4), enabled favicons in settings>tabs but the icon is stuck on its idle version regardless of kernel state.

Issue #1821, opened 2016-10-08
iopub rate limits should not be unlimited by default


jasongrout:
Since we live in a finite world with limited resources, we should pick reasonable defaults for these limits that were introduced in #931:

https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L918

And thus disarm yet another foot-gun.

Thanks to @williamstein for conversations about this.

jasongrout:
Also, it would be good to insert an iopub message that says that further iopub messages were limited, so that the error is stored in the notebook and it is clear to the end user. Right now only the server operator gets to see what the problem was.

williamstein:
Thanks.  I frequently have to manually edit ipynb files using vim to remove huge output that users of SageMathCloud generate.  These are often students who are first time Python/Jupyter users, so epically large unintentional output is not uncommon.

jasongrout:
Thanks for bringing that up. For the record, the rate limits here don't address performance problems that may come up with rendering files that already have many outputs.

gnestor:
@jasongrout @Carreau @minrk Any suggestions as to what a good default value would be?

minrk:
I agree 100%, but I don't know how to pick a sensible default other than human testing. My guess is that message rate limit should be between 100-1k per second. I have less of a sense of the flow rate limit. I'll play with some values and come back with something to start with.

minrk:
I've played around a bit, and I'm having difficulty coming up with reasonable default numbers. With almost any value, I can cause problems in the browser with certain output patterns, but also be annoyingly restrictive with others, using the same values.

Right now, I'm looking at:

- data_rate = ~1M
- msg_rate = 500-1000
- rate_limit_window = 2-3

williamstein:
>  I can cause problems in the browser with certain output patterns, but also be annoyingly restrictive with others, using the same values.

I don't know how you guys implemented things, but in SMC at least, whenever rate limitation happens, there is a big clear message saying it happened with instructions about how to raise/lift the limitation.

minrk:
Yup, we've got a big clear message saying that rate limit has been hit and how to raise it. It's still really unpleasant when it happens, so I don't want the default limit to be too low.

gnestor:
I guess that the best way to test this is to merge into master and start using it ourselves. I just submitted a PR. Have a look and let's merge and start testing...

PR #1846, opened 2016-10-19
Set finite rate limits for iopub

gnestor:
Closes https://github.com/jupyter/notebook/issues/1821

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -915,15 +915,15 @@ class NotebookApp(JupyterApp):
         help="Reraise exceptions encountered loading server extensions?",
     )
 
-    iopub_msg_rate_limit = Float(0, config=True, help="""(msg/sec)
+    iopub_msg_rate_limit = Float(1000, config=True, help="""(msg/sec)
         Maximum rate at which messages can be sent on iopub before they are
         limited.""")
 
-    iopub_data_rate_limit = Float(0, config=True, help="""(bytes/sec)
+    iopub_data_rate_limit = Float(1000000, config=True, help="""(bytes/sec)
         Maximum rate at which messages can be sent on iopub before they are
         limited.""")
 
-    rate_limit_window = Float(1.0, config=True, help="""(sec) Time window used to 
+    rate_limit_window = Float(3, config=True, help="""(sec) Time window used to 
         check the message and data rate limits.""")
 
     def parse_command_line(self, argv=None):


minrk:
Thanks! I've got some refinements to the rate measuring and messages I came up with while testing. I'll open another PR with those when I get a chance.

williamstein:
I'm going to comment here, since I don't know where to comment.  I just spent a while trying to use the rate limiting command line options for jupyter, and  it doesn't seem to solve the real problem.   I can't think of any useful way to even use it, which is a bummer, since I desperately need something like this (as students are constantly causing trouble, which leads to disaster).  

The basic test problem is this:

1. Drawing a plot should work.
2. Doing something dumb like this should be severely limited: `for i in range(50000): print('hi')`.

The currently-implemented functionality treats the output of 1 and 2 exactly the same.  Both result in a single message.  However, rendering and dealing with a 0.3MB plot (say), is completely different than dealing with a single output message that is 0.3MB of strings of text, which is very hard on the browser to render...  Ignoring this difference is not good.

In SageMathCloud, images are just sha1 hashes that reference a remote image store.  So rate limiting based on data transmitted works fine.

williamstein:
I'm going to run with rate limitations (as you implemented them) in production for a week on SMC and see what happens with user feedback.  Maybe this does work fine.  I could be completely wrong above. Data will tell.

gnestor:
@williamstein This is a very common problem for beginner Jupyter users. I think that rate limits should be imposed by default with available config flags to override them. Let us know how it goes and how this can be improved.

williamstein:
Having deployed the current released rate limiting code in practice for a week, there is one surprise way in which Jupyter's approach is far inferior to what we do with SageMathCloud worksheets.    In SMC, when you hit the rate limits, the error message says, basically "type the following (right here in this worksheet!) to increase the limit".  That works fine, and we've never had a complaint.  However, asking the users to restart the backend jupyter server with different parameters is disruptive and isn't even possible in some environments (e.g., SMC).

gnestor:
@williamstein Do you suggest that we handle limits the same that SMC does?

williamstein:
@gnestor -- YES.  If you can, please make it so users can increase the rate limit from some default by explicitly running code in their notebook.    As it is now, we are getting a continuous stream of support requests due to people hitting even rather large rate limits.  E.g., here's one from 30 minutes ago: "Hi, i am suddenly having trouble displaying a web map. Previously (3-4 weeks ago) it took a while to draw the map however it would display after a minute or two. Now i get an error when the "display(m)" function in cell 5 executes via the button press in cell 6. The error is as follows:  "iopub data rate exceeded. The notebook server will temporarily stop sending iopub  messages to the client in order to avoid crashing it.  To change this limit, set the config variable  `--NotebookApp.iopub_data_rate_limit`."  

The error message provides 0 help to my users.  They don't explicitly start Jupyter notebook servers -- they use Jupyter notebooks.  Instead, the error should be: "iopub data rate exceeded. The notebook server will temporarily stop sending iopub  messages to the client in order to avoid crashing it.  To change this limit, ... run this code ...; for more help and options related to iopub rate limitation, type ... (or visit http://...)"

Carreau:
Let's see if we can make an APi on the server that can allow us to change the limiting rate. 

As the rate limiting is on the server, it might not be super easy to have something from the kernel that change the rate limit, but we might have a toggle in the UI.

williamstein:
> As the rate limiting is on the server, it might not be super easy to have something from the kernel that change the rate limit, but we might have a toggle in the UI.

Presumably there is some way to send commands via code that goes directly to the server and not the kernel?  E.g., maybe
```
%%server blah
```
The communications channel exists to do this, since the client UI uses it...

Just something in the UI would only solve half the problem.  It definitely does not solve the entire problem, which is that some notebooks specific need different rate limitations than others, and this should be something one can set in the notebook itself.

Carreau:
> Presumably there is some way to send commands via code that goes directly to the server and not the kernel? E.g., maybe

Yes and no. 

No: The server act strictly as a proxy, it does not peek at the messages. 
Yes: %blah could return a javascript object that would trigger some API call to the frontend.
Yes: You could inject a magic which is aware of the server and can communicate to it via another method.

But in the general case the kernel is in the dark as to whether the server is running it, or not. It could even be connected to 2 "clients" one that would be rate limited and not the other and would still be unaware of this fact.

PR #1854, opened 2016-10-27
Improve performance of directory listing

takluyver:
- stat() the file once and reuse the result to save on syscalls. Not as good as using scandir(), but should still help.
- Don't re-check if parent directories are hidden for each file in the directory.

Could do with testing on Windows.

Prompted by discussion on the mailing list that listing medium size directories can be very slow.

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -4,9 +4,11 @@
 # Distributed under the terms of the Modified BSD License.
 
 
+import errno
 import io
 import os
 import shutil
+import stat
 import warnings
 import mimetypes
 import nbformat
@@ -22,7 +24,7 @@ from traitlets import Any, Unicode, Bool, TraitError, observe, default, validate
 from ipython_genutils.py3compat import getcwd, string_types
 from . import tz
 from notebook.utils import (
-    is_hidden,
+    is_hidden, is_file_hidden,
     to_api_path,
 )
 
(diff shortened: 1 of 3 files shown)


minrk:
👍! Do you think the existing tests provide sufficient Windows coverage, since we have Appveyor now? Or are there some manual tests we should have a go at?

takluyver:
I've pinged Oscar on the mailing list to try it - I think the existing tests can give us reasonable confidence that I haven't broken it, but I'm not sure if it will actually make a difference to the performance.

In my brief tests on Linux, listing a large-ish directory seemed to be maybe 10% faster, but it was already in the 50ms range, so it's not a big win.

rgbkrk:
I'm currently polling the notebook contents API so even a little improvement is helpful.

oscar6echo:
Cf. https://groups.google.com/forum/#!topic/jupyter/SRVKUyJYPtM

@takluyver Thanks for investigating

takluyver:
To summarise those results here, it was taking 14-24 seconds without this pull request, and seemed to take about 9s with this PR. I was hoping for a bit more of a speedup, but it's still a definite improvement - and we don't seem to have broken anything.

Carreau:
> To summarise those results here, it was taking 14-24 seconds without this pull request, and seemed to take about 9s with this PR. I was hoping for a bit more of a speedup, but it's still a definite improvement - and we don't seem to have broken anything.

Thanks !

I guess to speedup more we would need to actually profile see where the bottleneck is.

Merging.

PR #1855, opened 2016-10-27
iopub rate limit refinements

minrk:
- reset counter on status: idle (avoids Run All running into limits)
- fix rate measurements to exclude messages not sent due to rate limiting
- fix formatting of rate-exceeded messages, tweak text a bit
- after pausing output, don't resume sending until rate has dropped 20% below limit, to avoid rapid back & forth across the limit

cc @haraldschilly who brought up the Run All issue on gitter.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -915,7 +915,7 @@ class NotebookApp(JupyterApp):
         help="Reraise exceptions encountered loading server extensions?",
     )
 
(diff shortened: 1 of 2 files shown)


haraldschilly:
Hi, I made the effort to run this PR of the notebook inside an SMC project. The errors I was talking about previously do indeed disappear with this fix!

minrk:
@haraldschilly thanks!

Carreau:
Thanks !

PR #1856, opened 2016-10-27
Revert long path workaround for appveyor

takluyver:
It should work without this again now that Miniconda 4.2.11 is out, incorporating conda/conda#3349.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,16 +16,6 @@ build: off
 install:
   - appveyor DownloadFile "https://raw.githubusercontent.com/pelson/Obvious-CI/master/bootstrap-obvious-ci-and-miniconda.py"
   - cmd: python bootstrap-obvious-ci-and-miniconda.py %CONDA_INSTALL_LOCN% %platform% %CONDA_PY:~0,1% --without-obvci
-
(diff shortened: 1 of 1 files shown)


takluyver:
Seemingly not. Maybe the `-latest-` version hadn't been changed when that ran.

Carreau:
Bump to 5.1

takluyver:
I don't really know what to do about this one. It's supposed to be fixed in conda, but every time I've tried re-running it on Appveyor, it fails again. Let's give it another try...

blink1073:
conda-forge recipes explicitly upgrade conda: https://github.com/conda-forge/jupyterlab-feedstock/blob/799b17e8271081316117a523086bfcfe0504218e/appveyor.yml#L63, could that be it?

takluyver:
Interesting. I don't think that's it, because it happens after activating the environment, which is the step that fails for us. But it reminds me that Appveyor already has Miniconda installed, so I'll make another PR which tries using that rather than downloading a copy.

takluyver:
Closing as #2109 has landed, and that seems like a neater way of doing things.

PR #1858, opened 2016-10-28
Add hover styles to HTML table row

gnestor:
Side-port of PR merged with jupyterlab: https://github.com/jupyterlab/jupyterlab/pull/1089

Follow up on https://github.com/jupyter/notebook/pull/1776

screenshot


Carreau:
Thanks !

PR #1861, opened 2016-10-31
fix notebook mime-type on download links

minrk:
notebooks are not `application/json`, they are `application/x-ipynb+json`.

This fixes Safari adding the `.json` extension to notebooks on download.

### DIFF
--- a/notebook/files/handlers.py
+++ b/notebook/files/handlers.py
@@ -43,7 +43,7 @@ class FilesHandler(IPythonHandler):
         
         # get mimetype from filename
         if name.endswith('.ipynb'):
-            self.set_header('Content-Type', 'application/json')
+            self.set_header('Content-Type', 'application/x-ipynb+json')
         else:
             cur_mime = mimetypes.guess_type(name)[0]
             if cur_mime is not None:


willingc:
@minrk @takluyver There were discussions a few weeks ago about custom MIME types that resulted in this doc https://jupyter.readthedocs.io/en/latest/reference/mimetype.html We should make sure that we are not propagating more than one type for the same notebook. I'm assuming this PR only covers notebook downloads.

Ref: https://github.com/jupyter/jupyter/pull/209, https://github.com/jupyter/jupyter_client/issues/207

Carreau:
cc @rgbkrk . if we use jupyter, should register vnd.jupyter ?

minrk:
This was chosen as the mimetype for notebooks at [a dev meeting](https://ipython.hackpad.com/2014-IPython-Weekly-Development-Meetings-part-2-bxJTQTtMzh0#:h=August-14) [some years ago](https://github.com/jupyter/jupyter/issues/68#issuecomment-207520317), though admittedly almost nothing has used or documented it until now. I wasn't planning on starting a new discussion about changing the mimetype for notebooks, but we can have that discussion again if people prefer.

@willingc so far, we don't have a place for the mimetypes of notebooks themselves, just the outputs/interactive pieces used in output areas, etc. We should add this one to that page, though.

willingc:
@minrk I'm totally cool with other folks decide. Merely raising the question to avoid confusion now or down the road ;-) Agree we should add this one to the doc page.

rgbkrk:
`x-` or `vnd` are fine - they're both valid for these "custom" (non-IANA declared). I primarily care about ensuring consistency and documented specs. :smile:

`application/x-ipynb+json` was declared sometime ago, we can keep on with it.

PR #1865, opened 2016-11-01
don't store default editable/deletable metadata

minrk:
it fills notebooks with redundant flags

these should only be stored if defined and not-default

follow-up to #1482


takluyver:
:+1:

Carreau:
Thanks !

PR #1866, opened 2016-11-02
Add a `register_mime_type` method to OutputArea

gnestor:
This method will allow nbextensions to easily register a new mime type and a renderer for that mime type.

Usage:

```js
var append_json = function (json, md, element) {
  var type = ‘application/json’;
  var toinsert = this.create_output_subarea(md, 'output_json rendered_html', type);
  this.keyboard_manager.register_events(toinsert);
  ReactDOM.render(React.createElement(JSONTree, {data: json}), toinsert[0]);
  element.append(toinsert);
  return toinsert;
};
Object.getPrototypeOf(OutputArea).register_mime_type(‘application/json’, append_json, true);
```

See https://github.com/gnestor/notebook_json/tree/render-output for a working implementation  using this method.

Some considerations:

* Should this method not allow the replacement over append functions (e.g. registering a new append function for 'text/markdown' 'text/html')?
* We probably needs tests for this...

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -526,7 +526,8 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
      * @return {jQuery} A selector of all cell elements
      */
     Notebook.prototype.get_cell_elements = function () {
-        return this.container.find(".cell").not('.cell .cell');
+        var container = this.container || $('#notebook-container')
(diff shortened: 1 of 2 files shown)


rgbkrk:
Awesome to see this in the works!

Carreau:
Thanks ! We can write tests in a separate PR. 
I think it's ok to replace, kernels like haskell are already hacking around it.

Carreau:
Backported.

PR #1870, opened 2016-11-03
Fix link for rtd badge

willingc:
RTD badge in README.md is not properly updating. Changing link back to http as that is still what RTD is recommending for their markdown badge


takluyver:
Thanks Carol

PR #1874, opened 2016-11-03
Start documenting how to use Comms

takluyver:
See gh-1869

Ping @rgbkrk and @willingc .

I didn't see a good place to put this document in the landing page, so I've stuck it under user docs for now.

Issue #1869, opened 2016-11-02
Document comms


takluyver:
Answering a SO question, I realised that we don't appear to have any docs on how to use comms.

rgbkrk:
It is pretty open ended.

takluyver:
It is! I'd be happy to close it once we have

- [ ] some explanation of what comms are
- [ ] how to register targets in Python, connect them from JS, and send messages both ways.
- [ ] link to the messaging spec docs about comm messages for authors of other kernels.

willingc:
Hey @takluyver, Thinking this was the SO question: http://stackoverflow.com/questions/40352628/messaging-and-jupyter-notebooks

If you or others want to do a detailed outline, that would be a helpful start. Alternatively links to the above 3 points in the code or other docs. Thanks.

rgbkrk:
I'm _very_ happy to pair on this or even explore it from a protocol level (see https://github.com/jupyter/jupyter_client/issues/198). I'm not a good one to speak to for the JS APIs on jupyter/notebook for comms as we're taking a slightly different approach in nteract: https://github.com/nteract/nteract/pull/840

takluyver:
I'm going to try to make some basic docs now. If you want to take it further, that would be great.

willingc:
@takluyver If you make bullets, I'm happy to wordsmith. We can certainly iterate 👍

takluyver:
Closing as we now have some docs - that doesn't stop us improving them more!

PR #1875, opened 2016-11-04
Minor edits to comms doc

willingc:
@takluyver @rgbkrk Some clarifying edits to #1874 #1869

### DIFF
--- a/docs/source/comms.rst
+++ b/docs/source/comms.rst
@@ -7,10 +7,8 @@ update widget state.
 
 A comm consists of a pair of objects, in the kernel and the frontend, with an
 automatically assigned unique ID. When one side sends a message, a callback on
-the other side is triggered with that message data. For example, the kernel can
-send a message to the frontend which triggers a callback by the frontend to
-the kernel with the message data. Either side, the frontend or kernel, can
-open or close the comm.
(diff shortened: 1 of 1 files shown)


docs/source/comms.rst
```diff
@@ -7,8 +7,10 @@ update widget state.
 
 A comm consists of a pair of objects, in the kernel and the frontend, with an
 automatically assigned unique ID. When one side sends a message, a callback on
-the other side is triggered with that message data. Either side can open or
-close the comm.
+the other side is triggered with that message data. For example, the kernel can
+send a message to the frontend which triggers a callback by the frontend to
+the kernel with the message data. Either side, the frontend or kernel, can
```
takluyver: This doesn't quite match my understanding. If there is a message from the kernel to the frontend, the callback is in the frontend. Callbacks are just function calls, so they're not 'to' another part.

takluyver:
Other than that, :+1: thanks :-)

docs/source/comms.rst
willingc: Thanks. I'll strike that sentence.

Carreau:
Backported.

PR #1879, opened 2016-11-08
support routing updating outputs with display_id

minrk:
notebook implementation of jupyter/jupyter_client#209

- display(id=...) always adds something to the page
- update(id=...) updates only existing things on page

A single display_id can be used multiple times on the page, and all instances will be updated and kept in sync.

related: https://github.com/ipython/ipykernel/pull/204 , https://github.com/ipython/ipython/pull/10048

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -224,23 +224,13 @@ define([
             json.text = content.text;
             json.name = content.name;
             break;
+        case "execute_result":
+            json.execution_count = content.execution_count;
         case "update_display_data":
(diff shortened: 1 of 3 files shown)


minrk:
An example notebook for testing with these messages (doesn't require IPython PRs): https://nbviewer.jupyter.org/gist/minrk/1a1e56a611f1ff1e2645f733bdd0e381

minrk:
Only missing thing: updating execute_result.

rgbkrk:
Looking good!

minrk:
No longer WIP, this should be all set, now.

Issue #1886, opened 2016-11-11
list running servers occasionally gives wrong information


jim22k:
`notebook.notebookapp.list_running_servers()` will occasionally give wrong information, listing servers which are no longer running.

I traced this down to having outdated nbserver-xxxx.json info files in my runtime dir. It seems that there is a bug in the `os.unlink` call which tries to remove these old files.

Normally, having the old files hanging around isn't a problem, but on the rare chance that the pid becomes active again from another process, list_running_servers will show that outdated information as a currently running server. This becomes a problem for things like `nbopen()` which attempt to use existing servers rather than spawn new ones.

takluyver:
Fixed by #1887

PR #1887, opened 2016-11-11
Fix bug when attempting to remove old nbserver info files

jim22k:
When list_running_servers find an old nbserver-xxxx.json info file, it attempts to delete it. It needs to use the full path rather than the filename. An exception was being raised, but the error was suppressed silently.

This fixes issue #1886


takluyver:
Thanks! I had spotted the issue with pid reuse before, but I didn't realise we were failing to clean up files here.

Issue #1589, opened 2016-07-04
Selecting downwards to markdown cell is broken


takluyver:
1. Create a new notebook.
2. Make a bunch of cells, and convert them to markdown mode.
3. Select a cell other than the last cell, and ensure you are in command mode.
4. Press shift-down to extend the selection downwards.

Expected result: the cell selected in step 3 and the cell immediately below that are highlighted.

Actual result: the cursor moves down (as if I wasn't holding shift).
- Replicated on master with Firefox and Chromium
- It seems to be the visual highlighting going wrong; changing the cell type still affects all the cells that should be selected.
- The deciding thing seems to be if the cell the selection is extending _to_ is a markdown cell: I can select down from a markdown cell to a code cell, but not down from a code cell to a markdown cell.
- There is another bug when a range has been selected going up from the bottom, and then you press up or down without shift. I hope they're related and will both be fixed at once, but if not, I'll report that in more detail.

takluyver:
Who do I prod about broken UI stuff these days?

takluyver:
This is still broken, and I do consider it a blocker for 4.3.

michaelpacer:
@takluyver do you think this is something that we can work on on the hacktrain? Or will it likely require internet access to debug appropriately? 

I.e., do you feel that between you, @Carreau, @yuvipanda and I and the local codebase we will likely have the relevant knowledge needed to fix this?

takluyver:
I think there's a pretty good chance we can work out the problem on the hacktrain. :-)

minrk:
Looks like this bug was introduced in master by the attachments code, and 4.x is not affected (there is no `MarkdownCell.prototype.select` in 4.x)).

I just followed the repro steps on the 4.x branch, and it appears to be working as intended. Retargeted as 5.0. Feel free to send it back to 4.3 if you can reproduce it with the 4.x branch.

PR #1891, opened 2016-11-14
Fix extending range selection down to markdown cell

takluyver:
The 'moveanchor' argument was not getting passed down to the parent class' `select()` method.

Closes gh-1589

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,7 +14,7 @@ platform:
 build: off
 
 install:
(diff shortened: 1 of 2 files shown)


takluyver:
I've also now fixed the issues with appveyor in this project, by switching where we get ObviousCI from, as described in pelson/Obvious-CI#68

rgbkrk:
Yay for fixing appveyor while you were at it.

PR #1896, opened 2016-11-14
Accept JSON output data with mime type "application/*+json"

gnestor:
Currently, output data must be a string unless the mime type equals “application/json”. This doesn’t allow JSON output data to be rendered by mime types such as “application/vnd.plotly.v1+json” or “application/vnd.vega+json”. This change simply allows JSON output data to be rendered by a custom mime renderer if the mime type includes “application/“ and “json”.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -275,7 +275,7 @@ define([
         }
         var data = bundle.data;
         $.map(OutputArea.output_types, function(key){
-            if (key !== 'application/json' &&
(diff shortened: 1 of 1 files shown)


takluyver:
For the record, this follows on from jupyter/nbformat#43

gnestor:
Thanks @takluyver, I was getting validation errors when saving the notebook. I just did a dev install of nbformat and no more validation errors 👍

rgbkrk:
Yeah we definitely need an nbformat release soon, possibly to include the JSON Array ability /cc @minrk.

PR #1903, opened 2016-11-17
Allow websocket connections from scripts

minrk:
scripts don't set origin on connection

we allow these on API requests, websockets should match.

cc @rgbkrk: is there a reason we block 'missing origin' websockets when we allow the same for API requests? Is it possible for a request from a sensible browser to be missing Host and/or Origin?

### DIFF
--- a/notebook/base/zmqhandlers.py
+++ b/notebook/base/zmqhandlers.py
@@ -133,13 +133,9 @@ class WebSocketMixin(object):
         if origin is None:
             origin = self.get_origin()
         
-        # If no header is provided, assume we can't verify origin
-        if origin is None:
-            self.log.warning("Missing Origin header, rejecting WebSocket connection.")
-            return False
-        if host is None:
-            self.log.warning("Missing Host header, rejecting WebSocket connection.")
-            return False
+        # If no origin or host header is provided, assume from script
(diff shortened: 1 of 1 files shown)


blink1073:
"All browsers will send an Origin header. You can use this header for security (checking for same origin, whitelisting/ blacklisting, etc.) and send a 403 Forbidden if you don't like what you see. However, be warned that non-browser agents can just send a faked Origin. Most applications will reject requests without this header."

https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers

rgbkrk:
As @blink1073 pointed out, browsers are not supposed to be missing Host and Origin.

At the time we wrote this, I assumed scripted clients would have to set Host and Origin.

minrk:
Is it possible for browsers (or more precisely, ajax calls in a browser) to omit Host? If so, we should probably only be allowing missing Origin as a signal for scripts, and not allowing Origin set but no Host.

rgbkrk:
I agree with that logic. I'm ok with what's here too for what it's worth.

blink1073:
Ditto to what @rgbkrk said.

minrk:
Thanks, both!

PR #1904, opened 2016-11-17
json_errors should be outermost decorator on API handlers

minrk:
in order to apply to 403 raised by web.authenticated


takluyver:
Seems sensible, thanks.

PR #1755, opened 2016-09-10
Adds edit and view buttons, #1752

dsblank:
Puts [Edit] and [View] buttons before the trashcan. Allows editing and viewing of files. Addresses #1752

PR #1905, opened 2016-11-17
Adds edit and view buttons

### LINKED ISSUE #1752: separate edit and view links
Some files, e.g. `.html` files, can logically be both viewed (via `/files/`) and edited (via `/edit/`). Currently, we have just the one link that goes to `/edit/` or `/files/`, depending on the file type. We don't currently have separate links to these to URLs for files, but we probably should.

One option is to add view, edit to the list of actions along with delete, rename. Another is to add the links inline, perhaps on hover.

JupyterLab doesn't have this issue since it has an Open With... dialog, so if we take much longer to get around to this one, pointing to JupyterLab may end up being the solution.

I know this has been discussed a few places, but I can't find a dedicated issue for it, so opening here. If someone finds a duplicate, we can close this one.

cc @dsblank for interest, @ellisonbg for UI

gnestor:
Continuation of #1755

closes #1752

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -956,18 +956,18 @@ define([
     NotebookList.prototype.view_selected = function() {
         var that = this;
         that.selected.forEach(function(item) {
-            var item_path = item.path;
-            var w = window.open(utils.url_path_join('/files', item_path));
-	});
-    }
+            var item_path = utils.encode_uri_components(item.path);
+            var w = window.open(utils.url_path_join(that.base_url, '/files', utils.encode_uri_components(item_path)), IPython._target);
+      	});
+    };
 
     NotebookList.prototype.edit_selected = function() {
         var that = this;
         that.selected.forEach(function(item) {
-            var item_path = item.path;
-            var w = window.open(utils.url_path_join('/edit', item_path));
-	});
-    }
+            var item_path = utils.encode_uri_components(item.path);
+            var w = window.open(utils.url_path_join(that.base_url, '/edit', utils.encode_uri_components(item_path)), IPython._target);
+      	});
+    };
 
     NotebookList.prototype.duplicate_selected = function() {
         var message;
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -139,6 +139,8 @@ define([
             $('.download-button').click($.proxy(this.download_selected, this));
             $('.shutdown-button').click($.proxy(this.shutdown_selected, this));
             $('.duplicate-button').click($.proxy(this.duplicate_selected, this));
+            $('.view-button').click($.proxy(this.view_selected, this));
+            $('.edit-button').click($.proxy(this.edit_selected, this));
             $('.delete-button').click($.proxy(this.delete_selected, this));
 
             // Bind events for selection menu buttons.
@@ -597,6 +599,20 @@ define([
             $('.delete-button').css('display', 'none');
         }
 
+        // View is visible when an item is renderable or downloadable
+        if (selected.length > 0 && !has_directory) {
+            $('.view-button').css('display', 'inline-block');
+        } else {
+            $('.view-button').css('display', 'none');
+        }
+
+        // Edit is visible when an item is editable
+        if (selected.length > 0 && !has_directory) {
+            $('.edit-button').css('display', 'inline-block');
+        } else {
+            $('.edit-button').css('display', 'none');
+        }
+
         // If all of the items are selected, show the selector as checked.  If
         // some of the items are selected, show it as checked.  Otherwise,
         // uncheck it.
@@ -936,6 +952,22 @@ define([
             }
         });
     };
+
+    NotebookList.prototype.view_selected = function() {
+        var that = this;
+        that.selected.forEach(function(item) {
+            var item_path = item.path;
+            var w = window.open(utils.url_path_join('/files', item_path));
+	});
+    }
+
+    NotebookList.prototype.edit_selected = function() {
+        var that = this;
+        that.selected.forEach(function(item) {
+            var item_path = item.path;
+            var w = window.open(utils.url_path_join('/edit', item_path));
+	});
+    }
 
     NotebookList.prototype.duplicate_selected = function() {
         var message;
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -600,7 +600,9 @@ define([
         }
 
         // View is visible when an item is renderable or downloadable
-        if (selected.length > 0 && !has_directory) {
+        if (selected.length > 0 && !has_directory && selected.every(function(el) {
+            return el.path.match(/html?|jpe?g|png|gif|tiff?|svg|bmp|ico|pdf|doc|xls/);
+        })) {
             $('.view-button').css('display', 'inline-block');
         } else {
             $('.view-button').css('display', 'none');
@@ -965,7 +967,8 @@ define([
         var that = this;
         that.selected.forEach(function(item) {
             var item_path = utils.encode_uri_components(item.path);
-            var w = window.open(utils.url_path_join(that.base_url, '/edit', utils.encode_uri_components(item_path)), IPython._target);
+            var item_type = item_path.endsWith('.ipynb') ? 'notebooks' : 'edit';
+            var w = window.open(utils.url_path_join(that.base_url, item_type, utils.encode_uri_components(item_path)), IPython._target);
       	});
     };
 
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -959,7 +959,8 @@ define([
         var that = this;
         that.selected.forEach(function(item) {
             var item_path = utils.encode_uri_components(item.path);
-            var w = window.open(utils.url_path_join(that.base_url, '/files', utils.encode_uri_components(item_path)), IPython._target);
+            var item_type = item_path.endsWith('.html') ? 'view' : 'files';
+            var w = window.open(utils.url_path_join(that.base_url, item_type, utils.encode_uri_components(item_path)), IPython._target);
       	});
     };
 
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -560,9 +560,9 @@ define([
             $('.rename-button').css('display', 'none');
         }
 
-        // Move is visible iff at least one item is selected, and none of them
+        // Move is visible if at least one item is selected, and none of them
         // are a running notebook.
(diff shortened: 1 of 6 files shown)


gnestor:
@takluyver It's working for me. However, on Chrome, the user must enable pop-ups for the domain, otherwise only the first file in the selection will be opened and a pop-up warning will display in the address bar.

takluyver:
I think that's what Min expected on #1755. I'm inclined to make it a single-item action so that we're not dealing with browser pop-up blockers.

gnestor:
@minrk @Carreau What say you? If we decide that "View" (viewing a notebook opens the JSON file in a separate tab) and "Edit" (editing a notebook opens the JSON file in a text editor) should only work for single files vs. multiple files, then no action is required and this PR can be closed.

minrk:
@gnestor I think this is still useful. I'd just set the limit at 1, so that it doesn't run into pop-up blockers.

notebook/static/tree/js/notebooklist.js
```diff
@@ -595,6 +597,20 @@ define([
             $('.delete-button').css('display', 'none');
         }
 
+        // View is visible when an item is renderable or downloadable
+        if (selected.length > 0 && !has_directory) {
```
gnestor: @dsblank I'm just curious: How do you use "View" and "Edit" for grading? I assumed "View" would open the notebook, but it opens the raw JSON file in the browser and "Edit" opens the JSON file in a text editor...

dsblank: Sometimes, I just want to open the notebooks as notebooks (with kernel), be able to run, view, and print them. That was my intention by adding these buttons for the group selection operations.

gnestor: Ok, then I believe the expected behavior for "View" is to open notebooks as a notebook vs. raw file (see my screencap below for an example). Given that, we can remove "View" entirely because viewing and editing any file other than a notebook is the same (e.g. open it in a text editor) and editing a notebook is to open it.

gnestor:
"View": 

view

"Edit":

edit

notebook/static/tree/js/notebooklist.js
dsblank: What about HTML files?

gnestor: Ah, good point! HTML files seem like the one file type that should provide a "View" button. I'll try that now...

gnestor:
"Edit" notebook and markdown files:

edit

"View" HTML file:

view

notebook/static/tree/js/notebooklist.js
dsblank: Great... that looks useful! What about all other types that one can view in the browser, such image files (jpg, gif)? Could there be a way to register editors for particular type of files?

gnestor: I think the primary purpose of this PR is to allow "View" and "Edit" on multiple files at once. If we only allowed these button for single file selections, then there would be no need for "Edit" as it has the same effect of clicking on a file, and we are realizing that "View" is only needed for HTML and image files...so if we remove this for multiple files, we are left with only "View" for a handful of file types.

Carreau:
Sweet ! +1

minrk:
Nice!

Carreau:
Looking forward to have an occasion to use that !

notebook/static/tree/js/notebooklist.js
```diff
@@ -595,6 +597,22 @@ define([
             $('.delete-button').css('display', 'none');
         }
 
+        // View is visible when an item is renderable or downloadable
+        if (selected.length > 0 && !has_directory && selected.every(function(el) {
+            return el.path.match(/html?|jpe?g|png|gif|tiff?|svg|bmp|ico|pdf|doc|xls/);
```
gnestor: @dsblank I updated the "View" condition to a regex of common browser-renderable file types. I could create an API method on NotebookList to register additional file types, but it would probably be easier to just decide on them now...

takluyver:
I think it might be a bit surprising for new users that 'View' and 'Edit' on a notebook both show you the raw JSON. I understand that that fits in with editing other text-based file types, but I would naively expect 'edit' on a notebook to be equivalent to opening it normally, and 'view' to be equivalent to the nbconvert-based 'print preview' we offer from the file menu.

How hard would it be to make the buttons say something like 'View raw' and 'Edit raw' if you have a notebook selected?

gnestor:
@takluyver I've revised so that "View" is only visible for HTML, image, and PDF files and "Edit" will open notebooks as notebooks and all other files in the text editor.

notebook/static/tree/js/notebooklist.js
gnestor: You're welcome! If you take a look at JupyterLab and specifically https://github.com/jupyterlab/jupyterlab_json, you will see that lab allows for extensions to both render output by mime type and render files by extension, so there is a lot opportunity to create custom file handlers.

gnestor:
One more use case to consider: What if you wanted to look at or edit the JSON of a notebook file? There is currently no way to open a notebook in a text editor. Do we want to accommodate that with an "Edit Raw" or "Edit Metatdata" for selected notebook files?

dsblank:
I don't think we need a button for that.... Maybe be able to open from editor would be enough?

takluyver:
Thanks @gnestor, that sounds good to me. I don't think there's a pressing need for 'edit raw'.

I'd still probably agree with Min that I'd only show the buttons with a single file selected, and avoid popup blockers altogether. I think the main thing we wanted to fix is that there's no easy way to view an HTML file, because clicking on it opens an editor. I don't feel strongly about this, though.

We should think about the **security implementations**. An HTML file could contain Javascript, which can execute when you 'view' it. Because it's served from the notebook domain, that Javascript can presumably start a kernel and use it to execute code (unless I've overlooked something). That seems a bit unexpected for a 'view' action. Is there any way we can mitigate this? Ping @rgbkrk

rgbkrk:
> We should think about the security implementations. An HTML file could contain Javascript, which can execute when you 'view' it. Because it's served from the notebook domain, that Javascript can presumably start a kernel and use it to execute code

That's exactly correct. If the view needs to be supported it needs to be in a sandboxed iframe (since we don't have access to other subdomains).

dsblank:
As for showing the buttons when there is only 1 vs. more selected, often times the better option doesn't make itself clear until you try to write the documentation (or teach someone how to use it). If you start describing how to use this, it becomes complicated... buttons showing/not showing because different combinations of types (files, directories) and counts (one, two) are being selected. If you can make an option more flexible and easier to explain, then go with that.

takluyver:
Aha, so we would use that to disable JS scripts entirely for the HTML you're viewing?

I'm going to bump this PR to milestone 5.0 as it seems like a bigger thing to work out, and I'd like to get 4.3 released fairly soon.

rgbkrk:
> I'm going to bump this PR to milestone 5.0 as it seems like a bigger thing to work out, and I'd like to get 4.3 released fairly soon.

I certainly agree with that.

rgbkrk:
It's not that we need to disable JS so much as we want to disable access to the Jupyter REST APIs and websockets.

screen shot 2016-11-19 at 1 11 29 pm

```html
<iframe sandbox="allow-scripts" src="test.html">
</iframe>
```

However, you can disable JS by omitting the `allow-scripts` set in the sandbox options:

```html
<iframe sandbox src="test.html">
</iframe>
```

takluyver:
Ah, I see, a sandboxed iframe has a special origin, so any attempt to access the Jupyter REST APIs would be blocked by the cross-origin protection. Thanks, I didn't know about that :-)

rgbkrk:
The other security options that are available include disabling scripts on specific paths with a more stringent [content security policy](https://developer.mozilla.org/en-US/docs/Web/Security/CSP).

takluyver:
I think that the general 'treat this as a different origin' security measure should do the trick.

@gnestor does this make sense? I imagine we will need a separate handler at something like `/viewhtml/`, which renders pages something like:

```html
<!-- /viewhtml/foo.html -->
<iframe sandbox="allow-scripts" src="/files/foo.html">
</iframe>
```

And viewing an HTML file would take you to this page instead of the `/files/` URL. I don't see any way around serving the HTML insecurely at `/files/`, but at least we can make it a bit harder to get there.

gnestor:
@takluyver Makes sense. Thanks!

@rgbkrk Could we modify the CSP in lieu of redirecting to a page with an iframe? If so, that seems like a simpler solution...

@minrk I see that you created the [FileRedirectHandler](https://github.com/jupyter/notebook/blame/master/notebook/base/handlers.py#L593), how would you go about disallowing HTML files from accessing the Jupyter REST APIs? If we went with iframe approach, would it make sense to create a `file.html` template and do something like:

```py
self.render_template('file.html', source='/files/foo.html')
```

minrk:
> how would you go about disallowing HTML files from accessing the Jupyter REST APIs?

I think IFrames are the only way to do this. I'm not sure of the details, though. A properly configured IFrame should prevent access to the API via existing CORS, etc.

gnestor:
@minrk Ok, then we'll take that approach. Would you recommend creating a template and doing something like this in `FilesRedirectHandler`:

```py
self.write(self.render_template('file.html', source='/files/foo.html'))
```

minrk:
Yeah, a template for the iframe page sounds quite sensible.

gnestor:
@takluyver Yes, I will do this next. Thanks for the reminder 👌

gnestor:
Update: 

* Edit button: Edit functionality is currently broken: https://github.com/jupyter/notebook/issues/2071
* View button: HTML files use the `/view/filename.html` route which loads the files inside of an iframe to prevent access to the Jupyter API via existing CORS
* I added a test for the new view route

I would appreciate any feedback regarding tests and UX before merging.

notebook/static/tree/js/notebooklist.js
```diff
@@ -597,6 +599,22 @@ define([
...
+        })) {
+            $('.view-button').css('display', 'inline-block');
+        } else {
+            $('.view-button').css('display', 'none');
+        }
+
+        // Edit is visible when an item is editable
+        if (selected.length > 0 && !has_directory) {
```
takluyver: Do we want to whitelist/blacklist files for editability? E.g. you can't open a png in the text editor.

gnestor: We could use a regex pattern like above to include/exclude specific extensions. I just opened a PNG in the text editor and it's pretty useless and harmless, so I'm neutral.

takluyver: That was what I was thinking. :-) I'd probably go for a blacklist. Maybe there are too many types to be worth trying to list them all, though. Do we have mimetypes here? Maybe we could block edit on `image/*`, with an exception for `image/svg+xml`.

takluyver:
OK, let's land this, we can refine things later.

PR #1907, opened 2016-11-18
Add `Notebook.render_cell_output` method

gnestor:
Continuation of https://github.com/jupyter/notebook/pull/1866 and https://github.com/gnestor/notebook_json/pull/5 depends on this.

Since extensions are _usually_ loaded after the notebook has loaded and outputs have been rendered, a mime type renderer extension requires an API method for re-rendering the output of a cell after the extension has loaded and registered its mime type using `OutputArea.register_mime_type`. This method allows extensions to do that by replacing a cell with itself which will trigger a re-render.

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -1640,6 +1640,17 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
             first_inserted.focus_cell();
         }
     };
+    
+    /**
+     * Re-render the output of a CodeCell.
+     */
+    Notebook.prototype.render_cell_output = function (code_cell) {
+        var cell_data = code_cell.toJSON();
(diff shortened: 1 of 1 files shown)


gnestor:
@Carreau Care to review?

Carreau:
+1, would just calling `fromJSON` to an existing cell have the same effect. Do you need to delete and re-re-create ?

Carreau:
Also for information to anyone passing by, this should be safe as we don't call clear-output so it's not affected by the [widget security bug](https://blog.jupyter.org/2016/09/26/ipywidget-security-release/)

gnestor:
Thanks for clarifying that!

gnestor:
> +1, would just calling `fromJSON` to an existing cell have the same effect. Do you need to delete and re-re-create ?

@Carreau You're right, you don't need to! I will submit a new PR removing that bit!

PR #1908, opened 2016-11-18
Improve `Notebook.render_cell_output`

gnestor:
Continuation of https://github.com/jupyter/notebook/pull/1907

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -1647,7 +1647,9 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
     Notebook.prototype.render_cell_output = function (code_cell) {
         var cell_data = code_cell.toJSON();
         var cell_index = this.find_cell_index(code_cell);
+        var trusted = code_cell.output_area.trusted;
         this.clear_output(cell_index);
+        code_cell.output_area.trusted = trusted;
         code_cell.fromJSON(cell_data);
     };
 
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/notebook.js
```diff
@@ -1647,9 +1647,8 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
     Notebook.prototype.render_cell_output = function (code_cell) {
         var cell_data = code_cell.toJSON();
         var cell_index = this.find_cell_index(code_cell);
-        this.delete_cell(cell_index);
-        var new_cell = this.insert_cell_at_index(cell_data.cell_type, cell_index);
-        new_cell.fromJSON(cell_data);
+        this.clear_output(cell_index);
```
Carreau: That `clear_output()` could dangerous. It will mark the cells as trusted (even if it is not), so might render untrusted javascript IIUC.

gnestor: Ok, good call. So we have to either clear output, delete the cell, or something equivalent or `CodeCell.fromJSON` will add additional outputs to the cell. Do you have any other ideas?

Carreau: We can change `clear_output()` to not change the trust model, so it would be _safe by default_ , and add a `trust()` method, that you can possibly chain.

notebook/static/notebook/js/notebook.js
```diff
@@ -1647,9 +1647,10 @@ import {ShortcutEditor} from 'notebook/js/shortcuteditor';
     Notebook.prototype.render_cell_output = function (code_cell) {
         var cell_data = code_cell.toJSON();
         var cell_index = this.find_cell_index(code_cell);
-        this.delete_cell(cell_index);
-        var new_cell = this.insert_cell_at_index(cell_data.cell_type, cell_index);
-        new_cell.fromJSON(cell_data);
+        var trusted = code_cell.output_area.trusted;
+        this.clear_output(cell_index);
+        code_cell.output_area.trusted = trusted;
```
gnestor: This is not the most elegant solution but it should work, yes?

minrk: yup, should be fine

Carreau:
Great ! Thanks !

PR #1912, opened 2016-11-20
Fix highlighting of Python code blocks

takluyver:
The configuration on RTD doesn't default to Python highlighting, it seems.

http://jupyter-notebook.readthedocs.io/en/latest/comms.html


takluyver:
Self merging, since this is rather trivial.

PR #1917, opened 2016-11-23
load minified typeahead.min.js

### LINKED ISSUE #1154: requirejs config wrong in 4.1.0 - tries to load typeahead from wrong location
The requirejs config in the notebook page template in 4.1.0 is wrong, and tries to load the jquery-typeahead dependency from the wrong location. The line https://github.com/jupyter/notebook/blob/4.1.0/notebook/templates/page.html#L41 should be changed from:

```
            typeahead: 'components/jquery-typeahead/dist/jquery.typeahead'
```

to

```
            typeahead: 'components/jquery-typeahead/dist/jquery.typeahead.min'
```

(i.e., add `.min` to the end of the path)

As a result, `jupyter notebook --NotebookApp.ignore_minified_js=True` hangs when showing a notebook and just shows a blank screen.

minrk:
not needed on master, which already loads the right thing.

closes #1154

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -37,7 +37,7 @@
             moment: 'components/moment/moment',
             codemirror: 'components/codemirror',
             termjs: 'components/xterm.js/dist/xterm',
-            typeahead: 'components/jquery-typeahead/dist/jquery.typeahead'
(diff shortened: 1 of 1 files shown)


minrk:
ignore appveyor failure - appveyor shouldn't be running on 4.x.

rgbkrk:
Min adding in min

PR #1920, opened 2016-11-23
Change max-width for tables to 150px

gnestor:
Addresses https://github.com/jupyter/notebook/pull/1776 and follow up from https://github.com/jupyterlab/jupyterlab/pull/1274


minrk:
@gnestor this is marked as 4.3, but #1776 isn't. Presumably they should both have the same target, unless I misunderstood something. Are we doing this in 4.3 or 5.0?

gnestor:
Oh right, we had a discussion about this and decided to bump it to 5.0. Thanks @minrk!

Issue #759, opened 2015-11-26
Bug in formatting mathjax/LaTeX inconsistently depending on chosen delimiters, markdown interaction


michaelpacer:
I just discovered that the notebook formats the same LaTeX input in different ways depending upon your choice of equation delimiter for LaTeX/mathjax which traditionally should not matter.

An example of this bug (plus a png of the correctly rendered incorrect rendering, since github apparently has its own bugs regarding displaying LaTeX/mathjax embedded in notebooks) can be found at https://gist.github.com/michaelpacer/2b627ffe8f10cde53926.

Put briefly, it's not supposed to matter whether you choose to delimit LaTeX formatted math with `\(…\)` and `$…$` for inline equations or `\[…\]` or `$$…$$` for displayed equations. Of course, because `\` is used as an escape character we have to instead use`\\(…\\)` `\\[…\\]`, which in theory is fine. However, because LaTeX formatted subscripts use the `_` symbol, and that conflicts with markdown's interpretation of `_…_` as an indicator of italics, it seems that notebook parses these two different ways of delimiting equations differently. If you use `\\(…\\)` `\\[…\\]` then you need to escape underscores `\_` but that means that your LaTeX is now invalid outside of the notebook (as can be seen in the rendering of `\_` when using `$…$` and `$$…$$` as the delimiters). 

For now it seems safer to guide people to just use `$…$` and `$$…$$`, since then at least it will interpret valid LaTeX properly (this has the unfortunate side-effect of conflicting with one of the [LaTeX "deadly sins"](http://ctan.mirrorcatalogs.com/info/l2tabu/english/l2tabuen.pdf), but I think that's a less big issue than writing inaccurate LaTeX to get accurate formatting, imho). 

I'm not sure if there is going to be an easy or even possible fix for this (since I know this kind of parsing is incredibly difficult to manage) but I was befuddled with why my properly formatted LaTeX was not being rendered at all, so I figured I'd bring it to people's attention.

minrk:
It might be in how we escape math to avoid markdown mucking it up. Markdown shouldn't be touching the latex inside these blocks.

michaelpacer:
Closed by #2220

PR #1935, opened 2016-11-30
Add \[ \] and \( \) as maths delimiters

jfbercher:
As mentioned in [nbconvert/#477](https://github.com/jupyter/nbconvert/issues/477) and earlier in [notebook#759](https://github.com/jupyter/notebook/issues/759), \\[ \\]  and \\( \\) are not recognized as maths delimiters, and the equations not correctly displayed. 
This is not intuitive for LaTeX users (as it is now the recommended way to delimit maths) and causes problems when converting to eg html since the contents of those maths blocks is considered as markdown. 
The PR attempt to solve that by updating `mathjaxutils.js`
- the regex MATHSPLIT is updated to account for these new delimiters
- new start/end delimiters are matched

### DIFF
--- a/notebook/static/notebook/js/mathjaxutils.js
+++ b/notebook/static/notebook/js/mathjaxutils.js
@@ -58,7 +58,7 @@ define([
 
     // MATHSPLIT contains the pattern for math delimiters and special symbols
     // needed for searching for math in the text input.
-    var MATHSPLIT = /(\$\$?|\\(?:begin|end)\{[a-z]*\*?\}|\\[\\{}$]|[{}]|(?:\n\s*)+|@@\d+@@)/i;
+    var MATHSPLIT = /(\$\$?|\\[\[\(\)\]]|\\(?:begin|end)\{[a-z]*\*?\}|\\[\\{}$]|[{}]|(?:\n\s*)+|@@\d+@@)/i
 
     //  The math is in blocks i through j, so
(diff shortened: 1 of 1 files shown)


takluyver:
I'm inclined to say sorry, but no to further additions to Markdown syntax, even if they're now the recommended way to delimit maths in Latex. The more special exceptions we add to Markdown parsing, the messier it gets.

rgbkrk:
Thank you for taking the time to make a pull request.

The primary reason we can't support `\[` is that it is valid Markdown/CommonMark for an escaped bracket.

[As written](https://talk.commonmark.org/t/mathematics-extension/457/18?u=rgbkrk) by @jgm:

> ...the problem with that syntax is that `\[` and `\(` already have clearly defined meanings: they are escaped `[` and `(` characters, and it is very important that there be a way to escape these characters in certain circumstances.

We all want to support mathematics typesetting. For maintainers and operators of notebook server deployments, we can't support extensions to the syntax that are unambiguous. This does mean having to wait it out for others to solve within CommonMark or contributing there.

rgbkrk:
All that being said, it's worth looking at 

> `\[ \]` and `\( \)` are not recognized as maths delimiters, and the equations not correctly displayed.

as a usability issue where the notebook could detect likely LaTeX and provide a notification/suggestion to use `$$` and `$`. I'm not volunteering to write that though. :wink:

PR #1938, opened 2016-12-01
don't check origin on token-authenticated requests

minrk:
adds LoginHandler.should_check_origin classmethod API

While testing, I noticed that we were checking origin and authentication on CSP violation reports, but browsers are sending CSP reports with no Origin, meaning that no CSP report will ever be accepted. That doesn't make sense to me, so I disabled both - a CSP report can now be made from *anywhere* (no auth, no check). @rgbkrk does that sound right to you? Or should it be authenticated but no origin check?

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -295,8 +295,8 @@ class IPythonHandler(AuthenticatedHandler):
             # No CORS headers deny the request
             allow = False
         if not allow:
-            self.log.warning("Blocking Cross Origin API request.  Origin: %s, Host: %s",
-                origin, host,
+            self.log.warning("Blocking Cross Origin API request for %s.  Origin: %s, Host: %s",
+                self.request.path, origin, host,
             )
         return allow
     
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -79,6 +79,16 @@ class AuthenticatedHandler(web.RequestHandler):
             return 'anonymous'
         return self.login_handler.get_user(self)
 
+    def skip_check_origin(self):
+        """Ask my login_handler if I should skip the origin_check
+        
+        For example: in the default LoginHandler, if a request is token-authenticated,
+        origin checking should be skipped.
+        """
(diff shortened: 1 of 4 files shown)


rgbkrk:
It definitely sounds like the origin check shouldn't happen for CSP reports.

As for authentication, I'm unsure. I'd lean towards authenticated to mitigate any external API usage "lying".

minrk:
@rgbkrk thanks, I'll disable origin check, but keep auth

notebook/services/security/handlers.py
```diff
@@ -3,18 +3,22 @@
...
 
 from ...base.handlers import APIHandler, json_errors
 from . import csp_report_uri
 
 class CSPReportHandler(APIHandler):
     '''Accepts a content security policy violation report'''
+
+    def skip_origin_check(self):
+        """Don't check origin when reporting origin-check violations!"""
+        return True
```
rgbkrk: Nice, simple enough!

notebook/base/zmqhandlers.py
```diff
@@ -126,7 +126,9 @@ def check_origin(self, origin=None):
         
         Tornado >= 4 calls this method automatically, raising 403 if it returns False.
         """
-        if self.allow_origin == '*':
+
+        if self.allow_origin == '*' or (
+            hasattr(self, 'skip_check_origin') and self.skip_check_origin()):
```
rgbkrk: 👍

notebook/base/handlers.py
```diff
@@ -295,8 +306,8 @@ def check_origin(self, origin_to_satisfy_tornado=""):
             # No CORS headers deny the request
             allow = False
         if not allow:
-            self.log.warning("Blocking Cross Origin API request.  Origin: %s, Host: %s",
-                origin, host,
+            self.log.warning("Blocking Cross Origin API request for %s.  Origin: %s, Host: %s",
+                self.request.path, origin, host,
```
rgbkrk: Oooh, excellent addition of the path to the log

gnestor:
@minrk Does this need to be backported before releasing 4.3?

minrk:
Backported

PR #1939, opened 2016-12-01
Add debug log for static file paths

jasongrout:
Too often I'm trying to track down an errant extra static file path (like some javascript installed in some obscure place). This makes it easy to see exactly what is being served.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -40,6 +40,12 @@ non_alphanum = re.compile(r'[^A-Za-z0-9]')
 
 sys_info = json.dumps(get_sys_info())
 
+def log():
+    if Application.initialized():
+        return Application.instance().log
+    else:
+        return app_log
+
 class AuthenticatedHandler(web.RequestHandler):
     """A RequestHandler with an authenticated user."""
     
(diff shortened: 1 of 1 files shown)


jasongrout:
CC @dmadeka1

jasongrout:
Perhaps `log()` doesn't have to be a function? I'm not sure what the specific lifetime concerns are with the application initialization check. @minrk - would the following work just as well?

```
    if Application.initialized():
        log = Application.instance().log
    else:
        log = app_log
```
at the top level?

takluyver:
@jasongrout `FileFindHandler`, which you're modifying, inherits from `IPythonHandler`, which defines the `log` property. So you should be able to use `self.log.debug(...)`, I think, without needing an extra function.

The application won't be initialised at import time, so putting that check at the top level of the module wouldn't work.

minrk:
Arg, I wrote a whole comment last week, but apparently GitHub lost it. What @takluyver said - module-level initialization won't work because import happens too early. This function is fine as-is. @takluyver it's needed here because this method is required to be a classmethod.

gnestor:
@jasongrout Does this need to be backported before releasing 4.3?

minrk:
Backported

PR #1941, opened 2016-12-02
Catch and log handler exceptions in events.trigger

### LINKED ISSUE #1602: document is null error, prevents notebook saving
There appears to be a weird timing bug with showing the logo for the current kernel spec and with the DOM being ready. I typically see this when accessing a remote notebook server where the connection is a bit slow.

screen shot 2016-07-07 at 9 46 47 am

When it happens, the problem is undetectable (short of watching the JS console) until I go to save the notebook. Then I get the dreaded:

at which point I have to run  `Jupyter.notebook._fully_loaded = true` to save my work (if I know this cheat code) or cuss because I can't save whatever I've typed into the notebook since loading it (if I don't know the cheat code).

The problem seems to be coming from here:

https://github.com/jupyter/notebook/blob/fe1490fedbad456ccc8579e671024ed15823a842/notebook/static/notebook/js/kernelselector.js#L135

I don't mind taking a crack at a fix since it's really annoying when it bites. But I need a bit of guidance. Is there or should there be a DOM-ready guard somewhere earlier in the call stack before this method gets invoked, or does this method need to guard against DOM-readiness itself before calling img.show or hide? What about other places in the code?

minrk:
rather than throwing handler errors in the call to `.trigger()`,
which I think we don't expect to happen (at least our code suggest we don't).

Extensions can register buggy event handlers. These should not be able to cause failures in the event-triggering code paths.

This should fix several avenues by which extensions (or other bugs) could prevent a notebook from loading.

This closes #1602 in that the bug doesn't prevent notebook loading, but it doesn't fix the original error @parente is seeing. I don't fully understand the cause there, because notebook loading doesn't start before the page has finished loading.

### DIFF
--- a/notebook/static/base/js/events.js
+++ b/notebook/static/base/js/events.js
@@ -22,5 +22,16 @@ define(['base/js/namespace'], function(Jupyter) {
     Jupyter.Events = window._Events;
     Jupyter.events = window._events;
     
-    return $([window._events]);
+    var events = $([window._events]);
+
+    // catch and log errors in triggered events
+    events._original_trigger = events.trigger;
+    events.trigger = function (name, data) {
+        try {
+            this._original_trigger.apply(this, arguments);
+        } catch (e) {
+            console.error("Exception in event handler for " + name, e, arguments);
+        }
+    }
+    return events;
 });


takluyver:
Sounds like a good idea to me. I'll leave this for a little while for other people to look at.

Issue #1495, opened 2016-05-27
JS error with notebook master on Windows


dhirschfeld:
When trying to run the `jupyter notebook` from source I get a blank page and the following error in the console:
![image](https://cloud.githubusercontent.com/assets/881019/15594465/fd49aa0a-23f8-11e6-96ef-96484613df2d.png)

Being a JS noob I'm not sure how to debug this but am happy to provide any further info if required...

Carreau:
Hi there, sorry for the delay. Can you try to issue a `npm run build` from the root of the source directory ?

dhirschfeld:
Just tried with the latest:
```
git pull
npm run build
python setup.py js css
pip install -e .
```
...but it fails with the same error
![image](https://cloud.githubusercontent.com/assets/881019/16140896/ff86d6ac-3498-11e6-853c-29c75fbab364.png)

Carreau:
hum... can you force refresh, go to the `...` > `settings` of the developper panel and check  `enable source maps`. That might give more information on which file it comes from .

dhirschfeld:
Sorry, pretty low bandwidth for this at the moment - I *think* it may be something to do with incompatible settings or possible the fact that I'm running from a different base_url.  Hopefully I'll have time to revisit it shortly...

dhirschfeld:
Closing because this is pretty old so I doubt it's relevant anymore...

PR #1945, opened 2016-12-03
Fix dependency level of underscore.js in bower.json.  Fixes #1894

### LINKED ISSUE #1894: Blank screen when starting notebook
When running from the latest source from github ( as of 2016-11-14 ),  when starting up the notebook, instead of the main screen, I get a blank screen.  Examination of the JS debugging info yields the error:

```
TypeError: root is undefined
Looks to be in function function __webpack_require__(moduleId) {
```

Is there some piece that I am missing here that isn't documented? I've tried this on 2 different machines with the same result.

JCEmmons:
The bower.json and package.json are out of sync wrt underscore.js, which causes Windows based systems to produce a blank screen when running from source. The fix is simple and eliminates the problem. See also #1495, as I'm pretty sure it's the same issue there.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,6 +15,6 @@
     "MathJax": "components/MathJax#~2.6",
     "requirejs": "~2.1",
     "text-encoding": "~0.1",
-    "underscore": "components/underscore#~1.5"
+    "underscore": "components/underscore#~1.8.3"
   }
 }


takluyver:
Can I ask someone with more understanding of our JS infraststructure than me to double check this? Thanks!

rgbkrk:
Looks great.

Issue #1940, opened 2016-12-02
Variable referenced before assignment, stopping error info printing


MaximilianR:
`exception` can be referenced prior to assignment if `exc_info` isn't truth-y: https://github.com/jupyter/notebook/blame/master/notebook/base/handlers.py#L369

I just had this using jupyter-kernel-gateway:

```
jupyter-kernel_1  | [E 161202 08:08:47 web:996] Uncaught exception in write_error
jupyter-kernel_1  |     Traceback (most recent call last):
jupyter-kernel_1  |       File "/usr/local/lib/python2.7/site-packages/tornado/web.py", line 994, in send_error
jupyter-kernel_1  |         self.write_error(status_code, **kwargs)
jupyter-kernel_1  |       File "/usr/local/lib/python2.7/site-packages/notebook/base/handlers.py", line 358, in write_error
jupyter-kernel_1  |         exception=exception,
jupyter-kernel_1  |     UnboundLocalError: local variable 'exception' referenced before assignment
```

takluyver:
Thanks, #1947 should fix it.

PR #1947, opened 2016-12-04
Ensure variable is set if exc_info is falsey

takluyver:
Closes gh-1940


rgbkrk:
Seems like a good option.

gnestor:
@takluyver Does this need to be backported before releasing 4.3?

takluyver:
Sure. I think it's a minor bug, but it's also a simple fix.

minrk:
Backported

PR #1952, opened 2016-12-06
Set mimerender index in `OutputArea.display_order`

gnestor:
Usage:

```js
OutputArea.register_mime_type(MIME_TYPE, append_json, {
    safe: true,
    index: 1
});
```

Reference implementation: https://github.com/jupyterlab/mimerender-cookiecutter/pull/9


gnestor:
@Carreau @takluyver This is a non-breaking change that I'd like to squeeze into 4.3 before releasing. It makes nbextension registration more consistent with labextension.

takluyver:
Looks OK to me. @Carreau, you added this method, so I'll let you OK the change.

takluyver:
Oh, nvm. @gnestor made the PR to add it, @Carreau merged. I think this is fine, then. Go ahead and backport it.

gnestor:
Backported 👍

PR #1953, opened 2016-12-07
Polyfill promises from the page template.

jasongrout:
Doing this from webpack as a plugin caused https://github.com/ipython/ipywidgets/issues/844 (and probably other issues we haven’t noticed), and also forced us to use the es6-promise Promise object instead of the native one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,10 @@ env:
 before_install:
     - pip install --upgrade pip
     - pip install --upgrade setuptools wheel nose coverage codecov
-    - nvm install 5.6
-    - nvm use 5.6
+    - nvm install 6.9.2
+    - nvm use 6.9.2
+    - node --version
+    - npm --version
     - npm upgrade -g npm
     - npm install
     - 'if [[ $GROUP == js* ]]; then  npm install -g casperjs@1.1.0-beta5; fi'
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "webpack": "^1.12.13"
   },
   "dependencies": {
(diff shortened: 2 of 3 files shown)


jasongrout:
This fixes ipython/ipywidgets#844

ivanov:
👍 🍰  it works! Thanks @jasongrout !

ivanov:
just waiting for the travis tests to pass...

rgbkrk:
I don't think it's node + promises that are failing - I think it's the casper based tests.

jasongrout:
Thanks. You can tell I'm taking a lot of shots in the dark. At first I thought it was phantom not supporting promises, but Slimer didn't seem to help much.

jasongrout:
Ah, this might be the issue, since we are using the babel-loader for some reason: https://github.com/babel/babel-loader#note-transform-runtime--custom-polyfills-eg-promise-library

jasongrout:
I think this is ready to go now. Instead of putting the polyfill on the page, I'm still including it in the webpack bundle, but this time as an explicit polyfill file instead of trying to use the ProvidePlugin mechanism.

When we switch away from using the babel-loader, I think we can reconsider using this approach.

ivanov:
HUZZAH! 🍪 for @jasongrout !

minrk:
@gnestor I marked this as 5.0, because it's webkit-related. The webkit stuff isn't in the 4.x branch. @jasongrout is there still something to do for the 4.x branch?

jasongrout:
I think we're probably good - I wasn't seeing an issue on 4.2.x, at least.

PR #1956, opened 2016-12-07
Make the font size more robust against fickle browser values.

jasongrout:
Also change the line-height multiplier to 1.3, which is closer to what the CSS is (I think it is 1.21 something.)

Thanks to @williamstein and @haraldschilly for bringing this to our attention: https://github.com/sagemathinc/smc/issues/1262

We also re-enable auto scrolling for Firefox, since this might have been the issue.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -98,8 +98,8 @@ define([
             return false;
         }
         // line-height from http://stackoverflow.com/questions/1185151
-        var fontSize = this.element.css('font-size');
-        var lineHeight = Math.floor(parseInt(fontSize.replace('px','')) * 1.5);
+        var fontSize = this.element.css('font-size') || '14px';
+        var lineHeight = Math.floor((parseFloat(fontSize.replace('px','')) || 14) * 1.3);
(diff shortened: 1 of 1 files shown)


jasongrout:
@williamstein, @haraldschilly - can you check to see if this change fixes your issue?

williamstein:
+1 to doing this.  If this code works at all, it'll fix our issue, since on Firefox (when using fromJSON), we were getting fontSize is undefined tracebacks.   I'm not sure what the workflow would be for testing this properly...  

This issue on our side https://github.com/sagemathinc/smc/issues/1343 -- it would be nice if somebody could comment on https://github.com/sagemathinc/smc/issues/1343 when this PR goes into Jupyter master.

jasongrout:
> I'm not sure what the workflow would be for testing this properly...

Is it easy to hand-edit the javascript file with this change to check it? The relevant file should be /static/notebook/js/built/1.main.min.js if you're using notebook master. Search for `prototype._should_scroll` to find the code line.

minrk:
Thanks! I imagine we'll get feedback and tweaks related to other things in 4.3, so we can do a 4.3.1 with the inevitable polish (this included) in a week or two.

PR #1958, opened 2016-12-09
Add output callback overrides

jasongrout:
This allows us to override output callbacks to redirect output messages, and is used to implement the Output widget, for example. It does not redirect status messages or messages on other channels.

This also fixes an async bug where comm message handlers were not being waited on in the message processing loop.

### DIFF
--- a/notebook/static/services/kernels/kernel.js
+++ b/notebook/static/services/kernels/kernel.js
@@ -988,6 +988,7 @@ define([
         var callback_ids = this._msg_callbacks_overrides[msg_id];
         if (!callback_ids) {
             console.error("Popping callback overrides, but none registered", msg_id);
+            return;
         }
         return callback_ids.pop();
     }
--- a/notebook/static/services/kernels/kernel.js
+++ b/notebook/static/services/kernels/kernel.js
@@ -41,6 +41,7 @@ define([
         this.username = "username";
         this.session_id = utils.uuid();
         this._msg_callbacks = {};
+        this._msg_callbacks_overrides = {};
         this._display_id_to_parent_ids = {};
         this._msg_queue = Promise.resolve();
         this.info_reply = {}; // kernel_info_reply stored here after starting
@@ -301,6 +302,7 @@ define([
         this.events.trigger('kernel_restarting.Kernel', {kernel: this});
         this.stop_channels();
         this._msg_callbacks = {};
+        this._msg_callbacks_overrides = {};
         this._display_id_to_parent_ids = {};
 
         var that = this;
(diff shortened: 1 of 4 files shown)


jasongrout:
CC @minrk, @SylvainCorlay, @ivanov.

ivanov:
Thanks, @jasongrout - I'm starting to review this locally, but in the meantime, it looks like there are two tests failling on Travis for this:
```
In /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js:49
  Untitled suite in /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
    assertEquals: cell 1 has the right number of outputs
In /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js:70
  Untitled suite in /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
    assertEquals: Subject equals the expected value
```

notebook/static/services/kernels/kernel.js
```diff
@@ -931,11 +956,30 @@ define([
...
+        if (output_callbacks === void 0) {
+            this._msg_callbacks_overrides[msg_id] = output_callbacks = [];
+        }
+        output_callbacks.push(callback_id);
+    }
+
+    Kernel.prototype.output_callback_overrides_pop = function(msg_id) {
+        var callback_ids = this._msg_callbacks_overrides[msg_id];
+        if (!callback_ids) {
+            console.error("Popping callback overrides, but none registered", msg_id);
```
ivanov: don't we want to return here, or wrap the later return in an `else` so we don't have this message followed by the error.

ivanov:
It would be great to get some tests of this feature, too.

jasongrout:
> It would be great to get some tests of this feature, too.

I added some tests.

jasongrout:
I added some more tests showing the output callback overrides work correctly with the display_id things that @minrk and @rgbkrk added recently.

minrk:
Thanks @jasongrout, and thanks for reviewing, @ivanov!

rgbkrk:
Woo, thank you!

minrk:
@jasongrout I marked this as 5.0, since that's where the display_id is targeted and this touches some of the same things. If you feel strongly that this should be in 4.4, we can try to wrangle that, as well. It's my hope that 4.3 is the last 4.x-series release, short of small and/or important bugfixes. I don't think we are that far off from 5.0, either.

jasongrout:
Thanks. Hopefully notebook 5 will be ready to go in time for us to release widgets 6. If not, let's address it then.

minrk:
@jasongrout I hope so. Thanks, @jasongrout!

Issue #1655, opened 2016-08-04
Revise  markdown"attachment" spec.


Carreau:
It seem we made a few unfortunate decision with the way we use attachments. 
- It does not follow markdown spec: `![alt](/url "title")` imply that we :
  - need to use `attachment://` for the link to be an url
  - escape the filename as `)]` should **not** be present. 

Have to check if the attachment thing was released to know if we can fix it without backward compat. .

Carreau:
@julienr I guess this one will be of interest to you :-) 

> Have to check if the attachment thing was released to know if we can fix it without backward compat.

Attachments are only on master, so it should be safe to make think markdown compliant again !

minrk:
> Have to check if the attachment thing was released to know if we can fix it without backward compat.

It hasn't been, but it's part of why I [pinged people](https://github.com/jupyter/nbformat/issues/40) about releasing nbformat, as well as brought it up at the last two dev meetings.

However, the about-to-be-released nbformat spec in 4.1 *only* defines how attachments are included in the notebook data structure, not how markdown references to them should be made. That bit is in this repo, and only in master, so we are free to fix it if there's something wrong.

julienr:
@Carreau Ooops, that was an oversight on my part, sorry about that. Seems like your patch fixes it though or is there anything else that require changes ?

mgeier:
Sorry if I missed this in previous discussions, but why exactly is the double slash `//` necessary for URLs?

I think this looks quite ugly.
And since no host specification is following, it might not even be a valid URL?
If anything, shouldn't it be *three* slashes (which is even more ridiculous)?

I didn't find anything about this in the Markdown spec (https://daringfireball.net/projects/markdown/syntax).

The current CommonMark spec (http://spec.commonmark.org/0.27/#link-destination) doesn't require it.

RFC 3986, if I understand it correctly, doesn't require the double slash: https://tools.ietf.org/html/rfc3986#section-4.2

minrk:
`://` is not strictly required to be a valid URL, but it is conventional and clear. I don't have a strong preference between `attachment:name.ext` and `attachment://name.ext`, though I think the latter is clearer. I think you are correct that there wasn't a correctness issue, as `attachment:name.ext` is valid, according to markdown in particular and URLs in general.

> And since no host specification is following, it might not even be a valid URL?

I don't believe there is any requirement that hosts come after `://` (see `file://`)

> If anything, shouldn't it be three slashes (which is even more ridiculous)?
 
[...]

PR #1965, opened 2016-12-11
Use attachment:path instead of attachment://path

mgeier:
This partly reverts #1659, see also #1655 and https://github.com/jupyter/nbformat/pull/77.

### DIFF
--- a/notebook/static/notebook/js/textcell.js
+++ b/notebook/static/notebook/js/textcell.js
@@ -247,9 +247,9 @@ define([
                 marked(text, function (err, html) {
                     html = security.sanitize_html(html);
                     html = $($.parseHTML(html));
-                    html.find('img[src^="attachment://"]').each(function (i, h) {
+                    html.find('img[src^="attachment:"]').each(function (i, h) {
                         h = $(h);
-                        var key = h.attr('src').replace(/^attachment:\/\//, '');
+                        var key = h.attr('src').replace(/^attachment:/, '');
                         if (key in that.attachments) {
                             data.attachments[key] = JSON.parse(JSON.stringify(
                                 that.attachments[key]));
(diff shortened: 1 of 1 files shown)


takluyver:
What will happen with notebooks already saved with references to attachments including the `//`?

mgeier:
I guess those attachments wouldn't be shown until their URIs are fixed.

takluyver:
Do we want to try to automatically fix them? Or even just provide a script which people can run to fix them?

minrk:
> Do we want to try to automatically fix them?

since the spec has been out for a few versions, I would.

michaelpacer:
Would the automatic fix "change" them in memory (& so, as displayed) immediately when the notebook is loaded and then overwrite the file system version only when the file is saved? Or would it attempt to rewrite the file immediately upon loading?

takluyver:
Probably the former - that's already how we handle upgrading notebooks in an older format, for instance.

mgeier:
I hope you don't expect me to implement the auto-fix behavior ... I don't know enough JavaScript to do that!

takluyver:
I suspect it could be done in Python. Maybe it belongs in the nbformat package rather than here. @minrk would it be worth doing a minor version bump on the format and using the conversion machinery to implement that?

Probably the trickiest bit would be parsing the markdown to modify it. I proposed a [markdown passthrough](https://github.com/lepture/mistune-contrib/pull/9/files) for mistune, but it's not yet accepted. I expect we could get away with regexes, as `attachment://` is only likely to occur in this context.

minrk:
We could possibly do it in nbformat. In any case, it doesn't need to be in this PR.

michaelpacer:
Can you point to examples of the conversion machinery as it already exists in nbformat that would be used in this case? 

I'm trying to figure out how to handle the header id convention change across nbconvert and the notebook, and while I was planning on trying to do it without altering ipynb metadata, I'm gradually coming to the conclusion that that may be impossible to do without breaking things in unintuitive ways once file sharing comes into the picture.

takluyver:
Here's some of the conversion machinery:
https://github.com/jupyter/nbformat/blob/master/nbformat/v4/convert.py

None of it currently reaches into the markdown to modify things, though - it's all about modifying the JSON structure of the notebook. So it may not be all that useful.

minrk:
I think it will probably be simpler to have the warning and shim in this repo. We can do that in another PR.

Issue #165, opened 2015-06-22
Safer ways of deleting notebooks - or undoing delete?


psychemedia:
_[Picking up on a googlegroups thread: [Disable "Delete Notebooks" option...](https://groups.google.com/forum/#!topic/jupyter/x46rBLUeOHU).]_

After mistakenly deleting a bundle of (fortunately) backed up notebooks, I wondered if there might be safer ways of managing this interaction.

If I have currently selected a number a notebooks, clicking the red delete button currently prompts me for confirmation with the message _Are you sure you want to permanently delete the N files/folders selected?_. I'm often blond to these messages, and if for some reason I don't check which notebooks are ticked I could delete many of them by mistake.

One way to reinforce the confirmation message might be to show a list of notebooks to be deleted in the confirmation dialogue, along with pre-ticked checkboxes that show exactly which notebooks are about to be deleted, allowing you to uncheck ones mistakenly included in the deletion list.

Another option identified in the thread was an option to _undo_ a delete, with 'wastebin cacheing' enabled for thirty days, for example. (I wonder if the checkpointing mechanism can be appropriated for this, redirected to a wastebin area?)

For the paranoid educator in me, being able to disable or move the delete notebook button somehow (a `custom.js` tweak was suggested in the original thread, but I'm not sure what to disable?) could be a good option.

Carreau:
> along with pre-ticked checkboxes that show exactly which notebooks are about to be deleted, allowing you to uncheck ones mistakenly included in the deletion list.

That is IMHO too much or need a confirmation dialog to confirm that what you have ticked it correct.

jasongrout:
+1 for a trash folder somewhere you can restore from

Carreau:
This seem like generic enough that there should be a library for that. This would require a refactor of the content manager though if we want an undo in the UI.

damianavila:
>+1 for a trash folder somewhere you can restore from

:+1:  too... I think is a sensible option...

takluyver:
[Send2Trash](https://pypi.python.org/pypi/Send2Trash) is a way to use the OS-provided trash mechanism, which should probably at least be an option, and possibly the default (since server admins are more likely to configure things like this than desktop users).

I don't think desktop 'recycle bins' are very well designed - as I described before, I think users tend to end up either bypassing them entirely, or accumulating thousands of 'deleted' files. But they are there, and for desktop users, I think it's better to use the existing trash mechanism than come up with our own system.

takluyver:
[...]

PR #1968, opened 2016-12-12
Send files to OS trash mechanism on delete

takluyver:
This adds a config option, defaulting to use trash, and adds a dependency on the pure-Python 'send2trash' package. Linux, Mac and Windows should all be supported.

Alternatively, we could make it default to hard delete (the current behaviour), and let users opt in to trash behaviour. Then Send2Trash could be a soft dependency.

This doesn't touch the UI yet, so you still get a confirmation dialog which inaccurately says it will permanently delete' the file. If we want to do this, we'll need some way for the contents manager to pass
the UI a hint about whether deleting is permanent or not.

Closes gh-165


minrk:
I think a lot of people would be happy about this, so I would say that it makes sense as a default. One case where this might not be sufficient is a hosted service where the notebook server is the only window on the system (e.g. JupyterHub, cloud services). In these cases, without a *read* API for the trash, you get something in the middle: lost access to files, but the data isn't actually cleaned up. But then again, trash is really something along the lines of a `~/.trash` folder, so it's *possible* to dig around and find what you lost.

takluyver:
This still seems like there's enough to work out that I'm bumping it to 5.1.

takluyver:
5.1 is approaching, and there's still quite a bit of work to do. Bumping to backlog.

gnestor:
@takluyver It would be great to ship this! What more needs to be done?

takluyver:
To enable it by default, we'd need a way to get files back from the trash, I think. The major platforms have different trash mechanisms, so that's not straightforward.

To make it an off-by-default option, which offers much less benefit, we'd need a way for the frontend to know whether 'delete' is hard or soft, so that the UI can display appropriate information.

takluyver:
Looking at some of the issues open on send2trash (https://github.com/hsoft/send2trash/issues/9, https://github.com/hsoft/send2trash/issues/2), and thinking a bit about the complexity of the operation in general, I suspect that this would be an ongoing source of problems which we have little ability to fix. I'm not sure whether that outweighs the benefits.

One option would be to define our own trash mechanism separate from the system recycle bin. But I suspect that there's a bunch of complexity we haven't thought about, with things like network filesystems, folders shared with VMs/containers, and folders synced by things like Dropbox. It would also be annoying for people who actually want to delete things if those things were neither deleted nor placed in the normal trash.

gnestor:
I hear ya. It sounds like an elegant solution for the common user (because they can always recover the notebook from the trash) but I'm sue there are many edge cases where this fails.

minrk:
👍 on trying this out in 5.1, for me.

takluyver:
Do you think we should try to do UI changes for 5.1, or leave the UI as is, with no mention of trash?

minrk:
Since there's a server-side flag and we have no UI of *our own* for recovering things from the trash, I think it's okay as-is for now.

takluyver:
Travis failures are due to a bug in Send2Trash, fixed by hsoft/send2trash#12 . I've asked for a new release there.

takluyver:
Optimistically tagged as 5.1 - we're waiting on a new release of send2trash before we can do this. If we're ready to release first, bump this to 5.2.

takluyver:
There's a new release of send2trash - re-triggering tests.

takluyver:
There's another test failure with a unicode character. I can try to make a fix in send2trash, but I'm going to bed soon, so it won't be today.

gnestor:
Ok no prob, let's just mark as 5.2 for now...

minrk:
AOK with 5.2. Let's try to get this in master soon after shipping 5.1, so we start facing the bugs in daily usage.

takluyver:
Re-running tests with send2trash 1.4.1.

takluyver:
Tests are passing, but I think it's still best to merge this after the release, so we can test it in master for a while.

takluyver:
5.1 is now out, if we want to come back to this. As a reminder, this PR currently:

- Makes [Send2Trash](https://pypi.python.org/pypi/Send2Trash) a hard dependency.
- Adds a config option `FileContentsManager.delete_to_trash`.
  - False is the current behaviour (hard delete)
  - True (default) sends files to the platform trash location when you delete them through the contents API
- Does not provide any way to restore trashed files or permanently delete them. Users will have to use their system UI for this.

gnestor:
I just tested and it works for me on macOS 10.12.6. I think this is heaps better than permanently deleting files be default. I imagine it would be really painful to accidentally delete some important notebooks and files and not be able to recover them. It's much less painful to be required to go to the trash to permanently delete files. 

I vote to merge!

gnestor:
Would love to get this in 5.2, but I think it still requires a little more review/testing. Marking as 5.3 for now...

dsblank:
Imagine having a class of new users... please give us a countdown so that we can celebrate the coming of this feature :)

takluyver:
I think the plan was to merge it just after 5.1 so it was in master for a while. We missed doing that, and we're releasing 5.2 sooner than expected, so I guess the new plan is to merge it soon after 5.2 so that it's in master for a while before 5.3.

gnestor:
Roger that @takluyver!

takluyver:
Ping @gnestor @minrk - still happy to try putting this in for 5.3?

minrk:
Let's do it!

gnestor:
👏👏👏

willingc:
Nice addition @takluyver. Small issue with the merged PR: it broke the RTD build :(

takluyver:
Thanks Carol. I think #2964 should fix it.

jt7williams:
Hello, I have deleted a Jupyter notebook by accident. I was using the Hub interface through google chrome, is there a way for me to recover this file?

kevin-bates:
Hmm.  Can you describe the environment (configuration, operating system, etc.) in which Hub is running?  Do you have access to the machine which was/is hosting your spawned Notebook server instance?

On Windows, I'm assuming the deleted file would reside in your `RecycleBin`, on MacOs in your `Trash` and, it looks like, on Unix a `.Trash` directory (see the links on the send2trash README).  This functionality utilizes the [`send2trash`](https://github.com/arsenetar/send2trash) package.   There are some caveats listed on the repository's README.  Here's a link to the [notebook code](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L536-L550) in case that helps.

In the meantime, I'm copying @minrk since he would have insight on how this might behave across a Hub cluster - either single or multi-node and might also have other ideas.

I'm hoping your file is sitting in one of the aforementioned locations.  Good luck.

jt7williams:
Hello, I don't have access to the server, it is the school's server, but it's okay tho, I just recoded the notebook. Thank you 
very much.

st-bender:
This should never have been merged as it results in unexpected behaviour. It causes only trouble for no benefits. Make this a separate menu entry "move to trash" and not disguise it as "delete".

Issue #1830, opened 2016-10-11
Possible security vulnerability to connect to notebook from another tab in the browser or another browser and execute arbitrary code


frejanordsiek:
I saw that Guile just fixed a method that a browser could connect to it even if it only listened on localhost and possibly run arbitrary code (https://lists.gnu.org/archive/html/guile-user/2016-10/msg00007.html). It seems like Jupyter notebooks would be vulnerable to the same sort of attack.

The attack, which is described at http://bouk.co/blog/hacking-developers/ (cited in the guile bug report), is basically malicious a cross protocol scripting attack combined with DNS rebinding to change the attacking code's DNS record to localhost as the article puts it.

Given that Jupyter notebooks listen to a TCP port on localhost using http, it would seem that Jupyter notebooks might be even more vulnerable to the same sort of attack. I do hope I am wrong on this, though, and jupyter notebooks are immune.

The mitigation that Guile suggests, using a unix socket, would obviously not work for Jupyter notebooks. The whole point is to use them in the browser. Only thing I can think of, if it is an issue, would be to have the jupyter notebook, when it is run, generate some sort of random authentication information and print it out to the terminal and wait a bit before starting everything so that the user can copy it or make a pop-up window in a GUI environment that shows it.

Carreau:
Hi there, 

Thanks for the report. For information we prefer security vulnerability report to be disclosed on security(at)ipython.org in order to investigate without potential malicious listener to use the hypothetic vulnerability. We'll have a look.

frejanordsiek:
Oh, sorry about that. Do you think it is too late to just delete this issue report or do you think too many people have been notified about it?

Carreau:
> Oh, sorry about that. Do you think it is too late to just delete this issue report or do you think too many people have been notified about it?

No problem, now that's it's here there are little advantage of removing it.

frejanordsiek:
Didn't know ipython.org was still around. I should have looked for that at least for a bit first instead of only digging around on jupyter.org

Carreau:
So a couple of information, regardless of whether you run a localhost server or a remote one we always _strongly suggest_ setting up https/SSL and password. 

This is in [our docs](http://jupyter-notebook.readthedocs.io/en/latest/public_server.html#securing-a-notebook-server) and we are enforcing that a little more each time we make a new major release. We also provide [a tool ](https://github.com/jupyter/notebook/blob/master/tools/secure_notebook.py) to automaticlly secure notebooks. 

[...]

PR #1969, opened 2016-12-12
Mitigation for POST requests from malicious forms

takluyver:
In discussion on gh-1830, I found that HTML forms in Firefox do not send an Origin header. You can therefore submit a POST request with an empty body to trigger certain actions, such as starting a kernel, avoiding the origin check we do.

This mitigates that for creating files and starting kernels, by requiring a JSON body, even if there's no data. I cannot currently find a way to create a JSON body in a request sent from a form.

The other cases I've found are interrupting and restarting kernels. This does not affect those cases, but they are only possible if you have a kernel ID.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -346,7 +346,7 @@ class IPythonHandler(AuthenticatedHandler):
     def get_json_body(self):
         """Return the body of the request as JSON data."""
         if not self.request.body:
-            return None
+            raise web.HTTPError(400, u'No JSON in body of request')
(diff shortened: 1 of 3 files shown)


takluyver:
This doesn't feel like a proper fix, though, so I'd like to work out something better. Maybe we should block any requests without `Origin` set, and force scripts to use the token to bypass origin checks.

takluyver:
[Firefox issue about sending Origin](https://bugzilla.mozilla.org/show_bug.cgi?id=446344)

minrk:
Merged xsrf patch

PR #1972, opened 2016-12-13
better docs for token auth

### LINKED ISSUE #1944: Opening notebook on Chrome (Mac) asks for unknown password
Running notebook-4.2.3 installed via pip from master on Python 3.5.2 (Anaconda) on macOS 10.12.

I typically use Safari as my default browser, so notebook sessions open automatically with Safari. However, if I try to open a notebook using Chrome, I am confronted with a password input. Having never (knowingly) set a password, I have no idea how to log in and access my notebook session. This is a notebook being run locally on a MacBook Air. 

Safari never requests a password input.

minrk:
- clearer description of defaults in help output for token
- show how to get the token on the login page
- add details about token auth to the security doc

closes #1944

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -15,14 +15,14 @@ For more detailed information, see
 .. _release-4.3:
 
 4.3
------
+---
 
 4.3 is a minor release with many bug fixes and improvements.
 
 Highlights:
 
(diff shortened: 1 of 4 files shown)


takluyver:
The JS tests seem to be getting worse and worse. Restarted...

PR #1975, opened 2016-12-14
add Authorization to allowed CORS headers

minrk:
so that CORS requests can use token authentication in the Authorization header by default.

cf https://github.com/jupyterlab/jupyterlab/pull/1416

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -416,7 +416,7 @@ class APIHandler(IPythonHandler):
         return super(APIHandler, self).finish(*args, **kwargs)
 
     def options(self, *args, **kwargs):
-        self.set_header('Access-Control-Allow-Headers', 'accept, content-type')
+        self.set_header('Access-Control-Allow-Headers', 'accept, content-type, authorization')
         self.set_header('Access-Control-Allow-Methods',
                         'GET, PUT, POST, PATCH, DELETE, OPTIONS')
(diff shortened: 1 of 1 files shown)


takluyver:
Restarted some JS tests (any idea if there's anything we can do to make these more reliable?)

I have little understanding of the significance of this, so I'll let people who hopefully do have a look.

blink1073:
I confirmed that this allows us to run the JupyterLab tests using Karma and tokens.

blink1073:
I prefer this solution over sending the token as a URL param in plain text, per https://www.w3.org/Protocols/rfc2616/rfc2616-sec15.html#sec15.1.3.

minrk:
Background:

CORS severely limits what you can do from one site to another. Prior to making a CORS request, browsers make an OPTIONS request, to ask the server what should be allowed. One of the things that is limited is setting headers on those requests. The server specifies which headers should be permitted with Access-Control-Allow-Headers.

Since part of the point of the token authorization is to make authenticated cross-origin access *easier*, we should allow the Authorization header to be included, which is what this PR does.

gnestor:
@minrk Does this need to be backported for 4.3.1?

blink1073:
Yes, please, we are waiting for this in JupyterLab.

gnestor:
Backported! Thanks @minrk.

Issue #1970, opened 2016-12-12
Multiple carriage returns are not handled correctly in v4.3


auneri:
Steps to reproduce:
```
>>> print('1\r2')  # yas!
2
>>> print('1\r2\r3')  # wat?
23
```

Potential culprit: pull request #1836

takluyver:
Thanks, I can reproduce that.

gnestor:
@lgeiger Any ideas?

minrk:
@gnestor looks like it was the `+` on the end being too greedy. #1982, #1983 should fix it.

PR #1982, opened 2016-12-16
fix carriage return handling

minrk:
The only real fix is an errant `+` on the final replacement, which would end up skipping sequential replacements.

- leaves trailing `\r` on the text, if there is one
- use groups to avoid unnecessary replace calls (no change)
- includes test


minrk:
closes #1970

PR #1983, opened 2016-12-16
fix carriage return handling [backport]

### LINKED ISSUE #1970: Multiple carriage returns are not handled correctly in v4.3

minrk:
4.x backport of #1982

closes #1970

The only real fix is an errant `+` on the final replacement, which would end up skipping sequential replacements.

- leaves trailing `\r` on the text, if there is one
- use groups to avoid unnecessary replace calls (no change)
- includes test

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -385,11 +385,11 @@ define([
     // carriage return characters
     function fixCarriageReturn(txt) {
         txt = txt.replace(/\r+\n/gm, '\n'); // \r followed by \n --> newline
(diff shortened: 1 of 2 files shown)


rgbkrk:
The build is generally flaky and broken in notebook js right, not just for these carriage return PRs?

takluyver:
Yep, it's been really flakey lately.

rgbkrk:
Alright, I'll lessen my hesitation towards merging.

PR #1984, opened 2016-12-18
Rename 'resize' event to 'resizeOutput'

jasongrout:
Our convention for events has been to scope component events with `.Component`, but we didn't do it for some OutputArea events. This is also related to #1985.

This is a breaking change (changes the names of events), so targeting 5.x.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -112,7 +112,7 @@ define([
         this.prompt_overlay.dblclick(function () { that.toggle_output(); });
         this.prompt_overlay.click(function () { that.toggle_scroll(); });
 
-        this.element.resize(function () {
+        this.element.on('resizeOutput', function () {
             // maybe scroll output,
             // if it's grown large enough and hasn't already been scrolled.
             if (!that.scrolled && that._should_scroll()) {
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/outputarea.js
```diff
@@ -336,7 +336,7 @@ define([
             if (needs_height_reset) {
                 that.element.height('');
             }
-            that.element.trigger('resize');
+            that.element.trigger('resize.Output');
```
minrk: We scope events to components when they are global events (e.g. on `events.trigger('event.Component')`). These are events on a specific element, so scoping seems redundant, doesn't it? For instance, you still can't listen for all `resize.Output` events with this, can you? Is that what you are after?

jasongrout: What I was after was fixing the bug where 'resize' events trigger up, then are thrown on the window, and are considered *window* resize events.

We could just as easily change the name from 'resize' to 'resizeOutput'. Thoughts?

minrk: Works for me

jasongrout: done

takluyver:
@jasongrout This now has a merge conflict, can you rebase it?

jasongrout:
> @jasongrout This now has a merge conflict, can you rebase it?

@takluyver - done.

PR #1985, opened 2016-12-18
Ignore resize events that bubbled up and didn't come from window.

jasongrout:
See https://bugs.jquery.com/ticket/9841 - we were having a major slowdown in interact widgets because the output areas were triggering a 'resize' event, which was bubbling up to the window object and running a function which required a DOM read.

### DIFF
--- a/notebook/static/base/js/page.js
+++ b/notebook/static/base/js/page.js
@@ -52,9 +52,13 @@ define([
         this._resize_site();
     };
 
-    Page.prototype._resize_site = function() {
+    Page.prototype._resize_site = function(e) {
         // Update the site's size.
-        $('div#site').height($(window).height() - $('#header').height());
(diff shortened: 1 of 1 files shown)


notebook/static/base/js/page.js
```diff
@@ -52,9 +52,13 @@ define([
         this._resize_site();
     };
 
-    Page.prototype._resize_site = function() {
+    Page.prototype._resize_site = function(e) {
         // Update the site's size.
-        $('div#site').height($(window).height() - $('#header').height());
+        // only trigger if the event actually is the window's, not bubbling up.
+        // See https://bugs.jquery.com/ticket/9841#comment:8
+        if (!e.target.tagName) {
```
minrk: From Travis: `e.target` is undefined

PR #1993, opened 2016-12-19
add missing waits for output in display_id tests

minrk:
I think this will fix the recent increase in Travis' false failure rate.


minrk:
This passed without needing a restart, which is a good sign, but of course not proof. Self-merging, as it's just a test, and should help other PRs get through.

takluyver:
Thanks, the failures have been annoying me

PR #1994, opened 2016-12-19
Further highlight token info in log output

### LINKED ISSUE #1980: Make token URL clearer when no-browser is given
Or be more explicit on the login webpage that a token need to be passed, 
and can be found in the Cli that launched the notebook. 

Otherwise users can't log-in after upgrading to 4.3

(and print the token on the CLI in a more visible manner.)

minrk:
add critical-level log statement at the end of startup with token info

closes #1980

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1321,7 +1321,15 @@ class NotebookApp(JupyterApp):
                 b = lambda : browser.open(url_path_join(self.connection_url, uri),
                                           new=2)
                 threading.Thread(target=b).start()
-        
+
+        if self.token:
+            self.log.critical('\n'.join([
+                '\n',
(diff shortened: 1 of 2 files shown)


rgbkrk:
One alarming thing about us logging this in general - if people choose to export their logs to external storage, this is now exposed in those logs. This helps the localhost and SSH based user so I understand why this is here, just a note for the future.

minrk:
@rgbkrk good call. I've switched it so that it only logs the token if it's been automatically generated. Setting a token via config would result in nothing logged.

rgbkrk:
@minrk great, thank you!

gnestor:
@minrk Looks good. Is this ready to merge?

minrk:
Thanks, @gnestor and @rgbkrk!

PR #1996, opened 2016-12-19
Fix typo in documentation

montanalow:
Active is inverted.


takluyver:
Thanks, but I think it's correct as is; it says '*without* an active kernel'.

PR #2002, opened 2016-12-21
rename log.warn to log.warning as log.warn is deprecated

srinivasreddy:


takluyver:
Thanks.

minrk:
If this is backported, the changes in `setupbase.py` should not be included, as that's not a `logging.Logger`, but `distutils.log`, which doesn't have a `warning` function. This was fixed in the webpackify PR, which made several other setupbase changes, but that's not for backport. Not hard to do, but easy to forget, as these warning conditions don't come up often.

Issue #1967, opened 2016-12-12
Spacing in text cells on Safari 10


ivirshup:
I've been encountering some formatting strangeness in any text block (code and markdown) on Safari v10.0.1 using jupyter notebook v4.3.0.

Spaces in text are not being show. Once characters are placed the formatting looks correct. I've included an example below:

![jupytercode_spacing](https://cloud.githubusercontent.com/assets/8238804/21085541/7f410ade-c067-11e6-8dc1-f2c6660b1f47.gif)

If there are characters on the line after the cursor, there don't seem to be any issues. I haven't been able to replicate the issue in Chrome. 

Possibly related to #1943. Thanks for any help!

gnestor:
Related: https://github.com/jupyterlab/jupyterlab/issues/522

minrk:
I've definitely seen this, but can't seem to make it happen now (also Safari v10.0.1). @ivirshup does this happen with any text (i.e. cursor doesn't move across whitespace, only when characters are entered), or only in specific circumstances?

gnestor:
@minrk I'm only observing it when on single spaces. Double spaces and characters after single spaces update correctly. I'm testing against notebook at master on Safari 10.0.2.

ivirshup:
@minrk 

I'm haven't done very thorough testing, but I just spent about ten minutes in a fresh notebook and this is what I'm seeing:

## In code cells

Strange space formatting:

* Will occur immediately after any delimiter
  * e.g. `=`, `,`, `.`, `(`, `)`, `[`, `]`, `{`, `}`, `"`, `'`, etc.
* Will not occurs within closed delimiters
* Will not occur within compound statement if closed by `:`
  * e.g. it will occur here `if x =={cursor}`, but not here`if x =={cursor}:`

## In markdown cells

Strange space formatting occurs:

* Unless cursor is at start of a new line
* Unless cursor is placed immediately after closed markup delimiter (e.g. `*text*{cursor}`)
* Unless within markup delimiters (e.g. `*text{cursor}*`)

Hope this helps!

fonnesbeck:
I'm seeing it now on 10.0.2 also with notebook 4.3.0. Where I see it most prominently is when I type a space, but the cursor does not move, yet when I type a letter afterwards, it is properly spaced. So the space registered, but was not displayed until additional text was added.

minrk:
[...]

PR #2004, opened 2016-12-21
Downgrade to CodeMirror 5.16

gnestor:
Temporary fix for https://github.com/jupyter/notebook/issues/1967

I have tested and this resolves the strange space behavior in Safari 10 👌


minrk:
Thanks, @gnestor!

PR #2005, opened 2016-12-21
forward-port xsrf fixes

minrk:
this has already been released in 4.3.1

I resolved merge conflicts, but tests may not pass yet


takluyver:
Did we decide not to do the thing with enforcing the empty JSON body? I've lost track of all the different pieces.

minrk:
> Did we decide not to do the thing with enforcing the empty JSON body?

Yes, that would be redundant, and looking around at code, sending a JSON body for requests that take no args seemed more inconvenient/annoying than adopting standard xsrf techniques.

PR #2007, opened 2016-12-22
add `jupyter notebook password` entrypoint

minrk:
now that we are securing notebooks by default, increasing the convenience of defining a password is more important.

Mostly copied from the `tools/secure_notebook` script. I omitted the ssl-cert-generation from that script, since I think it's less portable to platforms like Windows, and there are many more ways to do it.

TODO:

- [x] update 'securing the notebook' docs with reference to this command
- [x] exercise with tests

cc @fperez

records, hashes, and stores password in json config

### DIFF
--- a/notebook/auth/security.py
+++ b/notebook/auth/security.py
@@ -9,8 +9,8 @@ import io
 import json
 import os
 import random
-import sys
 import traceback
+import warnings
 
 from ipython_genutils.py3compat import cast_bytes, str_to_bytes, cast_unicode
 from traitlets.config import Config, ConfigFileNotFound, JSONFileConfigLoader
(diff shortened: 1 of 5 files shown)


notebook/auth/security.py
```diff
@@ -99,3 +108,40 @@ def passwd_check(hashed_passphrase, passphrase):
...
+
+    yield config
+
+    with io.open(config_file, 'w', encoding='utf8') as f:
+        f.write(cast_unicode(json.dumps(config, indent=2)))
+
+    try:
+        os.chmod(config_file, mode)
+    except Exception:
+        print("Failed to set permissions on %s:" % config_file, file=sys.stderr)
```
takluyver: This needs `from __future__ import print_function` for Py2.

fperez:
Minor question above, but otherwise looks good, thx!

notebook/auth/security.py
minrk: switched to a regular warning

minrk:
docs and tests added

rgbkrk:
Great idea. I'd be comfortable with it being a top level command too, no strength behind that opinion.

takluyver:
Anything else to do on this? It looks ready to me.

minrk:
I'm all set here.

PR #2026, opened 2017-01-07
Throttle element resizing in output areas

ashaffer:
This massively improves performance for certain applications with no discernible cost to UX. I'm not sure how much of an issue this is for other applications, but for training Keras nns, jupyter is almost unusably slow right now. This fixes that completely.


minrk:
Cool!

PR #2033, opened 2017-01-10
simple overload of passwd_check

vlimant:
part of a solution to use the unix password as the jupyter password involves this change, which allows us to overload 

```python
class SyncLogin(LoginHandler):
    def passwd_check(self, h, t ):
        import crypt
        h = self.get_hashed_password()
        return crypt.crypt(t, h ) == h
```

no other method worked.
I'd be happy to be given an alternative solution.

### DIFF
--- a/notebook/auth/login.py
+++ b/notebook/auth/login.py
@@ -67,7 +67,7 @@ class LoginHandler(IPythonHandler):
     def hashed_password(self):
         return self.password_from_settings(self.settings)
 
-    def self.passwd_check(self, a, b):
+    def passwd_check(self, a, b):
         return passwd_check(a, b)
     
     def post(self):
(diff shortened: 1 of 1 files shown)


notebook/auth/login.py
```diff
@@ -67,10 +67,13 @@ def get(self):
     def hashed_password(self):
         return self.password_from_settings(self.settings)
 
+    def self.passwd_check(self, a, b):
```
minrk: I think you want `def passwd_check` here

minrk:
You could override the whole `post` handler in the meantime, but this is simpler and makes sense to me.

vlimant:
sorry about the typo in defining passwd_check

minrk:
Thanks!

PR #2041, opened 2017-01-16
Update button icon.

clarka34:
Changed button icon so it displays properly. Previously, it displayed an empty icon.


takluyver:
Thanks!

PR #2055, opened 2017-01-21
speed up casperjs tests

ivanov:
This code claimed to be required for SlimerJS, but our test suite does
not currently pass on SlimerJS (and does no better or worse without this
code).

With PhantomJS, however, this eliminates a ~10s startup cost for each
file which uses notebook_test, and all of the tests continue to pass.
There are ~40 files using this, the bult of them in the js/notebook
section, which should now run 5-6 minutes faster as a result.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,9 @@ env:
   global:
     - PATH=$TRAVIS_BUILD_DIR/pandoc:$PATH
   matrix:
+    - GROUP=js/notebook
     - GROUP=python
     - GROUP=js/base
-    - GROUP=js/notebook
     - GROUP=js/services
     - GROUP=js/tree
 
--- a/notebook/tests/util.js
+++ b/notebook/tests/util.js
@@ -78,15 +78,6 @@ casper.open_new_notebook = function () {
             });
         });
     });
-
-    // Because of the asynchronous nature of SlimerJS (Gecko), we need to make
-    // sure the notebook has actually been loaded into the IPython namespace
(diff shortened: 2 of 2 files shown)


ivanov:
There are likely other places where our testing code can be simplified, but this was such a clear quick win that it would be a shame to not make testing faster for everyone immediately.

ivanov:
Sweet, as expected, the `GROUP=js/notebook` tests which used to take 13 minutes to run on travis, now take 7.5 minutes!

Travis run time before:
screen shot 2017-01-20 at 5 39 28 pm


With this PR:
screen shot 2017-01-20 at 6 22 13 pm

I'm going to try one more minor thing, a reordering of how the tests are run on Travis which should further reduce the elapsed time.

rgbkrk:
Excellent! Thank you Paul.

ivanov:
I think it's unrelated, but I did just have to kick one of the builds as node threw this error:

``` 
subprocess.CalledProcessError: Command '['node', 'tools/build-main.js', 'notebook']' returned non-zero exit status 1
    Failed to build tree { Error: Error: ENOENT: no such file or directory, open '/tmp/pip-build-rKp5zF/notebook/notebook/static/components/bootstrap/js/bootstrap.min.js'
```

ivanov:
For what its worth, the elapsed time did go down to 11-12 minutes by running starting the longest running tests earlier. Pipelining optimization FTW!

blink1073:
Nice mathing!

PR #2061, opened 2017-01-23
add event for finished execution of cell

### LINKED ISSUE #2060: add event for finished execution of cell
At the moment, there are is no event that signals the completion of the evaluation/execution of a code cell (FYI, there is one, if the execution starts). The aim of this issue is to implement such an event.

dkrenn:
closes #2060 

At the moment, there are is no event that signals the completion of the evaluation/execution of a code cell (FYI, there is one, if the execution starts). The aim is to implement such an event.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -334,6 +334,12 @@ define([
         CodeCell.msg_cells[this.last_msg_id] = this;
         this.render();
         this.events.trigger('execute.CodeCell', {cell: this});
+        this.events.on('finished_iopub.Kernel', function (evt, data) {
+            if (that.kernel.id === data.kernel.id && that.last_msg_id === data.msg_id) {
+		console.log('finished_execute.CodeCell', {cell: that});
+		this.events.trigger('finished_execute.CodeCell', {cell: that});
+	    }
+        });
     };
     
     /**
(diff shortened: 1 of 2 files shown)


minrk:
Thanks, @dkrenn! Looks like this PR was made against the wrong branch. Can you update the base to `:master`? All PRs should be made against master, and we can consider backporting after merge if necessary.

dkrenn:
@minrk Set to master. (Bckporting was not meant by me; simply I wanted to show only correct diff, i.e. the changes that I made branching from 4.3.1. Now the PR is crowed again...)

takluyver:
I think you'll need to make the commits again starting from a recent commit on master (or copy them using `git cherry-pick` if you're comfortable doing that). Because you made the commits on top of the 4.x branch, targetting the PR to master doesn't work.

dkrenn:
@takluyver Sure, this will work (even merging in master on my branch should work). However, there are (as far as I have checked) no merge conflicts at all, so a from a git-workflow-perspective, no need to do such a thing as merge master in or rebase it on master. Not sure, what Github's problem is with this. 
Branch-update follows...

dkrenn:
Ok, I see the problem now: Your master builds up on 4.1.0, whereas my branch builds up on 4.3.1. This explains now all the merging issues.

takluyver:
We do 4.x (for instance) releases from master until we merge something which we only want to go into 5.0. At that point, we create a `4.x` branch for further releases, and backport changes into it. In this case, that happened after 4.1.

dkrenn:
@takluyver Thanks for the clarification. I cherry-picked now. Looks much better now :)

takluyver:
Thanks. This looks OK to me now. We'll leave it a while for other people to take a look, and merge if there are no objections.

Issue #1664, opened 2016-08-08
Add welcome message in JS console.


Carreau:
A user the open the JS console is likely going to poke around JS, and be a developer.
I think a small message that say, hi, and give them some information as well as encourage the to contribute would be nice !

michaelpacer:
I support this, strongly.

Carreau:
We could do the same on the jupyter website.

jinmehta13:
Hello, I'm new to open source but a big fan of Jupyter Notebooks and hoping to contribute if I can. Any tips to get started on this?

Carreau:
Hi @jinmehta13 welcome !

How experienced are you, and in which domain ?

Depending on how you want to get started I would do the following:

If you want to get started on the website, and add a message in the JS console there, the first step would be to get the website to work on you local machine. The website repo is [there](https://github.com/jupyter/jupyter.github.io), once this work we can see for modifications.

If you like to make that on the notebook, the simplest if first to get this current repo on your computer, and managed to install a development version of the notebook. You should be able to find the documentation [here](http://jupyter-notebook.readthedocs.io/). Once this is done you can start to make local modifications :-) 

One thing that would be great, is while you are doing this: when you find something unclear, hard, or just don't know where to search, take notes and tell us afterward, that will help us to improve the documentation a bit more for the next contributor. 

If you have more questions, feel fre to ask.

jinmehta13:
Hey @Carreau, I finished up my undergrad in Computer Science and Engineering this past May. Still looking for a job but my internship experiences worked with a single page web application and a lot of IPython Notebooks with pandas, NumPy, and matplotlib. Like I mentioned earlier, new to open source but looking to be an involved contributor to a project and this seemed like a good fit given my familiarity with Jupyter from projects.

So it took a little time but I managed to get the website working on my local machine. It'd be nice to work on the notebook too so I'll try to have that setup over the next few days.

Some notes on the setup process for the website (note: I had no prior experience with Ruby, Gems, etc.):

- Installing Ruby and the DevKit was straightforward and I had no issues following the docs

- Installing github-pages and some other dependencies on a Windows 64-bit machine required some work. It might help to mention in the [Readme](https://github.com/jupyter/jupyter.github.io#quick-local-testing) to run `bundle install` to execute the Gemfile already included in the project. At one point I made a Gemfile in the Devkit directory.

[...]

PR #2065, opened 2017-01-24
Welcome Message for Console and Terminal Output

jinmehta13:
This PR is in reference to #1664.

Added a welcome message similar to that implemented at https://github.com/jupyter/jupyter.github.io/pull/181 in two locations:

One message will be logged to the browser console for user's accessing the developer tools.
The second message will be logged to the user's terminal output.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -6,6 +6,7 @@
 
 from __future__ import absolute_import, print_function
 
+import notebook
 import binascii
 import datetime
 import errno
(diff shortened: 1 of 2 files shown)


minrk:
Great! I think the unconditional message in the terminal is too much, since this will be shown to all users every time the notebook server is launched. Perhaps this one should only be triggered for a dev-install? Or maybe only having the js message is best.

Carreau:
I pushed a commit that show the message in the terminal only on dev.

Carreau:
Thanks @jinmehta13

PR #2070, opened 2017-01-25
Add "Shutdown" item to Kernel menu

ivanov:
Right now, the only way to work on a notebook without a running kernel is to open it, and then shut down the session from the tree view.

This PR adds a Shutdown item to the Kernel menu. On top of that, it removes the confirmation dialog from the Restart menu item variants in cases where the kernel has already been shutdown. It also adds test cases for both the presence of a confirmation dialog when there is a live session, and the absence of confirmation when there is not a live session.  

shutdown demo1

### DIFF
--- a/notebook/tests/services/session.js
+++ b/notebook/tests/services/session.js
@@ -121,6 +121,9 @@ casper.notebook_test(function () {
         }
     );
 
+    this.thenEvaluate( function() {IPython.notebook.session.start()});
+    this.wait_for_kernel_ready();
+
     // check for events when restarting the session
     this.event_test(
         'restart_session',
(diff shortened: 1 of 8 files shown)


notebook/tests/services/session.js
```diff
@@ -164,6 +167,9 @@ casper.notebook_test(function () {
         20000
     );
 
+    this.thenEvaluate( function() {IPython.notebook.session.start()});
+    this.wait_for_kernel_ready();
+
```
ivanov: These tests were assuming that a `kernel_killed.Session` event will be triggered, even in cases where the kernel was already killed (which was the case here, the kernel is dead at the beginning of these two event tests). I've cleaned up the code so that `kernel_killed.Session` event is only triggered when the kernel is actually being killed. So here we first restore the notebook session so the kernel is running, restoring the expected events.

minrk:
Very nice, thanks @ivanov!

Issue #2035, opened 2017-01-10
movable `find and replace` modal box


BBischof:
I would really appreciate if the `find and replace` modal was movable. It pops up over the code I'm trying to modify very often, which sounds stupid, but I find myself closing the modal and having to reopen it a lot.

gnestor:
This is a good point! The notebook uses Bootstrap and jQuery, so this should be trivial...

BBischof:
🙌

PR #2076, opened 2017-01-26
Make dialogs draggable

gnestor:
Fixes https://github.com/jupyter/notebook/issues/2035

This simply calls `.draggable()` on `$(".modals")` when a modal receives a mousedown. I tried to add a `draggable` option and that didn't work. I also tried calling `.draggable()` on creation of a modal and that scrolled the page to the bottom.

screencap

### DIFF
--- a/notebook/static/base/js/dialog.js
+++ b/notebook/static/base/js/dialog.js
@@ -55,6 +55,9 @@ define(function(require) {
         dialog_content.append(
             $("<div/>")
                 .addClass("modal-header")
+                .mousedown(function() {
+                  $(".modal").draggable();
+                })
                 .append($("<button>")
                     .attr("type", "button")
                     .addClass("close")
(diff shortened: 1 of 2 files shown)


ivanov:
Sweet, @gnestor, thanks for this. I understand the motivation for having the whole modal be draggable, instead of just the header, but unfortunately that makes it impossible to select text from the modal (which makes editing the notebook metadata a rather painful experience). I sent you a patch to fix this - https://github.com/gnestor/notebook/pull/5

gnestor:
Thanks @ivanov! I also added a "move" cursor to the modal header to indicate that it is now draggable 👍

ivanov:
Oh wow, nice touch! 🎼 🍷 🎶 🍸  Very classy! :bowtie:

gnestor:
@ivanov 😊

PR #2078, opened 2017-01-28
Fix JS error on running notebooks tab

parente:
When a kernelspec is removed while a kernel spawned from it is in use,
the Running tab on the dashboards page errors and no longer shows
any results. Fall back on using the kernel name from the session
in this case so that users can see and terminate kernels.

### DIFF
--- a/notebook/static/tree/js/kernellist.js
+++ b/notebook/static/tree/js/kernellist.js
@@ -49,7 +49,7 @@ define([
             return; // wait for kernelspecs before first load
         }
         this.clear_list();
-        var item, path, session;
+        var item, path, session, info;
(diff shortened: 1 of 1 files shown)


parente:
Some context: We deploy shared kernel environments to our Jupyter Hubs. We retain kernel specs for the last N environment versions. When we clean up a kernel spec, users who are still using the associated kernel and environment on disk can continue to work. However, because the kernel spec is gone, the JS error happens and the Running tab fails to render any notebooks. The user is stuck killing off the entire notebook server or navigating around the file tree to find running notebooks to shut down.

takluyver:
Looks fine to me, thanks.

PR #2083, opened 2017-01-29
trigger event change.Cell

dkrenn:
Trigger an (new) event `change.Cell` when the input of a cell is changed.

At the moment, only a `set_dirty` is triggered; but neither the information, which cell caused the change and nor what exactly changed is missing. However all information is available, it just has to be passed on.


takluyver:
This looks fine to me, but I'd like someone more familiar with the notebook JS stuff to look over it too.

PR #2086, opened 2017-01-30
Using hooks together in content manager

dmvieira:
This is important because other hooks can be implemented following same guidelines

I believe that these hooks are important for simple customizations:

* pre_get_hook
* post_get_hook
* pre_delete_hook
* post_delete_hook
* pre_update_hook
* post_update_hook

I don't know if this is the best place to implement hooks, but I see that pre and post save hooks were implemented

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -19,9 +19,8 @@ from .filecheckpoints import FileCheckpoints
 from .fileio import FileManagerMixin
 from .manager import ContentsManager
 
-from ipython_genutils.importstring import import_item
-from traitlets import Any, Unicode, Bool, TraitError, observe, default, validate
-from ipython_genutils.py3compat import getcwd, string_types
+from traitlets import Unicode, Bool, TraitError, observe, default, validate
+from ipython_genutils.py3compat import getcwd
 
 from notebook import _tz as tz
 from notebook.utils import (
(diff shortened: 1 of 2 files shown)


notebook/services/contents/filemanager.py
```diff
@@ -94,43 +93,6 @@ def _update_save_script(self):
...
-        if not callable(value):
-            raise TraitError("post_save_hook must be callable")
-        return value
-
-    def run_post_save_hook(self, model, os_path):
-        """Run the post-save hook if defined, and log errors"""
-        if self.post_save_hook:
-            try:
-                self.log.debug("Running post-save hook on %s", os_path)
-                self.post_save_hook(os_path=os_path, model=model, contents_manager=self)
```
minrk: post_save hook was declared here because the signature requires that the result be a file on disk. The higher-level manager doesn't have a notion of `os_path`, so the same post_save hook doesn't make sense.

minrk:
The best place to implement code at most of these stages is in subclassed ContentsManagers, which is why we don't have these hooks defined. Subclasses can make more assumptions about their storage mechanism than the base classes, as indicated by the fact that only the FileContentsManager has a post_save hook, not the general base class.

notebook/services/contents/filemanager.py
dmvieira: ok, but save method has a model and path to be implemented:

`def save(self, model, path):`

Couldn't we find a way to do generic triggers for content manager?

dmvieira:
So, hooks are not a project aim?

notebook/services/contents/filemanager.py
takluyver: The `path` there is an API path, which the file manager has to translate into an OS path. Another file manager implementation might, for instance, store contents in a database, so it doesn't have a filesystem path.

dmvieira:
I think your hooks are under construction, so I'm following @minrk considerations doing my own subclass with these hooks: https://github.com/globocom/jupyterhub_hook_filemanager

I'm trying to follow **pre_save** and **post_save** patterns

Thx guys!

notebook/services/contents/filemanager.py
dmvieira: yes, but it has a path and the hook could ignore path type (filesystem, database, etc).

takluyver: The point of `post_save_hook` is that it can do something with the data that was just saved, though. To do that, it needs to access the filesystem/database/whatever. FileContentsManager's post_save_hook is passed the OS path so that it can do things with files.

dmvieira:
Closing here because I solve my problem and hooks are not full defined

PR #2091, opened 2017-01-31
fix and tests for the issue related with the trailing slash using 'nbextension install'

### LINKED ISSUE #2069: Errors on 'Custom front-end extensions' docs
Hi all,

Reference: http://jupyter-notebook.readthedocs.io/en/latest/extending/frontend_extensions.html#custom-front-end-extensions

I'm updating some of my stuff and I was re-reading some docs. When I arrive to this [example](http://jupyter-notebook.readthedocs.io/en/latest/extending/frontend_extensions.html#defining-and-registering-your-own-actions) I think there are some issues:

* In the line below I think `name` should be `action_name`:

```javascript
        var full_action_name = Jupyter.actions.register(action, name, prefix); // returns 'my_extension:show-alert'
```

* Also, the line above in conjunction with the line below are not adding the action and the button:

```javascript
        Jupyter.toolbar.add_buttons_group([full_action_name]);
```

I had to modify this:

```javascript
    function load_ipython_extension() {

        var handler = function () {
            alert('this is an alert from my_extension!');
        };

        var action = {
            icon: 'fa-comment-o', // a font-awesome class used on buttons, etc
            help    : 'Show an alert',
            help_index : 'zz',
            handler : handler
        };
        var prefix = 'my_extension';
        var action_name = 'show-alert';

        var full_action_name = Jupyter.actions.register(action, name, prefix); // returns 'my_extension:show-alert'
        Jupyter.toolbar.add_buttons_group([full_action_name]);
    }
```
with
```javascript
    function load_ipython_extension() {

        var handler = function () {
            alert('this is an alert from my_extension!');
        };

//        var action = {
//            'icon': 'fa-comment-o', // a font-awesome class used on buttons, etc
//            'callback' : handler,
//            'label': 'Kaka kaka kaka',
//            'id': 'idKaka'
//        };
//        var prefix = 'my_extension';
//        var action_name = 'show-alert';

//        var full_action_name = Jupyter.actions.register(action, action_name, prefix); // returns 'my_extension:show-alert'
//        Jupyter.toolbar.add_buttons_group([{full_action_name}]);
        Jupyter.toolbar.add_buttons_group(
              [{
                  'icon': 'fa-comment-o',
                  'callback' : handler,
                  'label': 'Kaka kaka kaka',
                  'id': 'idKaka'
              }]
        );
    }
```
to make it work.

Also, in this [section](https://github.com/jupyter/notebook/blob/master/docs/source/extending/frontend_extensions.rst#installing-and-enabling-extensions), this line:

`jupyter nbextension install path/to/my_extension/ [--user|--sys-prefix]`

should be:

`jupyter nbextension install path/to/my_extension [--user|--sys-prefix]`

i.e., without the final '/' in the path.

I can prepare a quick PR to solve these issues in case I'm not wrong.

kikocorreoso:
closes #2069

It seems there is a problem when one wants to install a nbextension contained in a folder. This:

`jupyter nbextension install my_extension/ --sys-prefix`

Installs the content of the `my_extension/` folder but not the folder itself. It is prone to overwrite other files already present in the `nbextensions` folders.

With this PR, the next two lines would be similar:

`jupyter nbextension install my_extension/ --sys-prefix`
`jupyter nbextension install my_extension --sys-prefix`

And the instructions [here](https://jupyter-notebook.readthedocs.io/en/latest/extending/frontend_extensions.html#installing-and-enabling-extensions) would behave correctly.

### DIFF
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -165,7 +165,7 @@ def install_nbextension(path, overwrite=False, symlink=False,
         full_dest = None
     else:
         if not destination:
-            destination = basename(path)
+            destination = basename(normpath(path))
         destination = cast_unicode_py2(destination)
         full_dest = normpath(pjoin(nbext, destination))
         if overwrite and os.path.lexists(full_dest):
--- a/notebook/tests/test_nbextensions.py
+++ b/notebook/tests/test_nbextensions.py
@@ -179,10 +179,13 @@ class TestInstallNBExtension(TestCase):
         self.assert_installed(self.files[-1])
     
     def test_single_dir_trailing_slash(self):
-        d = [u'∂ir/', u'∂ir\\']
-        for dest in d:
-            install_nbextension(pjoin(self.src, dest))
-            self.assert_installed(self.files[-1])    
+        d = u'∂ir/'
+        install_nbextension(pjoin(self.src, d))
+        self.assert_installed(self.files[-1])
+        if os.name == 'nt':
+            d = u'∂ir\\'
+            install_nbextension(pjoin(self.src, d))
+            self.assert_installed(self.files[-1])
 
     def test_destination_file(self):
         file = self.files[0]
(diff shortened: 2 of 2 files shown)


takluyver:
Thanks, this looks good to me.

PR #2108, opened 2017-02-01
merge nbserver_extensions

### LINKED ISSUE #2063: server extensions not loaded when installed to different locations
If two extensions are installed, one with flag `--user` and another with `--sys-prefix`, then they are both listed by `jupyter nbextension list`, but only one is actually loaded when the server starts.  I suspect the configuration objects are being over-written rather than merged.

See the following issue for more details: https://github.com/OpenGeoscience/geonotebook/issues/70

minrk:
- [x] test
- [x] handle config_file_name and config_dir Application settings

Load with ConfigManager so that merge happens recursively, unlike normal config values.

Makes loading more consistent with frontend extensions, which is good.

It is a bit icky because the same files are loaded and read twice, and one key in a traitlets-defined config file is loaded differently than all others.

This does behave more like we intend and people expect for extensions, though.

closes #2063

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -86,6 +86,7 @@ from traitlets.config.application import catch_config_error, boolean_flag
 from jupyter_core.application import (
     JupyterApp, base_flags, base_aliases,
 )
+from jupyter_core.paths import jupyter_config_path
 from jupyter_client import KernelManager
 from jupyter_client.kernelspec import KernelSpecManager, NoSuchKernel, NATIVE_KERNEL_NAME
 from jupyter_client.session import Session
@@ -1232,9 +1233,24 @@ class NotebookApp(JupyterApp):
             # in the new traitlet
             if not modulename in self.nbserver_extensions:
                 self.nbserver_extensions[modulename] = True
-        
-        for modulename in sorted(self.nbserver_extensions):
-            if self.nbserver_extensions[modulename]:
+
+        # Load server extensions with ConfigManager.
+        # This enables merging on keys, which we want for extension enabling,.
+        # Regular config loading only merges at the class level,
+        # so each level (use > env > system) clobbers the previous.
+        manager = ConfigManager(read_config_path=jupyter_config_path())
+        section = manager.get('jupyter_notebook_config')
+        extensions = section.get('NotebookApp', {}).get('nbserver_extensions', {})
+
+        for modulename, enabled in self.nbserver_extensions.items():
+            if modulename not in extensions:
+                # not present in `extensions` means it comes from Python config,
+                # so we need to add it.
+                # Otherwise, trust ConfigManager to have loaded it.
+                extensions[modulename] = enabled
+
+        for modulename, enabled in sorted(extensions.items()):
+            if enabled:
                 try:
                     mod = importlib.import_module(modulename)
                     func = getattr(mod, 'load_jupyter_server_extension', None)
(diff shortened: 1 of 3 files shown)


notebook/notebookapp.py
```diff
@@ -1232,9 +1233,24 @@ def init_server_extensions(self):
...
                 self.nbserver_extensions[modulename] = True
-        
-        for modulename in sorted(self.nbserver_extensions):
-            if self.nbserver_extensions[modulename]:
+
+        # Load server extensions with ConfigManager.
+        # This enables merging on keys, which we want for extension enabling,.
+        # Regular config loading only merges at the class level,
+        # so each level (use > env > system) clobbers the previous.
+        manager = ConfigManager(read_config_path=jupyter_config_path())
```
blink1073: This will not pick up any overrides given to the `NotebookApp` - `config_file_name`, `config_dir`.

minrk: Good call, and further evidence that this is probably too fiddly to be worth doing in-place.

minrk: Should work with config_dir and config_file_name now.

notebook/notebookapp.py
```diff
@@ -1232,9 +1233,28 @@ def init_server_extensions(self):
             # in the new traitlet
             if not modulename in self.nbserver_extensions:
                 self.nbserver_extensions[modulename] = True
-        
-        for modulename in sorted(self.nbserver_extensions):
-            if self.nbserver_extensions[modulename]:
+
+        # Load server extensions with ConfigManager.
+        # This enables merging on keys, which we want for extension enabling,.
```
blink1073: Extra comma at the end

notebook/notebookapp.py
```diff
@@ -1232,9 +1233,28 @@ def init_server_extensions(self):
...
+        # Regular config loading only merges at the class level,
+        # so each level (use > env > system) clobbers the previous.
```
blink1073: User*

notebook/notebookapp.py
Carreau: removed.

notebook/notebookapp.py
Carreau: r'd

Carreau:
Thanks Steve !

Madhu94:
@minrk @Carreau @blink1073 This was supposed to merge nbserver_extensions for ``jupyter_notebook_config.py`` files as well, not just ``jupyter_notebook_config.json`` files ?

blink1073:
Looking down the rabbit hole it looks like it only applies to `.json` [files](https://github.com/ipython/traitlets/blob/a33176faa103e013a3cff0903217138cd4c1da5b/traitlets/config/manager.py#L51-65).

Madhu94:
Thanks @blink1073! Was that done on purpose ? Any reason the notebook doesn't do the merge for nbserver_extensions in ``.py`` files ?

blink1073:
I'm not sure if that was a conscious decision historically or not, @minrk would be the best person to know for sure.

minrk:
`jupyter nbextension enable/disable` always happens in .json files, so that's the mechanism by which this is generally happening. Including .py files would further complicate this already complicated band-aid. My answer would be that nbextension config shouldn't ever occur in `.py` config files because of this special handling.

.py config files also already have a clearer, explicit merge mechanism, so shouldn't have to get mixed up in this:

```python
c.Class.trait.update({'x': 5}) # update existing config, if any
# vs
c.Class.trait = {'x': 5} # override
```

Madhu94:
Thanks @minrk

PR #2114, opened 2017-02-01
Isolate directionality of a string from surrounding

ebraminio:
Hey guys, this little patch will make life of people dealing with text containing RTL text easier. It simply turns this:
`print('\n'.join(['متن متن', '123']))`

before

Into this:
after

To understand the difference, look carefully at the strings sequence on the image. On the first one, which is broken, an Arabic string which logically should be first (as it is printed first on the output the code) came after the number visually. After the change however, highlighting changed in a way that strings visual order actually matches its logical order (the printed text)

This type of workaround is somehow common on other projects, for example have a look at these patches on Chrome browser DevTools:
* http://crbug.com/178635 https://codereview.chromium.org/212653002
* http://crbug.com/330387 https://codereview.chromium.org/109053003

### DIFF
--- a/notebook/static/notebook/less/highlight.less
+++ b/notebook/static/notebook/less/highlight.less
@@ -27,6 +27,9 @@ Adapted from GitHub theme
 
 .highlight-string{
   color: #BA2121;
+  /* isolate strings unicode direction from surrounding text */
+  unicode-bidi: isolate;
+  unicode-bidi: -webkit-isolate;
 }
 
 .highlight-comment{


blink1073:
I opened https://github.com/jupyterlab/jupyterlab/issues/1598 to track this in JupyterLab as well.

Carreau:
Does that work better as well when using "toggle rtl layout" (https://github.com/jupyter/notebook/pull/1863)  ?

+1

ebraminio:
@Carreau: It won't interfere with such cases, if not improve them, at least I can say. Interesting that I wanted to go for such feature after this, where can I test that?

ebraminio:
Sorry guys, my further testing revealed this regresses editing experience of the same case, I should come with another solution which perhaps some of JS of CodeMirror should be altered on the way (so I should try to upstream my patch there), so for now, here is the revert: #2119

But I still like to know where to see #1863? Is it not exposed to the users yet? I perhaps can help on test and improving that also.

takluyver:
It's exposed to the users in master, not yet a released version... however, I've just gone to try it, and found that the command palette which you'd use to access it seems to be broken (#2121, #2122).

ebraminio:
> It's exposed to the users in master, not yet a released version...

No problem, I have a working version of the master locally. Ok, will look at them, perhaps will help on them if be able to :) Thanks.

Issue #2113, opened 2017-02-01
Terminals broken


takluyver:
When I start a terminal, I don't see any prompt or other text in the black area, and I can't type anything, but this appears in the lower left corner:

![screenshot from 2017-02-01 17-44-14](https://cloud.githubusercontent.com/assets/327925/22518868/2c60498a-e8a6-11e6-8e67-a0c94ac98282.png)

I've rebuilt the JS and CSS, made sure I'm not hitting the browser cache, and tested it in both Firefox and Chromium, with the same results. I also tried upgrading xtermjs to 2.2.3, the latest release, and still saw the same. I guess that maybe some of our CSS/JS is clashing with that of xtermjs? @yuvipanda , you've done stuff with terminals recently, do you have a chance to look at it?

There are no errors in the terminal, and nothing that looks relevant in the JS console:

```
Using //@ to indicate sourceMappingURL pragmas is deprecated. Use //# instead[Learn More]
resize to : 57 rows by  138 columns
Use of getPreventDefault() is deprecated.  Use defaultPrevented instead.
resize to : 41 rows by  138 columns
```

takluyver:
$SHELL is set to `/bin/bash`.

Carreau:
I think that's more than CSS, the `div.xtem-rows` is actually outside of the `div.xterm-scroll-area`, while it should be in it.

gnestor:
I can confirm that 4.x branch does not have this issue...

gnestor:
Likely culprit: https://github.com/jupyter/notebook/commit/47d44519584f894c0b8bac55a2e234d532bb14db

gnestor:
It looks like we were requiring `xterm/src/xterm.css` but now that we don't have webpack, how should we include xterm's CSS?

blink1073:
IIRC they have a bower package.

blink1073:
I'll look into this tomorrow morning.

blink1073:
Nevermind, Hrant already fixed it. ;)

PR #2116, opened 2017-02-01
Add back xterm.css

gnestor:
Closes https://github.com/jupyter/notebook/issues/2113


Carreau:
Thanks !

We're closing in on 5 ! Do one of you want to send a message to the mailing list saying that we are closing in on 5.0, that we need a last look on critical issue and a hand for a couple of last PR ? 

Then we can cut a beta?

takluyver:
Don't forget about the issues marked 4.4. I don't think we're doing 4.4
now, so they should be considered for 5.0.

Carreau:
> Don't forget about the issues marked 4.4. I don't think we're doing 4.4
> now, so they should be considered for 5.0.

I'm seeing a few we can bump to 5.1 !

PR #2119, opened 2017-02-02
Revert "Isolate directionality of a string from surrounding"

ebraminio:
This, while fixes visual ordering issue, but breaks editing experience of the same case.

For example trying to double click selection of the first word from the case available on #2114 shows the regression:
screen shot 2017-02-02 at 11 43 50 am

Can be considered a CM issues but we should come with another solution for this.

### DIFF
--- a/notebook/static/notebook/less/highlight.less
+++ b/notebook/static/notebook/less/highlight.less
@@ -27,9 +27,6 @@ Adapted from GitHub theme
 
 .highlight-string{
   color: #BA2121;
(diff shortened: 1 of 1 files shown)


takluyver:
That's a  pity; thanks for contributing. Have you tested it on [plain Codemirror](http://codemirror.net/mode/python/index.html)? Is it worth opening an issue on Codemirror itself?

ebraminio:
Yes it does, just've filed https://github.com/codemirror/CodeMirror/issues/4532

PR #2124, opened 2017-02-02
Upgrade xtermjs to 2.2.3

takluyver:
I tested very briefly and this seemed to work. 2.2.3 is the latest release at present. Release notes are here:
https://github.com/sourcelair/xterm.js/releases


takluyver:
Thanks Steve, that's three times as many browsers as I had tested ;-)

PR #2128, opened 2017-02-03
remove token query parameter from pages on load

### LINKED ISSUE #2127: Clear the one-time token from the URL when the notebook is launched
Right now it's confusing that the token in the URL of a browser that has been launched by Jupyter notebook only works one time. We could work around the confusion by clearing it from the URL address immediately.

Thanks to @minrk for the suggestion and @blink1073 for also bringing it up.

minrk:
remove token query parameter from pages on load

Once the page has been loaded, the token has done its job and can be discarded. Its presence can lead to confusion, especially copying/pasting URLs with one-time tokens present.

closes #2127

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -156,6 +156,32 @@
 {% block script %}
 {% endblock %}
 
+<script type='text/javascript'>
+  function _remove_token_from_url() {
+    if (window.location.search.length <= 1) {
+      return;
+    }
+    var search_parameters = window.location.search.slice(1).split('&');
+    var found = false;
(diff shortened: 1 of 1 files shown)


notebook/templates/page.html
```diff
@@ -156,6 +156,32 @@
 {% block script %}
 {% endblock %}
 
+<script type='text/javascript'>
+  function _remove_token_from_url() {
+    if (window.location.search.length <= 1) {
+      return;
+    }
+    var search_parameters = window.location.search.slice(1).split('&');
+    var found = false;
```
takluyver: unused variable

Issue #1852, opened 2016-10-24
Proposal for fixing mirroring issues in Jupyter Notebook


samarsultan:
## Jupyter Notebook
Arabic Data Handling

### Overview
Mirroring is essential in case the application interface is translated to any language with a Bidirectional script (e.g. Arabic, Hebrew). The Arabic and Hebrew script natural Global Orientation is from right-to-left, hence, the native speakers of these languages expect to see the GUI layout mirrored so that the whole GUI can be read from right-to left as well. This includes viewing menu buttons on the right, the navigation tree on the right, and indentation to the left. GUI mirroring is necessary for any application which has GUIs meant to be translated to a language based on bidi (i.e. Bidirectional) scripts such as Arabic and Hebrew.

## High-level Design
Whenever the Arabic or Hebrew translation is loaded, mirroring of the UI will occur, this may depend on web browser setting or UI language setting.

### Preferences
No special preference needed for mirroring

### Terminology and Acronyms
•	LTR: Left to Right
•	RTL: Right to Left
•	GUI Mirroring: Change the Graphical User Interface directionality from a Left to Right orientation to a Right to Left orientation. (The effect is a horizontal position swapping of the GUI elements)

### Mirroring Issues
In this section, we highlight mirroring issues found when enforcing right directionality (RTL) on Jupyter Notebook.

### 1- Main Page :
Comparing the below two images, it is apparent that mirroring issues with the editor. The following is a list of issues found:
•	Header logo should be on the right.
•	Tabbed items should be on the right (menu tabs)
•	Toolbar items should be swapped 

![1](https://cloud.githubusercontent.com/assets/20149645/19651141/fcfb2090-9a0b-11e6-91da-bfae1754d91e.png)
Figure:1 Main Page in RTL Mode

![2](https://cloud.githubusercontent.com/assets/20149645/19651161/0d775e0c-9a0c-11e6-8b16-ce857909a2f7.png)
Figure:2 Main Page in LTR Mode

### 2- Dropdown Menu:
•	Menu items should be aligned right.
•	Tree-selector buttons should be swapped.
The following images will highlight these issues.

![3](https://cloud.githubusercontent.com/assets/20149645/19651209/3d212a02-9a0c-11e6-8ae5-fca7cceab951.png) 
Figure:3 Dropdown Menu in RTL mode  

![4](https://cloud.githubusercontent.com/assets/20149645/19651234/5435024a-9a0c-11e6-9829-7833674c4038.png)
[...]

PR #2132, opened 2017-02-03
Remove double border in Find and replace UI.

Carreau:
There is a less-than optimal double border between the last button and
the input field.

Remove it by making the right border of the last button be None

---- 

Before/after:
screen shot 2017-02-03 at 13 25 33

### DIFF
--- a/notebook/static/notebook/less/searchandreplace.less
+++ b/notebook/static/notebook/less/searchandreplace.less
@@ -8,7 +8,16 @@
     border-width: 1px;
     border-radius: 0px;
   }
-    
+
+
+  [dir="ltr"] & .input-group-btn + .form-control {
+    border-left: none;
+  }
+
+  [dir="rtl"] & .input-group-btn + .form-control {
+     border-right: none;
+  }
+
   #replace-preview .replace {
     
     & .match {
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -107,7 +107,7 @@
     data-jupyter-api-token="{{token | urlencode}}"
   {% endif %}
  {% endblock params %}
->
+dir="ltr">
 
 <noscript>
     <div id='noscript'>


Carreau:
Hum, it may be better to remove the `border-left` of the text field for when the button is pressed.

blink1073:
Yeah, that seems right.

Carreau:
> Yeah, that seems right.

Done. Also trying to use the same trick with RTL layout. Not perfect, we should inject dir='rtl' by default on the `<body>` tag.

Carreau:
Done. 

Body now have `ltr` by default so we can have rules that apply only to LTR or rules that can apply only to RTL instead of waiting for the first switch to have LTR/RTL flags on the body.

I'm going to assume this may be on interest for @ebraminio and @samarsultan i'm going in unknow territory though so Appreciate input.

ebraminio:
Honeslty I don't know how to see RTL UI for this yet, a quick step-by-step guide would be nice I guess.

Carreau:
> Honeslty I don't know how to see RTL UI for this yet, a quick step-by-step guide would be nice I guess.

I've said it in the other issue but I'll repeat here for completeness and future reader. 
Use the Command palette (Ctrl-Shift-P), and choose "Toggle RTL Layout")

The RTL layout is relatively bad as it was added recently.

ebraminio:
Aha, I didn't know about command palette at all :) Now I see, this patch is OK but IMHO the functionality offered by "Toggle RTL Layout" is not very correct. Is there any plan to translate whole notebook UI to an RTL language? Does notebook is going to have alternative languages, like Arabic and Hebrew? If so this makes of course.

But what I really like to have here, (and a friend requested me to do on this project) is the ability to *just* result cells direction. Of course this feature won't collide with that.

So in nutshell. This is OK to merge. Thanks.

Carreau:
> Aha, I didn't know about command palette at all

Ok, so we're not publicising it well enough! 

> Now I see, this patch is OK but IMHO the functionality offered by "Toggle RTL Layout" is not very correct. Is there any plan to translate whole notebook UI to an RTL language? Does notebook is going to have alternative languages for UI, like Arabic and Hebrew? If so this makes sense of course for testing proposes, otherwise, not very much.

There have been some requests to translate in general, and proposal for RTL (see https://github.com/jupyter/notebook/issues/1852). If the usage of notebook keep increasing then we'll have to do it at some point. The open source project does not have enough resources, or knowledge to it properly though so it's not moving forward fast. We're probably not even in contact with people using RTL language to properly understand what we should be doing. So any help welcome. 

> What I really like to have here, (and a friend requested me to do on this project) is the ability to just toggle result cells' direction. Of course merge of this won't collide with that.

Would that be on a per-cell basis ? Or on all notebook ? with a better understanding we might be able to pull that off.

ebraminio:
> There have been some requests to translate in general, and proposal for RTL

Great. Even the fact I am using a well mirrored software (MediaWiki) on every day basis, as a programmer, I personally don't see much value on making jupyter completely translated to have a RTL UI. However, I understand and appreciate the efforts there. I like to add that I see only one real and effective way making a currently LTR UI to RTL, and that is use of a CSS mirror tool, like [R2](https://github.com/ded/R2) or [CSSJanus](https://github.com/cssjanus/cssjanus), described [here](https://www.mediawiki.org/wiki/ResourceLoader/Features#Flipping) (reading this will be very insightful, specially the excepting rule part). Also [here](https://commons.wikimedia.org/w/index.php?title=Template:Dir&action=edit) is a complete list of RTL languages are listed that you probably like to complete isMirroringEnabled function there, "ar|arc|arz|azb|bcc|ckb|bqi|dv|fa|glk|he|kk-arab|kk-cn|ks|ku-arab|mzn|pnb|ps|sd|ug|ur|ydd|yi".

> Would that be on a per-cell basis ? Or on all notebook ? with a better understanding we might be able to pull that off.

Just per result cells. That would be really super useful. For example you are doing a RTL web scraping  thing and you just want to see a part of the page\ in RTL.

One other good to have RTL support is applying dir="auto" to all languages markdown paragraph, like what I did on https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/6296

PR #2134, opened 2017-02-03
Better handeling or RTL prompts

Carreau:
Basically avoid this:

screen shot 2017-02-03 at 14 49 32

for this:

screen shot 2017-02-03 at 14 51 31

Which is not great, but it's the only option I found to ave consistent breaking regardless of wether we have space or a number in the prompt.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -465,7 +465,7 @@ define([
         } else {
             ns = encodeURIComponent(prompt_value);
         }
-        return 'In&nbsp;[' + ns + ']:';
+        return '<bdi>In</bdi>&nbsp;[' + ns + ']:';
     };
 
     CodeCell.input_prompt_continuation = function (prompt_value, lines_number) {
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -455,7 +455,14 @@ define([
         var toinsert = this.create_output_area();
         this._record_display_id(json, toinsert);
         if (this.prompt_area) {
-            toinsert.find('div.prompt').addClass('output_prompt').text('Out[' + n + ']:');
+            toinsert.find('div.prompt')
+                    .addClass('output_prompt')
+                    .empty()
+                    .append(
+                      $('<bdi>').text('Out')
+                    ).append(
+                      '[' + n + ']:'
+                    );
         }
         var inserted = this.append_mime_type(json, toinsert);
         if (inserted) {
--- a/notebook/tests/notebook/prompt_numbers.js
+++ b/notebook/tests/notebook/prompt_numbers.js
@@ -21,16 +21,16 @@ casper.notebook_test(function () {
         var a = 'print("a")';
         var index = this.append_cell(a);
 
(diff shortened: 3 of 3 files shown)


ebraminio:
How can I repro the issue? Please don't use isolate-override, it is intended just for very very rare cases, using `unicode-bidi: isolate; unicode-bidi: -webkit-isolate;` is usually enough.

Carreau:
> How can I repro the issue

Use the command palette (Ctrl-Shift-P) and choose `toggle-rtl-layout`, you can use the `Help>Edit keyoard shortcuts` menu to bind it to a custom shortcut. 

I'll be happy to get indications on how to fix properly. Thanks !

ebraminio:
Ok thanks. I didn't know about command palette at all :) I see fundamental issue with toggle-rtl-layout feature, explained [here](https://github.com/jupyter/notebook/pull/2132#issuecomment-277473368)

But without considering that, what really can make this better is to instead wrap "In" "Out" with `<span dir="rtl">In</span>` tag, is that possible?

If not, the second preferred way would be applying dir="ltr" (the attribute) or direction: rtl (CSS, less preferred) to the element.

ebraminio:
In a nutshell, consider use of `<bdi>` around "In" and "Out", if is possible.

Carreau:
> But without considering that, what really can make this better is to instead wrapping "In" "Out" with <bdi>In</bdi> tag, is that possible?

Yes it should be possible. We'll have to coordinate with @michaelpacer to do this in nbconvert as well. 
 I also really appreciate your input as well, I don't think anyone regularly involved with IPython/Jupyter is an expert on internationalisation and/or RTL languages even less mix of both. So feedback welcome. 
I was unaware of `<bdi>`/`<bdo>`.

Carreau:
`In` and `Out` should now be wrapped in `<bdi>`.

ebraminio:
Sorry. This probably doesn't work, what I meant was to put just "In" and "Out" text on `<bdi>` and leave the rest :-) (also rule of thumb, never use `<bdo>` and other override things, they are unnecessary really :-) ) So something like `<bdi>In</bdi> [*]:` and `<bdi>Out</bdi>[*]:`.

Carreau:
> Sorry. This probably doesn't work, what I meant was to put just "In" and "Out" text on <bdi> and leave the rest :-) (also rule of thumb, never use <bdo> and other override things, they are unnecessary really :-) ) So something like <bdi>In</bdi> [*]: and <bdi>Out</bdi>[*]:.

Sorry I dropped the ball for a couple of days. This is now done. And thanks for taking the time to explain to me.

blink1073:
image

👍

Carreau:
Which one is the opening bracket and which one is the closing one ?

blink1073:
Full context:

image

Carreau:
We'll still have to make a better job in codemirror...

ebraminio:
I believe for CM you can apply `dir="ltr"` on code parent element.

ebraminio:
Great! Considering the title of this PR, this is now perfect IMO :)

Please have a look at my comment on https://github.com/jupyter/notebook/pull/2132#issuecomment-277501236 after this also. Probably two bug can be filed by it.

Carreau:
Ok, self merging.

Issue #650, opened 2015-10-24
Large notebook fails to save


glider-gun:
When I was using IPython notebook to analyze our experiment data, I noticed I could not save notebook.
The console (from which I started `ipython notebook`) stated:

```
[I 17:37:11.736 NotebookApp] Malformed HTTP message from ::1: Content-Length too long
```

So I guess this problem comes from notebook size.
I was using "bokeh" library to plot my data, and the notebook file was about 100 MB on disk.

To reproduce, I prepared a new notebook and did many plot to produce a large-filesize notebook.

<img width="961" alt="2015-10-24 17 30 24" src="https://cloud.githubusercontent.com/assets/10886262/10711289/926f3e06-7a76-11e5-9d2f-281e18e11247.png">
This does 30001-point plot repeatedly (e.g. 100 plots in above screen shot,)
I could not save the notebook above: when I repeated saving with increasing number of plots, again above about 100MB, I could not save the notebook (with the same console message).

In a little more detail, I could save notebook until 88 plots, when notebook file size was 104756892 byte or 99.904 MB. And I could not save it with 89 plots. By increasing number of plots, file size increased about 1.1 MB per one plot.

I searched issue list, but could not find about this.
Is this limit intentional? Is there some work around for this problem (without removing cells from notebook)?

---

My environments are:

```
$ python -c "import IPython; print(IPython.sys_info())"
{'commit_hash': u'2d95975',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/IPython',
 'ipython_version': '3.2.1',
 'os_name': 'posix',
 'platform': 'Darwin-13.4.0-x86_64-i386-64bit',
 'sys_executable': '/opt/local/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python',
 'sys_platform': 'darwin',
 'sys_version': '2.7.10 (default, Aug 26 2015, 18:15:57) \n[GCC 4.2.1 Compatible Apple LLVM 5.1 (clang-503.0.40)]'}
```

OS: Mac (OSX, 10.9.5 Mavericks)
Browser: Safari 9.0 (9537.86.1.56.2)
matplotlib python library ver. 1.4.3
numpy python library ver. 1.9.2
bokeh python library ver. 0.10.0

willingc:
[...]

PR #2139, opened 2017-02-06
Bump up tornado buffer sizes to allow saving large notebooks

takluyver:
Sort of addresses gh-650

I haven't actually tested with a large notebook

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -224,6 +224,11 @@ class NotebookWebApplication(web.Application):
             iopub_msg_rate_limit=jupyter_app.iopub_msg_rate_limit,
             iopub_data_rate_limit=jupyter_app.iopub_data_rate_limit,
             rate_limit_window=jupyter_app.rate_limit_window,
+
+            # maximum request sizes - support saving larger notebooks
+            # tornado defaults are 100 MiB, we increase it to 0.5 GiB
+            max_body_size = 512 * 1024 * 1024,
+            max_buffer_size = 512 * 1024 * 1024,
             
             # authentication
             cookie_secret=jupyter_app.cookie_secret,


royitaqi:
Sorry for the newbie question: how do I know if this change has been released? I'm running jupyter notebook 5.0.0 (from About menu), and seem to still have the "Content-length too long" [problem](https://github.com/jupyter/notebook/issues/650).

takluyver:
If you've got 5.0, you should have this change. You can look for the file `notebookapp.py` on your computer and push these numbers even further up if your notebook is too big. But if it's over 500MB, it's probably not going to work terribly well anyway - the format is not really designed for storing large amounts of data.

gnestor:
@royitaqi Have you had a chance to test this?

royitaqi:
@gnestor I just tried and it seemed not working. But I'm confused because I see two different version numbers when I was double-checking my version number.

From the About menu in the Jupyter notebook page, I see:
```
The version of the notebook server is 5.0.0
```

But from the command line where I started my jupyter notebook, I see
```
$ jupyter --version
4.3.0
```

Which version number should I look at, @takluyver?


I will also try to upgrade jupyter (if it can be done) to see if it makes any difference.

gnestor:
```
$ jupyter notebook --version
5.0.0
```

royitaqi:
These are inside my `site-packages` directory.

```
notebook
notebook-5.0.0.dist-info
...
jupyter-1.0.0.dist-info
jupyter_client
jupyter_client-5.0.1.dist-info
jupyter_console
jupyter_console-5.1.0.dist-info
jupyter_core
jupyter_core-4.3.0.dist-info
jupyter.py
```

This seems to explain the two version numbers I saw (5.0.0 is from `notebook` and 4.3.0 is from `jupyter_core`).

My jupyter notebook version is indeed 5.0.0.
```
$ jupyter notebook --version
5.0.0
```

I have the following in my `~/.jupyter/jupyter_notebook_config.py`:
```
c.NotebookApp.tornado_settings = {
    'max_body_size': 1073741824,
    'max_buffer_size': 1073741824,
}
```
The number equals to 1GB .

However, I still cannot save my notebook, which should typically be about 200MB large. The error from the jupyter notebook (running in a console ) is:
```
Malformed HTTP message from 127.0.0.1: Content-Length too long
```

Any idea?

gnestor:
I looked at the tornado source at this config should take care of it: https://github.com/tornadoweb/tornado/blob/28e631ae242dec1ae51877fed92d1b16e7a07dd4/tornado/http1connection.py#L558

So no ideas here... If you can share your notebook, I will download and try to debug.

royitaqi:
@gnestor  I uploaded my ipynb: https://expirebox.com/download/418668a409c00c3b1af69fa7e313c449.html. It's 183.33 MB.

This is what I just tried:

1. Start jupyter notebook server.
2. Open this ipynb file.
3. Make a modification in the notebook's code, but don't run (it won't run since it depends on a lot of local libraries).
4. Save the notebook. See the yellow "Error" notification show up in the notebook webpage. See the "Content-Length" error in the jupyter notebook server output.

gnestor:
@royitaqi Sorry, the files expired. Can you update?

jherbel:
@gnestor @royitaqi 
Has there been any progress on this lately? I still seem to have the problem of not being able to store large notebooks. 

Here is an example notebook which produces many times the same plot, such that the notebook
size eventually exceeds 100 MB, causing an error for me:

[https://expirebox.com/download/9fb8b577e271bd91b8d76ba8050f7f90.html](https://expirebox.com/download/9fb8b577e271bd91b8d76ba8050f7f90.html)

The error message I get in the console is:

`[I 18:02:37.663 NotebookApp] Malformed HTTP message from ::1: Content-Length too long`

Here is some information about my setup:

```
$ python` -c "import IPython; print(IPython.sys_info())"
{'commit_hash': u'b467d487e',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/joergherbel/Library/Python/2.7/lib/python/site-packages/IPython',
 'ipython_version': '5.5.0',
 'os_name': 'posix',
 'platform': 'Darwin-15.6.0-x86_64-i386-64bit',
 'sys_executable': '/opt/local/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python',
 'sys_platform': 'darwin',
 'sys_version': '2.7.14 (default, Sep 22 2017, 00:05:22) \n[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)]'}
```

These are the versions of jupyter and jupyter notebook:

```
$ jupyter --version
4.4.0
$ jupyter notebook --version
5.2.1
```

I am running jupyter notebook without any custom configuration, i.e. I don't have a file 
`~/.jupyter/jupyter_notebook_config.py`
and I just use 
`jupyter notebook` 
to start the server.

I don't know if it is relevant, but I have the following file:
```
$ cat /Users/joergherbel/.jupyter/nbconfig/notebook.json 
{
  "load_extensions": {
    "jupyter-js-widgets/extension": true
  }
```

gnestor:
Hi @jherbel! I tested the notebook and I am able to open it, run the cells, and save:

screenflow

```bash
$ jupyter --version
4.4.0
$ jupyter notebook --version
5.3.0.dev0
```

I am also just using the default tornado config from `notebookapp.py`. 

I suspect that the jupyter notebook installation that you are using to run this notebook is not `~/Library/Python/2.7/lib/python/site-packages/notebook/notebookapp.py`. Is that possible?

jherbel:
Hi @gnestor , thanks for your reply! How can I be sure that I indeed run the notebook using `~/Library/Python/2.7/lib/python/site-packages/notebook/notebookapp.py` ? What I did up to now is:

`$ which jupyter
/Users/joergherbel/Library/Python/2.7/bin/jupyter`

and

```
In [1]: import jupyter

In [2]: jupyter.__file__
Out[2]: '/Users/joergherbel/Library/Python/2.7/lib/python/site-packages/jupyter.pyc'

In [3]: import notebook

In [4]: notebook.__file__
Out[4]: '/Users/joergherbel/Library/Python/2.7/lib/python/site-packages/notebook/__init__.pyc'
```

I also noticed that you use Python 3, while I use Python 2, can that make a difference?

gnestor:
> $ which jupyter /Users/joergherbel/Library/Python/2.7/bin/jupyter

Ya, it looks right.

> I also noticed that you use Python 3, while I use Python 2, can that make a difference?

@takluyver @Carreau Is it possible that Python version could somehow affect tornado behavior?

takluyver:
> Is it possible that Python version could somehow affect tornado behavior?

Possible but not likely.

jherbel:
@gnestor One thing I just noticed is that when you tried saving the notebook, you did not re-run, right? Could you please re-run and then try to save it (it will take a while thouhg....)? Otherwise we cannot check whether the error actually occurs for you, since the notebook I uploaded has less than 100 MB (I am not able to save the notebook after it finished running).

aquirdTurtle:
I have the same problem (use many large interactive plots, files get >100MB) although (the windows equivalent of) which jupyter gives me
C:\ProgramData\Anaconda3\Scripts\jupyter.exe
And I'm using jupyter lab instead of the classic notebook. My only workaround right now is to manually break up large notebooks into multiple small notebooks, which is pretty annoying :(

HonestShrimp:
Get the same problem: can't save the Notebook when its size exceed 100 MB (Notebook with numerous plots). I get the message "Auto save failed" in the Notebook and the following in the prompt:

`[I 23:11:32.000 NotebookApp] Malformed HTTP message from ::1: Content-Length too long`

I am using:
```
jupyter --version
4.4.0
jupyter notebook --version
5.5.0
python --version
Python 3.6.5 :: Anaconda, Inc.
```
Same as @aquirdTurtle , I break my notebooks or limit the number of figures, which is not really convinient.

PR #2140, opened 2017-02-06
Preliminary i18n implementation as outlined in Jep16

JCEmmons:
I wanted to get a preliminary PR out here for you all to take a look at so you can see the current progress and also so I can get some help on some of the issues.  Please read the i18n [README](https://github.com/JCEmmons/notebook/blob/jep16/notebook/i18n/README.md) on this branch for instructions on how to get started.

This is still a work in progress, and I am not a professional translator, so I don't claim everything to be perfect just yet. Take a look and let me know what you think.


takluyver:
Thanks for working on this!

I'm aiming to release 5.0 soon (hopefully this week), so I think this will probably miss that release, but I'd like to try to get it in for 5.1.

JCEmmons:
What's the time schedule re: 5.1?   I would like to have some real target 
dates to shoot for....


Regards,

John C. Emmons
Globalization Architect & Unicode CLDR TC Chairman
IBM Software Group
Internet: emmo@us.ibm.com

takluyver:
We don't really have a schedule for releases, I'm afraid. If the changes in 5.0 cause some problems we need to fix, there might be a 5.1 in a couple of weeks, but in that case most of the new features planned for 5.1 would be targeted to 5.2 instead.

If you want some target dates, I'd hope to have the JEP and this infrastructure merged by the end of the month.

JCEmmons:
How do we go about making this so it can be merged?  I've still got a few small things that can be worked on here, but the vast majority of the work here is complete.  Please advise.

JCEmmons:
Still waiting....  @takluyver can you comment pls?

takluyver:
- We're still getting notebook 5.0 out, so this has been waiting for that to happen.
- There are merge conflicts now - can you rebase the branch on master? If you're not familiar with rebasing, I can provide a brief outline of how I usually do it.

JCEmmons:
Any help you can give would be appreciated......


Regards,

John C. Emmons
Globalization Architect & Unicode CLDR TC Vice Chairman
IBM Globalization Team
e-mail: emmo@us.ibm.com

takluyver:
To rebase and resolve merge conflicts, I usually:

- Ensure master is up to date - `git checkout master` and `git pull` (from jupyter/notebook, rather than your fork on Github)
- Check out this branch again - `git checkout jep16`
- Run `git rebase master`
- Each time it reaches a conflict, it will stop and tell you about it. Run `git status` to see which files have conflicts. Fix them in a text editor (look for markers like `>>>>>` and `<<<<<`). Then `git add` the files you've fixed, and run `git rebase --continue`.
- Once it has got through all the commits, it will stop without a message about conflicts.
- Then you need to force push the rebased branch - `git push -f <my_remote> jep16` (replace `<my_remote>` with the name it has for your Github fork of the repo.

JCEmmons:
I'll work on the travis-ci failures next week.

notebook/notebookapp.py
```diff
@@ -969,7 +972,7 @@ def _notebook_dir_validate(self, proposal):
             # If we receive a non-absolute path, make it absolute.
             value = os.path.abspath(value)
         if not os.path.isdir(value):
-            raise TraitError("No such notebook dir: %r" % value)
+            raise TraitError(_("No such notebook dir: %r") % value)
```
takluyver: Unfortunately here the convention of using `_` for translatable strings has clashed with the Python convention of using `_` for a variable that won't be used.

JCEmmons:
I'm looking at the logs from the js tests in Travis CI, and I have no idea how to go about debugging here. Can someone shed some light on what might be going on?

takluyver:
😕 the errors look like this - anyone got any ideas?

```
Test file: /home/travis/build/jupyter/notebook/notebook/tests/services/session.js
FAIL "#kernel-python2 a, #kernel-python3 a" still did not exist in 10000ms
#    type: uncaughtError
#    file: /home/travis/build/jupyter/notebook/notebook/tests/services/session.js
#    error: "#kernel-python2 a, #kernel-python3 a" still did not exist in 10000ms

#    stack: not provided
```

takluyver:
Aha, I've found the issue. You've got `utils.i18n` in a few places where it should be just `i18n`.

JCEmmons:
Thanks @takluyver for your help! That's what happens when you try to refactor things.....

takluyver:
The tests are now producing a whole lot of `ReferenceError: Can't find variable: utils [object Arguments]`

JCEmmons:
It's tough to test this myself, since the tests don't seem to be very 
friendly on Windows.  I'll keep working it, thanks.

takluyver:
Even without running the tests, you may be able to trigger some of these errors by running the notebook in a browser with the Javascript console open.

JCEmmons:
It would be really nice if we could go ahead and get this merged, as I'm getting tired of resolving merge conflicts.

takluyver:
Sorry for the delay @JCEmmons . It's the end of the day over here now, but I'll try to have a look at this tomorrow.

takluyver:
Sorry, most of today got taken up working on Windows packaging of some magnetic simulation code. I haven't forgotten about this, though.

notebook/i18n/de/LC_MESSAGES/notebook.po
```diff
@@ -0,0 +1,416 @@
...
+"Content-Type: text/plain; charset=utf-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Globalization Pipeline\n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"Language: de\n"
+"X-Generator: Poedit 1.8.11\n"
+
+msgid "The Jupyter Notebook requires tornado >= 4.0"
+msgstr "Der Jupyter Notizblock erfordert Tornado >= 4.0"
```
takluyver: I think I'd treat 'Notebook' as part of a name and not translate it. Likewise 'Widgets' a few lines below.

notebook/i18n/de/LC_MESSAGES/notebook.po
```diff
@@ -0,0 +1,416 @@
...
+msgid "Allow the notebook to be run from root user."
+msgstr "Erlauben Sie, dass der Notizblock von Root geführt wird."
+
+msgid ""
+"The Jupyter HTML Notebook.\n"
+"    \n"
+"    This launches a Tornado based HTML Notebook Server that serves up an "
+"HTML5/Javascript Notebook client."
+msgstr ""
+"Die Jupyter HTML Notebook.n n dies startet einen tornadobasierten HTML "
```
takluyver: The newlines have been converted to `n`.

notebook/i18n/de/LC_MESSAGES/notebook.po
```diff
@@ -0,0 +1,416 @@
...
+
+msgid ""
+"Deprecated: Use minified JS file or not, mainly use during dev to avoid JS "
+msgstr ""
+"Missbilligt: Verwenden Sie minified JS Akte oder nicht vermeiden Sie während "
+"dev hauptsächlich gewöhnlich JS recompilation"
+
+msgid "Set the Access-Control-Allow-Credentials: true header"
+msgstr ""
+"Setzen Sie die Zugangskontrolle erlauben Sie Ausweispapiere: Wahre Überschrift"
```
takluyver: `Access-Control-Allow-Credentials` shouldn't be translated again - the HTTP protocol uses English words whatever the locale of the user is.

I know the translations aren't the focus here, but if we're going to merge it with German translations in, I think it's worth doing a bit of basic QA.

notebook/i18n/de/LC_MESSAGES/notebook.po
```diff
@@ -0,0 +1,416 @@
...
+
+msgid "Whether to allow the user to run the notebook as root."
+msgstr "Ob, dem Benutzer zu erlauben, den Notizblock als Wurzel zu führen."
+
+msgid "The default URL to redirect to from `/`"
+msgstr "Die versäumte URL, zu der man umleiten kann, von"
```
takluyver: The `/` has gone missing.

notebook/i18n/de/LC_MESSAGES/notebook.po
```diff
@@ -0,0 +1,416 @@
...
+"authentication."
+msgstr ""
+"Der volle Pfad zu einem Zertifikatautoritätszertifikat für SSL/ TLS "
+"Kundenbestätigung."
+
+msgid "The file where the cookie secret is stored."
+msgstr "Die Akte, wo das Keksgeheimnis gespeichert wird."
+
+msgid "Writing notebook server cookie secret to %s"
+msgstr "Notizblockserverkeksgeheimnis auf %s schreiben"
```
takluyver: My German coworker is laughing uproariously at this translation. I think that while 'keks' is the translation of an edible cookie, an HTTP cookie is called a cookie.

notebook/static/base/js/dialog.js
```diff
@@ -86,7 +86,7 @@ define(function(require) {
             var button = $("<button/>")
                 .addClass("btn btn-default btn-sm")
                 .attr("data-dismiss", "modal")
-                .text(label);
+                .text(i18n.translate(label).fetch());
```
takluyver: Should these labels be translated where they're defined, rather than here?

notebook/static/base/js/i18n.js
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+// Module to handle i18n ( Internationalization ) and translated UI
+
+define([
+    'jed',
+    'moment',
+    'json!../../../i18n/nbjs.json',
+    'json!../../../i18n/de/LC_MESSAGES/nbjs.json',
```
takluyver: Is there a way to only load the data for languages that are needed, rather than pre-loading everything before selecting one?

notebook/static/base/js/namespace.js
```diff
@@ -1,82 +1,82 @@
...
-    jglobal('SessionList','tree/js/sessionlist');
-
-    Jupyter.version = "5.0.0.dev";
-    Jupyter._target = '_blank';
-    return Jupyter;
-});
-
-// deprecated since 4.0, remove in 5+
-var IPython = Jupyter;
+// Copyright (c) Jupyter Development Team.
```
takluyver: Why is the diff showing that this whole file has been changed, when it looks like most of it hasn't? This would mess up tools like `git blame`.

notebook/static/notebook/js/actions.js
```diff
@@ -90,139 +94,164 @@ define(function(require){
...
+            cmd: i18n._('enter command mode'),
+            help    : i18n._('enter command mode'),
             help_index : 'aa',
             handler : function (env) {
                 env.notebook.command_mode();
             }
         },
         'insert-image': {
-            help      : 'insert image',
+            cmd: i18n._('insert image'),
```
takluyver: This PR seems to be adding a `cmd` key to lots of these action definitions. Maybe this is a rebasing mistake from something that was removed in master?

notebook/static/notebook/js/celltoolbarpresets/rawcell.js
```diff
@@ -5,20 +5,21 @@ define([
...
-    ["HTML", "text/html"],
-    ["Markdown", "text/markdown"],
-    ["Python", "text/x-python"],
-    ["Custom", "dialog"],
+    [i18n._("None"), "-"],
+    [i18n._("LaTeX"), "text/latex"],
+    [i18n._("reST"), "text/restructuredtext"],
+    [i18n._("HTML"), "text/html"],
+    [i18n._("Markdown"), "text/markdown"],
+    [i18n._("Python"), "text/x-python"],
```
takluyver: I think these 5 (LaTeX, reST, HTML, Markdown, Python) are names that shouldn't be translated.

notebook/static/notebook/js/codecell.js
```diff
@@ -467,7 +469,7 @@ define([
         } else {
             ns = encodeURIComponent(prompt_value);
         }
-        return '<bdi>In</bdi>&nbsp;[' + ns + ']:';
+        return '<bdi>'+i18n._('In')+'</bdi>&nbsp;[' + ns + ']:';
```
takluyver: I suspect that this is going to look messy with anything other than 2 characters. That's not up to you to fix, though.

notebook/static/notebook/js/commandpalette.js
```diff
@@ -151,7 +151,7 @@ define(function(require){
           }
 
           src[group].data.push({
-            display: humanize_action_id(action_id),
+            display: action.cmd || humanize_action_id(action_id),
```
takluyver: I think this is part of the same `cmd` change noted above.

notebook/static/notebook/js/kernelselector.js
```diff
@@ -265,15 +266,19 @@ define([
...
+        // will pick up the strings.  The actual setting of the text
+        // for the button is in dialog.js.
+        var button_labels = [ i18n._("Continue Without Kernel"), i18n._("Set Kernel"), i18n._("OK") ];
         
         dialog.modal({
-            title : 'Kernel not found',
+            title : i18n._('Kernel not found'),
             body : body,
             buttons : {
                 'Continue Without Kernel' : {
```
takluyver: We can't translate dictionary keys? That's annoying?

notebook/static/notebook/js/notebook.js
```diff
@@ -2710,10 +2728,10 @@ define([
...
                             notebook: that,
                             keyboard_manager: that.keyboard_manager,
-                            title: "Notebook changed",
-                            body: "The notebook file has changed on disk since the last time we opened or saved it. "+
-                                  "Do you want to overwrite the file on disk with the version open here, or load "+
-                                  "the version on disk (reload the page) ?",
+                            title: i18n._("Notebook changed"),
+                            body: i18n._("The notebook file has changed on disk since the last time we opened or saved it.") + " " +
+                                  i18n._("Do you want to overwrite the file on disk with the version open here, or load "+
+                                  "the version on disk (reload the page) ?"),
```
takluyver: Is there a reason to separate these two sentences rather than translating them as one unit? I assume that one chunk of text can be translated more naturally.

notebook/static/notebook/js/notebook.js
```diff
@@ -2812,25 +2830,21 @@ define([
...
-                    $("<em>").text("when you open it")
-                )
-            ).append(".").append(
-                " Selecting trust will immediately reload this notebook in a trusted state."
+            .text(i18n._("A trusted Jupyter notebook may execute hidden malicious code when you open it."))
+            .append(
+                i18n._("Selecting trust will immediately reload this notebook in a trusted state.")
             ).append(
-                " For more information, see the "
+                i18n._(" For more information, see the Jupyter security documentation: ")
```
takluyver: Ditto here - would it be better to translate this message as one piece?

notebook/static/notebook/js/notificationarea.js
```diff
@@ -157,22 +159,26 @@ define([
...
 
-                var msg = 'The kernel has died, and the automatic restart has failed.' +
-                        ' It is possible the kernel cannot be restarted.' +
-                        ' If you are not able to restart the kernel, you will still be able to save' +
+                var msg = i18n._('The kernel has died, and the automatic restart has failed.' +
+                        ' It is possible the kernel cannot be restarted.') + ' ' +
+                        i18n._('If you are not able to restart the kernel, you will still be able to save' +
                         ' the notebook, but running code will no longer work until the notebook' +
-                        ' is reopened.';
+                        ' is reopened.');
```
takluyver: Again, better translated as one?

notebook/static/notebook/js/searchandreplace.js
```diff
@@ -367,7 +374,8 @@ define(function(require){
 
   var load = function(keyboard_manager){
     var action_all = {
-        help: 'find and replace',
+        cmd: i18n._('find and replace'),
```
takluyver: cmd again

notebook/static/notebook/js/textcell.js
```diff
@@ -546,9 +548,9 @@ define([
         highlight_modes : {
             'diff'         :{'reg':[/^diff/]}
         },
-        placeholder : "Write raw LaTeX or other formats here, for use with nbconvert. " +
-            "It will not be rendered in the notebook. " +
-            "When passing through nbconvert, a Raw Cell's content is added to the output unmodified.",
+        placeholder : i18n._("Write raw LaTeX or other formats here, for use with nbconvert. " +
+            "It will not be rendered in the notebook.") + " " +
+            i18n._("When passing through nbconvert, a Raw Cell's content is added to the output unmodified."),
```
takluyver: Combine into one piece for translation?

notebook/static/notebook/js/tour.js
```diff
@@ -127,7 +127,7 @@ define(['jquery', 'bootstraptour'], function($, Tour) {
             reflex: true, // click on element to continue tour
             animation: false,
             duration: this.step_duration,
-            onStart: function() { console.log('tour started'); },
+            onStart: function() { console.log(i18n._('tour started')); },
```
takluyver: This looks like a debugging message, I wouldn't bother translating it.

notebook/static/notebook/js/tour.js
```diff
@@ -141,7 +141,7 @@ define(['jquery', 'bootstraptour'], function($, Tour) {
     };
 
     NotebookTour.prototype.start = function () {
-        console.log("let's start the tour");
+        console.log(i18n._("let's start the tour"));
```
takluyver: Same here, I think it's a debugging message.

notebook/static/base/js/dialog.js
takluyver: I see why this is in place now. I'd still like to find a way to avoid duplicating the button labels in the files where they are defined, but that needn't be solved in this PR.

takluyver:
Overall this is looking good, and I'd like to get it in soon.

JCEmmons:
Thanks @takluyver for reviewing this in detail.  Much of what you suggest is sound, and I can incorporate a lot of your suggestions.  I'm going to be out next week, and I'm not sure I'm going to get to this before I leave.  Please realize that I am not a professional translator, so I would suggest consulting with one before actually claiming full support for German or any other language, for that matter.  There is still some follow up work that will need to be done, but I'm pretty happy with where we are at this point.

notebook/i18n/de/LC_MESSAGES/notebook.po
JCEmmons: Machine translation can be amusing at times...

takluyver:
Yeah, I don't think we need to worry too much about the quality of the translations for merging this; I'm sure that once the infrastructure is in place, there will be people willing to do the translation. But I wonder if maybe we should merge it without the German translation files, and let native speakers create the translations from scratch rather than correcting machine-translated content.

JCEmmons:
We can certainly do that, and I think it's probably a good idea ( i.e. not include the machine translated German, and update the supported languages file accordingly ).  That way, the structure is in place to do what we want, when someone has a language they need to do a real translation on.

I can update this PR to accommodate that....

notebook/i18n/de/LC_MESSAGES/notebook.po
JCEmmons: Moot since we're removing the machine translated German.

notebook/i18n/de/LC_MESSAGES/notebook.po
JCEmmons: Moot since we're removing the machine translated German. This was an artifact of our old MT engine.

notebook/i18n/de/LC_MESSAGES/notebook.po
JCEmmons: Moot since we're removing the machine translated German.

notebook/i18n/de/LC_MESSAGES/notebook.po
JCEmmons: Moot since we're removing the machine translated German.

notebook/i18n/de/LC_MESSAGES/notebook.po
JCEmmons: Moot since we're removing the machine translated German.

notebook/static/base/js/dialog.js
JCEmmons: Yeah, it's kind of confusing until you understand the dialog.js code, but it does work properly, as long as you have the button labels defined properly.

notebook/static/base/js/i18n.js
JCEmmons: Yes, there is, but it usually requires writing a JS plugin in order to do it. I've run into a similar problem in other projects I have worked on.  What you really want here is a synchronous load, which is kind of a no-no in JS these days.   It can be done, and I'm willing to contribute it if/when people actually need a real translation done.

notebook/static/base/js/namespace.js
JCEmmons: Not sure... Looks like just formatting diffs between spaces/tabs.  I will replace with the old version in my next commit.

notebook/static/notebook/js/actions.js
JCEmmons: No, it's actually a necessary addition. There are some cases where the command name tries to show up ( I think on the keyboard shortcuts dialog ), but was not translatable because the displayed name was just being synthesized from the internal name ( i.e. "insert-image" becomes "insert image" ).  And in some cases, you need both a short command name, and a longer help message, which is why I have 2 separate fields for them now.

notebook/static/notebook/js/celltoolbarpresets/rawcell.js
JCEmmons: OK, you're probably correct about  that.  I will fix accordingly.

notebook/static/notebook/js/codecell.js
JCEmmons: Agreed  - it needs to be short.

notebook/static/notebook/js/commandpalette.js
JCEmmons: Right - if there's a "cmd" string associated, then use that, otherwise, do as before.

notebook/static/notebook/js/kernelselector.js
JCEmmons: I'm not sure I understand the issue here.

notebook/static/notebook/js/notebook.js
JCEmmons: We can translate them as a single unit. I think when I was doing the preliminary development, the MT engine had problems with really long keys, which is why I broke it up.  I can merge it back into a single paragraph.

notebook/static/notebook/js/notebook.js
JCEmmons: Yes, agreed.

notebook/static/notebook/js/notificationarea.js
JCEmmons: Yes, agreed.

notebook/static/notebook/js/searchandreplace.js
JCEmmons: Yes, without this in place, the user of a translated version would always see "find and replace" instead of translated text.

notebook/static/notebook/js/textcell.js
JCEmmons: Yes, agreed - and fixed.

notebook/static/notebook/js/tour.js
JCEmmons: ok - no problem to remove.

notebook/static/base/js/i18n.js
takluyver: Maybe it's possible to use require in a way that it doesn't fail if it can't find a module? It looks like we can get the language code from `navigator` before we call `define()`, so we could stick it into a URL - but we don't know at that point if we're asking for a supported language.

I think I'd like to get this worked out before we merge the translation machinery - sorry for another delay, but loading the messages seems pretty fundamental, and the current mechanism is one we know we'll need to replace.

The `json!` thing is pretty neat, though - I didn't know about that before.

notebook/static/notebook/js/actions.js
takluyver: Hmmm. Could we maybe skip one where they're both the same? @Carreau , thoughts?

notebook/static/notebook/js/kernelselector.js
takluyver: I'm guessing that because JS syntax allows `{a: 1}`, where the key becomes the string 'a', it can't also allow `{i18n._('continue'): 1}` to translate the key in the dictionary in place? In Python there would be no problem witht that. Just an annoyance.

notebook/services/edit.html
```diff
@@ -34,7 +34,7 @@
     <div id="menus" class="navbar navbar-default" role="navigation">
```
takluyver: Is this moved into `services` by mistake?

JCEmmons:
I don't think it was intended. I'll look into it after I get the plugin 
working.


Regards,

John C. Emmons
Globalization Architect & Unicode CLDR TC Vice Chairman
IBM Globalization Team
e-mail: emmo@us.ibm.com

notebook/static/base/js/i18n.js
JCEmmons: OK - it required a bit of reworking, but I finally got this to work.  What you will see is a new plugin called i18nload that loads the appropriate JSON data, and then a promise called msg_promise that is returned. Once msg_promise is resolved, then you can retrieve i18n messages as normal. If a supported translation is not requested or not available, then the promise will resolve with just the default data, and everything works the same as before....

JCEmmons:
Hello @takluyver !   I think I've addressed ALL of the issues you raised in your previous review, including:
1. Removal of the machine translated German.
2. Implementation of the i18nload plugin that will correctly load the appropriate translated messages from the JSON based on the language set in the browser ( i.e. it is not hard coded German, as before ).
3. Squashing some of the messages that previously had been split apart into a single message.
4. Remove translation for some of the console only messages ( Things like "tour started..." ).

All merge conflicts have been resolved and the build is passing successfully. I am pretty convinced that this PR is ready to merge. Please go ahead and merge it unless you find some strange problem.....

Thanks!

JCEmmons:
@minrk @takluyver Can we please get this merged asap? I'm getting asked about it regularly...

takluyver:
Sorry, we are keen to get this in, but there's lots of other things to do as well. I'll have a look now.

notebook/services/edit.html
takluyver: Was this deliberately moved? It looks like it still belongs in `templates`?

notebook/static/base/js/i18n.js
```diff
@@ -0,0 +1,54 @@
...
+    	return msg_promise;
+    }
+    var i18n = new Jed(nbjs);
+    init().then(function (msg) {
+    	i18n.msg = msg;
+    	i18n.msg._ = i18n.msg.gettext;
+    });
+    
+    i18n._ = i18n.gettext;
+    i18n.msg = i18n; // Just a place holder until the init promise resolves.
```
takluyver: I think we should move this up a few lines to before the `init().then(...)` - AIUI, an already resolved promise can call the following function immediately, in which case this would reset `i18n.msg` after it has been correctly set above.

notebook/static/base/js/i18nload.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Plugin to load a single locale.
+ */
+define([
+	"require",
+	"module",
```
takluyver: What is `module` and `module.id` here?

notebook/static/base/js/i18nload.js
```diff
@@ -0,0 +1,26 @@
...
+	 // These are only here so that the optimizer knows which ones we MIGHT load.
+	 // We will actually only load the ones we need. There should be one entry
+	 // here for each language you want to support.
+	 // For example, for German....
+	 // "json!base/../../i18n/de/LC_MESSAGES/nbjs.json"
+    ], function (require, module) {
+	return {
+		id: module.id,
+
+		load: function (locale, callerRequire, onload, loaderConfig) {
```
takluyver: Am I right that this function is not used? `i18n.js` appears to use `i18nload.id` and call require directly.

notebook/static/base/js/i18nload.js
```diff
@@ -0,0 +1,26 @@
...
+ * Plugin to load a single locale.
+ */
+define([
+	"require",
+	"module",
+	 // These are only here so that the optimizer knows which ones we MIGHT load.
+	 // We will actually only load the ones we need. There should be one entry
+	 // here for each language you want to support.
+	 // For example, for German....
+	 // "json!base/../../i18n/de/LC_MESSAGES/nbjs.json"
```
takluyver: Does the optimiser look at commented values, or is this comment describing something that happens if it's uncommented? If we need uncommented values here, what stops them all being loaded when `i18nload` is loaded?

takluyver:
This looks like JS running before the promise has resolved will use the untranslated messages, so you could end up with a mixture of translated and untranslated messages on the page. Is there something I'm not seeing that prevents this? I can't believe this isn't a solved problem.

notebook/services/edit.html
JCEmmons: This looks like a mistake... I will move it back.

notebook/static/base/js/i18nload.js
JCEmmons: module is part of requirejs, which allows us to define the name of the module for the plugin. Remember, a plugin has two parts, the javascript module ( in this case i18nload ), and the argument ( the stuff after the ! ). So the plugin defines its module id in the object it passes back, and then i18n.js does a require on the module.id.

notebook/static/base/js/i18nload.js
JCEmmons: The load function is used as it's what actually loads the right data ( i.e. it's the guts of the plugin ).  As discussed before, the plugin has two parts, the module ID ( stuff before the ! ) and the argument ( which in this case is "locale" the first argument passed to the load function.

takluyver: Aha, I haven't come across requirejs plugins before.

notebook/static/base/js/i18nload.js
JCEmmons: The optimizer doesn't look at commented values.  You will need to comment them back in in order to support a particular language.  The answer to your 2nd question is, nothing stops them, as they will all get loaded if there are a lot of them.   The current implementation is not optimal, and I understand that.   However, since we don't have real translations yet, I don't want this to be a showstopper issue for getting the merge done.  I can work on improving the plugin so that it is more robust.   One approach that I have used in the past is to allow the set of translations you want to load to be configurable via require.config(), and then have the plugin read that.

notebook/static/base/js/i18n.js
JCEmmons: A good idea... This makes it a lot more obvious what we are doing as well...

notebook/services/edit.html
JCEmmons: I moved it back where it belongs, and regenerated i18n/nbui.pot so that it picks up the strings appropriately.

takluyver:
I'm getting ready to go to a conference, so I won't be able to look at this much in the near future. I'd like to ask one of the other developers to have a look (@gnestor @Carreau @minrk ). We're also getting close to another release, so if we don't think this is going to be mergeable soon

I know this is frustrating, and I'm sorry it's taking so long to process it.

notebook/static/base/js/utils.js
```diff
@@ -237,6 +237,7 @@ define([
         "ansi-cyan-intense",
         "ansi-white-intense",
     ];
+
```
JCEmmons: Not sure how this got there, but I'll clean it up....

JCEmmons:
I suppose that in theory you could be right, but in practice I'm not seeing it. Usually the promise is resolved quickly and the translations come in as expected.

minrk:
@JCEmmons thanks for addressing the review! I think this is in good shape to merge, with one exception I just noticed while testing this out locally: Some of the changes here have mixed in some tab indentation instead of spaces, and added Windows CRLF line endings in a few places. This is part of why the diff on a few files looks so large. You can tell git to [handle CRLF automatically](https://help.github.com/articles/dealing-with-line-endings/#platform-windows) with:

    git config --global core.autocrlf true

to avoid this sort of problem in the future. If you check the "Allow edits from maintainers" box on the right:

screen shot 2017-07-11 at 10 43 25

I can do the last conflict resolution and fix these little things and get this merged. Or you can pull from [my branch](https://github.com/minrk/notebook/tree/2140), if you prefer.

JCEmmons:
Hi @minrk  and thanks.... Allow edits from maintainers is already enabled, so if you can handle the last merge conflict and do the merge I would appreciate it greatly!!! Thanks......

minrk:
We can continue to iterate on this in master as things come up. Thanks for your work and patience!

minrk:
Due to the issues with line endings and whitespace changes, I chose to squash this PR rather than deal with a complicated rebase. Hopefully we can catch similar issues earlier in subsequent PRs.

gnestor:
Hi @JCEmmons! I'm about to publish a release candidate for notebook 5.1 and I noticed the following error when opening a new terminal (in the Jupyter Notebook):

image

Does this make sense to you?

JCEmmons:
Somewhat.... It appears that it's not finding Jed.js, which is now a 
requisite package. It appears that jed 1.1.1 never made it into the 
bower.json, which would explain the error.


Regards,

John C. Emmons
Globalization Architect & Unicode CLDR TC Vice Chairman
IBM Globalization Team
e-mail: emmo@us.ibm.com

takluyver:
I installed RC1 and I can't reproduce this - @gnestor did you get it worked out.

It does [look like](https://github.com/jupyter/notebook/blob/b099303a96dcc9c905f944d27dfc030ff9832fbc/bower.json#L12) jed is listed in bower.json - maybe it was a matter of clearing out some cache?

gnestor:
I'm even seeing this in master 🤔 I don't see this when opening notebooks, just terminals.

jcb91:
I'm seeing the same missing jed error breaking the jupyter_nbextensions_configurator standalone page at https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/issues/43, using a pip-installed notebook 5.1.0 :cry:

PR #2145, opened 2017-02-07
add base_url to view page

### LINKED ISSUE #2143: The new view button for html file is not working
screen shot 2017-02-06 at 9 33 12 pm

while the url `..../user/randy3k/files/foo.html` still works.

It is possibly due to https://github.com/jupyter/notebook/commit/224294c2743df1df2855ea511ad4f612080472b0
or it is an issue for jupyterhub.

minrk:
closes #2143

### DIFF
--- a/notebook/templates/view.html
+++ b/notebook/templates/view.html
@@ -23,7 +23,7 @@
     }
   </style>
   <div id="container">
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, I suspected it was something like this, but wasn't sure where to look.

PR #2148, opened 2017-02-07
Work on changelog for 5.0

takluyver:
We seem to have missed the extension that defines Github issue & PR links until now.


Carreau:
should we turn the sphinx ext into a package at some point ?

Carreau:
Thanks for all the screenshots !

takluyver:
Yeah, I suspect that making it a package will make sense. I just had a look, there's [something similar](https://pypi.python.org/pypi/sphinxcontrib-github/0.1.3) on PyPI, but I prefer how ours works.

PR #2159, opened 2017-02-08
attachment mimebundle values are strings, not lists

### LINKED ISSUE #2158: Markdown Cell attachment not persisting
Hopefully I'm missing something obvious.

Running on latest master (https://github.com/jupyter/notebook/commit/038eda825f77a48c1c057c906691a04385fd9b1c), it looks like markdown attachments are not getting persisted.

markdown_attachment

minrk:
When first added, the first item of the list contained the attachment. After roundtripping to a file, the list was correctly converted to a single string, the first element of which is a single character, not the full data.

closes #2158

### DIFF
--- a/notebook/static/notebook/js/textcell.js
+++ b/notebook/static/notebook/js/textcell.js
@@ -125,7 +125,7 @@ define([
          * Add a new attachment to this cell
          */
         this.attachments[key] = {};
-        this.attachments[key][mime_type] = [b64_data];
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks Min, that's working for me.

minrk:
I also found an issue where copying cells wouldn't copy attachments. I fixed that one and added a test for it as well.

PR #2168, opened 2017-02-09
Add test dependencies to `devDependecies`

gnestor:
It’s more conventional within the JS community to include test dependencies in `devDependencies` of `package.json` vs. [requiring developers to install them before runnings tests](https://github.com/jupyter/notebook/blob/master/CONTRIBUTING.rst#javascript-tests). `devDependecies` are only installed on `npm install` inside a project vs. `npm install [package name]`, so it will only be installed by developers.

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,9 @@
   },
   "devDependencies": {
     "bower": "*",
+    "casperjs": "^1.1.3",
     "less": "~2",
+    "phantomjs": "^2.1.7",
     "requirejs": "^2.1.17"
   },
   "dependencies": {


package.json
```diff
@@ -15,7 +15,9 @@
   },
   "devDependencies": {
     "bower": "*",
+    "casperjs": "^1.1.3",
     "less": "~2",
+    "phantomjs": "^2.1.7",
```
minrk: installing phantomjs produces a warning that the package has been renamed 'phantomjs-prebuilt'

blink1073:
Isn't this a private package?

blink1073:
i.e., it could have the `private: true` flag and only include `dependencies`.

gnestor:
Ok, so we should depend on `phantomjs-prebuilt` instead? 

Regarding path stuff, if we run an npm script, npm will add `node_modules/.bin` to the path). So we could just call `npm run test:[test name]` from `notebook/jstest.py` or possibly make `npm run test` the entry point.

minrk:
@blink1073 yes, that makes sense.

@gnestor yes, using `npm run` would be the thing to do. I mainly mean that it would be appropriate to also do it in this PR.

One difference, though: in general, our javascript tests can be run against an *install* of the notebook package. Switching to `npm run` would mean that the tests cannot be run except from the repo.

gnestor:
@minrk Hmmm. So users who have installed the notebook using `pip install notebook` vs. cloning the repo can run the tests?

minrk:
Yes, providing they have casper and phantom installed.

gnestor:
Ok, so  I guess we have make a call: leave the test setup as is and allow users to run tests against a user install of the notebook or change it so that it's more Node package-like, with test dependencies in the devDependencies and a `npm run test` script... 🤔

minrk:
In favor of changing:

- install js test dependencies in one command
- doesn't require any global installation of phantom or casper

Against:

- can't run tests against an install without at least *also* doing a checkout and build from source, ...
- ...unless `jstest` handles both `npm run` and global cases, which may not be worth maintaining.

Not either way:

- The javascript here is not an npm package in any way–we just happen use npm to grab a few build dependencies, so following npm package practices doesn't seem to point one way or the other.

I'm fine either way - none of those points seem too strong one direction or another. using local phantom & casper would probably have saved me a few headaches running tests locally. Given that this package is heading toward "maintenance mode" and isn't an npm package, I might be inclined to leave it alone unless the change solves problems people are actually facing ("It's too hard to setup and run tests locally" totally counts).

If you do make the change, do make sure to update travis and `jstest.py` to use the bundled casper, rather than the global one.

gnestor:
Ok, it sounds like a "if it ain't broke don't fix it" situation. I'll close this and we can always re-open if it comes back up.

PR #2173, opened 2017-02-10
Add a clear_output code cell event

jasongrout:
This is triggered when a code cell clears its output. It is needed for the widgets to be able to clear themselves when the cell output is cleared, fixing https://github.com/ipython/ipywidgets/issues/637

PR #2174, opened 2017-02-10
Backport #2173 to 4.x

jasongrout:
This is triggered when a code cell clears its output.

This backports #2173 to 4.x.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -506,7 +506,7 @@ define([
     CodeCell.prototype.clear_output = function (wait) {
         this.output_area.clear_output(wait);
         this.set_input_prompt();
-        this.events.trigger('clearOutput.CodeCell', {cell: this});
+        this.events.trigger('clear_output.CodeCell', {cell: this});
     };
 
 
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -506,6 +506,7 @@ define([
     CodeCell.prototype.clear_output = function (wait) {
         this.output_area.clear_output(wait);
         this.set_input_prompt();
+        this.events.trigger('clearOutput.CodeCell', {cell: this});
     };
 
 


minrk:
I take it this is a request for a 4.5 release?

jasongrout:
A request for another change in 4.x, regardless the version it makes it. And I think we should hold off on releasing until we make sure there are no more changes needed to get widgets working with 4.x :).

minrk:
Does this mean that ipywidgets 6 won't work with 4.4?

jasongrout:
ipywidgets 6 will work with 4.4. This enables us to clear the widgets in a cell when the cell's output is cleared, which resolves a longstanding feature request/frustration with widgets (see https://github.com/ipython/ipywidgets/issues/637)

takluyver:
If it's a new feature, I don't think it warrants another 4.x release. There are a lot of other improvements which are about to go out in 5.0. If this is backported, it's not clear why we wouldn't just backport a load of other improvements to the 4.x branch.

jasongrout:
Since 5.0 is coming soon, I'm fine with saying that this is a new feature of widgets on 5.0. That said, 

> If this is backported, it's not clear why we wouldn't just backport a load of other improvements to the 4.x branch.

This backport is already done, so I think it's different than planning a lot more time to backport lots of other things.

takluyver:
Backporting the commit is not the hard bit: making sure that everything works together, cutting releases and helping users who run into issues with the new release is what takes time and effort.

This bugs me a bit because by asking for a backport and a new release, you're implicitly saying that the change you want is more important than the other changes people have made for 5.0. If you believe that it is, then please put a justification of that when proposing a backport. But when we're approaching a release, the default is that new features won't get backported.

jasongrout:
> This bugs me a bit because by asking for a backport and a new release, you're implicitly saying that the change you want is more important than the other changes people have made for 5.0. If you believe that it is, then please put a justification of that when proposing a backport. But when we're approaching a release, the default is that new features won't get backported.

Thanks for sharing this out loud. I'm not asking for a backport (I contributed it, here), and I agreed with you that we consider this a 5.0 feature (good call!) - perhaps I should have closed this PR at that point. Sorry, I don't mean to set your priorities or imply your priorities are wrong - on the contrary, I really applaud your and others' focus and work on moving forward with notebook 5.0. We're also working hard on trying to get an overdue ipywidgets 6.0 out the door too.

I really appreciate the 4.4 release, which was a real blocker for ipywidgets (thus a really high priority for us), and I'm of course willing to help with potential problems from the changes we contributed.

takluyver:
Sorry, I should have said asking for the backport to be merged and a new
release made. Either way, thanks for taking my somewhat irate earlier
message well.

If you (or anyone) think a PR should be backported, generally I'd like to
see that raised in comments on the original pr, rather than opening a
backport pr straight away. I can see that you're trying to make it easier,
but from my end it feels a bit pushy.

I'd also like to work out what our version numbers actually mean. I don't
have a clear sense of which repos use the model from IPython (feature
releases are major, bug fix roll ups minor), and which follow semver
(breaking releases major, feature releases minor).

Issue #2172, opened 2017-02-10
Terminal background wrong size until resized


takluyver:
Ping @blink1073 - is it possible this is related to #2171 which I just merged?

When I open a terminal page, the black background of the terminal shows up only as a narrow bar across the top of the terminal area. If I resize the window and trigger a terminal resize, it assumes the correct size and continues to work correctly. I see a similar issue on Firefox and Chromium, though not exactly the same (the bar is smaller in Chromium). The terminal still appears to be working, but light coloured text is not easily readable.

Here's what I see in Firefox:

![screenshot from 2017-02-10 11-53-59](https://cloud.githubusercontent.com/assets/327925/22826050/48b9f0f2-ef88-11e6-8fe0-9a09f861f5f7.png)

blink1073:
Drat, I had seen that at first but then it mysteriously went away and never resurface across the three browsers.

takluyver:
That suggests caching issue, but I did make an effort to check that it wasn't that...

takluyver:
I made some further effort, and I'm still seeing it consistently. If it is something in cache, it's so far resisted every trick I can think of to avoid loading it.

blink1073:
I tried for the past half hour and can't figure out what to do about this, shall I make a PR to revert the version?

takluyver:
Can you still not reproduce it? Or can you see it but aren't sure of the cause?

blink1073:
I see it but can't remove it programmatically.

blink1073:
It doesn't happen in the JupyterLab terminal [example](https://github.com/jupyterlab/jupyterlab/tree/master/examples/terminal), but I'm not sure of the exact implementation difference that could cause it.

takluyver:
If we can't come up with a better resolution, I think we'll need to revert it before 5.0, but let's not do that just yet.

- Is it worth reporting this to xtermjs? As far as I know, we're using their library as intended.
- Might other people on our team have some ideas? @yuvipanda and @rgbkrk have done stuff with the terminal in the past, IIRC.

blink1073:
FWIW I pulled latest master on xterm.js and their included demo works fine as well.

takluyver:
Filed sourcelair/xterm.js#539 - at least if we're doing something daft, it might be obvious to them what it is.

blink1073:
👍

PR #2184, opened 2017-02-13
Upgrade xterm.js to newly released 2.3.2

takluyver:
Closes gh-2172


takluyver:
That was a satisfyingly quick response from the xterm.js devs. :-)

blink1073:
Indeed.

PR #2186, opened 2017-02-14
Support underline and inverse in ANSI escape codes

tonycpsu:
I was dumping output from [cdiff](https://github.com/ymattw/cdiff) into a notebook and noticed that its use of underline and inverse escape codes wasn't being reflected in the notebook output.  A couple simple tweaks to the ANSI parsing and some added CSS classes and it now does.

### DIFF
--- a/tools/tests/ANSI Test.ipynb
+++ b/tools/tests/ANSI Test.ipynb
@@ -48,22 +48,22 @@
      "text": [
       "This is normal text\n",
       "\n",
-      "\u001b[01mThis is bold text\n",
+      "\u001b[01mThis is bold text\u001b[00m\n",
       "\n",
-      "\u001b[04mThis is underlined text\n",
+      "\u001b[04mThis is underlined text\u001b[00m\n",
       "\n",
-      "\u001b[07mThis is inverse text\n"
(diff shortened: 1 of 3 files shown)


tools/tests/ANSI Test.ipynb
```diff
@@ -34,6 +34,42 @@
...
+    }
+   ],
+   "source": [
+    "print (\"This is normal text\")\n",
+    "print()\n",
+    "print (\"{ESC}01mThis is bold text\".format(**locals()))\n",
+    "print()\n",
+    "print (\"{ESC}04mThis is underlined text\".format(**locals()))\n",
+    "print()\n",
+    "print (\"{ESC}07mThis is inverse text\".format(**locals()))"
```
takluyver: Is it intentional that you don't reset the style between lines, so this actually generates:
- normal
- bold
- bold + underlined
- bold + underlined + inverse

tonycpsu: Nope, good catch. I've updated the test to reset.

takluyver:
Great, thanks. This looks good to me.

PR #2195, opened 2017-02-15
Don't force notebooks to be opened in a new tab

joar:
New tabs can be convenient, but we shouldn't force users to open their notebooks in a new tab.
If the users would like to open their notebook in a new tab they may use the [Meta]-click shortcut.

As far as I can see, there are no technical limitation prohibiting notebooks from being opened in the same tab.

[When to use target=”_blank”](https://css-tricks.com/use-target_blank/) has more reasoning regarding `target="_blank"`.

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -689,12 +689,6 @@ define([
 
         item.find(".item_buttons .running-indicator").css('visibility', running ? '' : 'hidden');
 
-        // directory nav doesn't open new tabs
-        // files, notebooks do
-        if (model.type !== "directory") {
-            link.attr('target',IPython._target);
-        }
-
         // Add in the date that the file was last modified
         item.find(".item_modified").text(utils.format_datetime(modified));
         item.find(".item_modified").attr("title", moment(modified).format("YYYY-MM-DD HH:mm"));


Carreau:
Thanks for your PR. Though this actually _removes_ the ability to configure this.
If you do not like default behavior you should set `IPython._target` in your `custom.js`.

This was originally left empty, set to `_blank` on purpose after many requests from user to _not_ open in the same tab by default, then made configurable hereafter to fulfill your use case. 

This would then just revert to original behavior for which we _did_ had a strong pushback. 

So -1.

ivanov:
Hey there @joar, I'm going through issues and it seems to me that it makes sense to close this one. As @Carreau mentioned, we believe the `IPython._target` mechanism is sufficient for the use case you speicify and we do not wish to remove that feature unconditionally.

Thanks everyone and happy hacking! :bowtie:

joar:
@Carreau Sorry about that, I did not find any of those discussions after some quick searches, I might have missed a case of `["jupyter", "IPython"] * ["don't open new tab", "new tab",  "target", "_blank"]`.

I did not know about the setting of variables in `custom.js`. :sun_behind_rain_cloud:

Carreau:
@joar no worries, we could do a better job at documentation. 
Where would you expect to find such information ? 
Do you want to send a PR that add this information in such a place ? 

Thanks !

Issue #2153, opened 2017-02-07
Keyboard shortcut editor documentation. 


Carreau:
Review and document how to set shortcuts with space, comma, dash, minus, esc .... etc as they are special characters and need to be spelled out.

gnestor:
@Carreau Would you like to tackle this before we release 5.0.0beta1 today?

Carreau:
> @Carreau Would you like to tackle this before we release 5.0.0beta1 today?

I can try, but I have a couple things that have been requested to be done fast. 
Also, as I know how to do it, I might miss the main point on how to explain it. 
I'll try to give it a shot though.

gnestor:
Ok, if you explain here then I can make the changes and submit a PR 👍

Carreau:
Most of the info are already on the dialog: 

```
This dialog allows you to modify the keymap of the command mode, and persist the changes. You can define many type of shorctuts and sequences of keys.
Use dashes - to represent keys that should be pressed with modifiers, for example Shift-a, or Ctrl-;.
Separate by commas if the keys need to be pressed in sequence: h,a,l,t.
You can combine the two: Ctrl-x,Meta-c,Meta-b,u,t,t,e,r,f,l,y. Casing will have no effects: (e.g:  ; and : are the same on english keyboards). You need to explicitelty write the Shift modifier. Valid modifiers are Cmd, Ctrl, Alt ,Meta, Cmdtrl. Refer to developer docs for the corresponding keys depending on the platform.You can hover on the name/description of a command to see its exact internal name and differentiate from actions defined in various plugins. Changing the keybindings of edit mode is not yet possible.
```

things that need to be added: Modifier are (IIRC) case sensitive. Cmdtrl is Command on mac, ctrl on windows/linux. Meta is Alt/Option (IIRC). space comma dash/minus need to be explicitly spelled out.

michaelpacer:
I can work on this today, it happened to arise in talking with @Carreau about documentation and keyboard shortcuts, so I'll be looking at this feature today anyway.

gnestor:
Thanks @michaelpacer

michaelpacer:
What does "close pager" mean?

[...]

PR #2198, opened 2017-02-16
More complete & accurate keybinding editor documentation

michaelpacer:
This should address #2153, though there is still room for improvement. 

In particular, I would like to add a shorter default display, with a "see more" button that will expand into the current text. 

That feature is not a blocking feature though, as it is a inconvenience only if you need to scroll through it frequently (which I've only noticed because of the amount of time I spent testing the editor).

### DIFF
--- a/notebook/static/notebook/js/shortcuteditor.js
+++ b/notebook/static/notebook/js/shortcuteditor.js
@@ -141,8 +141,8 @@ var KeyBindingList = createClass({
             "<kbd>l</kbd> <kbd>t</kbd> but not <kbd>h</kbd> <kbd>a</kbd> <kbd>a</kbd> <kbd>l</kbd> "+
             "<kbd>t</kbd> or <kbd>a</kbd>  <kbd>h</kbd> <kbd>l</kbd> <kbd>t</kbd>.\n"+
             "   - Sequences can include the same key multiple times (e.g., `d,d`).\n"+
-            "   - You cannot include a sequence that is a 'prefix' of another sequence.\n"+
-            "     - E.g., `d,d,d` cannot be used a the same time as `d,d`).\n"+ 
+            "   - You cannot include any pairs of sequences where one is a 'prefix' the other.\n"+
+            "     - E.g., `d,d,d` cannot be used a the same time as `d,d`.\n"+ 
             "   - Key combinations are unique elements that can be used in a sequence.\n"+ 
             "     - E.g., `Ctrl-d,d` and `d,d` can exist at the same time and are both valid key sequences.\n"+
             "\n"+
--- a/notebook/static/notebook/js/shortcuteditor.js
+++ b/notebook/static/notebook/js/shortcuteditor.js
@@ -106,10 +106,19 @@ var KeyBindingList = createClass({
              }
           }));
       });
-      children.unshift(createElement('div', {className:'well', key:'disclamer', dangerouslySetInnerHTML:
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/shortcuteditor.js
```diff
@@ -106,27 +106,55 @@ var KeyBindingList = createClass({
...
+            "This dialog allows you to modify the keyboard shortcuts available in command mode. "+ 
+            "Any changes will be persisted between sessions and across environments. "+
+            "You can define two kinds of shorctuts: **key combinations** and **key sequences**.\n"+
+            "\n"+
+            " - **Key Combinations**:\n"+
+            "   - Use hyphens `-` to represent keys that should be pressed at the same time.\n"+
+            "   - This is designed for use with *modifier* keys: `Cmd`, `Ctrl`, `Alt` ,`Meta`, "+
+            "`Cmdtrl`, and `Shift`.\n"+ 
+            "   - At most, one non-modifier key can exist in a key combination.\n"+
+            "   - Multiple non-modifier key can exist in a key combination.\n"+
```
takluyver: This line appears to directly contradict the one above it. Did you mean to remove 'non-'?

takluyver:
This is great, but:

- We're still missing an explanation of 'Cmdtrl'
- I am a bit concerned that is now so long that I don't see the actual shortcut editing area without scrolling down. Maybe an easy way to get round this for 5.0 would be to move the explanation to the end of the dialog, and have just the first bit at the top, like:

> Here you can modify the keyboard shortcuts available in command mode. Your changes will be stored for later sessions. See more **details of defining keyboard shortcuts** below.

(with a link from the bold text to scroll the help info into view)

notebook/static/notebook/js/shortcuteditor.js
michaelpacer: Yes!

michaelpacer:
I like that idea, but how should I handle the text scrolling? Just an anchor tag with an id inside the beginning of the detailed explanation and a hyperlink?

takluyver:
I think that should work. Try it and see :-)

notebook/static/notebook/js/shortcuteditor.js
```diff
@@ -106,27 +106,55 @@ var KeyBindingList = createClass({
...
+            "   - *Invalid Example*s: `a-b` and `a-Ctrl-Shift`. \n"+
+            " - **Key Sequences**:\n"+
+            "   - Use commas `,` to represent keys that should be pressed in sequence.\n"+
+            "   - The order in which keys must be pressed exactly matches the left-to-right order of "+
+            "the characters in the sequence, with no interruptions.\n"+
+            "     - E.g., `h,a,l,t` would be triggered by typing <kbd>h</kbd> <kbd>a</kbd> "+
+            "<kbd>l</kbd> <kbd>t</kbd> but not <kbd>h</kbd> <kbd>a</kbd> <kbd>a</kbd> <kbd>l</kbd> "+
+            "<kbd>t</kbd> or <kbd>a</kbd>  <kbd>h</kbd> <kbd>l</kbd> <kbd>t</kbd>.\n"+
+            "   - Sequences can include the same key multiple times (e.g., `d,d`).\n"+
+            "   - You cannot include a sequence that is a 'prefix' of another sequence.\n"+
```
Carreau: I read the the opposite way: `if d,d,d is bound you cannot bind d,d`. Both are true.

takluyver:
Nice, thanks Mike. :-)

Issue #226, opened 2015-07-23
TypeError: IPython.load_extensions is not a function


ofenerci:
Dear All,

I am some problems about extensions. I have installed Jupyter as described in "https://github.com/jupyter/notebook" but when I try to install some extensions installed inside "~/.jupyter/nbextensions", I get an error 
--------------------------------------------->--------------------------------------------------------
ln [1]: %%javascript
IPython.load_extensions('IPython-notebook-extensions-master/usability/python-markdown/main');

Javascript error adding output!
TypeError: IPython.load_extensions is not a function
See your browser Javascript console for more details.

------------------------------------------<------------------------------------------------------------------------
I don' know what could be a problem and am lost in the ocean of Jupyter. 

Ozhan

takluyver:
We're still getting Jupyter released - if you're feeling lost with the development version, please install the stable version of IPython, which includes the notebook, using:

    pip install ipython[notebook]

It looks like we no longer put `load_extensions()` in the interactive namespace, so now you'll need to do something like:

```javascript
require(['base/js/utils'],
function(utils) {
    utils.load_extensions('...');
});
```

PR #2215, opened 2017-02-22
Cull idle kernels

kevin-bates:
(Updated to include review recommendations below...)

This change introduces the ability to cull kernels that have been idle for a specified amount of time.  Off by default, it is enabled by setting `--MappingKernelManager.cull_idle_timeout` to a positive value representing the number of seconds a kernel must remain inactive to be culled.  This feature leverages [activity tracking's](https://github.com/jupyter/notebook/pull/1827) `last_activity` value to determine when the last activity for a given kernel occurred.  Typical values for `cull_idle_timeout` will be on the order of 43200 (12 hours) or 86400 (24 hours) seconds as described in this [JKG issue](https://github.com/jupyter/kernel_gateway/issues/96).  These changes originally targeted Jupyter Kernel Gateway per [issue #226](https://github.com/jupyter/kernel_gateway/issues/226) but really belongs in the Notebook server.

An optional property has been introduced to adjust the interval in which idle kernels are culled provided they have exceeded the aforementioned timeout property.  This property is adjusted by setting `--MappingKernelManager.cull_interval` to a positive value.  If not set or non-positive, the default value of 300 (5 minutes) seconds will be used.

JKG [Issue 227](https://github.com/jupyter/kernel_gateway/issues/227) has been opened to remove the activity tracking that was also in place in the JKG.  Prior to that, there could be false reports of active kernels that have been culled by the underlying MappingKernelManager.  If not removed, we should minimally adjust the JKG's activity tracking handler to consult the active kernel manager class on each request.

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -264,6 +264,10 @@ class MappingKernelManager(MultiKernelManager):
         if not self._initialized_culler and self.cull_kernels_after_minutes > 0:
             if self._culler_callback is None:
                 loop = IOLoop.current()
+                if self.kernel_culling_interval_seconds <= 0: #handle case where user set invalid value
+                    self.log.warn("Invalid value for 'kernel_culling_interval_seconds' detected (%s) - using default value (%s).",
+                        self.kernel_culling_interval_seconds, self.kernel_culling_interval_seconds_default)
+                    self.kernel_culling_interval_seconds = self.kernel_culling_interval_seconds_default
                 self._culler_callback = PeriodicCallback(
                     self.cull_kernels, 1000*self.kernel_culling_interval_seconds, loop)
                 self.log.info("Culling kernels with idle durations > %s minutes at %s second intervals ...",
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -58,8 +58,11 @@ class MappingKernelManager(MultiKernelManager):
             raise TraitError("kernel root dir %r is not a directory" % value)
         return value
 
+    cull_idle_timeout_minimum = 300 # 5 minutes
     cull_idle_timeout = Integer(0, config=True,
-        help="""Timeout (in seconds) after which a kernel is considered idle and ready to be culled."""
+        help="""Timeout (in seconds) after which a kernel is considered idle and ready to be culled.  Values of 0 or
+        lower disable culling. The minimum timeout is 300 seconds (5 minutes). Positive values less than the minimum value
+        will be set to the minimum."""
     )
 
     cull_interval_default = 300 # 5 minutes
@@ -252,6 +255,10 @@ class MappingKernelManager(MultiKernelManager):
         """
         if not self._initialized_culler and self.cull_idle_timeout > 0:
             if self._culler_callback is None:
+                if self.cull_idle_timeout < self.cull_idle_timeout_minimum:
+                    self.log.warning("'cull_idle_timeout' (%s) is less than the minimum value (%s) and has been set to the minimum.",
+                        self.cull_idle_timeout, self.cull_idle_timeout_minimum)
+                    self.cull_idle_timeout = self.cull_idle_timeout_minimum
                 loop = IOLoop.current()
                 if self.cull_interval <= 0: #handle case where user set invalid value
                     self.log.warning("Invalid value for 'cull_interval' detected (%s) - using default value (%s).",
@@ -268,8 +275,13 @@ class MappingKernelManager(MultiKernelManager):
     def cull_kernels(self):
         self.log.debug("Polling every %s seconds for kernels idle > %s seconds...",
             self.cull_interval, self.cull_idle_timeout)
+        """Create a separate list of kernels to avoid conflicting updates while iterating"""
         for kernel_id in list(self._kernels):
-            self.cull_kernel_if_idle(kernel_id)
+            try:
+                self.cull_kernel_if_idle(kernel_id)
+            except Exception as e:
+                self.log.exception("The following exception was encountered while checking the idle duration of kernel %s: %s",
+                    kernel_id, e)
 
     def cull_kernel_if_idle(self, kernel_id):
         kernel = self._kernels[kernel_id]
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -11,14 +11,16 @@ import os
 
 from tornado import gen, web
 from tornado.concurrent import Future
-from tornado.ioloop import IOLoop
+from tornado.ioloop import IOLoop, PeriodicCallback
 
 from jupyter_client.multikernelmanager import MultiKernelManager
-from traitlets import Dict, List, Unicode, TraitError, default, validate
+from traitlets import Dict, List, Unicode, TraitError, Integer, default, validate
 
 from notebook.utils import to_os_path
 from notebook._tz import utcnow, isoformat
 from ipython_genutils.py3compat import getcwd
+
+from datetime import datetime, timedelta
 
 
 class MappingKernelManager(MultiKernelManager):
@@ -33,6 +35,10 @@ class MappingKernelManager(MultiKernelManager):
     root_dir = Unicode(config=True)
     
     _kernel_connections = Dict()
+
+    _culler_callback = None
+
+    _initialized_culler = False
 
     @default('root_dir')
     def _default_root_dir(self):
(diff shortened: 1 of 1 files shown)


notebook/services/kernels/kernelmanager.py
```diff
@@ -52,6 +58,26 @@ def _update_root_dir(self, proposal):
             raise TraitError("kernel root dir %r is not a directory" % value)
         return value
 
+    cull_kernels_after_minutes_env = 'CULL_KERNELS_AFTER_MINUTES'
```
parente: Providing a way to set traitlets with environment variables is done in the kernel gateway code base, but it's not common in the notebook codebase.

notebook/services/kernels/kernelmanager.py
```diff
@@ -225,3 +256,42 @@ def record_activity(msg_list):
...
+            self.cull_kernel(kId, kernel)
+
+    def cull_kernel(self, kId, kernel):
+        activity = kernel.last_activity
+        name = kernel.kernel_name
+        self.log.debug("kId=%s, name=%s, last_activity=%s", kId, name, activity)
+        if activity is not None:
+            dtNow = utcnow()
+            #dtActivity = datetime.strptime(activity,'%Y-%m-%dT%H:%M:%S.%f')
+            dtIdle = dtNow - activity
```
minrk: In Python, we try to use `_` to separate words, such as `kernel_id`, `idle_time` instead of `camelCase`

notebook/services/kernels/kernelmanager.py
```diff
@@ -52,6 +58,26 @@ def _update_root_dir(self, proposal):
             raise TraitError("kernel root dir %r is not a directory" % value)
         return value
 
+    cull_kernels_after_minutes_env = 'CULL_KERNELS_AFTER_MINUTES'
+    cull_kernels_after_minutes_default = 0
```
kevin-bates: Thank you - this has been addressed in the subsequent commit.

notebook/services/kernels/kernelmanager.py
```diff
@@ -225,3 +256,42 @@ def record_activity(msg_list):
...
+    def initialize_culler(self):
+        """Start idle culler if 'cull_kernels_after_minutes' is greater than zero.
+
+        Regardless of that value, set flag that we've been here.
+        """
+        if not self._initialized_culler and self.cull_kernels_after_minutes > 0:
+            if self._culler_callback is None:
+                loop = IOLoop.current()
+                if self.kernel_culling_interval_seconds <= 0: #handle case where user set invalid value
+                    self.log.warn("Invalid value for 'kernel_culling_interval_seconds' detected (%s) - using default value (%s).",
```
kevin-bates: +1

notebook/services/kernels/kernelmanager.py
```diff
@@ -225,3 +256,42 @@ def record_activity(msg_list):
...
+
+        self._initialized_culler = True
+
+    def cull_kernels(self):
+        self.log.debug("Polling every %s seconds for kernels idle > %s minutes...",
+            self.kernel_culling_interval_seconds, self.cull_kernels_after_minutes)
+        for kId, kernel in self._kernels.items():
+            self.cull_kernel(kId, kernel)
+
+    def cull_kernel(self, kId, kernel):
```
kevin-bates: Good point - thanks.

notebook/services/kernels/kernelmanager.py
kevin-bates: I had noticed the convention previously but am finding this to be a very difficult habit to break - thanks for pointing it out.

notebook/services/kernels/kernelmanager.py
```diff
@@ -225,3 +245,40 @@ def record_activity(msg_list):
...
+
+        Regardless of that value, set flag that we've been here.
+        """
+        if not self._initialized_culler and self.cull_idle_timeout > 0:
+            if self._culler_callback is None:
+                loop = IOLoop.current()
+                if self.cull_interval <= 0: #handle case where user set invalid value
+                    self.log.warning("Invalid value for 'cull_interval' detected (%s) - using default value (%s).",
+                        self.cull_interval, self.cull_interval_default)
+                    self.cull_interval = self.cull_interval_default
```
Carreau: Should we also warn for cull_interval < 60s ? it's technically not wrong, but it's likely that user might have thoughs it was minutes. Even anything below 300s seem it can be a mistake.

notebook/services/kernels/kernelmanager.py
```diff
@@ -52,6 +58,15 @@ def _update_root_dir(self, proposal):
             raise TraitError("kernel root dir %r is not a directory" % value)
         return value
 
+    cull_idle_timeout = Integer(0, config=True,
+        help="""Timeout (in seconds) after which a kernel is considered idle and ready to be culled."""
```
Carreau: Nitpicking, I think a "Values of 0 or lower will disable culling" could be nice. But we can improve later in a subsequent PR.

notebook/services/kernels/kernelmanager.py
```diff
@@ -225,3 +245,40 @@ def record_activity(msg_list):
...
+                self.log.info("Culling kernels with idle durations > %s seconds at %s second intervals ...",
+                    self.cull_idle_timeout, self.cull_interval)
+                self._culler_callback.start()
+
+        self._initialized_culler = True
+
+    def cull_kernels(self):
+        self.log.debug("Polling every %s seconds for kernels idle > %s seconds...",
+            self.cull_interval, self.cull_idle_timeout)
+        for kernel_id in list(self._kernels):
```
Carreau: I'm guessing the `list()` here is to reify the things you want to try to cull and make sure you don't modify the list while culling. Is that right ? If so it seem like something easily broken. 

What do @minrk and @parente think about building the list of kernel to shutdown and shut them al down at once ? 
That should make the flow a little bit easier to follow (and also to change the predicate if necessary).

notebook/services/kernels/kernelmanager.py
kevin-bates: +1

notebook/services/kernels/kernelmanager.py
kevin-bates: Yeah, I agree.  I thought implementing a minimum timeout would be good (now that I've tested it :-)).  300 strikes me as good minimum.  If folks are making changes in these areas and need shorter timeouts for testing, they can temporarily adjust the constant. 

Positive values less than the minimum would result in a warning and auto-adjustment to the minimum (similar to what happens if the interval is <= 0 and culling is enabled).

So, to your comment above, the description would be...

> "Timeout (in seconds) after which a kernel is considered idle and ready to be culled. Values of 0 or lower disable culling. The minimum timeout is 300 seconds.  Positive values less than the minimum will be adjusted to the minimum."

(or is that too verbose?)

minrk: @Carreau a warning is a good idea.

@kevin-bates that sounds perfect. You might add `(5 minutes)` next to 300 seconds.

notebook/services/kernels/kernelmanager.py
minrk: Either way is fine. I don't think getting a list of dictionary keys is easily broken.

Would this be clearer?

```python
# identify the idle kernels to be culled
to_cull = [ kid for kid in self._kernels if self.should_cull(kid) ]
# cull them
for kernel_id in to_cull:
    self.shutdown_kernel(kernel_id)
```

parente: > What do @minrk and @parente think about building the list of kernel to shutdown and shut them al down at once ?

Sounds good to me as long as there's appropriate checks in place so that an problem getting the status of any one kernel (if that's possible) doesn't prevent cleanup of others that should be culled.

notebook/services/kernels/kernelmanager.py
kevin-bates: Will do.

notebook/services/kernels/kernelmanager.py
kevin-bates: I guess I'm not really following the benefit of obtaining the list of cullable kernels, then walking that list to perform the cull when we could just as easily encounter an issue determining the cullable kernels.  So, to address the point raised by @parente, what about something like the following?  (Is that the right way to handle an unknown set of exceptions in python?)
```python
        for kernel_id in list(self._kernels):
            try:
                self.cull_kernel_if_idle(kernel_id)
            except:
                self.log.error("The following error was encountered while checking the idle duration of kernel %s: %s",
                    kernel_id, sys.exc_info()[0])
```

minrk: That works, too. There are plenty of valid ways to do it. A few notes:

- a comment about copying the kernel id list so we don't modify the dict during iteration might help clarity
- you can use `log.exception` to automatically log a traceback, so you don't need to use `exc_info`.
- always use at least `except Exception:`, because `except:` catches things like KeyboardInterrupts that shouldn't be caught.

kevin-bates: Thank you @minrk - very helpful.  Here's what I've got...
```python
        """Create a separate list of kernels to avoid conflicting updates while iterating"""
        for kernel_id in list(self._kernels):
            try:
                self.cull_kernel_if_idle(kernel_id)
            except Exception as e:
                self.log.exception("The following exception was encountered while checking the idle duration of kernel %s: %s",
                    kernel_id, e)
```

minrk:
@kevin-bates excellent, thanks! We're in the process of finishing up the 5.0 release from the master branch, but this can be merged once that release has been pushed out.

minrk:
5.0 is out, long live 5.1!

parente:
After finally getting a test bed setup for this change, I realize that it does not take into account whether the kernel is busy or not, or whether there are connections to the kernel or not. I'm going to submit a follow-on PR that adds settings for whether these attributes should affect the culling decision or not.

bigquant:
when will 5.1 go release?

kevin-bates:
@bigquant - Sounds like its very soon per the latest Jupyter Weekly Summary (2017, week 29): https://groups.google.com/forum/#!topic/jupyter/NgR4XpUf66E
```
Notebook (Grant)

    Grant is closing out outstanding issues marked with 5.1 milestone
    Planning on releasing a 5.1 release candidate by EOW
```

takluyver:
Yup, we're hoping to put the release candidate out any day now.

birdstar:
This is exactly the feature I wanted！

PR #2231, opened 2017-02-26
Adjust HTML table style (GH2209)

### LINKED ISSUE #2209: New table style makes it impossible to view data (5.0 beta)
I know this has been raised and discussed before (https://github.com/jupyterlab/jupyterlab/issues/961, https://github.com/jupyterlab/jupyterlab/pull/910#issuecomment-248599489, and actually the max-width has been increased due to that discussion), but, with the risk of trolling, I would like to raise this again (and opening an issue here instead of at jupyterlab, as with the 5.0 beta it seems more urgent here). 
To be clear: I really like the new table style! My only concern is regarding the hard-coded max-width and hiding of data.

**Problem**: due to the hard-coded `max-width: 150px` combined with `white-space: nowrap`, actual data gets hidden without any alternative way to look at the actual data. 
For me, the notebook is for a large part about "interacting with my data". An essential aspect of this, just looking at what is in my dataframe, is now impossible once you have a bit longer cell contents.

And you very easily get to this limit of 150px, eg timestamps with subsecond digits (common in certain fields), a bit longer species names, review comments, ...

**Possible solution**: My preferred solution would be to rework this part of the table styling to *always display all provided data* (and there are some different options to obtain this, like no max-width, or a max-width but with text wrapping over multiple lines, ...) 
Other way of putting it is this principle: let the library that generates the table control what data is visible, and let jupyter just display it 'unopinionated'.

As I said, the above is my preferred solution because this would let pandas (and other libraries generating tables) be able to control what is actually displayed (as we already currently do with `pd.options.display.max_colwidth`). 
But of course, there can also be other solutions like:

- make this easily configurable from jupyter
- display full data on hover
- ..

To be clear: it doesn't *has* to be my provided solution, but at least I hope you are open to reconsidering this aspect of the table styling (irregardless of the outcome of the discussion in the end :-)).

---

Note: we can override this from the pandas side, by including a style tag with every generated table. Relevant issue is here: https://github.com/pandas-dev/pandas/issues/15379 (and we would consider a bug-fix 0.19.3 release if necessary).
I will put some illustration of the interaction of pandas with the new table style in a next comment.

jorisvandenbossche:
Closes #2209 

@ellisonbg I now set line-height, white-space and max-width to new values. But I think, I could also remove those lines since they are the default values anyway.

### DIFF
--- a/notebook/static/notebook/less/renderedhtml.less
+++ b/notebook/static/notebook/less/renderedhtml.less
@@ -74,11 +74,9 @@
         text-align: right;
         vertical-align: middle;
         padding: 0.5em 0.5em;
-        line-height: 1.0;
-        white-space: nowrap;
-        max-width: 150px;
-        text-overflow: ellipsis;
-        overflow: hidden;
+        line-height: normal;
+        white-space: normal;
+        max-width: none;
         border: none;
     }
     th {


jorisvandenbossche:
Regarding line-height / padding. 

Current master:

screenshot from 2017-02-26 12-25-29

With this PR, it set line-height from 1.0 to 'normal', because otherwise multi-line cells looked weird (see the issue), but that also causes a normal table without wrapped cells look much more spacious (on my laptop: text is 20px height with 2x 6px padding -> 32 px):

screenshot from 2017-02-26 12-23-45

If I adjust padding from 0.5 to 0.3, it becomes more similar to master (same text height of 20px but with 2x 3.6px padding -> 27.2 px):

screenshot from 2017-02-26 12-38-57

On master it has a text line height of 12 px, with padding of 2x 6px, giving 24px in total.

ellisonbg:
Thanks! Merging...

jorisvandenbossche:
@ellisonbg Thanks for merging! 

I personally like the one above with a bit less padding (third figure) to have it more similar with master before this PR (first picture), but that is more a matter of personal taste, so leave that decision up to you :-) 
(and I think such small adjustments can always be made after in a next small release after using it a bit more in practice?)

takluyver:
Yep, I think it's totally fine to make adjustments in 5.1

PR #2233, opened 2017-02-27
Updated docs on installed phantomjs

louise-davies:
I had trouble trying to run the notebook tests on master and realised it was because I had v1.9.8 of PhantomJS installed and that when I updated to a later version it worked fine. Since they changed the npm package name from phantomjs to phantomjs-prebuilt that meant that the contributing doc was outdated. So a quick one-liner fix to help future contributors.


takluyver:
Thanks @louise-davies :cookie:

Issue #2245, opened 2017-03-02
function doc takes too long to appear - beta v


ggrrll:
In the beta version (`Version: 5.0.0b1`):

-  you have to type **4 times** shift + tab to have the doc opened (before it was only 2)
- the doc appears directly as a window at the bottom, instead of a window at the cursor position - the one with a yellowish background (not a big deal, even if  I prefer the one that appears 'in line')

takluyver:
The tooltip is still showing up for me, but not in the right place:

![screenshot from 2017-03-02 14-36-21](https://cloud.githubusercontent.com/assets/327925/23511621/a6252e96-ff55-11e6-8521-91fc61fdad66.png)

We should fix this for 5.0

ggrrll:
whats is the shortcut to get is ? (still `shift + tab`?)

takluyver:
Yep, still shift-tab

takluyver:
I think I've worked out what's causing this - when we display the tooltip, we get the location of the current cell so that the tooltip can float over it, but I think that location is coming relative to the wrong parent element.

ggrrll:
(ok, no rush... - I still get no tooltip at all)

takluyver:
Are you trying it while scrolled all the way to the top of a notebook? It appears at the top of the document for me, so if I'm scrolled down, I don't see it.

takluyver:
#2246 fixes it for me.

ggrrll:
ah, now I see it :) - should I upgrade `notebook` ?

takluyver:
#2246 isn't merged yet, but if you've got time to [install from source](http://jupyter-notebook.readthedocs.io/en/latest/contributing.html#setting-up-a-development-environment) to test it, it would be good to have someone else checking that it fixes the issue.

PR #2246, opened 2017-03-02
Add cell position when calculating location of cursor for tooltip

takluyver:
Closes gh-2245, I hope, but I don't know if there's a better way to fix it.

I think we need to calculate the location relative to the 'offset parent' of the tooltip. Adding one extra parent level from CodeMirror seems to work at the moment.


minrk:
Fix confirmed, thanks!

PR #2249, opened 2017-03-03
Test the cell tags toolbar

minrk:
- [x] test adding tags separated by comma and space, including duplicate items
- [x] test that adding tags affects metadata
- [x] test that clicking X removes tags from metadata
- [x] test that clicking X removes tags from the toolbar

### DIFF
--- a/notebook/static/notebook/js/celltoolbarpresets/tags.js
+++ b/notebook/static/notebook/js/celltoolbarpresets/tags.js
@@ -46,8 +46,8 @@ define([
     };
 
     var preprocess_input = function(input) {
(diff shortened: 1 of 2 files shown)


minrk:
@vidartf when I right-click, the `click()` callback is not called. Do you see this? It doesn't seem right that the 'X' should be the only element on the page with this behavior, in any case (the add tags button next to it doesn't have this).

vidartf:
@minrk There has been some changes recently: https://discourse.wicg.io/t/new-event-for-non-primary-button-click/1527, maybe this is not important anymore.

PR #2257, opened 2017-03-07
Tell user how to avoid token authentication.

Carreau:
Most user don't care how to get the token each time (if they see the
message). They want to avoid using a token.

Make it clearer that Password should be configured.

Be more tricky, say that password not configured **yet**, to make the
user think they __have to__.

The text is not technically correct as as far as I remember you can have
PW and Token at the same time.  But I'm going to say that in 99% of case
where users read that, the password is unset.

### DIFF
--- a/notebook/templates/login.html
+++ b/notebook/templates/login.html
@@ -45,16 +45,20 @@
     {% endif %}
     {% if token_available %}
     {% block token_message %}
-    <div class="col-sm-6 col-sm-offset-3 text-left">
-      <p class="warning">
-        Token authentication is enabled.
-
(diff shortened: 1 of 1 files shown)


takluyver:
Setting and then remembering a password isn't great UX either, unfortunately (it's easier right now, but it's either a small-cost-every-time, or irritation when you can't remember your password, especially if you run notebook servers on several machines).

Our SHA-hashed password in the config file is also pretty poor, security-wise.

Carreau:
we also do have [that](https://github.com/jupyter/notebook/blob/master/tools/secure_notebook.py) to make user life easier.

PR #2262, opened 2017-03-07
Add 'new_browser_window' config option

### LINKED ISSUE #2248: Feature Request: Open notebook in a new window instead of a new tab.
As the title says, i'd like to be able to open notebooks in a new window by default. I frequently have several browser windows opened on different workspaces, and the current behaviour will open
the notebook in whatever window I last used (I think). This turns out to be somewhere where I don't want it to be quite often.

A quick look at the code suggests that the following might do the trick (untested, though):

```diff
diff --git a/notebook/notebookapp.py b/notebook/notebookapp.py
index 498405a6d..38655c1c6 100755
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -678,6 +678,10 @@ class NotebookApp(JupyterApp):
                       standard library module, which allows setting of the
                       BROWSER environment variable to override it.
                       """)
+    new_browser_window = Bool(False, config=True,
+                              help="""Whether to open the notebook in a new 
+                              browser window. The default is to use a new tab.
+                              """)
     
     webapp_settings = Dict(config=True,
         help="DEPRECATED, use tornado_settings"
@@ -1387,7 +1391,7 @@ class NotebookApp(JupyterApp):
                 uri = url_concat(uri, {'token': self.one_time_token})
             if browser:
                 b = lambda : browser.open(url_path_join(self.connection_url, uri),
-                                          new=2)
+                                          new=1 if new_browser_window else 2)
                 threading.Thread(target=b).start()
 
         if self.token and self._token_generated:
```

However, I'm not familiar with the codebase. So if it is possible to merge this change, I'd very much appreciate if someone who is could polish the patch and open a pull request. Thanks!

wordofglass:
On startup, this optionally opens the notebook in a new browser window instead of a new tab.

Closes #2248

A quick test shows that this works at least for an up-to-date Firefox on Arch. Motivation, copied from the issue:
> As the title says, i'd like to be able to open notebooks in a new window by default. I frequently have several browser windows opened on different workspaces, and the current behaviour will open
the notebook in whatever window I last used (I think). This turns out to be somewhere where I don't want it to be quite often.

I grepped the repository for existing related config options, and found no additional mentions of them in the docs. Did I miss some place where this should be added?

Thanks for considering this request!

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -678,11 +678,18 @@ class NotebookApp(JupyterApp):
                       standard library module, which allows setting of the
                       BROWSER environment variable to override it.
                       """)
-    new_browser_window = Bool(False, config=True,
-                              help="""Whether to open the notebook in a new
-                              browser window. The default is to use a new tab.
-                              """)
-    
+
+    webbrowser_open_new = Integer(2, config=True,
+        help="""Specify Where to open the notebook on startup. This is the
+        `new` argument passed to the standard library method `webbrowser.open`.
+        The behaviour is not guaranteed, but depends on browser support. Valid
+        values are:
+            2 opens a new tab,
+            1 opens a new window,
+            0 opens in an existing window.
+        See the `webbrowser.open` documentation for details.
+        """)
+
     webapp_settings = Dict(config=True,
         help="DEPRECATED, use tornado_settings"
     )
@@ -1391,7 +1398,7 @@ class NotebookApp(JupyterApp):
                 uri = url_concat(uri, {'token': self.one_time_token})
             if browser:
                 b = lambda : browser.open(url_path_join(self.connection_url, uri),
-                                          new=1 if self.new_browser_window else 2)
+                                          new=self.webbrowser_open_new)
                 threading.Thread(target=b).start()
 
         if self.token and self._token_generated:
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -678,6 +678,10 @@ class NotebookApp(JupyterApp):
                       standard library module, which allows setting of the
                       BROWSER environment variable to override it.
                       """)
+    new_browser_window = Bool(False, config=True,
+                              help="""Whether to open the notebook in a new
+                              browser window. The default is to use a new tab.
+                              """)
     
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for the PR!

This seems to do the opposite for me (Safari). When new_browser_window=True, it always opens the *same* tab for repeated calls (which I like!), whereas with new_browser_window=False it opens a new tab. Safari might be blocking new window creation, hence the "if possible" bit in the docs. From reading [the docs](https://docs.python.org/3/library/webbrowser.html), this might be what I should expect, but the config name makes it surprising that enabling a `new_thing` flag causes it to open *fewer* new things.

Maybe we should expose the browser.new argument directly, and copy their docs, so pass `1` for window, `2` for tab, etc. e.g.

```python
webbrowser_open_new = Integer(2, config=True,
    help="""new argument passed to webbrowser.open.
    2 opens a new tab.
    1 opens a new window.
    0 opens in an existing window.

    See the webbrowser.open docs for details.
    """
```

wordofglass:
This appears sensible, I updated the branch accordingly.

I'm not sure, whether the `0` option is really relevant here, though. To my understanding, it only has an effect when using the same controller (obtained through `webbrowser.get()`) several times. Or is this what you meant with "repeated calls"?

minrk:
Thanks! We'll get this merged as soon as we push out 5.0 final.

EgZvor:
Just a note that this doesn't work in Windows at all due to the webbrowser.open behaviour.

ggrrll:
Hi,


thanks for the nice feature. 

Maybe I am doing something wrong, but

adding 
`c.NotebookApp.webbrowser_open_new = 1` 
in `jupyter_notebook_config.py` 

 doesn't seem to work for me (both on firefox and safari)

Thanks for the support.

(on notebook 5.6.0 / safari 11.1.2 and firefox  61.0.1)

gpano:
Hi,

unable to get this to work either.

* jupyter 4.4.0
* notebook 5.6.0

tried on: 
* safari 12.0.1
* chrome 70.0.3538.110

grayfall:
Same issue here.

PR #2279, opened 2017-03-12
Refactoring Page object

John-Pap:
I spotted some TODO comments for some hardcoded selectors in the page.js file, so I refactored the constructor of class Page to accept the two selectors as arguments. I hope I got it right.


Carreau:
That seem reasonable to me, but I'll let other more up-to-date with NB developpement pitch in. 
I'm tagging as 5.1 are we are close to 5.0 release and I don't want to delay 5.0 more. 

Thanks !

minrk:
Makes sense. Given that 100% of Page invocations use the same argument, we could probably also remove the TODO and accept that these aren't actually variable parameters.

PR #2280, opened 2017-03-13
Fix deprecated decodestring warning

adimitrova:
While running the nose tests a warning was thrown saying that 
`notebook/services/contents/largefilemanager.py:60: DeprecationWarning: decodestring() is a deprecated alias, use decodebytes()
 bcontent = base64.decodestring(b64_bytes)` 

As per the documentation of [base64.decodestring](https://docs.python.org/3/library/base64.html#base64.decodestring) this is deprecated since version 3.1 and the new function to use is decodebytes().

Thus, the changes were made in the source code in file:
`notebook/services/contents/largefilemanager.py`

### DIFF
--- a/notebook/services/contents/largefilemanager.py
+++ b/notebook/services/contents/largefilemanager.py
@@ -57,7 +57,7 @@ class LargeFileManager(FileContentsManager):
                 bcontent = content.encode('utf8')
             else:
                 b64_bytes = content.encode('ascii')
-                bcontent = base64.decodestring(b64_bytes)
+                bcontent = base64.decodebytes(b64_bytes)
         except Exception as e:
             raise web.HTTPError(
                 400, u'Encoding error saving %s: %s' % (os_path, e)
--- a/notebook/services/contents/largefilemanager.py
+++ b/notebook/services/contents/largefilemanager.py
@@ -57,7 +57,7 @@ class LargeFileManager(FileContentsManager):
                 bcontent = content.encode('utf8')
             else:
                 b64_bytes = content.encode('ascii')
-                bcontent = base64.decodebytes(b64_bytes)
+                bcontent = base64.b64decode(b64_bytes)
         except Exception as e:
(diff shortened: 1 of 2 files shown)


adimitrova:
Python 2 tests failed and I've updated the code and it's now compatible with both python 2 & 3.

Carreau:
Sweet ! 

Do you want to also send a patch like the following to make the use of decodestring an error ?

```diff
diff --git a/setup.cfg b/setup.cfg
index 622c74f59..8960ebeee 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,4 +8,5 @@ warningfilters=module   |.*            |DeprecationWarning |notebook.*
                ignore   |.*schema.*    |UserWarning        |nbfor.*
                ignore   |The 'warn' method is deprecated, use 'warning' instead     | DeprecationWarning |notebook.*
                error    |encodestring\(\) is a deprecated alias, use encodebytes\(\)| DeprecationWarning | notebook.*
+               error    |decodestring\(\) is a .*| DeprecationWarning | notebook.*
```

That will prevent future code modification to re-introduce it. You can add it here if you like to, or even make a separate PR later. 

I'm going to approve this as is, but as we are close to releasing 5.0 I'm going to wait for someone else to also validate. 

Thanks !

gnestor:
@Carreau I just pushed your change. Have one more look before I merge.

BTW, how would I apply your patch exactly? Copy/paste it to a `.patch` file and run `git apply file.patch`? I tried that and got `fatal: corrupt patch at line 10`.

Carreau:
> BTW, how would I apply your patch exactly? Copy/paste it to a .patch file and run git apply file.patch? I tried that and got fatal: corrupt patch at line 10.

Hum, unsure, I just gave an example, And did not even tested it. Plus I was on a custom fork, so the exact patch would likely fail because of commit range. Sorry about that.

PR #2281, opened 2017-03-13
Remove redundant sort from back-end

ajayaadhikari:
While I was looking for fixme's in the code I found the following. In the file [notebook/services/contents/handlers.py](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/handlers.py#L126-L129) a fixme comment mentions that the sorting of the contents of a directory should be done in the front-end. I checked the code on the front-end where this is supposed to be done. 

I found out that it is already done [there](https://github.com/jupyter/notebook/blob/420715e3e5685d9216f09c25d644b420d11fe109/notebook/static/tree/js/notebooklist.js#L372-L386). So the sorting is done redundantly on the front-end and the back-end. I removed the sorting from the back-end.

### DIFF
--- a/notebook/services/contents/handlers.py
+++ b/notebook/services/contents/handlers.py
@@ -18,17 +18,6 @@ from notebook.base.handlers import (
 )
 
 
-def sort_key(model):
-    """key function for case-insensitive sort by name and type"""
-    iname = model['name'].lower()
-    type_key = {
-        'directory' : '0',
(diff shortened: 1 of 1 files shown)


Carreau:
I'm +1 I think we should wait post 5.0 though, just not to introduce a potential change so close to a release.

minrk:
Thanks!

PR #2283, opened 2017-03-13
Add pandoc to travis

John-Pap:
While inspecting the code coverage on Codecov, I saw that some packages had suspiciously low coverage. I checked the Travis build logs and I realized that some tests are skipped because pandoc is missing. I added it as an apt addon to the Travis configuration file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,10 @@ env:
     - GROUP=js/services
     - GROUP=js/tree
 
+addons:
+    apt_packages:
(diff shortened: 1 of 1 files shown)


blink1073:
Great catch, @John-Pap!

michaelpacer:
The pandoc on apt is woefully out of date. It probably will not be able to handle most of our tests.

I'll make a change to what we use in nbconvert to handle this.

PR #2286, opened 2017-03-14
Make mime bundle accessible to renderers

gnestor:
This change simply pushes outputs to the mime bundle before displaying data so that renderers can access the entire mime bundle. This is useful for injecting additional outputs into a mime bundle (e.g. injecting a static image or static HTML so that notebooks render properly on Github). This functionality is available in JupyterLab and this PR provides feature parity for renderer extensions.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -324,16 +324,16 @@ define([
                 console.log("unrecognized output type: " + json.output_type);
                 this.append_unrecognized(json);
         }
+        
+        if (record_output) {
+            this.outputs.push(json);
+        }
 
         if (json.output_type === 'display_data') {
             var that = this;
             this.append_display_data(json, this.handle_appended);
         } else {
             this.handle_appended();
-        }
-        
-        if (record_output) {
-            this.outputs.push(json);
         }
     };
 
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -621,7 +621,9 @@ define([
             targets = this._display_id_targets[display_id] = [];
         }
         targets.push({
-            index: this.outputs.length,
+            // output has been appended when this is called,
+            // so use the index of the last item.
+            index: this.outputs.length - 1,
             element: element,
         });
     };
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -324,16 +324,16 @@ define([
                 console.log("unrecognized output type: " + json.output_type);
                 this.append_unrecognized(json);
         }
-        
-        if (record_output) {
-            this.outputs.push(json);
-        }
 
         if (json.output_type === 'display_data') {
             var that = this;
             this.append_display_data(json, this.handle_appended);
         } else {
             this.handle_appended();
+        }
+
+        if (record_output) {
+            this.outputs.push(json);
         }
     };
 
@@ -621,9 +621,7 @@ define([
             targets = this._display_id_targets[display_id] = [];
         }
         targets.push({
-            // output has been appended when this is called,
-            // so use the index of the last item.
-            index: this.outputs.length - 1,
+            index: this.outputs.length,
             element: element,
         });
     };
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -335,6 +335,11 @@ define([
         if (record_output) {
             this.outputs.push(json);
         }
+
+        this.events.trigger('output_added.OutputArea', {
+            output: json,
+            output_area: this,
+        });
     };
 
     OutputArea.prototype.handle_appended = function () {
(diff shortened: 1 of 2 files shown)


Carreau:
I don't see any issues from the top of my head.

gnestor:
It looks like we have some failed tests:

```
In /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
  Untitled suite in /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
    assertEquals: Subject equals the expected value
In /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
  Untitled suite in /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
    assertEquals: correct number of callback outputs
In /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
  Untitled suite in /home/travis/build/jupyter/notebook/notebook/tests/notebook/display_id.js
    assertEquals: correct number of cell outputs
```

@rgbkrk Any idea why this change is causing these `display_id` tests to fail?

rgbkrk:
Not sure, @minrk did the implementation for the jupyter notebook. nteract's is situated a bit differently.

rgbkrk:
Since silence is consent, I should probably speak up by saying that I'm not a big fan of passing the entire mimebundle through to a renderer. It seems like there's a mismatch between the way we declare the notebook format and the way it's being interpreted here. Can't quite put my finger on it though.

jasongrout:
> It seems like there's a mismatch between the way we declare the notebook format and the way it's being interpreted here. Can't quite put my finger on it though.

I can't quite put my finger on it either, but it has smelled a bit off to me too to isolate the mimetypes, have renderers say they can handle specific mimetypes, but then let the renderers change the state from the server however they wanted. I know the use-case of having the frontend render a static image and saving that in the notebook, for example, but it still feels a bit off.

gnestor:
Ok, interesting. @rgbkrk hinted at this the other day but I didn't quite understand the issue... 

Some high-level questions:

* What is the rationale for allowing an output to have multiple mime types?
* What is the rationale for isolating mime types?
* What are the dangers of allowing a renderer to access data/metadata of other mime types?

Some things to consider:

* Jupyterlab has supported mime bundle _injection_ for a while and @blink1073 recently refactored the rendermime interface to make output data and metadata observable and editable by a renderer: https://github.com/jupyterlab/jupyterlab/pull/1709
* According to @ellisonbg, including a static image in the mime bundle is a requirement for jupyterlab_vega (as ipyvega provided this and it's proven to be an important feature): https://github.com/altair-viz/jupyterlab_vega/issues/4

My thoughts:

A renderer is responsible for rendering output of a specific mime type in the notebook. The notebook is no longer a single environment but many (classic notebook, jupyterlab, nteract, nbviewer, Github's nbviewer, etc.). Ideally, a renderer will work across environments, but it doesn't. As a result, instead of seeing a nice plot on Github, you end up seeing `<altair.Vega object>`. How can this be improved? Well, all notebook environments can render `image/png` and `text/html`. The renderer on the client-side is well-suited to provide `image/png` and `text/html` representations of data that's already been renderered vs. a library of the kernel-side. Additionally, if the renderer is responsible for providing fallback mime types, then all kernels benefit from this feature, not just IPython. Now, it's not perfect (I'm seeing many edge cases where both static images and HTML fall short, especially on Github), but it solves a legitimate problem. 

In my ideal world, a user would not be required to install an extension for every mime type they would like to render because this adds a lot of complexity to a user's workflow and makes reproducibility less guaranteed. I did [a little experiment](https://github.com/gnestor/jupyterlab_react) that lazy loads remote renderers, meaning that as soon as an output is available, the renderer fetches the required package for that mime type from npm, dynamically imports it using SystemJS, and renders it using using that package (which is a React component). Now don't get me wrong, this raises **huge** security concerns, but it demonstrates how renderers could be more dynamic and less coupled with the environment.

minrk:
One important consideration for this PR as-is: it's giving different behavior to execute_result when compared with display_data. `display_data` will have access to the 'current' mime-bundle as `outputs[-1]`, whereas `execute_result` (and other output types) will not. `outputs.length` will have a different meaning for display_data output vs all other outputs, and the display_id target relies on this value having the same meaning. I've fixed the test above, but I know that display_id will now no longer work for execute_results. The test coverage doesn't check that case yet, though.

If renderers should have access to the whole mime-bundle, I think passing the whole mime-bundle to the renderer makes more sense than ensuring it is stored on the outputs list prior to rendering. Relying on the outputs list means async code can't use it, as other outputs can have been appended in the meantime.

Perhaps the simplest version of this is to add an `output_added.OutputArea` event with the output data, which could allow it to be modified.

minrk:
@gnestor I updated the PR to add the `output_added.OutputArea` event, which I think should satisfy the need for "do something with this output", instead of relying on the `.outputs` list.

blink1073:
Sounds good, @minrk.

gnestor:
Thanks, @minrk! I will test this out with the extensions that depend on this functionality now.

gnestor:
Ok, let's merge this and we can iterate if need be. Thanks again, @minrk!

PR #2293, opened 2017-03-15
only consider stream output for data rate limit

### LINKED ISSUE #2287: iopub rate limits are too low by default, for visualization-heavy projects
After upgrading to Jupyter version 5.0.0b1 on Mac OS X, I'm getting errors like:

```
IOPub data rate exceeded.
The notebook server will temporarily stop sending output
to the client in order to avoid crashing it.
To change this limit, set the config variable
`--NotebookApp.iopub_data_rate_limit`.
```

for just about every notebook I run, all of which have various visualizations.  From that message I had no idea what units the rate limits might be expressed in, but by adding enough zeros I eventually figured out that the problems go away if I launch jupyter with:

```
jupyter notebook --NotebookApp.iopub_data_rate_limit=10000000000
```

As mentioned in #1821, it is not obvious what limits are reasonable to set for this value by default, but I would like to put in my vote that the current value is at least an order of magnitude too low.  A sample matplotlib script that usually causes the problem is below, but I mostly use HoloViews, where I see the problem for just about every visualization when using the Bokeh backend, which requires a lot of JavaScript to be sent to the browser.  

As someone who in part maintains HoloViews, I now have to recommend that all of our users start Jupyter in this way, which makes it very awkward for them, for what I think is a very reasonable way to be using a Jupyter notebook for visualizations.  So I would strongly plead with you to increase this value by default so that it is only reached for actual error cases, not for ordinary visualization workflows.

```
import numpy as np
import matplotlib.pyplot as plt

freq = 100 
def sine(x, phase=0, freq=freq):
    return np.sin((freq * x + phase))

dist = np.linspace(-0.5,0.5,256)
x,y = np.meshgrid(dist, dist)
grid = (x**2+y**2)

testpattern = sine(grid, freq=freq)
methods = [None, 'none', 'nearest', 'bilinear', 'bicubic', 'spline16',
           'spline36', 'hanning', 'hamming', 'hermite', 'kaiser', 'quadric',
           'catrom', 'gaussian', 'bessel', 'mitchell', 'sinc', 'lanczos']

fig, axes = plt.subplots(3, 6, figsize=(96, 48))
fig.subplots_adjust(hspace=0.3, wspace=0.05)

for ax, interp_method in zip(axes.flat, methods):
    ax.imshow(testpattern , interpolation=interp_method, cmap='gray')
    ax.set_title(interp_method)

plt.show()
```

minrk:
large stream outputs cause much more problems than images or binary widget data, which seem to hit the rate limit more than anything.

this does open us up to large HTML and/or displayed text output potentially causing issues,
but most large outputs that have been hitting this limit do *not* cause issues (#2287).

The main throttle that's important is DOM elements to be added to the page. We can't feasibly distinguish between a massive million-element SVG and HTML with one big image in it.

This does explicitly remove the ability to do throttling based on *bandwith*. If people do want that, I can add the stream limit to a new config value. In that case, I would change the default all-output rate limit to no limit, so the same behavior as this PR, just with an extra knob.

closes #2287

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -997,7 +997,7 @@ class NotebookApp(JupyterApp):
         limited.""")
 
     iopub_data_rate_limit = Float(1000000, config=True, help="""(bytes/sec)
-        Maximum rate at which messages can be sent on iopub before they are
+        Maximum rate at which stream output can be sent on iopub before they are
         limited.""")
 
     rate_limit_window = Float(3, config=True, help="""(sec) Time window used to 
--- a/notebook/services/kernels/handlers.py
+++ b/notebook/services/kernels/handlers.py
@@ -360,7 +360,12 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
                     The notebook server will temporarily stop sending output
                     to the client in order to avoid crashing it.
                     To change this limit, set the config variable
-                    `--NotebookApp.iopub_msg_rate_limit`."""))
+                    `--NotebookApp.iopub_msg_rate_limit`.
+                    
+                    Current values:
+                    NotebookApp.iopub_msg_rate_limit={} (msgs/sec)
+                    NotebookApp.rate_limit_window={} (secs)
+                    """.format(self.iopub_msg_rate_limit, self.rate_limit_window)))
             else:
                 # resume once we've got some headroom below the limit
                 if self._iopub_msgs_exceeded and msg_rate < (0.8 * self.iopub_msg_rate_limit):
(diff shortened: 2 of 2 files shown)


jbednar:
This sounds like a good solution.  Thanks so much for the quick response!

rgbkrk:
It just occurred to me that this was slated for 5.1 and I merged it on to master. Is this ok or do I need to revert it?

takluyver:
I think we should probably revert it, or make a `5.0.x` branch from the previous commit to make the release from. We're really close to 5.0 now. @minrk what do you want to do?

minrk:
Reverting's probably easiest for now.

takluyver:
This was reverted in #2326, so it will need to be repeated for 5.1.

rgbkrk:
Thank you, my apologies. Glad this PR was made.

PR #2297, opened 2017-03-16
Trigger 'cleared' event when output area is cleared

gnestor:
This would be helpful to have for unmounting React-based renderers when outputs are cleared and prevent duplicate instances of the components.

Usage within a renderer:

```js
this.element.on('cleared', () => {
   ReactDOM.unmountComponentAtNode(toinsert[0]);
});
```

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -969,6 +969,7 @@ define([
 
             // Notify others of changes.
             this.element.trigger('changed');
+            this.element.trigger('cleared');
             
             this.outputs = [];
             this._display_id_targets = {};


takluyver:
This looks fine to me. Squeeze it in for 5.0?

There's some kind of error with installing npm on Travis - I don't think it's related to this, but it's stopping the tests from running.

jasongrout:
I thought I implemented a signal for this for clearing widgets. Let me check...

jasongrout:
Yeah: https://github.com/jupyter/notebook/pull/2173. Is that not sufficient?

jasongrout:
I suppose they are slightly different. My event is a notebook-wide event for components to know when a cell clears its output - useful for something outside of the output area (and something that possibly doesn't have a handle on the output area). This event is a per-outputarea event, so more suitable for something rendered directly in that output area. Another distinction is that a single cell may have more than one output area - my event is per cell, and this is per output area.

gnestor:
@jasongrout Thanks for this! I did some experimenting and I think your `'clear_output.CodeCell'` suffices. It's more consistent with Min's recently added `'output_added.OutputArea'` event and since it no longer depends on a specific `output_area.element`, I can take it out of the `append` function and avoid duplicating event handlers.

```js
// When an output is cleared,
output_area.events.on('clear_output.CodeCell', handleClearOutput);
// When an output is added to the output area
output_area.events.on('output_added.OutputArea', handleAddOutput);
```

I'm going to close this and release 5.0.0rc!

PR #2302, opened 2017-03-17
Enforce version number at build time.

Carreau:
Prevent making releases with wrong version number that will confuse pip,
or make sdists to appear more recent than wheels.

cc @gnestor

### DIFF
--- a/notebook/static/base/js/namespace.js
+++ b/notebook/static/base/js/namespace.js
@@ -73,7 +73,7 @@ define(function(){
     // tree
     jglobal('SessionList','tree/js/sessionlist');
 
-    Jupyter.version = "5.0.0-rc.1";
+    Jupyter.version = "5.0.0.dev";
     Jupyter._target = '_blank';
     return Jupyter;
 });
--- a/setupbase.py
+++ b/setupbase.py
@@ -14,6 +14,7 @@ This includes:
 from __future__ import print_function
 
 import os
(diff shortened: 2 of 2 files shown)


gnestor:
Thanks @Carreau! This is very helpful.

gnestor:
It works!

```
grant:notebook grant$ python setup.py sdist
  File "setup.py", line 51, in <module>
    from setupbase import (
  File "/Users/grant/Sites/jupyter/notebook/setupbase.py", line 76, in <module>
    raise ValueError('Invalid version number `%s`, please follow pep440 convention or pip will get confused about which package is more recent.' % version)
ValueError: Invalid version number `5.0.0-rc.1`, please follow pep440 convention or pip will get confused about which package is more recent.
```

Do you think the error message should include a link to the version scheme? https://www.python.org/dev/peps/pep-0440/#version-scheme

rgbkrk:
😄

minrk:
If we switch to always building with setuptools, this should also become unnecessary as it will do the check. I've started doing this in most new things, adding only the `bdist_egg_disabled`. I think we should stop supporting raw distutils install/sdist in general.

blink1073:
Cough, cough: https://github.com/jupyter/jupyter-packaging

Carreau:
> 😄

We know you are a semver guy :-)

Issue #2307, opened 2017-03-18
preact, preact-compat, proptypes as bower dependencies vs. npm


gnestor:
Follow up from https://github.com/jupyter/notebook/issues/2056: Users are seeing `error: [Errno 2] No such file or directory: '/PATH/TO/NOTEBOOK/node_modules/proptypes'` in 5.0.0rc1.

The notebook's only npm dependencies after being [unwebpackified](https://github.com/jupyter/notebook/pull/2011) is preact, preact-compat, and proptypes. Is there any reason why we can't get these dependencies from bower? There probably is but I just want to make sure. 

1. Install preact, etc. from npm: https://github.com/jupyter/notebook/blob/271c4b1f279901e06f522a22a2521eb9a9e6c20b/package.json#L22-L23
2. Copy npm packages from `node_modules` to `components` directory: https://github.com/jupyter/notebook/blob/master/setupbase.py#L383
3. Import preact, etc. using script tags: https://github.com/jupyter/notebook/blob/326077fa2d7addd646de2302217f8cff4e9c8d5d/notebook/templates/page.html#L19-L21
4. Use preact, etc. in ShortcutEditor: https://github.com/jupyter/notebook/blob/c305d8fda3b5f8bfb8c1ce520c0a6927f56261a1/notebook/static/notebook/js/shortcuteditor.js#L36-L169

cc @minrk @blink1073

blink1073:
I'd say use the bower versions.

minrk:
> Is there any reason why we can't get these dependencies from bower?

There aren't bower versions of these, as far as I can tell. That's the only reason they are dealt with as they are. If someone has a bower version, that works for me.

blink1073:
Ah, apologies, I had thought I had found preact on bower, it was just the misleading libraries.io site.

gnestor:
I submitted a PR that manages to do this via bower. Can you review?

PR #2319, opened 2017-03-21
Move preact, preact-compat, and proptypes to bower

gnestor:
Closes https://github.com/jupyter/notebook/issues/2307

Only preact is available via bower, but bower can also install packages via unpkg.org, which allows bower to manage preact-compat and proptypes.


rgbkrk:
Yay for unpkg.

minrk:
Cool, thanks @gnestor!

PR #2323, opened 2017-03-21
Explain how to use cell tags in changelog

takluyver:


takluyver:
Self-merging, because this is a pretty straightforward addition.

PR #2329, opened 2017-03-22
Add latex button to dropdown

mpacer:
In response to a question from https://github.com/jupyter/nbconvert/issues/552 about why we don't have a LaTeX dropdown button. 

Note: it doesn't return a zip file with the associated outputs, but neither does the markdown. Fixing that will need to be part of a larger project.

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -204,6 +204,10 @@ define([
         this.element.find('#download_pdf').click(function () {
             that._nbconvert('pdf', true);
         });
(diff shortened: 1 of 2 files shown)


takluyver:
Seems reasonable.

The zip file thing is working for me with download as markdown, at least. I think it may only zip it if there are any outputs that become separate files (images, usually).

mpacer:
@takluyver yep — I wasn't being careful when testing to make sure that it would have produced an output dir. Apologies for raising unnecessary alarms.

PR #2347, opened 2017-03-29
Fix typo setting write_config_dir in config manager

### LINKED ISSUE #2316: Traitlets 4.3.1 gives me: DeprecationWarning: object.__init__() takes no parameters
Traitlets 4.3.1 gives me:
```
DeprecationWarning: object.__init__() takes no parameters
  super(HasTraits, self).__init__(*super_args, **super_kwargs)
```
which seems to come from https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L1046-L1050 where a `ConfigManager` is created with a `config_dir ` argument, which is passed on to `LoggingConfigurable` and `Configurable`, neither of which support it.

minrk:
There is no `config_dir` attribute so it is ignored, triggering a warning in traitlets. Typos like this will eventually become errors (as they should be).

closes #2316

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1046,7 +1046,6 @@ class NotebookApp(JupyterApp):
         self.config_manager = self.config_manager_class(
             parent=self,
             log=self.log,
-            config_dir=os.path.join(self.config_dir, 'nbconfig'),
         )
 
     def init_logging(self):


minrk:
Hm, test is failing which is super weird. Need to investigate further.

gnestor:
@minrk It sounds like this should make it in 5.0.0?

minrk:
Sure. It's just fixing a deprecation message that's been there for a while, removing a line of code that didn't do anything.

PR #2349, opened 2017-03-29
include metadata when copying a cell

### LINKED ISSUE #2273: splitting a slide cell results in one non-slide cell and one slide cell
After using 'split cell' in the middle of a cell marked as 'slide', what I end up with is

* a first cell **not** marked as slide
* and a second cell *marked** as slide

I would expect it the other way around; but maybe that's just me ?

I'm using 'jupyter notebook' with jupyter-4.3.0  on macos

parmentelat:
Following up on #2273, I came up with this proposal.

I was not able to test this though, because I do not know how to try out my modified code in a running jupyter

Also I chose to copy the original cell's metadata using json, because I believe a copy is required in this context, but there might be better ways to do that copy

This is my first pull request ever and I am not too comfy in javascript either.. 

closes #2273

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -1697,11 +1697,16 @@ define([
         if (cell.is_splittable()) {
             var texta = cell.get_pre_cursor();
             var textb = cell.get_post_cursor();
+            // current cell becomes the second one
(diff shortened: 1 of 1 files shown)


minrk:
Thanks, this should work!

Issue #2352, opened 2017-03-30
finished_execute.CodeCell event triggered multiple times


juhasch:
If I execute this snippet with notebook *5.0.0rc2*, I get a log whenever a codecell is executed:
```
%%javascript

require(['base/js/events'], function(events) 
    {
    events.on('finished_iopub.Kernel', function(evt){ console.log('finished_iopub.Kernel', evt)})
    events.on('finished_execute.CodeCell', function(evt){ console.log('finished_execute.CodeCell', evt)})
    })
```
If I execute any codecell now, `finished_execute.CodeCell` is triggered an increasing number of times.
So first execution of any codecell, `finished_execute.CodeCell` is triggered once. Second execution twice, etc.

Any idea why this is happening ?

gnestor:
It looks like a new event listener is added on `CodeCell.execute` which would explain why you get _n_ number of events triggered for _n_ number of executes called. I think the simplest fix is to use `events.one` which will remove the event listener after the first event is handled.

gnestor:
FYI @dkrenn:

Using [`events.on`](http://api.jquery.com/on/) will create a new event listener every time a cell is executed. [`events.one`](http://api.jquery.com/one/) will also create a new event listener, but then dispose itself after the first event is handled.

gnestor:
@juhasch If you get a chance, can you checkout this PR and confirm the it fixes this issue:

```
git fetch jupyter pull/2354/head:pr/2354
git checkout pr/2354
pip install -e .
````

dkrenn:
@gnestor Ok, thank you (I didn't know).

juhasch:
Tried it, looks good 👍 
Thanks for the quick fix!

PR #2354, opened 2017-03-30
Use `events.one` vs `events.on` inside of `CodeCell.execute`

gnestor:
Patches https://github.com/jupyter/notebook/pull/2061

Closes https://github.com/jupyter/notebook/issues/2352

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -335,11 +335,13 @@ define([
         this.render();
         this.events.trigger('execute.CodeCell', {cell: this});
         var that = this;
-        this.events.on('finished_iopub.Kernel', function (evt, data) {
+        function handleFinished(evt, data) {
             if (that.kernel.id === data.kernel.id && that.last_msg_id === data.msg_id) {
-		that.events.trigger('finished_execute.CodeCell', {cell: that});
-	    }
-        });
+            		that.events.trigger('finished_execute.CodeCell', {cell: that});
+                that.events.off('finished_iopub.Kernel', handleFinished);
+      	    }
+        }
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/codecell.js
```diff
@@ -335,10 +335,10 @@ define([
         this.render();
         this.events.trigger('execute.CodeCell', {cell: this});
```
gnestor: @rgbkrk When a CodeCell is executed, `'execute.CodeCell'` is triggered, but apparently no event is triggered when execution completes.

notebook/static/notebook/js/codecell.js
```diff
@@ -335,10 +335,10 @@ define([
...
         var that = this;
-        this.events.on('finished_iopub.Kernel', function (evt, data) {
+        this.events.one('finished_iopub.Kernel', function (evt, data) {
             if (that.kernel.id === data.kernel.id && that.last_msg_id === data.msg_id) {
```
gnestor: When `'finished_iopub.Kernel'` is triggered and its data matches that of the CodeCell, then we know that execution is complete.

notebook/static/notebook/js/codecell.js
```diff
@@ -335,10 +335,10 @@ define([
         this.render();
         this.events.trigger('execute.CodeCell', {cell: this});
         var that = this;
-        this.events.on('finished_iopub.Kernel', function (evt, data) {
+        this.events.one('finished_iopub.Kernel', function (evt, data) {
```
gnestor: However, I just realized that if another cell's execution completes in between this cell's execution starting and completing, then this event listener will be disposed before it can handle execution completion...

PR #2361, opened 2017-04-03
Fix typo in cell tags toolbar screenshot

gnestor:


gnestor:
@takluyver Does this commit need to be in the release? It seems like RTD is using master, so I think that's all that matters.

takluyver:
Yeah, I think that's fine.

PR #2368, opened 2017-04-05
only consider stream outputs for data-rate limit

minrk:
reissue #2293


takluyver:
This was already looked at and merged before, so I'm re-merging it after a relatively quick read.

PR #2385, opened 2017-04-07
ensure a trailing slash on the base_url

rgbkrk:
For some reason the `/` wasn't coming on the terminals endpoint for v5, resulting in:

screen shot 2017-04-05 at 12 11 39 pm

I went ahead and ensured a trailing slash on the baseURL. If there's a better way to do this or a more core issue (python side insurance that basepath ends with a slash), I'm all ears.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -759,7 +759,7 @@ class NotebookApp(JupyterApp):
         value = proposal['value']
         if not value.startswith('/'):
             value = '/' + value
-        elif not value.endswith('/'):
+        if not value.endswith('/'):
             value = value + '/'
         return value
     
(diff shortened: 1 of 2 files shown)


rgbkrk:
Looking at the terminal template perhaps `ws_path` needs to be put through `urlencode`?

```
{% block params %}

data-base-url="{{base_url | urlencode}}"
data-ws-url="{{ws_url | urlencode}}"
data-ws-path="{{ws_path}}"

{% endblock %}
```

rgbkrk:
To reproduce this:

```
jupyter notebook --NotebookApp.base_url="somebody/toldme"
```

Versions:

```
traitlets==4.3.2
terminado==0.6
notebook==5.0.0
```

rgbkrk:
Actually, I'm noticing this behavior with the `custom/custom.js` paths as well. This is looking like a regression on `base_url`, since our docs for the option state:

> Leading and trailing slashes can be omitted, and will automatically be added.

minrk:
@rgbkrk thanks! I see the origin of the bug: adding the trailing slash was in an `elif` rather than `if`, so if the base_url config was missing both leading and trailing slash, only the leading one would be added. Pushed a fix to your branch.

rgbkrk:
Thanks for the patch!

PR #2388, opened 2017-04-08
NbserverStopApp: stop notebooks through cli - jupyter notebook stop <…

brookisme:
- usage: `jupyter notebook stop <PORT>`
- <PORT> defaults to 8888
- no tests (sorry)

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -359,7 +359,8 @@ class NbserverStopApp(JupyterApp):
             self.port=int(self.extra_args[0])
 
     def start(self):
-        server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
+        servers=list_running_servers(self.runtime_dir)
+        server=next((server for server in servers if server.get('port')==self.port),None)
         if server: os.kill(server.get('pid'), signal.SIGTERM)
         else:
             ports=[s.get('port') for s in list_running_servers(self.runtime_dir)]
@@ -369,7 +370,7 @@ class NbserverStopApp(JupyterApp):
                 for port in ports: print("\t* {}".format(port))
             else:
                 print("There are currently no running servers")
-                self.exit(1)
+            self.exit(1)
 
 
 class NbserverListApp(JupyterApp):
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -349,19 +349,14 @@ class NotebookPasswordApp(JupyterApp):
 class NbserverStopApp(JupyterApp):
     version = __version__
     description="Stop currently running notebook server for a given port"
-    kill_cmd='kill'
-    kill_signal='-3'
-
 
     port = Integer(8888, config=True,
         help="Port of the server to be killed. Default 8888")
-
 
     def parse_command_line(self, argv=None):
         super(NbserverStopApp, self).parse_command_line(argv)
         if self.extra_args:
             self.port=int(self.extra_args[0])
-
 
     def start(self):
         server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -16,7 +16,6 @@ import json
 import logging
 import mimetypes
 import os
-import subprocess
 import random
 import re
 import select
@@ -366,8 +365,7 @@ class NbserverStopApp(JupyterApp):
 
     def start(self):
         server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
-        if server:
-            subprocess.Popen([self.kill_cmd,self.kill_signal,str(server.get('pid'))],stdout=subprocess.PIPE).communicate()
+        if server: os.kill(str(server.get('pid')), signal.SIGQUIT)
         else:
             ports=[s.get('port') for s in list_running_servers(self.runtime_dir)]
             if ports:
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -353,14 +353,6 @@ class NbserverStopApp(JupyterApp):
     kill_cmd='kill'
     kill_signal='-3'
 
-    flags = dict(
-        json=({'NbserverStopApp': {'json': True}},
-              "Produce machine-readable JSON output."),
-    )
-
-    json = Bool(True, config=True,
-        help="If True, each line of output will be a JSON object with the "
-                  "details from the server info fpyile.")
 
     port = Integer(8888, config=True,
         help="Port of the server to be killed. Default 8888")
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -365,7 +365,7 @@ class NbserverStopApp(JupyterApp):
 
     def start(self):
         server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
-        if server: os.kill(str(server.get('pid')), signal.SIGQUIT)
+        if server: os.kill(server.get('pid'), signal.SIGQUIT)
         else:
             ports=[s.get('port') for s in list_running_servers(self.runtime_dir)]
             if ports:
@@ -375,6 +375,7 @@ class NbserverStopApp(JupyterApp):
             else:
                 print("There are currently no running servers")
                 self.exit(1)
+
 
 class NbserverListApp(JupyterApp):
     version = __version__
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -374,7 +374,7 @@ class NbserverStopApp(JupyterApp):
                 for port in ports: print("\t* {}".format(port))
             else:
                 print("There are currently no running servers")
-
+                self.exit(1)
 
 class NbserverListApp(JupyterApp):
     version = __version__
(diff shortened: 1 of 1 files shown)


gnestor:
Thanks @brookisme! I tested and it's working for me!

@minrk Can you review?

notebook/notebookapp.py
```diff
@@ -346,6 +347,45 @@ def start(self):
...
+    description="Stop currently running notebook server for a given port"
+    kill_cmd='kill'
+    kill_signal='-3'
+
+    flags = dict(
+        json=({'NbserverStopApp': {'json': True}},
+              "Produce machine-readable JSON output."),
+    )
+
+    json = Bool(True, config=True,
```
minrk: json is unused and can be removed

notebook/notebookapp.py
```diff
@@ -346,6 +347,45 @@ def start(self):
         self.log.info("Wrote hashed password to %s" % self.config_file)
 
 
+class NbserverStopApp(JupyterApp):
+    version = __version__
+    description="Stop currently running notebook server for a given port"
+    kill_cmd='kill'
+    kill_signal='-3'
```
minrk: signal should be an integer from the `signal` module, e.g.`signal.SIGTERM`

notebook/notebookapp.py
```diff
@@ -346,6 +347,45 @@ def start(self):
...
+    def parse_command_line(self, argv=None):
+        super(NbserverStopApp, self).parse_command_line(argv)
+        if self.extra_args:
+            self.port=int(self.extra_args[0])
+
+
+    def start(self):
+        server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
+        if server:
+            subprocess.Popen([self.kill_cmd,self.kill_signal,str(server.get('pid'))],stdout=subprocess.PIPE).communicate()
```
minrk: Rather than Popen(['kill', ...]), it would be better to use [os.kill](https://docs.python.org/3/library/os.html#os.kill) here.

notebook/notebookapp.py
```diff
@@ -346,6 +347,45 @@ def start(self):
...
 
+class NbserverStopApp(JupyterApp):
+    version = __version__
+    description="Stop currently running notebook server for a given port"
+    kill_cmd='kill'
+    kill_signal='-3'
+
+    flags = dict(
+        json=({'NbserverStopApp': {'json': True}},
+              "Produce machine-readable JSON output."),
```
minrk: flags here are also unused and can be removed

notebook/notebookapp.py
```diff
@@ -346,6 +347,45 @@ def start(self):
...
+
+
+    def start(self):
+        server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
+        if server:
+            subprocess.Popen([self.kill_cmd,self.kill_signal,str(server.get('pid'))],stdout=subprocess.PIPE).communicate()
+        else:
+            ports=[s.get('port') for s in list_running_servers(self.runtime_dir)]
+            if ports:
+                print("There is currently no server running on port {}.".format(self.port))
```
minrk: This print statement can be moved up one level, and only print ports below.

minrk: These print statements should also go on stderr, since they are error output.

notebook/notebookapp.py
```diff
@@ -346,6 +347,45 @@ def start(self):
...
+                print("Ports currently in use:")
+                for port in ports: print("\t* {}".format(port))
+            else:
+                print("There are currently no running servers")
```
minrk: This should also end with `self.exit(1)` when no server can be found to stop.

notebook/notebookapp.py
brookisme: using `signal.SIGQUIT` since `signal.SIGTERM` requests confirmation

notebook/notebookapp.py
brookisme: It we move this above `if ports:` the output for no running servers will be
```bash
There is currently no server running on port 8888
There are currently no running servers
```

Its seems the 

```bash
There are currently no running servers
```

might be better.  Thoughts?

brookisme:
@minrk  thanks for the feedback.  This should be complete apart from the minor _no servers running_ issue commented on above.

gnestor:
Should to docs be updated to reflect this new subcommand?

notebook/notebookapp.py
```diff
@@ -346,6 +346,32 @@ def start(self):
...
+        help="Port of the server to be killed. Default 8888")
+
+    def parse_command_line(self, argv=None):
+        super(NbserverStopApp, self).parse_command_line(argv)
+        if self.extra_args:
+            self.port=int(self.extra_args[0])
+
+    def start(self):
+        server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
+        if server: os.kill(server.get('pid'), signal.SIGQUIT)
```
minrk: I think SIGTERM is probably the right signal here, instead of SIGQUIT.

brookisme: So strange... I could swear my initial tests requested confirmation after sending sigterm (which is why I switched to sigquit) but I just double checked and sigterm worked fine.  Anyway its fixed now.

takluyver: Is it possible you tested with SIGINT initially? IIRC that's the one that asks for confirmation.

brookisme: probably something like that.  apologies for confusion though - glad it worked in the end :)

notebook/notebookapp.py
```diff
@@ -346,6 +346,32 @@ def start(self):
...
+
+    def start(self):
+        server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
+        if server: os.kill(server.get('pid'), signal.SIGTERM)
+        else:
+            ports=[s.get('port') for s in list_running_servers(self.runtime_dir)]
+            if ports:
+                print("There is currently no server running on port {}.".format(self.port))
+                print("Ports currently in use:")
+                for port in ports: print("\t* {}".format(port))
```
brookisme: fixed.

notebook/notebookapp.py
```diff
@@ -346,6 +346,32 @@ def start(self):
...
+    port = Integer(8888, config=True,
+        help="Port of the server to be killed. Default 8888")
+
+    def parse_command_line(self, argv=None):
+        super(NbserverStopApp, self).parse_command_line(argv)
+        if self.extra_args:
+            self.port=int(self.extra_args[0])
+
+    def start(self):
+        server=next((server for server in list_running_servers(self.runtime_dir) if server.get('port')==self.port),None)
```
brookisme: i pulled `list_running_servers` out of the list comprehension so its a little bit shorter but i'm not sure if you'd like more.

notebook/notebookapp.py
```diff
@@ -346,6 +346,33 @@ def start(self):
...
+        super(NbserverStopApp, self).parse_command_line(argv)
+        if self.extra_args:
+            self.port=int(self.extra_args[0])
+
+    def start(self):
+        servers=list_running_servers(self.runtime_dir)
+        server=next((server for server in servers if server.get('port')==self.port),None)
+        if server: os.kill(server.get('pid'), signal.SIGTERM)
+        else:
+            ports=[s.get('port') for s in list_running_servers(self.runtime_dir)]
```
takluyver: You can now reuse the `servers` variable here.

brookisme: ha - yeah, that was the very first thing i did, back before my initial commit, but `list_running_servers` returns an iterator so it doesn't work...

```bash
In [1]: from notebook.notebookapp import *

In [2]: nbstop=NbserverStopApp()

In [3]: servers=list_running_servers(nbstop.runtime_dir)

In [4]: [s for s in servers]
Out[4]: 
[{u'base_url': u'/',
  u'hostname': u'localhost',
  u'notebook_dir': u'/Users/brook/code/jupyter/notebook',
  u'password': False,
  u'pid': 27105,
  u'port': 8888,
  u'secure': False,
  u'token': u'48019c94f11b76fa4690c1a7d431bf58e4b2fb2d0cd65ae5',
  u'url': u'http://localhost:8888/'},
 {u'base_url': u'/',
  u'hostname': u'localhost',
  u'notebook_dir': u'/Users/brook/code/jupyter/notebook',
  u'password': False,
  u'pid': 27213,
  u'port': 8889,
  u'secure': False,
  u'token': u'4c1588c1139d214b9d1eb432fd000270e2645a226633fde5',
  u'url': u'http://localhost:8889/'},
 {u'base_url': u'/',
  u'hostname': u'localhost',
  u'notebook_dir': u'/Users/brook/code/jupyter/notebook',
  u'password': False,
  u'pid': 27238,
  u'port': 8890,
  u'secure': False,
  u'token': u'db40ac85ddd5694cc4f67d9c4a18f4c066b63887f6d9fb9b',
  u'url': u'http://localhost:8890/'}]

In [5]: [s for s in servers]
Out[5]: []
```

takluyver: Aha, OK then. Thanks, merging :-)

isabeaups:
starting from which version of jupyter is this feature available? 
When I try to use it I get the error

> NotebookApp] No such file or directory: /Users/default/Workspace/project/stop

takluyver:
It should be available from notebook 5.1.

PR #2394, opened 2017-04-10
Trigger 'clear_output.CodeCell' before clearing output

gnestor:
Follow up from https://github.com/jupyter/notebook/pull/2297

This will trigger the `clear_output_CodeCell` event before clearing the output area, allowing extensions to clean up before clearing the output element. This also triggers the event on `CodeCell.execute`.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -313,7 +313,7 @@ define([
             stop_on_error = true;
         }
 
-        this.output_area.clear_output(false, true);
+        this.clear_output(false, true);
         var old_msg_id = this.last_msg_id;
         if (old_msg_id) {
             this.kernel.clear_callbacks_for_msg(old_msg_id);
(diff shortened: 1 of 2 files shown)


gnestor:
@jasongrout Care to review?

jasongrout:
Can you look at #2411? That would supersede this. 

In #2411, I kept the event on the output area since in widgets, we use the output area outside of a cell. Having the event there lets us listen for events on our own output areas.

notebook/static/notebook/js/codecell.js
```diff
@@ -313,7 +313,7 @@ define([
             stop_on_error = true;
         }
 
-        this.output_area.clear_output(false, true);
+        this.clear_output(false, true);
```
gnestor: @jasongrout I think that `this.clear_output` should be used instead of `this.output_area.clear_output` because `this.clear_output` triggers events whereas `this.output_area.clear_output` doesn't.

gnestor: So I think this diff should make it into the notebook via the PR or https://github.com/jupyter/notebook/pull/2411

jasongrout: Agreed!

PR #2398, opened 2017-04-11
use dedicated Session for activity monitor

minrk:
instead of re-using kernel.session, which can cause hash collisions if other objects use `kernel.session`.

This fixes the duplicate digest errors seen in https://github.com/jupyter/kernel_gateway/pull/230

cc @parente


parente:
LGTM. Thanks @minrk for tracking this down (a second time!)

Issue #2397, opened 2017-04-11
Jupyter Notebook /files/ endpoint doesn't correctly send UTF-8 text files


bertjwregeer:
I have various log files that are generated by an outside process that are accessible by the Jupyter notebook. We would like users to be able to use these log files to verify that their actions are correct, however as soon as they click on them the file is downloaded as ASCII `text/plain`, which causes the fancy quotes  to be replaced with:

from:

`‘text’`

to

`â€˜textâ€™`

Is there some way to force Jupyter to correctly identify files ending with `.log` as UTF-8 plain text files?

takluyver:
There's nothing wrong with `text/plain` for UTF-8. It's the browser that chooses how to display it, and in that case it looks like it's using latin-1 or cp1252. We could do encoding detection and tell the browser to use a specific encoding with the HTTP Content-Type header, but it's basically just educated guessing, and I'm not sure it's worth adding dependencies for this.

bertjwregeer:
Sending `text/plain` according to the HTTP spec means the browser should treat it as `ISO-8859-1` due to the HTTP standard specifying that is the default encoding.

I would argue that sending `text/plain; charset=UTF-8` is better than `text/plain` for files Jupyter doesn't recognise, as `ISO-8859-1` documents will mostly render correctly as `UTF-8`, and browsers already fall back to `ISO-8859-1` for files that don't correctly decode as `UTF-8`.

----

Another option would be to check the `LANG` and `LC_ALL` environment variables and based upon the setting using that as the default `charset` returned in the `Content-Type` header.

gnestor:
@bertjwregeer I'm looking for where this change may need to happen. Does this look right?

```diff
diff --git a/notebook/files/handlers.py b/notebook/files/handlers.py
index c54151125..be3197ce7 100644
--- a/notebook/files/handlers.py
+++ b/notebook/files/handlers.py
@@ -52,7 +52,7 @@ class FilesHandler(IPythonHandler):
                if model['format'] == 'base64':
                    self.set_header('Content-Type', 'application/octet-stream')
                else:
                    self.set_header('Content-Type', [-'text/plain')-]{+'text/plain; charset=UTF-8')+}

        if include_body:
            if model['format'] == 'base64':
```

bertjwregeer:
The diff format you used is not one I recognise, but yes, changing that header should work to change the default.

gnestor:
[...]

PR #2402, opened 2017-04-12
Specify `charset=UTF-8` when serving non-base64 files

gnestor:
Closes https://github.com/jupyter/notebook/issues/2397

### DIFF
--- a/notebook/files/handlers.py
+++ b/notebook/files/handlers.py
@@ -46,7 +46,9 @@ class FilesHandler(IPythonHandler):
             self.set_header('Content-Type', 'application/x-ipynb+json')
         else:
             cur_mime = mimetypes.guess_type(name)[0]
-            if cur_mime is not None:
+            if cur_mime == 'text/plain':
+                self.set_header('Content-Type', 'text/plain; charset=UTF-8')
+            elif cur_mime is not None:
                 self.set_header('Content-Type', cur_mime)
             else:
                 if model['format'] == 'base64':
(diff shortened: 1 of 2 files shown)


takluyver:
This handles the fallback case where we don't find a mimetype for a file. Do we also want to do the same if `guess_type` returns `text/plain` as the mimetype of e.g. a `.txt` file?

bertjwregeer:
@takluyver does mimetype return the charset for a file? I wouldn't think it would, but for text/plain sending charset=utf-8 wouldn't be that bad of a guess in most cases.

When you have a .txt file and click on it in the browser, you open up an editor that correctly shows the contents as a UTF-8, it is only when you download the raw file at the moment, for example when you rename a file with UTF-8 characters within it to `somefile.log` for example, because that won't cause the editor to be opened when clicked.

gnestor:
@takluyver It sounds like a good idea to me. Does that look right?

takluyver:
Thanks, that looks good. There's a test failure, though, because we have a test that expects `text/plain`:

```
FAIL: make sure ContentsManager returns right files (ipynb, bin, txt).
  File "C:\projects\notebook\notebook\tests\test_files.py", line 98, in test_contents_manager
    self.assertEqual(r.headers['content-type'], 'text/plain')
AssertionError: 'text/plain; charset=UTF-8' != 'text/plain'
- text/plain; charset=UTF-8
+ text/plain
```

I don't think mimetype - or anything in the standard library - does encoding detection.

gnestor:
Ok, tests are passing! Let's wait to see if these resolves https://github.com/jupyter/notebook/issues/2397 before merging...

bertjwregeer:
This resolves the issue for me and log files are now downloaded correctly.

PR #2411, opened 2017-04-13
Pass the output area object in events specific to the output area.

jasongrout:
Also, add a new event triggered right before the output area is cleared. This is useful if an output renderer has cleanup work to do before the DOM element is removed off of the page.

@gnestor and I both want to add mimebundle renderers that should have a method called when the rendered value needs to be cleaned up.

Passing the output area in also makes it easy to listen for this event at the notebook level once and perform some action on whatever output area triggers the event.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -348,7 +348,7 @@ define([
             this._needs_height_reset = false;
         }
 
-        this.element.trigger('resizeOutput');
+        this.element.trigger('resizeOutput', {output_area: this});
     };
 
     OutputArea.prototype.create_output_area = function () {
@@ -451,7 +451,7 @@ define([
         }
 
         // Notify others of changes.
-        this.element.trigger('changed');
+        this.element.trigger('changed', {output_area: this});
     };
 
 
@@ -965,11 +965,12 @@ define([
             // Remove load event handlers from img tags because we don't want
             // them to fire if the image is never added to the page.
             this.element.find('img').off('load');
+            this.element.trigger('clearing', {output_area: this});
             this.element.html("");
 
             // Notify others of changes.
-            this.element.trigger('changed');
-            this.element.trigger('cleared');
+            this.element.trigger('changed', {output_area: this});
+            this.element.trigger('cleared', {output_area: this});
             
             this.outputs = [];
             this._display_id_targets = {};


jasongrout:
It would be even better if the cell could keep track of the rendered values, and take an optional callback for when that rendered value would be cleared. But we can hack around that with an event like this.

gnestor:
This looks like a good solution to me. Are there advantages to attaching event handlers to the output element vs. the output area?

jasongrout:
I'm not sure I understand your question. We trigger on the output element because that is what has the jquery event system because it is a jquery object. The bubbling then goes up through the DOM, so you can listen to all of these events by doing something like `notebook.container.on('clearing', '.output', myhandler);`.

gnestor:
> This looks like a good solution to me. Are there advantages to attaching event handlers to the output element vs. the output area?

What I mean is `this.element.on('cleared', handler)` vs. `this.events.on('clear_output.CodeCell', handler)` (which is actually the code cell, not output area). 

Since the output element already has `changed`, `cleared`, and `resizeOutput` events, adding a `clearing` event makes sense.

jasongrout:
1. We want to trigger for output areas, not code cells, since we might have output areas not associated with code cells (as in output widgets).

2. I can't find it now, but I thought at one point @minrk preferred having output area events on the output area, rather than global events. It doesn't matter to me either way - I'm essentially doing the same handler either way. @minrk, do you have a preference between events on the outputarea element instance vs. a global (but scoped) event?

jasongrout:
I also figured keeping the events on the output area instance was more backwards compatible....

notebook/static/notebook/js/codecell.js
```diff
@@ -509,10 +509,10 @@ define([
     };
 
 
-    CodeCell.prototype.clear_output = function (wait) {
-        this.output_area.clear_output(wait);
-        this.set_input_prompt();
+    CodeCell.prototype.clear_output = function (wait, ignore_queue) {
```
rgbkrk: Should `ignore_queue` have a default argument since prior calls to this would not have had an argument? Perhaps `undefined` is ok to pass through to `this.output_area.clear_output`?

notebook/static/notebook/js/outputarea.js
```diff
@@ -942,7 +942,7 @@ define([
     };
 
 
-    OutputArea.prototype.clear_output = function(wait, ignore_que) {
+    OutputArea.prototype.clear_output = function(wait, ignore_clear_queue) {
```
rgbkrk: Certainly better naming. :smile:

rgbkrk:
> do you have a preference between events on the outputarea element instance vs. a global (but scoped) event?

I prefer global events that are scoped too.

jasongrout:
> I prefer global events that are scoped too.

I prefer that too, all else being equal, but was avoiding making too many huge changes, especially backwards-incompatible ones.

notebook/static/notebook/js/outputarea.js
jasongrout: It took me a while of staring at it to figure out the double negative logic and what it was supposed to say :).

notebook/static/notebook/js/codecell.js
jasongrout: Good call. I'll check.

jasongrout: It's fine the way it is. `undefined` is treated as `false`, which means things should proceed as normal in the `clear_output` function.

PR #2418, opened 2017-04-18
editor: remember user language choices per file extension

minrk:
When the language is selected manually, record the file extension and choice in config so that future files opened with the same extension use the same mode.

This allows users to teach the editor about file extensions CodeMirror doesn't know about.

cc @mebersole

### DIFF
--- a/notebook/static/edit/js/editor.js
+++ b/notebook/static/edit/js/editor.js
@@ -32,6 +32,8 @@ function(
         this.base_url = options.base_url;
         this.file_path = options.file_path;
         this.config = options.config;
+        this.file_extension_modes = options.file_extension_modes || {};
+
         this.codemirror = new CodeMirror($(this.selector)[0]);
         this.codemirror.on('changes', function(cm, changes){
             that._clean_state();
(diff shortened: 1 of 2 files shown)


takluyver:
Looks OK, but there are a bunch of test failures on Travis. I can't see what caused them at the moment, because I'm on free wifi which blocks requests to Amazon S3, where Travis stores logs.

takluyver:
I see a later PR should have fixed that. Reopening to test again.

mebersole:
@minrk Very cool, thanks! Can you tell me where the config file this modifies is stored normally?

minrk:
@mebersole it will be in `~/.jupyter/nbconfig/edit.json` by default.

notebook/static/edit/js/editor.js
```diff
@@ -54,6 +56,14 @@ function(
             );
             that._set_codemirror_options(cmopts);
             that.events.trigger('config_changed.Editor', {config: that.config});
+            if (cfg.file_extension_modes) {
+                // check for file extension in user preferences
+                var modename = cfg.file_extension_modes[that._get_file_extension()];
+                if (modename) {
+                    var modeinfo = CodeMirror.findModeByName(modename);
+                    that.set_codemirror_mode(modeinfo);
```
takluyver: Does this need to handle CodeMirror not finding the mode? That might be possible if you switch between two versions of the notebook, for instance.

minrk: Good catch. Done.

PR #2419, opened 2017-04-18
minor cleanup in 'jupyter notebook stop'

minrk:
simplify the logic a bit, fewer one-liners.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -366,6 +366,7 @@ class NbserverStopApp(JupyterApp):
             if server['port'] == self.port:
                 self.log.debug("Shutting down notebook server with PID: %i", server['pid'])
                 os.kill(server['pid'], signal.SIGTERM)
+                return
         else:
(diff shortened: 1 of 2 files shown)


notebook/notebookapp.py
```diff
@@ -359,17 +359,18 @@ def parse_command_line(self, argv=None):
...
-        servers=list_running_servers(self.runtime_dir)
-        server=next((server for server in servers if server.get('port')==self.port),None)
-        if server: os.kill(server.get('pid'), signal.SIGTERM)
+        servers = list(list_running_servers(self.runtime_dir))
+        if not servers:
+            self.exit("There are no running servers")
+        for server in servers:
+            if server['port'] == self.port:
+                self.log.debug("Shutting down notebook server with PID: %i", server['pid'])
+                os.kill(server['pid'], signal.SIGTERM)
```
takluyver: I think you need a `break` in this branch so that it doesn't take the `else` below.

minrk:
Thanks for checking. I'll add some tests for this before it's done.

PR #2420, opened 2017-04-18
accept jquery's ajax({url: '...'}) signature

minrk:
in `utils.ajax`.

to ease extensions updating from `$.ajax` to `utils.ajax`

cc @damianavila who ran into this

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -772,7 +772,12 @@ define([
     };
     
     var ajax = function (url, settings) {
-        // like $.ajax, but ensure Authorization header is set
+        // like $.ajax, but ensure XSRF or Authorization header is set
+        if (typeof url === "object") {
+            // called with single argument: $.ajax({url: '...'})
+            settings = url;
+            url = settings.url;
+        }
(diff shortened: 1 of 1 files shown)


notebook/static/base/js/utils.js
```diff
@@ -772,7 +772,12 @@ define([
     };
     
     var ajax = function (url, settings) {
-        // like $.ajax, but ensure Authorization header is set
+        // like $.ajax, but ensure XSRF or Authorization header is set
+        if (typeof url === "object") {
+            // called with single argument: $.ajax({url: '...'})
+            settings = url;
+            url = settings.url;
```
damianavila: Don't you think it is necessary to remove `settings.url` from `settings` after defining the `url` here? I guess it is not harmful but I think it would complete the "split" of the two objects.
In any case, just a minor comment, LGTM as is as well.

takluyver:
:+1: once you've considered Damian's comment.

notebook/static/base/js/utils.js
minrk: Makes sense, thanks.

PR #2422, opened 2017-04-18
pin setuptools to < 35

minrk:
setuptools 35 may be causing js tests to fail presumably due to a change in manifest files (?).

This PR is mostly to verify that setuptools 35 is actually the cause,
and serves as a temporary fix to avoid blocking other work, if so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
 before_install:
     - pip install --upgrade pip
     - pip install --upgrade setuptools wheel nose coverage codecov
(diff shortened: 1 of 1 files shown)


minrk:
Looks like a setuptools release was unrelated because pinning to 34 doesn't change anything. This is super weird.

minrk:
On track again! Looks like it was tornado 4.5 breaking our creaky websocket backport needed only for phantomjs 1.x #2424

PR #2424, opened 2017-04-18
stop using old websocket implementation in tests

minrk:
since we are using phantomjs 2, we don't need the old websocket implementation anymore, which is broken with tornado 4.5, anyway.

This appears to be the cause of the recent failures on Travis.


rgbkrk:
Excellent

fperez:
+1, thx!

PR #2427, opened 2017-04-19
demote allow-root message to warning from an error

minrk:
rather than refusing to run, warn about safety. There are plenty of cases, especially in containers, where this is a fine thing to do.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1395,8 +1395,7 @@ class NotebookApp(JupyterApp):
             except AttributeError:
                 uid = -1 # anything nonzero here, since we can't check UID assume non-root
             if uid == 0:
-                self.log.critical("Running as root is not recommended. Use --allow-root to bypass.")
-                self.exit(1)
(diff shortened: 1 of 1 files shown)


takluyver:
-0.5 to this, because I think a lot of people just won't read it. Making people add a flag to confirm that it's what they want is a much clearer signal, and it's not much of an imposition for the people who do want to do that - especially if the command is baked into a docker file or similar, not manually typed in each time.

minrk:
I don't think refusing to run because a user makes a certain choice is something we should be doing in the first place. I think it only increases frustration and doesn't solve any problems.

takluyver:
I seem to remember we added it after seeing users launching the notebook with 'sudo' to work around something that was easily resolved. I think it's a price worth paying to indicate to users who don't fully understand what they're doing that running the server as root is generally a bad idea.

minrk:
OK. I don't agree that it's a price worth paying, but I'll defer to the previous resolution.

PR #2430, opened 2017-04-19
fix some absolute URL paths

minrk:
URLs always need to be relative to base_url in the javascript, because they will 404 when run on a non-empty URL prefix (e.g. JupyterHub).

closes https://github.com/jupyterhub/jupyterhub/issues/1086

I also removed a couple of outdated and incorrect examples while finding other instances of mistaken absolute paths.

### DIFF
--- a/notebook/static/notebook/js/notificationarea.js
+++ b/notebook/static/notebook/js/notificationarea.js
@@ -250,7 +250,7 @@ define([
             link.id = 'favicon';
             link.type = 'image/x-icon';
             link.rel = 'shortcut icon';
-            link.href = src;
+            link.href = utils.url_path_join(utils.get_body_data('baseUrl'), src);
             if (oldLink) document.head.removeChild(oldLink);
             document.head.appendChild(link);
         };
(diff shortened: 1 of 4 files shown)


notebook/static/notebook/js/notificationarea.js
```diff
@@ -250,7 +250,7 @@ define([
             link.id = 'favicon';
             link.type = 'image/x-icon';
             link.rel = 'shortcut icon';
-            link.href = src;
+            link.href = utils.url_path_join(utils.get_body_data('baseUrl'), src);
```
takluyver: The paths passed into this function have a leading `/` - is this OK, or do they need to be changed as you've changed the other path below?

takluyver: @minrk ping! If you decide this isn't a problem, feel free to push the green button.

minrk: url_path_join strips intermediate `/`, so this is fine.

PR #2431, opened 2017-04-20
Redirect paste menu items to call actions

### LINKED ISSUE #2415: Paste menu items do not trigger paste actions
I'm building an extension that listens for actions in the notebook so I track how notebooks are used and evolve over time.

One of the actions I listen for is `paste-cell-below`. This action is triggered when the user hits the paste button in the toolbar, but not when they select the "Paste Cell Below" item in the Edit menu. I believe this is due to the `enable_paste` function on [line 1566 of Notebook.js](https://github.com/jupyter/notebook/blob/f5f2eadd923f689336345ecacfdfd2592685a8e0/notebook/static/notebook/js/notebook.js#L1566) which links the menu items directly to functions. I think this may need to set each of the onclicks to call the relevant action (which will in turn call the function) rather than call the paste functions directly.

acrule:
Closes #2415 regarding paste menu items calling the notebook's paste functions directly rather than the relevant actions.

My apologies for all the extraneous changes in notebook.js and notificationarea.js. It seems my text editor automatically removed trailing whitespace on a number of lines. The substantial changes in notebook.js are at line 1570.

### DIFF
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -283,6 +283,12 @@ define(function(require){
                 env.notebook.copy_cell();
             }
         },
+        'paste-cell-replace' : {
+            help: 'paste cells replace',
+            handler : function (env) {
+                env.notebook.paste_cell_replace();
+            }
+        },
         'paste-cell-above' : {
             help: 'paste cells above',
             help_index : 'eg',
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! Can we ask you to redo it avoiding the whitespace changes, though? They can cause merge conflicts and make things like git-blame less useful. It should be possible to configure your editor not to trim trailing whitespace automatically.

acrule:
Certainly. The changes should now be free of the trailing whitespace edits.

Thanks for bearing with me.

takluyver:
Thanks, this looks good to me. :-)

Issue #1129, opened 2016-02-22
Long title moves on save


jakirkham:
Sadly I have some notebooks with long titles to keep track of what data they are working on and what I am trying to do with them. If the filename is very long, then I sometimes see this minor UI annoyance on saves where the title moves. Depending on the title length, this behavior can become time dependent on the time of the last save. One thought might be to move the location of language icon so the title can't run into its space. Though there are plenty of other reasonable ideas. Sorry to trouble with such a minor issue.

![long_notebook_name](https://cloud.githubusercontent.com/assets/3019665/13226424/538ae47c-d95f-11e5-8a2c-4fde879d7a64.gif)

JamiesHQ:
cc: @gnestor

gnestor:
I just submitted a PR that should fix this: https://github.com/jupyter/notebook/issues/2448

It enforces a `max-width: 100%` and uses `display: flex` on the save_widget container and uses a `text-overflow: ellipsis` to clip the title if it overflows.

PR #2448, opened 2017-04-27
Fix save widget layout for long notebook titles

gnestor:
Closes https://github.com/jupyter/notebook/issues/1129

This enforces a `max-width: 100%` and uses `display: flex` on the `.save_widget` container and uses a `text-overflow: ellipsis` to clip the title if it overflows.

Before:

After:

### DIFF
--- a/notebook/static/notebook/less/savewidget.less
+++ b/notebook/static/notebook/less/savewidget.less
@@ -1,5 +1,7 @@
 span.save_widget {
     margin-top: 6px;
+    max-width: 100%;
+    display: flex;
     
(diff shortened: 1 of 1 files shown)


takluyver:
This looks OK to me, but I'd like someone more familiar with CSS to take a look.

rgbkrk:
Looks phenomenal.

rgbkrk:
I retriggered the failed bits of the build, they pass now. ZeroMQ wasn't able to get the port it wanted, now it's a-ok.

Issue #2404, opened 2017-04-12
Sandboxed iframe in the ViewHandler not rendering PDFs


hsbarbosa:
Hey,
The new ViewHandler in v5.0 doesn't seem to be rendering pdfs either with Chrome or Safari.
Chrome 57.0.2987.133
Safari  10.1 (12603.1.30.0.34)

Actually it seems to be a security issue that was recently corrected but now they are not allowing sandboxed iframes to execute scripts. Apparently it's only when the pdf is loaded in a different tab. 
Removing the sandbox directive from the iframe makes it render properly (but the iframe is sandboxed for a reason, I guess).

takluyver:
The sandboxing is so that HTML files you view cannot talk to kernels the server is running - it should mean that cross-domain protection kicks in between the content in the iframe and the notebook server.

We do sandbox it [with the allow-scripts flag](https://github.com/jupyter/notebook/blob/d00b7e3e17274bd879be4c23012ce5f6fbce896a/notebook/templates/view.html#L26), but maybe browsers aren't respecting that?

We should only need the sandboxing for files that can run code in the browser, like html and svg (?), so you can workaround this for a PDF by changing `/view/` to `/files/` in the URL. But PDF.js is probably complex enough that I'd have some reservations about letting it out of the sandbox by default, in case a malicious PDF can cause it to run Javascript and then talk to kernels.

hsbarbosa:
That's what I thought.  
Many thanks! 
Cheers,

gnestor:
I just submitted a PR to address this: https://github.com/jupyter/notebook/issues/2449

PR #2449, opened 2017-04-27
Open all files with `/files` path except for `.html` and .svg`

gnestor:
Closes https://github.com/jupyter/notebook/issues/2404


gnestor:
@takluyver Can you review?

takluyver:
`.htm` should be treated this way as well, it's a common extension for html files. I wonder if we should be doing it on mime type instead of extension - I think anything that our server offers with mimetype `text/html` can have the same effect, and that can vary by system. In some cases, installed apps can affect what extensions are recognised for a mimetype.

@minrk @rgbkrk do you know of any file types other than html and svg that we should be displaying in the iframe?

minrk:
In general, it's best to avoid a 'not_safe' blacklist, because it's easy to miss things (like the many synonyms for html) instead having a 'safe' list for things that can be trusted to be displayed raw. It seems like it makes more sense to specify a list of things that *don't* work in the iframe (are there examples other than PDF?)

gnestor:
I updated so that `is_viewable` is used in place of `is_not_safe` (just includes `htm` and `html`). Every other file extensions that isn't editable will be opened using the `/files` path and handled by the browser. I also updated the `is_editable` list to be more comprehensive (using the list from https://github.com/sindresorhus/text-extensions/blob/master/text-extensions.json).

takluyver:
Isn't that the same blacklisting approach just with the name 'viewable' in place of 'not_safe'?

I see `xht`, `xhtml`, `mht` and `mhtml` in the editable list - in certain circumstances, any of these could be rendered as HTML in a browser, so viewing them should open them in the iframe. This rather validates Min's point about the problems with a blacklist.

gnestor:
I have updated this PR. The way it works now:

* By default, all files (not directories) are opened using `/files` prefix and display the "View" button
* If a file is html (according to the list of html extensions and `text/html` mime type), it's opened using the `/view` prefix
* If a file is editable (according to the list of editable extensions and `text/*` and `application/*` mime types), it's opened using the `/edit` prefix and displays the "Edit" button
* If a file is a notebook, it's opened using the `/notebooks` prefix
  * Notebooks can be edited in the text editor using the "Edit" button

Issue #2468, opened 2017-05-03
Jupyter notebook not launching - WindowsError[2] : The System cannot find the file specified


vnnsrk:
I had installed Anaconda x64 python 2.7 recently, and tried to launch jupyter notebook from my command line. I get the below error.

```
C:\Users\srinath8n>jupyter notebook
Traceback (most recent call last):
  File "C:\Users\srinath8n\Anaconda2\Scripts\jupyter-notebook-script.py", line 3
, in <module>
    import notebook.notebookapp
  File "C:\Users\srinath8n\Anaconda2\lib\site-packages\notebook\notebookapp.py",
 line 59, in <module>
    from .base.handlers import Template404, RedirectWithParams
  File "C:\Users\srinath8n\Anaconda2\lib\site-packages\notebook\base\handlers.py
", line 41, in <module>
    sys_info = json.dumps(get_sys_info())
  File "C:\Users\srinath8n\Anaconda2\lib\site-packages\notebook\_sysinfo.py", li
ne 94, in get_sys_info
    return pkg_info(path)
  File "C:\Users\srinath8n\Anaconda2\lib\site-packages\notebook\_sysinfo.py", li
ne 76, in pkg_info
    src, hsh = pkg_commit_hash(pkg_path)
  File "C:\Users\srinath8n\Anaconda2\lib\site-packages\notebook\_sysinfo.py", li
ne 52, in pkg_commit_hash
    cwd=pkg_path, shell=True)
  File "C:\Users\srinath8n\Anaconda2\lib\subprocess.py", line 390, in __init__
    errread, errwrite)
  File "C:\Users\srinath8n\Anaconda2\lib\subprocess.py", line 640, in _execute_c
hild
    startupinfo)
WindowsError: [Error 2] The system cannot find the file specified
```

The path variables (COMSPEC and PATH) are correctly set, the kernel specs are correct and point to the python.exe in the Anaconda2 folder, but still it seems that it cannot find path to the file. Git is properly installed, and I tried to change the subprocess call to 'git.cmd', which did not work. I tried reinstalling anaconda and jupyter, and also removing all other python environments to prevent file name clashes. I tried many of the solutions in stack overflow, and in github, but none seems to solve the issue.

Any help is appreciated. Thanks.

takluyver:
Hopefully #2471 should fix this.

vnnsrk:
It did. Thanks a ton.

PR #2471, opened 2017-05-03
Make sys_info more robust

takluyver:
- Don't call it at import time
- Don't fail if `git` isn't available
- Don't use `shell=True` to invoke git

Closes gh-2468

### DIFF
--- a/notebook/_sysinfo.py
+++ b/notebook/_sysinfo.py
@@ -46,10 +46,10 @@ def pkg_commit_hash(pkg_path):
     while cur_path != par_path:
         cur_path = par_path
         if p.exists(p.join(cur_path, '.git')):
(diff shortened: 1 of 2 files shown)


minrk:
👍, though I have a vague recollection that `shell=True` was needed for it to work on Windows.

takluyver:
I dug into git history, and it appeared to have been `shell=True` from when it was originally written in IPython - we don't seem to have started with `shell=False` and then changed it in response to an issue. That doesn't guarantee that it works on Windows, and I also vaguely recall some issues with invoking other things this way, but I think it's worth a try.

sarathuniverse:
where to change this shell=true
I mean in which file i can change

takluyver:
This is a pull request that was already merged. You can see the files that it changed under the 'Files changed' tab at the top.

PR #2479, opened 2017-05-04
Adjust positioning of filename in save widget

takluyver:
I noticed that the filename was partly hidden on Firefox, and badly aligned vertically on both Firefox and Chromium).

Before (L Firefox, R Chromium):

screenshot from 2017-05-04 13-27-54

After:

screenshot from 2017-05-04 13-29-28

### DIFF
--- a/notebook/static/notebook/less/savewidget.less
+++ b/notebook/static/notebook/less/savewidget.less
@@ -6,7 +6,6 @@ span.save_widget {
     span.filename {
         height: 1em;
         line-height: 1em;
-        padding: 3px;
         margin-left: @padding-large-horizontal;
         border: none;
         font-size: 146.5%;


takluyver:
Ping @ellisonbg for design related PR.

minrk:
Weird, I'm not sure how this got in there. The baseline for the title and checkpoint (and logo) should match, which they appear to still not do:

4.2.3 (all correct):

screen shot 2017-05-04 at 15 03 13


4.4.1 (logo text changed size):

screen shot 2017-05-04 at 14 59 35

minrk:
Looks like the key change is setting `display: flex` on span.save_widget in #2448, which causes all of the alignments to go wrong. Removing just that line appears to fix all of the alignments (but also undoes the ellipsis-overflow in that PR). Ideally, we can achieve that goal another way without changing the `display` property.

cc @gnestor

ellisonbg:
I agree with Min that the baselines should be aligned, which doesn't quite appear to the be the case in the after shots (but it is closer).

gnestor:
@takluyver I just made a change to the original PR (https://github.com/jupyter/notebook/pull/2448) which a) uses `align-items` to set the baseline for text and b) uses a calculated width for `.save_widget` to that it always fits in the top row of the header vs. wrapping below it:

image

Care to check it out and let me know if it resolves the issue on Firefox?

```
git remote add gnestor https://github.com/gnestor/notebook.git
git fetch gnestor
git checkout gnestor/issue-1129
npm run build
```

takluyver:
That fixes the alignment. Descenders are still getting clipped, however. I see this in both Firefox and Chromium, though it's more severe in Firefox:

screenshot from 2017-05-08 11-45-02

(L Firefox, R Chromium, both zoomed in to make the issue clearer)

In Chromium, just the hook of the letter **y** is getting sliced.

Turning off `overflow: hidden` solves it in both browsers.

gnestor:
@takluyver Good catch! If you check out the same branch, it should be fixed now. I also clean up some of the layout (got rid of some `float`s).

takluyver:
Thanks @gnestor, that looks good. I take it you'll make a PR of that branch?

gnestor:
Closing in favor of https://github.com/jupyter/notebook/pull/2532

PR #2492, opened 2017-05-11
NotebookApp.websocket_compression_options config

### LINKED ISSUE #2490: Performance issue due to websocket compression
After benchmarking the transfer of large amounts of (binary) data ( https://github.com/jupyter-widgets/ipywidgets/issues/1339 ) I found out profiling the notebook server, that quite some time is spend on compressing the data in tornado. For 4MB of data (1 million 32bit floats), 0.7 second is spend on compression. 
The compression was enabled in this PR: https://github.com/jupyter/notebook/pull/1261
I think this should be reverted, since most use cases will involve running the notebook and the browser on the same machine, and then compression will only lead to a performance decrease.
I think compression should only be enabled for notebook servers that are expected to be only approached over the web.
cc @minrk @jasongrout @Carreau

minrk:
expose passthrough for WebSocketHandler.get_compression_options

adds `NotebookApp.websocket_compression_options` configurable as a passthrough for setting the return value for [WebSocketHandler.get_compression_options()](http://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler.get_compression_options) with a default of None (no compression).

This works, but marked WIP while waiting for a decision on default from #2490. Other options:

- default: 'auto' so that compression is off for localhost by default and on for remote connections (as identified via `remote_ip`)
- allow setting a callable instead of a scalar, so that config could enable custom logic (such as 'auto' based on remote_ip)

closes #2490

### DIFF
--- a/notebook/base/zmqhandlers.py
+++ b/notebook/base/zmqhandlers.py
@@ -296,5 +296,4 @@ class AuthenticatedZMQStreamHandler(ZMQStreamHandler, IPythonHandler):
         self.session = Session(config=self.config)
 
     def get_compression_options(self):
-        # use deflate compress websocket
-        return {}
+        return self.settings.get('websocket_compression_options', None)
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -93,7 +93,7 @@ from jupyter_client.kernelspec import KernelSpecManager, NoSuchKernel, NATIVE_KE
 from jupyter_client.session import Session
 from nbformat.sign import NotebookNotary
 from traitlets import (
-    Dict, Unicode, Integer, List, Bool, Bytes, Instance,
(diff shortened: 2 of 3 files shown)


blink1073:
Revert ccba3bd in favor of #2501?

jasongrout:
+1 to the default being no compression, like the current PR has.

maartenbreddels:
If this goes into a release, we can 'advertise' with a 100x data transfer rate improvement for ipywidgets.

takluyver:
@minrk I think there's a consensus in favour of what you've implemented here. Do you want to do anything else on this before we merge it?

minrk:
@takluyver no, sorry, I should have taken WIP off a while back.

maartenbreddels:
great 🎉

PR #2494, opened 2017-05-13
Fix test not to expect notebooks sorted

takluyver:
PR gh-2281 removed the sorting step from the server. This can result in non-deterministic test failures, because the test checked against a list. Comparing sets instead makes the order not matter.


takluyver:
Merging this to sort tests out.

PR #2498, opened 2017-05-17
Add cull_busy, cull_connected options

parente:
Additional logic to support culling or skipping kernels that are busy or that have connections.

Follow-on to #2215.

/cc @kevin-bates

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -76,7 +76,7 @@ class MappingKernelManager(MultiKernelManager):
         Only effective if cull_interval is not 0."""
     )
 
-    cull_busy = Bool(True, config=True,
+    cull_busy = Bool(False, config=True,
         help="""Whether to consider culling kernels which are busy.
         Only effective if cull_interval is not 0."""
     )
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -71,7 +71,7 @@ class MappingKernelManager(MultiKernelManager):
         help="""The interval (in seconds) on which to check for idle kernels exceeding the cull timeout value."""
     )
 
-    cull_connected = Bool(True, config=True,
+    cull_connected = Bool(False, config=True,
         help="""Whether to consider culling kernels which have one or more connections.
         Only effective if cull_idle_timeout is not 0."""
     )
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -73,12 +73,12 @@ class MappingKernelManager(MultiKernelManager):
 
     cull_connected = Bool(True, config=True,
         help="""Whether to consider culling kernels which have one or more connections.
-        Only effective if cull_interval is not 0."""
+        Only effective if cull_idle_timeout is not 0."""
     )
 
     cull_busy = Bool(False, config=True,
         help="""Whether to consider culling kernels which are busy.
-        Only effective if cull_interval is not 0."""
+        Only effective if cull_idle_timeout is not 0."""
     )
 
     #-------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rgbkrk:
What a nice self cleaning notebook server.

parente:
✨ 📓 ✨

kevin-bates:
@parente - thanks for adding this.  I'm pretty sure I understand the use case for the cull-busy flag (to allow a single cell's execution to exceed the idle period - correct?) but I guess I'm not following the cull-connected case.  I figured the kernel was always in a connected state.  Seems like this case would be covered by the prior culling feature - in conjunction with cull-busy.  Could you please clarify the use cases for these flags?  Just trying to minimize knobs here.  Thanks!

parente:
The connection count tracked by the API is actually how many *clients* are connected to the kernel via the notebook API. My thinking is that in some situations, you may not want to cull a kernel that has a browser tab open pointing at the associated notebook, even if the kernel is idle (e.g., kiosk mode). In others, you may want to cull regardless of whether there are connected clients or not (e.g., users at work forget they have tabs open). `cull_connected` is the knob to handle these two use cases. `cull_busy` handles the situation you described above.

At work, I plan to configure our JupyterHub instances to `cull_busy=False`, `cull_connected=True` such that users with long running jobs (e.g., Spark) won't lose work if they exceed the idle period, but browser tabs left open over the weekend won't keep kernels alive once they go idle.

kevin-bates:
Thank you for the great explanation.  In light of this, I'm wondering if `cull_busy=False` isn't a better default.  I understand that that would be contrary to the current functionality, but given culling isn't officially available it seems like a good time to make that kind of change.

parente:
I was thinking the same thing about the new `cull_busy` default and agree since this isn't even released yet, now's the time to change it. I'll send a new commit in a moment.

parente:
Thanks for the review @rgbkrk and @kevin-bates.

I'm not up to speed on how we're merging for 5.1 (or not) in the notebook repo. Label it 5.1 and wait for a merge?

takluyver:
@parente yes, that's pretty much it. And if it goes a few days without further review, remind people about it.

If you're confident enough in a change and you have merge permissions, you can also ask for more comments and announce that you'll merge it yourself in a couple of days unless people ask you to wait longer.

notebook/services/kernels/kernelmanager.py
```diff
@@ -71,6 +71,16 @@ def _update_root_dir(self, proposal):
         help="""The interval (in seconds) on which to check for idle kernels exceeding the cull timeout value."""
     )
 
+    cull_connected = Bool(True, config=True,
+        help="""Whether to consider culling kernels which have one or more connections.
+        Only effective if cull_interval is not 0."""
```
takluyver: Do you mean `cull_idle_timeout` in both of these messages? Looking at the code, setting `cull_interval` to 0 produces a warning and resets it to 300.

kevin-bates: Good catch - yes, the descriptions of each should reference `cull_idle_timeout`.

parente:
Thanks for the process heads-up. I've fixed the command line help. I'll leave the PR open for a couple days like you suggest and then merge it.

notebook/services/kernels/kernelmanager.py
```diff
@@ -71,6 +71,16 @@ def _update_root_dir(self, proposal):
         help="""The interval (in seconds) on which to check for idle kernels exceeding the cull timeout value."""
     )
 
+    cull_connected = Bool(True, config=True,
```
minrk: Maybe default to the most conservative False for both of these?

kevin-bates: @minrk I understand taking a conservative approach but I believe that a majority of administrators that enable culling in the first place (the feature is off by default) would want connected kernels culled - provided they've exceeded the idle duration and are not busy for that period.  That said, I suppose administrators will review the four culling parameters when enabling, but still feel `cull_connected = True` is the appropriate default - IMHO.

minrk:
Not executing code, but with active users still looking at it doesn't seem idle, to me. Killing a kernel for a currently active page seems pretty aggressive for a default.

kevin-bates:
Given that the idle period will typically be on the order of multiple hours I seriously doubt users will still be looking at the page having not done anything during that period.  However, since we don't default the idle time (only recommend) and given that admins may choose a timeout value in line with a human's window of patience, your example makes sense.
Thanks for the discussion.

PR #2500, opened 2017-05-17
Use function to construct empty message in command palette

takluyver:
This takes care of correctly escaping user input so it doesn't become HTML.

### DIFF
--- a/notebook/static/notebook/js/commandpalette.js
+++ b/notebook/static/notebook/js/commandpalette.js
@@ -165,8 +165,8 @@ define(function(require){
 
         input.typeahead({
           emptyTemplate: function(query) {
-            return $('<div>').text("No results found for").append(
-                $('<pre>').text(query)
+            return $('<div>').text("No results found for ").append(
+                $('<code>').text(query)
             );
           },
           maxItem: 1e3,
--- a/notebook/static/notebook/js/commandpalette.js
+++ b/notebook/static/notebook/js/commandpalette.js
@@ -164,7 +164,11 @@ define(function(require){
         // now src is the right structure for typeahead
 
         input.typeahead({
-          emptyTemplate: "No results found for <pre>{{query}}</pre>",
+          emptyTemplate: function(query) {
+            return $('<div>').text("Nao results found for").append(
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/commandpalette.js
```diff
@@ -164,7 +164,11 @@ define(function(require){
         // now src is the right structure for typeahead
 
         input.typeahead({
-          emptyTemplate: "No results found for <pre>{{query}}</pre>",
+          emptyTemplate: function(query) {
+            return $('<div>').text("Nao results found for").append(
```
ssanderson: Should this be "No results found for"?

yuvipanda: No, rather than Nao (although I'd love if we just start saying "nah", "nao", "NOPE", etc in place of 'No' throughout the code)

takluyver:
Yep, i changed it to check that I was using the updated code, and forgot to
change it back. My computer is off for the night now, so anyone with
access, feel free to fix that.

ssanderson:
:+1: from me

notebook/static/notebook/js/commandpalette.js
```diff
@@ -164,7 +164,11 @@ define(function(require){
         // now src is the right structure for typeahead
 
         input.typeahead({
-          emptyTemplate: "No results found for <pre>{{query}}</pre>",
+          emptyTemplate: function(query) {
+            return $('<div>').text("No results found for").append(
```
damianavila: Maybe you need to add an space after `for`?

takluyver:
At least on my system, the `<pre>` block was on a separate line, so it didn't need a space. However, I've adjusted it to use a `<code>` tag, which shows up on the same line:

screenshot from 2017-05-22 13-28-14

damianavila:
> However, I've adjusted it to use a <code> tag, which shows up on the same line:

Great!

damianavila:
Funny... I broke Github with the `<code>` above, it should be:

> However, I've adjusted it to use a `<code>` tag, which shows up on the same line:

>Great!

Issue #2502, opened 2017-05-18
Enhancement: Add line break after The Jupyter Notebook is running at:


GadgetSteve:
[I 13:47:40.278 NotebookApp] The Jupyter Notebook is running at: http://localhost:8888/?token=b5d714ae8bd101fd251053ddc55add86fafb8e0c1ddd690f
is displayed in the console on starting jupyter notebook and if the browser has not been set or cannot be found the user needs to copy and paste this into a browser.

The trouble is that the standard console or command window width is 80 characters while the message is 142 which, on many systems, results in the text wrapping which is not convenient for copy and paste. 

Splitting the message over 2 lines should enable the entire text that needs to be copied to appear on a single line.  If some other tools use the output format adding a backslash at the end of the first line should satisfy most if not all of them.

takluyver:
That sounds reasonable, do you want to make a PR?

GadgetSteve:
PR #2505 Added.

remram44:
This is fixed!

takluyver:
Thanks for the reminder!

PR #2505, opened 2017-05-18
Issue #2502: Add line break after The Jupyter Notebook is running at:

GadgetSteve:
Updated both notebook/notebookapp.py and docs/source/security.rst to add a newline before the URL - testing showed that the URL still did not fit in 80 columns because of the `[I 11:59:16.597 NotebookApp]` leader - changed to a `\n\r` to get the correct results. - Dropped the \ at the end idea as the logger output makes this pointless.

Tested that the clipboard gives the correct results. 

Minor change to currently untested feature so no test cases added. Existing tests still pass with one exception - `ERROR: Failure: ImportError (No module named fcntl)` which is as before for me.

Note: A number of lines in notebook/notebookapp.py were changed due to trailing spaces the only actual changes are on lines 1356-1357.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -187,7 +187,7 @@ class NotebookWebApplication(web.Application):
         jenv_opt.update(jinja_env_options if jinja_env_options else {})
 
         env = Environment(loader=FileSystemLoader(template_path), **jenv_opt)
-
+        
         sys_info = get_sys_info()
         if sys_info['commit_source'] == 'repository':
             # don't cache (rely on 304) when working from master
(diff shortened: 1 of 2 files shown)


notebook/notebookapp.py
```diff
@@ -226,7 +226,7 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
             },
             version_hash=version_hash,
             ignore_minified_js=jupyter_app.ignore_minified_js,
-
```
takluyver: Could we ask you to redo this with your editor set not to strip whitespace? Sorry for the hassle, but the changes can cause unnecessary merge conflicts, and make tools like `git blame` less useful.

Other than this, it looks great, thanks :-)

GadgetSteve:
Will do and I might raise a separate ticket for stripping whitespace.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

takluyver:
Thanks! Be aware that we don't generally go for 'cleanup' PRs.

takluyver:
Thanks :-)

PR #2532, opened 2017-05-30
Re: Adjust positioning of filename in save widget

gnestor:
Continuation of https://github.com/jupyter/notebook/pull/2479


takluyver:
Thanks!

PR #2534, opened 2017-05-30
Enforce right-alignment for cell in markdown tables

gnestor:
Fixes https://github.com/jupyterlab/jupyterlab/issues/2151

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -710,7 +710,6 @@ define([
           return start_tag + content + end_tag;
         };
         marked(text, { renderer: renderer }, function (err, html) {
-        // marked(text, function (err, html) {
             html = mathjaxutils.replace_math(html, math);
             toinsert.append(html);
         });
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -701,7 +701,16 @@ define([
         var text_and_math = mathjaxutils.remove_math(markdown);
         var text = text_and_math[0];
         var math = text_and_math[1];
-        marked(text, function (err, html) {
+        // Prevent marked from returning inline styles for table cells
+        var renderer = new marked.Renderer();
+        renderer.tablecell = function (content, flags) {
+          var type = flags.header ? 'th' : 'td';
+          var start_tag = '<' + type + '>';
+          var end_tag = '</' + type + '>\n';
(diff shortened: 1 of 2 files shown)


takluyver:
I'm wary of adding `!important` rules because they're hard for anyone else to override. Are we so confident in our opinion of how tables should look that we're willing to more-or-less force it on everyone?

cc @ellisonbg

gnestor:
@takluyver I suppose not because in https://github.com/jupyterlab/jupyterlab/issues/2151, @tdaff was actually looking to override the jupyterlab default styles for tables. We do want consistency across notebook and lab though and this would enforce that in notebook. I just don't understand why they're rendering differently given that they're both using marked... I suppose that it could be related to version inconsistency:

* lab: `"marked": "^0.3.6"`
* notebook: `"marked": "~0.3"`

takluyver:
I've moved this to 'backlog' so we can get on with 5.1, but left it open in case people want to work out where the inconsistency is coming from and what we can do about it.

gnestor:
Ya, I've explored the marked repo and there haven't been any changes to the markdown table style for 4 years, so the cause must be a discrepancy in the way marked is used in notebook vs. lab.

datasouvik:
Hi gentlemen,

Has this issue been resolved? Do we now get the table content as left-aligned or centrally aligned by default, instead of the right alignment?

Thanks,
Souvik

gnestor:
I have just updated this PR to prevent marked from adding inline style for table cells. As a result, there will be no need for `!important` and this will allow users to specify custom styles in custom.css.

gnestor:
@takluyver Care to review/merge?

notebook/static/notebook/js/outputarea.js
```diff
@@ -701,7 +701,16 @@ define([
...
+        // Prevent marked from returning inline styles for table cells
+        var renderer = new marked.Renderer();
+        renderer.tablecell = function (content, flags) {
+          var type = flags.header ? 'th' : 'td';
+          var start_tag = '<' + type + '>';
+          var end_tag = '</' + type + '>\n';
+          return start_tag + content + end_tag;
+        };
+        marked(text, { renderer: renderer }, function (err, html) {
+        // marked(text, function (err, html) {
```
takluyver: Do you want to clear away these commented out lines now that the new code is working?

gnestor:
@takluyver Thanks, merge away!

Issue #2379, opened 2017-04-06
cant type right bracket ']'


steboc:
i cant type right bracket 

Jupyter 5.01
Windows 7 ( French version ) and  windows 10 ( french)
Chrome 57.0.2987.133

edit : work with firefox

flothesof:
I have the same problem and incidentally the same setup: Windows 7 French, Notebook 5.0.0, Chrome 57.0.2987.133. Typing "]" inside a cell does not work.

gnestor:
Hmmm....

@Carreau @minrk

Carreau:
Can you remind me what's the key combinaison on French keyboard for ] ?
AltGR-0 ? Does it do nothing ? Or insert something else? Can you try on
codemirror.net demo?

steboc:
its AltGr + (key below f9 on the picture )
http://www.vallons.fr/cours-pc/Clavier_Azerty.jpg

KevinBaudin:
Got the same issue with Jupyter (Anaconda 2) since last update 2 days ago
on Google Chrome Version 57.0.2987.133 (64-bit)

French keyboard (and yes, it s indeed AltGr + ")", two keys left from backspace)

Doing ] with Alt+93 from now, but quite annoying.

Ask me anything if you want further informations to fix this

Carreau:
I've tried with my mac, setup with french keyboard and unfortunately I can't reproduce. So I'll have to find an actual windows machine and likely a Physical french keyboard layout (with altgr) and see what we can do.

How does it behave in codemirror directly ? 

http://codemirror.net/

KevinBaudin:
Nothing wrong on codemirror for me.
Still got the issue on Jupyter though.

gatsu:
Got the exact same problem, french azerty layout, win10, chrome. Using firefox "solve" the problem.

KevinBaudin:
Anything new? 
Any clue to fix this ?

Carreau:
> Anything new?
> Any clue to fix this ?

No, without a windows machine with the bug to actually track it down I can't narrow down the issue. 
@SylvainCorlay you are in France  maybe you have access to that ? I should be back in France this summer, I might have one machine available to do that. Unfortunately I stopped using French keyboard a couple of years ago and there is none available here.

SmirnovAnton:
[...]

PR #2535, opened 2017-05-30
Remove keymaps that conflict with non-English keyboards

gnestor:
Fixes https://github.com/jupyter/notebook/issues/2379

### DIFF
--- a/notebook/static/notebook/js/cell.js
+++ b/notebook/static/notebook/js/cell.js
@@ -135,8 +135,8 @@ define([
                 "Shift-Tab" : "indentLess",
                 // "Cmd-Alt-[" : "indentAuto",
                 // "Ctrl-Alt-[" : "indentAuto",
-                // "Cmd-/" : "toggleComment",
-                // "Ctrl-/" : "toggleComment",
+                "Cmd-/" : "toggleComment",
+                "Ctrl-/" : "toggleComment",
             }
         }
     };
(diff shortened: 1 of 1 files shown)


gnestor:
@minrk Care to review? This reverts some of https://github.com/jupyter/notebook/pull/963 which caused conflicts on non-English keyboards.

minrk:
Is it just the ctrl-alt shortcuts or also the cmd-alt shortcuts that cause this problem? I would certainly be sad to lose those.

gnestor:
@minrk This is only removing the <kbd>Cmd-Alt-[</kbd> shortcut, not all <kbd>Cmd-Alt-*</kbd>

minrk:
Sorry. My question was whether it was `ctrl-alt-[` or `cmd-alt-[` or both that cause the issue.

gnestor:
I inquired about that in the issue. Let's see what people say 👍 I think we just need to avoid using shortcuts that involve `[` and `]`, but I saw a one comment about `\` and @Carreau suggested that we also disable ones involving `/`.

Carreau:
I believe that's anything that touch [,], and `alt`.  Alt is heavily use on french keyboard to type punctuation. You just can't bind by default to anything that uses alt. 
That one of the main reason why I'm not an emacs user as most alt-based shortcut were conflicting with typing. 

The MacOS keyboard is slightly different, but same battle. Alt is a critical key.

gnestor:
Ok, so this PR just comments out the 2 commands that involve `alt` so I think we can merge this and iterate if necessary.

Carreau:
> Ok, so this PR just comments out the 2 commands that involve alt so I think we can merge this and iterate if necessary.

Yep, good point.

PR #2549, opened 2017-06-06
allow raises-exception to continue

### LINKED ISSUE #2515: Errors when importing pandas.
I have pandas installed in my default python site-packages folder. When importing pandas in Pycharm, everything goes smoothly: 
in: 
    `import pandas as pd`
out:
     
      `C:\Users\sx449_000\AppData\Local\Programs\Python\Python36-32\python.exe 
       C:/Users/sx449_000/PycharmProjects/untitled/ML.py

Process finished with exit code 0`

However, problem occurs when importing pandas in Jupyter Notebook:
In: `import numpy as np
import matplotlib as plt
import pandas as pd`
Out: 

     `---------------------------------------------------------------------------
     ModuleNotFoundError                       Traceback (most recent call last)
     C:\Users\sx449_000\AppData\Roaming\Python\Python36\site-packages\pandas\__init__.py in 
      <module>()
     25 try:
      ---> 26     from pandas._libs import (hashtable as _hashtable,
     27                              lib as _lib,

      C:\Users\sx449_000\AppData\Roaming\Python\Python36\site-packages\pandas\_libs\__init__.py in 
      <module>()
      2 
       ----> 3 from .tslib import iNaT, NaT, Timestamp, Timedelta, OutOfBoundsDatetime
      4 

      ModuleNotFoundError: No module named 'pandas._libs.tslib'

      ImportError                               Traceback (most recent call last)
      <ipython-input-8-79633c9e0bf8> in <module>()
      1 import numpy as np
      2 import matplotlib as plt
      ----> 3 import pandas as pd

     C:\Users\sx449_000\AppData\Roaming\Python\Python36\site-packages\pandas\__init__.py in 
      <module>()
     33                       "pandas from the source directory, you may need to run "
     34                       "'python setup.py build_ext --inplace --force' to build "
     ---> 35                       "the C extensions first.".format(module))
     36 
     37 from datetime import datetime

      ImportError: C extension: No module named 'pandas._libs.tslib' not built. If you want to import pandas 
        from the source directory, you may need to run 'python setup.py build_ext --inplace --force' to build 
     the C extensions first.`

I tried to uninstall then reinstall pandas through Pycharm Interpreter setting several times but the problem remains.

### LINKED ISSUE #2516: Use tag to set stop_on_error for CodeCell (Enhancement)
I use Jupyter notebook w/ Python a lot for writing documentation (which I later use nbconvert on).  One of the problems that arises is that I want to run all cells to compare output to previous output, but I also like to demonstrate "what *not* to do", which often raises errors.  These errors mean that Run All does not fully execute the code.

I think that if a standardized way were defined that marked a code cell as expected to error, it would be a big help to me.  I think that Jupyter 5.0's tags could do so.

For instance, in js/codecell.js we find this method fragment:

```
    CodeCell.prototype.execute = function (stop_on_error) {
        if (!this.kernel) {
            console.log("Can't execute cell since kernel is not set.");
            return;
        }

        if (stop_on_error === undefined) {
            stop_on_error = true;
        }
```

I believe that a fix might be as simple as something like:

```
        if (stop_on_error === undefined) {
            if (cell.metadata.tags !== undefined && cell.metadata.tags.indexOf('expected-failure') !== -1) {
                stop_on_error = false;
            } else {
                stop_on_error = true;
            }
        }
```

Would this be of interest and on the right track?  If so I'm happy to make a pull request.

mscuthbert:
Fixes #2516  — allows for code execution to continue if a tag “raises-exception” or “nbval-raises-exception” is encountered

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -310,7 +310,18 @@ define([
         }
 
         if (stop_on_error === undefined) {
-            stop_on_error = true;
+            if (this.metadata !== undefined && 
+                    this.metadata.tags !== undefined) {
+                if (this.metadata.tags.indexOf('raises-exception') !== -1) {
+                    stop_on_error = false;
+                } else if (this.metadata.tags.indexOf('nbval-raises-exception') !== -1) {
+                    stop_on_error = false;
+                } else {
+                    stop_on_error = true;
+                }
+            } else {
+               stop_on_error = true;
+            }
         }
 
         this.clear_output(false, true);
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -314,8 +314,6 @@ define([
                     this.metadata.tags !== undefined) {
                 if (this.metadata.tags.indexOf('raises-exception') !== -1) {
                     stop_on_error = false;
-                } else if (this.metadata.tags.indexOf('nbval-raises-exception') !== -1) {
-                    stop_on_error = false;
                 } else {
                     stop_on_error = true;
                 }


takluyver:
This looks OK, but let's not check for the `nbval-` version of the tag here; leave that to be specific to nbval. We should add the generic one to nbval, though.

mscuthbert:
Done -- and patch to `nbval` is also being reviewed as https://github.com/computationalmodelling/nbval/pull/64

takluyver:
Thanks!

PR #2552, opened 2017-06-06
catch and log event errors on outputarea.element

minrk:
we already do this on global events, but the new events on output_area.element were unsafe, allowing extension code to prevent actions such as clear_output, etc. from completing.

cc @jasongrout @vidartf


vidartf:
Can confirm that the fix works locally for my use case.

PR #2560, opened 2017-06-08
Protect against hypothetical future where javascript stops using surrogate pairs

minrk:
to avoid introducing the inverse of https://github.com/jupyter/jupyter_client/issues/259

and check surrogate pairs more strictly, in case of wonky unicode that shouldn't happen but is technically possible.

bumps protocol version to 5.2 for explicit unicode fix (https://github.com/jupyter/jupyter_client/pull/262)

follow-up to #2509

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -1031,7 +1031,8 @@ define([
             if (char_code >= 0xD800 && char_code <= 0xDBFF) {
                 var next_char_code = text.charCodeAt(i+1);
                 if (next_char_code >= 0xDC00 && next_char_code <= 0xDFFF) {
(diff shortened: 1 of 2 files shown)


takluyver:
This looks sensible to me.

Do you think there's any possibility that CodeMirror could switch to using the index in code points rather than code units, even if it's not fixed at the language level? I just realised that's still not something that this code would catch, but I suspect it's unlikely.

stevengj:
@takluyver, almost certainly not, unless some hypothetical future version of JavaScript switches its indexing behavior (ala Python 2 vs Python 3, but this seems unlikely):  codemirror/CodeMirror#4750

takluyver:
Ok, this is fine by me then.

PR #2572, opened 2017-06-12
Change comment identifier to be valid JavaScript

finaiized:
Copying and pasting the snippet with `#` as a comment identifier caused an error, so use `//` instead.


takluyver:
Thanks!

PR #2574, opened 2017-06-13
Workaround for Windows Containers

dhirschfeld:
On Windows `os.stat` treats host mapped volumes as broken symlinks

xref: jupyterlab/jupyterlab#2426

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -9,6 +9,7 @@ import io
 import os
 import shutil
 import stat
+import sys
 import warnings
 import mimetypes
 import nbformat
@@ -18,6 +19,7 @@ from tornado import web
 from .filecheckpoints import FileCheckpoints
 from .fileio import FileManagerMixin
 from .manager import ContentsManager
+from ...utils import f_stat, WINDOWS_CONTAINER, exists
 
 from ipython_genutils.importstring import import_item
 from traitlets import Any, Unicode, Bool, TraitError, observe, default, validate
@@ -220,12 +222,12 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         """
         path = path.strip('/')
         os_path = self._get_os_path(path=path)
-        return os.path.exists(os_path)
+        return exists(os_path)
 
     def _base_model(self, path):
         """Build the common base of a contents model"""
         os_path = self._get_os_path(path)
-        info = os.stat(os_path)
+        info = f_stat(os_path)
         last_modified = tz.utcfromtimestamp(info.st_mtime)
         created = tz.utcfromtimestamp(info.st_ctime)
         # Create the base model.
@@ -275,7 +277,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                     continue
 
                 try:
-                    st = os.stat(os_path)
+                    st = f_stat(os_path)
                 except OSError as e:
                     # skip over broken symlinks in listing
                     if e.errno == errno.ENOENT:
@@ -284,7 +286,9 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                         self.log.warning("Error stat-ing %s: %s", os_path, e)
                     continue
 
-                if not stat.S_ISREG(st.st_mode) and not stat.S_ISDIR(st.st_mode):
+                if (not WINDOWS_CONTAINER
+                        and not stat.S_ISREG(st.st_mode)
+                        and not stat.S_ISDIR(st.st_mode)):
                     self.log.debug("%s not a regular file", os_path)
                     continue
 
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -19,7 +19,7 @@ from tornado import web
 from .filecheckpoints import FileCheckpoints
 from .fileio import FileManagerMixin
 from .manager import ContentsManager
-from ...utils import f_stat, WINDOWS_CONTAINER, exists
+from ...utils import exists
 
 from ipython_genutils.importstring import import_item
 from traitlets import Any, Unicode, Bool, TraitError, observe, default, validate
@@ -227,7 +227,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
     def _base_model(self, path):
         """Build the common base of a contents model"""
         os_path = self._get_os_path(path)
-        info = f_stat(os_path)
+        info = os.lstat(os_path)
         last_modified = tz.utcfromtimestamp(info.st_mtime)
         created = tz.utcfromtimestamp(info.st_ctime)
         # Create the base model.
@@ -277,7 +277,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                     continue
 
                 try:
-                    st = f_stat(os_path)
+                    st = os.lstat(os_path)
                 except OSError as e:
                     # skip over broken symlinks in listing
                     if e.errno == errno.ENOENT:
@@ -286,7 +286,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                         self.log.warning("Error stat-ing %s: %s", os_path, e)
                     continue
 
-                if (not WINDOWS_CONTAINER
+                if (not stat.S_ISLNK(st.st_mode)
                         and not stat.S_ISREG(st.st_mode)
                         and not stat.S_ISDIR(st.st_mode)):
                     self.log.debug("%s not a regular file", os_path)
--- a/notebook/services/contents/tests/test_manager.py
+++ b/notebook/services/contents/tests/test_manager.py
@@ -108,7 +108,7 @@ class TestFileContentsManager(TestCase):
         self.assertEqual(cp_dir, os.path.join(root, cpm.checkpoint_dir, cp_name))
         self.assertEqual(cp_subdir, os.path.join(root, subd, cpm.checkpoint_dir, cp_name))
 
-    @dec.skip_win32
+    @dec.skipif(sys.platform == 'win32' and sys.version_info[0] < 3)
     def test_bad_symlink(self):
         with TemporaryDirectory() as td:
             cm = FileContentsManager(root_dir=td)
@@ -120,9 +120,16 @@ class TestFileContentsManager(TestCase):
             # create a broken symlink
             self.symlink(cm, "target", '%s/%s' % (path, 'bad symlink'))
             model = cm.get(path)
-            self.assertEqual(model['content'], [file_model])
 
-    @dec.skip_win32
+            contents = {
+                content['name']: content for content in model['content']
+            }
+            self.assertTrue('untitled.txt' in contents)
+            self.assertEqual(contents['untitled.txt'], file_model)
+            # broken symlinks should still be shown in the contents manager
+            self.assertTrue('bad symlink' in contents)
+
+    @dec.skipif(sys.platform == 'win32' and sys.version_info[0] < 3)
     def test_good_symlink(self):
         with TemporaryDirectory() as td:
             cm = FileContentsManager(root_dir=td)
@@ -224,7 +231,7 @@ class TestContentsManager(TestCase):
                 self.assertEqual(entry['name'], "nb.ipynb")
                 complete_path = "/".join([api_path, "nb.ipynb"])
                 self.assertEqual(entry["path"], complete_path)
-    
+
     def setUp(self):
         self._temp_dir = TemporaryDirectory()
         self.td = self._temp_dir.name
--- a/notebook/utils.py
+++ b/notebook/utils.py
@@ -24,6 +24,24 @@ from ipython_genutils import py3compat
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
+# Replace `os.stat` which can't stat a host mapped volume from inside a Windows Container.
+WINDOWS_CONTAINER = sys.platform == 'win32' and os.environ.get('USERNAME') == 'ContainerAdministrator'
+if WINDOWS_CONTAINER:
+    f_stat = os.lstat
+else:
+    f_stat = os.stat
+
+
+def exists(path):
+    """Replacement for `os.path.exists` which works for host mapped volumes
+    on Windows containers
+    """
+    try:
+        f_stat(path)
+    except OSError:
+        return False
+    return True
+
 
 def url_path_join(*pieces):
     """Join components of url into a relative url
@@ -58,10 +76,10 @@ def url2path(url):
     pieces = [ unquote(p) for p in url.split('/') ]
     path = os.path.join(*pieces)
     return path
-    
+
 def url_escape(path):
     """Escape special characters in a URL path
-    
+
     Turns '/foo bar/' into '/foo%20bar/'
     """
     parts = py3compat.unicode_to_str(path, encoding='utf8').split('/')
@@ -69,7 +87,7 @@ def url_escape(path):
 
 def url_unescape(path):
     """Unescape special characters in a URL path
-    
+
     Turns '/foo%20bar/' into '/foo bar/'
     """
     return u'/'.join([
@@ -77,7 +95,6 @@ def url_unescape(path):
         for p in py3compat.unicode_to_str(path, encoding='utf8').split('/')
     ])
 
-_win32_FILE_ATTRIBUTE_HIDDEN = 0x02
 
 def is_file_hidden_win(abs_path, stat_res=None):
     """Is a file hidden?
@@ -98,22 +115,15 @@ def is_file_hidden_win(abs_path, stat_res=None):
     if os.path.basename(abs_path).startswith('.'):
         return True
 
-    # check that dirs can be listed
-    if os.path.isdir(abs_path):
-        # can't trust os.access on Windows because it seems to always return True
-        try:
-            os.stat(abs_path)
-        except OSError:
-            # stat may fail on Windows junctions or non-user-readable dirs
-            return True
-
+    win32_FILE_ATTRIBUTE_HIDDEN = 0x02
     try:
         attrs = ctypes.windll.kernel32.GetFileAttributesW(
-            py3compat.cast_unicode(abs_path))
+            py3compat.cast_unicode(abs_path)
+        )
     except AttributeError:
         pass
     else:
-        if attrs > 0 and attrs & _win32_FILE_ATTRIBUTE_HIDDEN:
+        if attrs > 0 and attrs & win32_FILE_ATTRIBUTE_HIDDEN:
             return True
 
     return False
@@ -164,12 +174,12 @@ else:
 
 def is_hidden(abs_path, abs_root=''):
     """Is a file hidden or contained in a hidden directory?
-    
+
     This will start with the rightmost path element and work backwards to the
     given root to see if a path is hidden or in a hidden directory. Hidden is
-    determined by either name starting with '.' or the UF_HIDDEN flag as 
+    determined by either name starting with '.' or the UF_HIDDEN flag as
     reported by stat.
-    
+
     Parameters
     ----------
     abs_path : unicode
@@ -191,12 +201,12 @@ def is_hidden(abs_path, abs_root=''):
     # is_file_hidden() already checked the file, so start from its parent dir
     path = os.path.dirname(abs_path)
     while path and path.startswith(abs_root) and path != abs_root:
-        if not os.path.exists(path):
+        if not exists(path):
             path = os.path.dirname(path)
             continue
         try:
             # may fail on Windows junctions
-            st = os.stat(path)
+            st = f_stat(path)
         except OSError:
             return True
         if getattr(st, 'st_flags', 0) & UF_HIDDEN:
@@ -244,7 +254,7 @@ def to_os_path(path, root=''):
 
 def to_api_path(os_path, root=''):
     """Convert a filesystem path to an API path
-    
+
     If given, root will be removed from the path.
     root must be a filesystem path already.
     """
--- a/notebook/utils.py
+++ b/notebook/utils.py
@@ -24,20 +24,13 @@ from ipython_genutils import py3compat
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
-# Replace `os.stat` which can't stat a host mapped volume from inside a Windows Container.
-WINDOWS_CONTAINER = sys.platform == 'win32' and os.environ.get('USERNAME') == 'ContainerAdministrator'
-if WINDOWS_CONTAINER:
-    f_stat = os.lstat
-else:
-    f_stat = os.stat
-
 
 def exists(path):
     """Replacement for `os.path.exists` which works for host mapped volumes
     on Windows containers
     """
     try:
-        f_stat(path)
+        os.lstat(path)
     except OSError:
         return False
     return True
@@ -206,7 +199,7 @@ def is_hidden(abs_path, abs_root=''):
             continue
         try:
             # may fail on Windows junctions
-            st = f_stat(path)
+            st = os.lstat(path)
         except OSError:
             return True
         if getattr(st, 'st_flags', 0) & UF_HIDDEN:
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -17,7 +17,7 @@ from jupyter_client.session import Session
 from jupyter_client.multikernelmanager import MultiKernelManager
 from traitlets import Bool, Dict, List, Unicode, TraitError, Integer, default, validate
 
-from notebook.utils import to_os_path
+from notebook.utils import to_os_path, exists
 from notebook._tz import utcnow, isoformat
 from ipython_genutils.py3compat import getcwd
 
@@ -55,7 +55,7 @@ class MappingKernelManager(MultiKernelManager):
         if not os.path.isabs(value):
             # If we receive a non-absolute path, make it absolute.
             value = os.path.abspath(value)
-        if not os.path.exists(value) or not os.path.isdir(value):
+        if not exists(value) or not os.path.isdir(value):
             raise TraitError("kernel root dir %r is not a directory" % value)
         return value
 


notebook/services/contents/filemanager.py
```diff
@@ -275,17 +289,17 @@ def _dir_model(self, path, content=True):
...
                     if e.errno == errno.ENOENT:
                         self.log.warning("%s doesn't exist", os_path)
                     else:
-                        self.log.warning("Error stat-ing %s: %s", os_path, e)
+                        self.log.warning("Error stat-ing %s: %s", (os_path, e))
                     continue
 
                 if not stat.S_ISREG(st.st_mode) and not stat.S_ISDIR(st.st_mode):
-                    self.log.debug("%s not a regular file", os_path)
+                    self.log.warning("%s not a regular file", os_path)
```
dhirschfeld: Just made this change to confirm why the mapped volume *still* isn't showing.
Results in warnings being printed to the terminal:
```
[W 17:07:19.450 LabApp] C:\jupyter\notebooks not a regular file
[W 17:07:29.451 LabApp] C:\jupyter\notebooks not a regular file
[W 17:07:40.452 LabApp] C:\jupyter\notebooks not a regular file
```

notebook/services/contents/filemanager.py
```diff
@@ -275,17 +289,17 @@ def _dir_model(self, path, content=True):
...
                     continue
 
                 if self.should_list(name) and not is_file_hidden(os_path, stat_res=st):
```
dhirschfeld: `is_file_hidden(os_path, stat_res=st)` returns True for a host mapped volume.

dhirschfeld:
Working around the fact that the mapped directory is hidden results in
```
[I 17:32:32.350 LabApp] Refusing to serve hidden directory 'C:\\jupyter\\notebooks', via 404 Error
[W 17:32:32.350 LabApp] directory does not exist: 'notebooks'
[W 17:32:32.350 LabApp] 404 GET /api/contents/?content=1&1497339152344 (172.29.208.1) 4.00ms referer=http://b7c8e663c8c1:8888/lab?
```

dhirschfeld:
:rabbit2: :hole:

dhirschfeld:
For reference:
image

dhirschfeld:
...however:
```python
[19]: from notebook.utils import is_file_hidden_win

[20]: st = os.stat(r"C:\jupyter\notebooks", follow_symlinks=False)
... : st
Out[20]: os.stat_result(st_mode=41471, st_ino=281474976710694, st_dev=113970060, st_nlink=1, st_uid=0, st_gid=0, st_size=0, st_atime=1497345137, st_mtime=1497345137, st_ctime=1497345137)

[21]: st.st_file_attributes
Out[21]: 1040

[22]: bool(st.st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN)
Out[22]: False

[23]: is_file_hidden_win(r"C:\jupyter\notebooks")
Out[23]: True

**Comment by dhirschfeld** [comment]
Cleaned up the implementation and (hopefully) made it py27 compatible

**Comment by dhirschfeld** [comment]
**RFR:** @takluyver, @minrk 

@takluyver - I've included you in the request for review as you appear to be the [author](https://github.com/dhirschfeld/notebook/commit/1de04234485b13baa0005f6a22c5b68a2cc61f22) of the `is_file_hidden_win` function.  

The use of `os.lstat` which doesn’t follow symlinks may render that code block redundant, but it will only occur inside a Windows Container where that code block wouldn’t work at all so it’s likely an improvement over the current situation!

**Comment by dhirschfeld** [comment]
I'm not sure why travis has failed - I can't see anything that should be the case since the changes should only apply on Windows (and when inside a container). Random failure?

File: notebook/utils.py [hunk 3]
```diff
@@ -24,6 +24,13 @@
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
+# Replace `os.stat` which can't stat a host mapped volume from inside a Windows Container.
+WINDOWS_CONTAINER = sys.platform == 'win32' and os.environ.get('USERNAME') == 'ContainerAdministrator'
+if WINDOWS_CONTAINER:
+    f_stat = os.lstat
```
minrk: should we just use `lstat` on windows in general, rather than trying to check if it's a container?

minrk:
I restarted Travis. Intermittent failure unrelated to the PR.

notebook/services/contents/filemanager.py
```diff
@@ -275,16 +281,18 @@ def _dir_model(self, path, content=True):
...
                 try:
-                    st = os.stat(os_path)
+                    st = f_stat(os_path)
                 except OSError as e:
                     # skip over broken symlinks in listing
                     if e.errno == errno.ENOENT:
                         self.log.warning("%s doesn't exist", os_path)
                     else:
-                        self.log.warning("Error stat-ing %s: %s", os_path, e)
+                        self.log.warning("Error stat-ing %s: %s", (os_path, e))
```
minrk: Adding these parentheses will cause a formatting error:

    TypeError: not enough arguments for format string

minrk:
Is there a way we can avoid detecting and special-casing Windows containers, and do something that works on Windows and/or catch errors appropriately, instead?

notebook/services/contents/filemanager.py
dhirschfeld: Sorry, not sure how that change crept in - I'll blame it on auto-formatting!

dhirschfeld:
I agree it's a pretty ugly hack - I was aiming to just fix my usecase without changing the existing behaviour for anyone else.

There's probably a better fix, but I'm not 100% sure why all the existing checks (some of which fail on Windows containers) are there so I just tried a minimal fix.

Specifically, the `is_file_hidden_win` provides a py2/3 compatible check for the hidden attribute:
```python
ishidden = bool(st.st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN)  # py3 only
```
...but it also tries to call `os.stat` on the path with the comment that it's checking if the directory can be listed:
```python
    # check that dirs can be listed
    if os.path.isdir(abs_path):
        # can't trust os.access on Windows because it seems to always return True
        try:
            os.stat(abs_path)
        except OSError:
            # stat may fail on Windows junctions or non-user-readable dirs
            return True
```
I assume by calling `os.stat` it's just checking if the path is a broken symlink which then implies it's not listable... unless it's a host mapped directory from within a Windows Container.

My initial reaction was to simply delete that block entirely as I'm not sure why its necessary. I think it would be better to show a *"broken"* symlink dir than to pretend it never existed. When trying to change to a broken directory the notebook could then just let the underlying `OSError` bubble up to a dialog for the user. In the case of a host mapped directory there would be no underlying `OSError` and changing to the directory would Just Work.

dhirschfeld:
IIUC by just using `lstat` on Windows you would be forgoing the checks for broken symlinks. I'm not too sure how important that is and as above I think it may even be better to show broken symlinks rather than hide them.

dhirschfeld:
Other than `is_file_hidden_win` there are a fewother checks in the `FileContentsManager` which are problematic in a Windows Container:

The [`os.exists`](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L223) check will fail on a host mapped dir. This was resolved by using `lstat`

The `stat` check in [`_dir_model`](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L278) was similarly converted to an `lstat`

And finally, the check for a [*regular*](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L287-L289) file:
```python
if not stat.S_ISREG(st.st_mode) and not stat.S_ISDIR(st.st_mode):
    self.log.debug("%s not a regular file", os_path)
    continue
```
This was resolved by protecting with a `if not WINDOWS_CONTAINER` clause.

Again though I'm not sure why there's a requirement to protect against listing non-regular (??) files. I think for a file browser the requirement could just be if it shows up with ls/dir then display it in the file browser?

If that's acceptable you could then just delete the whole checking block in `_dir_model`:
```python
try:
    st = os.stat(os_path)
except OSError as e:
    # skip over broken symlinks in listing
    if e.errno == errno.ENOENT:
        self.log.warning("%s doesn't exist", os_path)
    else:
        self.log.warning("Error stat-ing %s: %s", os_path, e)
    continue

if not stat.S_ISREG(st.st_mode) and not stat.S_ISDIR(st.st_mode):
    self.log.debug("%s not a regular file", os_path)
    continue
```

dhirschfeld:
So, my recommendation would actually be to remove all checking from both functions and list all files whether regular or not or broken symlinks or not. If the user then tries to use a "broken" file or directory they should get the same OS error as if they tried to do the same from the terminal.

Not sure if that makes sense or is reasonable. It's certainly a change in behaviour and thus a design question for the core devs.

dhirschfeld:
...maybe expand the contents definition to include attributes such as hidden/symlink/broken and have a UI element which would allow the user to choose whether or not to display files based on those attributes.

minrk:
I'd rather not include hidden files in the listing, since hidden files should be...hidden. Broken links are different, though. I think we hid broken links because they caused errors elsewhere when `stat` is called, but if switching to lstat (probably across the board on Windows) causes the stat to not fail, then using that instead makes sense to me, and broken links can be included in the listing. We should just make sure that opening the URL for a broken link properly 404s instead of raising an uncaught error.

dhirschfeld:
I'll look into what happens with broken links on my branch and report back...

dhirschfeld:
On my branch, when run from JupyterHub I get an internal server error dialog when I double-click a directory symbolic link. Will investigate without JupyterHub to try and get more info from the terminal.

In contrast an NTFS Junction Point works fine in the file browser.

dhirschfeld:
Trying to start a kernel in a host mapped dir seems to cause an issue:
```python-traceback
Error Starting Kernel
  File "C:\Miniconda3\lib\site-packages\traitlets\traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'root_dir'

  File "C:\Miniconda3\lib\site-packages\notebook\base\handlers.py", line 516, in wrapper
    result = yield gen.maybe_future(method(self, *args, **kwargs))
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "C:\Miniconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "C:\Miniconda3\lib\site-packages\notebook\services\sessions\handlers.py", line 75, in post
    type=mtype))
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "C:\Miniconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "C:\Miniconda3\lib\site-packages\notebook\services\sessions\sessionmanager.py", line 79, in create_session
    kernel_id = yield self.start_kernel_for_session(session_id, path, name, type, kernel_name)
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "C:\Miniconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "C:\Miniconda3\lib\site-packages\notebook\services\sessions\sessionmanager.py", line 92, in start_kernel_for_session
    self.kernel_manager.start_kernel(path=kernel_path, kernel_name=kernel_name)
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "C:\Miniconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 307, in wrapper
    yielded = next(result)
  File "C:\Miniconda3\lib\site-packages\notebook\services\kernels\kernelmanager.py", line 92, in start_kernel
    kwargs['cwd'] = self.cwd_for_path(path)
  File "C:\Miniconda3\lib\site-packages\notebook\services\kernels\kernelmanager.py", line 66, in cwd_for_path
    os_path = to_os_path(path, self.root_dir)
  File "C:\Miniconda3\lib\site-packages\traitlets\traitlets.py", line 556, in __get__
    return self.get(obj, cls)
  File "C:\Miniconda3\lib\site-packages\traitlets\traitlets.py", line 535, in get
    value = self._validate(obj, dynamic_default())
  File "C:\Miniconda3\lib\site-packages\traitlets\traitlets.py", line 593, in _validate
    value = self._cross_validate(obj, value)
  File "C:\Miniconda3\lib\site-packages\traitlets\traitlets.py", line 599, in _cross_validate
    value = obj._trait_validators[self.name](obj, proposal)
  File "C:\Miniconda3\lib\site-packages\traitlets\traitlets.py", line 907, in __call__
    return self.func(*args, **kwargs)
  File "C:\Miniconda3\lib\site-packages\notebook\services\kernels\kernelmanager.py", line 52, in _update_root_dir
    raise TraitError("kernel root dir %r is not a directory" % value)
traitlets.traitlets.TraitError: kernel root dir 'C:\\jupyter\\dhirschf' is not a directory
```

dhirschfeld:
When double-clicking on a symlink directory JupyterLab pops up an Internal Server Error dialog and I get the below error in the terminal:
```
[E 15:45:06.909 LabApp] Unhandled error in API request
      File "C:\Miniconda3\lib\site-packages\notebook\base\handlers.py", line 516, in wrapper
        result = yield gen.maybe_future(method(self, *args, **kwargs))
      File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1055, in run
        value = future.result()
      File "C:\Miniconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "<string>", line 4, in raise_exc_info
      File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 307, in wrapper
        yielded = next(result)
      File "C:\Miniconda3\lib\site-packages\notebook\services\contents\handlers.py", line 124, in get
        path=path, type=type, format=format, content=content,
      File "C:\Miniconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 390, in get
        model = self._dir_model(path, content=content)
      File "C:\Miniconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 275, in _dir_model
        for name in os.listdir(os_dir):
    FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\jupyter\\dhirschf\\lnk2tmp'
[E 15:45:07.048 LabApp] {
      "Host": "101026def9c6:8888",
      "Connection": "keep-alive",
      "X-Xsrftoken": "2|04f222c8|c265413a0b7b6a0a666875450653facd|1497591882",
      "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36",
      "Accept": "*/*",
      "Referer": "http://101026def9c6:8888/lab?",
      "Accept-Encoding": "gzip, deflate, sdch",
      "Accept-Language": "en-US,en;q=0.8",
      "Cookie": "_xsrf=2|04f222c8|c265413a0b7b6a0a666875450653facd|1497591882"
    }
[E 15:45:07.048 LabApp] 500 GET /api/contents/lnk2tmp?content=1&1497591906904 (172.29.208.1) 141.06ms referer=http://101026def9c6:8888/lab?
```
Note: the directory `tmp` *does* exist so the symlink isn't broken
image

dhirschfeld:
...well, that's just odd - `os.listdir` and `cd` in the container both fail but when run on the host they work. I set the permissions so everyone has access / full control over both `tmp` and `lnk2tmp` folders but that made no difference so it seems like symlink directories aren't supported

dhirschfeld:
If I create instead a junction point it works fine. If I then purposely break the junction point by deleting `tmp` which it's pointing to I get the same Internal Server Error and traceback in the terminal
```
[E 16:17:59.923 LabApp] Unhandled error in API request
      File "C:\Miniconda3\lib\site-packages\notebook\base\handlers.py", line 516, in wrapper
        result = yield gen.maybe_future(method(self, *args, **kwargs))
      File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 1055, in run
        value = future.result()
      File "C:\Miniconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "<string>", line 4, in raise_exc_info
      File "C:\Miniconda3\lib\site-packages\tornado\gen.py", line 307, in wrapper
        yielded = next(result)
      File "C:\Miniconda3\lib\site-packages\notebook\services\contents\handlers.py", line 124, in get
        path=path, type=type, format=format, content=content,
      File "C:\Miniconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 390, in get
        model = self._dir_model(path, content=content)
      File "C:\Miniconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 275, in _dir_model
        for name in os.listdir(os_dir):
    FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\jupyter\\dhirschf\\lnk2tmp'
[E 16:17:59.957 LabApp] {
      "Host": "22d7b7bbfcd9:8888",
      "Connection": "keep-alive",
      "X-Xsrftoken": "2|0703062b|ad483203768cff5faa31b2a9ba65f6d1|1497593375",
      "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36",
      "Accept": "*/*",
      "Referer": "http://22d7b7bbfcd9:8888/lab?",
      "Accept-Encoding": "gzip, deflate, sdch",
      "Accept-Language": "en-US,en;q=0.8",
      "Cookie": "_xsrf=2|0703062b|ad483203768cff5faa31b2a9ba65f6d1|1497593375"
    }
[E 16:17:59.957 LabApp] 500 GET /api/contents/lnk2tmp?content=1&1497593879920 (172.29.208.1) 35.01ms referer=http://22d7b7bbfcd9:8888/lab?
```

dhirschfeld:
My suggestion would be for symlinks (which don't work) and for broken junction points to display the error text:
```
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\jupyter\\dhirschf\\lnk2tmp'
```
...rather than just the opaque Internal Server Error.

But I suppose that's an issue for JupyterLab rather than the notebook.

dhirschfeld:
File soft-links also fail, but hard links seem to work:
image
```
[W 16:31:36.620 LabApp] Cannot read non-file C:\jupyter\dhirschf\hw-softlink.txt
[W 16:31:36.620 LabApp] 400 GET /api/contents/hw-softlink.txt?format=text&type=file&content=1&1497594696615 (172.29.208.1) 1.96ms referer=http://22d7b7bbfcd9:8888/lab?
```

dhirschfeld:
The symlink situation on Windows seems to be a known issue:
https://github.com/docker/docker.github.io/pull/34/files

dhirschfeld:
> I'd rather not include hidden files in the listing, since hidden files should be...hidden

The 👍's on [this](https://github.com/jupyter/notebook/issues/1532#issuecomment-244159091) comment show that there is some support for having this configurable. As a relative power user on Windows one of the first things I do is to tick the Show Hidden Files box in the Explorer config.

I'm not advocating it in this PR though - I think it deserves a PR of its own if anyone is interested enough. I'll be happy to just get this working on Windows Containers.

dhirschfeld:
The latest push fixes the error when trying to start the server in a host mapped directory - https://github.com/jupyter/notebook/pull/2574#issuecomment-308938402

As it is, this PR allows me to use the Notebook/JupyterLab on NanoServer containers.

In this version I've deleted the broken symlink checking from the `is_file_hidden_win` function as I think the current behaviour of pretending it doesn't exist isn't helpful. Otherwise the change to using `lstat` is still protected by the `WINDOWS_CONTAINER` flag. If you think it's better to just always use `lstat` on Windows I'm happy to make that change but will let a core developer make that call.

If possible it would be great to get this into the next release as without these changes you can't use the Notebook/JupyterLab for anything serious on Windows Containers as there's no way to persist any user data outside of the container.

cc: @minrk, @takluyver

notebook/utils.py
```diff
@@ -98,22 +115,15 @@ def is_file_hidden_win(abs_path, stat_res=None):
...
-    # check that dirs can be listed
-    if os.path.isdir(abs_path):
-        # can't trust os.access on Windows because it seems to always return True
-        try:
-            os.stat(abs_path)
-        except OSError:
-            # stat may fail on Windows junctions or non-user-readable dirs
-            return True
-
+    win32_FILE_ATTRIBUTE_HIDDEN = 0x02
```
dhirschfeld: Made the module level global local to this function as it isn't used elsewhere

notebook/utils.py
```diff
@@ -98,22 +115,15 @@ def is_file_hidden_win(abs_path, stat_res=None):
...
         return True
 
-    # check that dirs can be listed
-    if os.path.isdir(abs_path):
-        # can't trust os.access on Windows because it seems to always return True
-        try:
-            os.stat(abs_path)
-        except OSError:
-            # stat may fail on Windows junctions or non-user-readable dirs
-            return True
```
dhirschfeld: Removed checking for a broken symlink as I think it's better to let the user know and have them deal with it rather than pretend it doesn't exist

notebook/utils.py
minrk: Can we do this for Windows in general, and remove the WINDOWS_CONTAINER constant?

notebook/services/contents/filemanager.py
```diff
@@ -284,7 +286,9 @@ def _dir_model(self, path, content=True):
                         self.log.warning("Error stat-ing %s: %s", os_path, e)
                     continue
 
-                if not stat.S_ISREG(st.st_mode) and not stat.S_ISDIR(st.st_mode):
+                if (not WINDOWS_CONTAINER
```
minrk: just `win32` instead of WINDOWS_CONTAINER?

dhirschfeld: As it stands the `WINDOWS_CONTAINER` flag both tests that the platform is `win32` *and* that you're running from inside a container by checking for the default container user `ContainerAdministrator`.

It's certainly possible to make the changes for all windows users but I was hesitant to potentially break existing code.

minrk: I certainly appreciate aiming for minimal impact. I'm coming from the perspective that anything WINDOWS_CONTAINER specific is going to get minimal testing, so if we can reduce the number of cases it will be easier for us to maintain / catch regressions. Plus, I think showing broken links should be fine, as long as it doesn't cause an error in directory listing (using lstat fixes this, I believe).

Instead of gating on Windows, since it can be a link, maybe it should be adding `stat.S_ISLNK` to the checks instead of making it conditional on the platform. Does that return True for these mounts? What is `os.stat(path).st_mode` for these mounts?

dhirschfeld: ` os.stat(path)` would throw a `FileNotFound` error. You'd need to either use `lstat` or pass the `follow_symlinks=False` arg (py3 only).

`lstat` gives a 41471 mode:
```python
[20]: st = os.stat(r"C:\jupyter\notebooks", follow_symlinks=False)
... : st
Out[20]: os.stat_result(st_mode=41471, st_ino=281474976710694, st_dev=113970060, st_nlink=1, st_uid=0, st_gid=0, st_size=0, st_atime=1497345137, st_mtime=1497345137, st_ctime=1497345137)

[21]: st.st_file_attributes
Out[21]: 1040
```

If we don't care about checking if a symlink is broken we could just globally use `lstat`. The only other issue is then the check for a *regular* file:
```python
if not stat.S_ISREG(st.st_mode) and not stat.S_ISDIR(st.st_mode):
    self.log.debug("%s not a regular file", os_path)
    continue
```
In that case doesn't recognise the mapped volume as a directory but sees it as a link instead:
```python
In [2]: stat.S_ISREG(41471)
Out[2]: False

In [3]: stat.S_ISDIR(41471)
Out[3]: False

In [4]: stat.S_ISLNK(41471)
Out[4]: True
```
...so the `if not WINDOWS_CONTAINER` could be replaced with `and not stat.S_ISLNK(st.st_mode)`

Happy to go ahead and make those changes if you think that's the right way forward...

dhirschfeld:
The CI error is below:
```
  File "/home/travis/build/jupyter/notebook/notebook/services/contents/tests/test_manager.py", line 123, in test_bad_symlink
    self.assertEqual(model['content'], [file_model])
AssertionError: Lists differ: [{'mimetype': 'text/plain', 'w... != [{'mimetype': 'text/plain', 'w...
```
The function which failed is:
```python
    @dec.skip_win32
    def test_bad_symlink(self):
        with TemporaryDirectory() as td:
            cm = FileContentsManager(root_dir=td)
            path = 'test bad symlink'
            _make_dir(cm, path)

            file_model = cm.new_untitled(path=path, ext='.txt')

            # create a broken symlink
            self.symlink(cm, "target", '%s/%s' % (path, 'bad symlink'))
            model = cm.get(path)
            self.assertEqual(model['content'], [file_model])
```
IIUC this function is testing that if you create a broken symlink the `FileContentsManager` ignores it - i.e pretends it never existed. The changes in the PR deliberately break that behaviour so that the `FileContentsManager` *will* list broken symlinks and an `OSError` will only be thrown later if one attempts to use the directory. So, if these changes get the go-ahead the test will also need to be changed... if I've got it correct.

minrk:
I think you are exactly right. Since we're changing how we treat bad symlinks (they show up as regular files, then 404 when one tries to open them), updating the test is appropriate.

If the contents manager sees your mounts as bad symlinks, do they look like files in the file listing, or directories? Can you open them?

dhirschfeld:
Host mounted volumes have a directory icon as they should. Opening them in the JupyterLab File Browser works fine and correctly lists the contents. It's just that stat fails for whatever reason on them. Not sure if that's a python issue or expected/correct behaviour.

In contrast a broken symlink directory shows up as a directory icon (in this PR) but double-clicking on it in the JupyterLab File Browser pops up an Internal Server Error dialog in the app and logs a FileNotFound error in the terminal.

I'll look to finish off this PR by updating the test for the new behaviour...

Thanks for the review @minrk!

notebook/services/contents/tests/test_manager.py
```diff
@@ -108,7 +108,7 @@ def test_checkpoint_subdir(self):
         self.assertEqual(cp_dir, os.path.join(root, cpm.checkpoint_dir, cp_name))
         self.assertEqual(cp_subdir, os.path.join(root, subd, cpm.checkpoint_dir, cp_name))
 
-    @dec.skip_win32
+    @dec.skipif(sys.platform == 'win32' and sys.version_info[0] < 3)
```
dhirschfeld: Prior to version 3.2 [`os.symlink`](https://docs.python.org/3.5/library/os.html#os.symlink) which is required for this test only existed for unix. This enables testing on Python 3 for Windows as all supported versions have this functionality now.

dhirschfeld:
For reference, this seems like a bug in docker. I still prefer the new behaviour though as hiding errors from the user doesn't give them a chance to do anything about it.

https://github.com/moby/moby/issues/31089
https://go-review.googlesource.com/c/41834/

notebook/services/contents/tests/test_manager.py
takluyver: `os.symlink` exists now, but I think that in most situations it still raises a permission error, because normal users don't have the right to create symlinks. So I suspect that when we next do a build on Windows, this might fail.

dhirschfeld: I wondered if that might be the case but I figured if the tests passed on appveyor it would be fine. 

That was based on the assumption that most normal Windows users probably wouldn't be running the tests. In that case it would only affect a Windows developer trying to run the tests without admin privileges where it would fail with a permission error. I think that having some testing on Windows/appveyor is probably worth any hassle that might cause?

OT: In the future it should be possible for developers without admin privileges to create symlinks on Windows 10 - https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/#oWkhfiQvGP6wte4V.97

takluyver: Ah, nice. I guess our build servers have got the 'creators update' and therefore are able to create symlinks. :-)

Issue #1387, opened 2016-04-22
Add robots.txt


wernight:
Just like https://github.com/jupyter/nbviewer/commit/316665e5d7eee4cba9010d12803df94bf993fc40 it'd be good to have a default `robots.txt`.

Carreau:
I'm unsure, the notebook is supposed to be behind login, so no robot should be able to crawl it.
Should not be too hard though. 
I guess a no-index nofollow could be good too. Just to avoid indexing potential open notebook on the web.

wernight:
It's not to allow crawling but for two things:

 * Because it's a good practice to just be explicit.
 * Because it can be used for example to get the Kubernetes Ingress controller working.

Possibly together, a [`humans.txt`](http://humanstxt.org/) wouldn't be a bad idea for you guys.

go-bears:
Is this issue still open? I can help. Do you want a  robots.txt file and/or a <meta> tag on the page.html template?

wernight:
I'm using Nginx Ingress Controller which doesn't require those files.

`robots.txt` and/or `humans.txt` may help others, or just as a common practice.

minrk:
It is. It would be good to add these (possibly not as files, just hardcoded Handlers).

takluyver:
Fixed by #2579.

PR #2579, opened 2017-06-15
added robots.txt

Krista:
This commit fixes issue #1387 by adding a file `robots.txt`
that makes robots go away.

This commit is also similar to
https://github.com/jupyter/nbviewer/commit/316665e5d7eee4cba9010d12803df94bf993fc40
as it also adds the `favicon.ico`.


takluyver:
Thanks

PR #2589, opened 2017-06-20
Fix logo alignment on dashboard

gnestor:
Fixes https://github.com/jupyter/notebook/issues/2575


takluyver:
I confirm that this fixes #2575, but the header of the notebook page looks like this:

screenshot from 2017-06-21 09-52-24

i.e. the notebook name is moved to the right of where it used to be.

gnestor:
Sorry that I missed that! I usually have my header hidden... Should be fixed now.

PR #2592, opened 2017-06-21
Add id=favicon in templates

takluyver:
The new dynamic favicon-changing code looks for an existing favicon in the DOM with id 'favicon'. This uses that id in the templates, so we don't end up with two favicon links in the page.

I was hoping that this would fix some issues I'm having with the favicon getting stuck, but it doesn't seem to. :-( It seems like the right thing to do anyway.


rgbkrk:
Seems wise.

Issue #1227, opened 2016-03-17
Ctrl-c doesn't seem to work in web terminal


praveen-srinivasan:
This is Jupyter notebook server 4.1.0

Ctrl-c doesn't seem to do what I would expect, e.g. if I do something like "tail -f foo" and then hit "ctrl-c", it doesn't exit "tail". It just prints "^C" on the screen.

takluyver:
Works for me. What OS & browser are you on?

praveen-srinivasan:
Mac OS X, Chrome

yuvipanda:
Same as https://github.com/jupyter/notebook/issues/104 I think?

takluyver:
I think #104 is about copy/paste, and this is about interrupting programs using Ctrl-C. Both are likely terminado issues rather than something we can fix, though there's a small chance that this is something on our side.

alouisos:
Reporting the same issue on Mac OS Chrome terminals. Cannot interrupt process in terminal with Ctrl+C

kingjr:
Same problem with with Ubuntu 14.04 & Firefox.

gnestor:
I have verified that `^C` is successfully interrupting a process using Chrome on OS X. 

Screenshot:

![screenshot](http://g.recordit.co/zCgVI7GZmM.gif)

ricardobarroslourenco:
I'm under the same issue, using MacOS El Capitan...

oolongknits:
I had the same problem and (at least in my case) this is a problem with the terminal shortcuts. 

I'm on Linux and in the terminal I did:
Edit -> Preferences -> Shortcuts
I found Ctr+C was a shortcut to Copy. 
I changed it to something else and now it works as it is supposed to.

oschuett:
Maybe the Jupyter terminal should overwrite some of the browser's default shortcuts:
https://stackoverflow.com/questions/3680919/overriding-browsers-keyboard-shortcuts

gnestor:
I should add that it's working for me using Chrome on MacOS Sierra.

I understand that <kbd>CTRL+C</kbd> on Windows is copy and I don't think there is anything we can do to override that, but I'm curious what shortcut do you use to kill a process in the Window's terminal? Linux?

oschuett:
I'm using Chrome under Linux and there it's not working either for the same reason. Hence, my proposal to prevent the default event handling in these cases.

gnestor:
But what if you want to copy something from the terminal output?

[...]

PR #2601, opened 2017-06-23
Upgrade bower dependencies

gnestor:
Follow up from https://github.com/jupyter/notebook/issues/1227

This upgrades xterm.js and codemirror dependencies in bower.json. 

I also removed the codemirror patch code in favor of installing codemirror from a fork repo because it seems like it will be easier to maintain that way (and it’s slightly less of a hack). Using the patch code, it isn’t obvious to any contributor except myself which part of codemirror needs patched and this patching needs to happen every time that codemirror is upgraded. Using a fork repo, we can simply rebase the patch branch to the most recent release commit on codemirror and replay the patch commit over it. For efficiency’s sake. the fork is under my name, but I can transfer it to the jupyter org or any other contributor can fork codemirror themselves and perform the same patch on their own branch.


takluyver:
Thanks @gnestor . Looks like there's an issue on CI with building the JS with your codemirror fork.

takluyver:
Also, have we re-tested for the issue with the latest Codemirror? It might have been fixed, in which case we could drop our patch.

gnestor:
I tested vanilla codemirror 5.27.2 and I can't reproduce any of those bugs from https://github.com/jupyter/notebook/issues/1967. I tested in Safari and Chrome. Can you give it a try to double-check?

takluyver:
I'll have a go with it, but I think that bug only ever affected Safari, so I can't check for it. If we can get another Mac user to test and find it working, I'd say ship the vanilla version and cross our fingers.

takluyver:
I had a very brief play with it in Firefox, it seems to be working OK.

gnestor:
Ya, it looks good to me too. Merge at will 👍

takluyver:
I'd like to get one more safari user (or at least, someone with a Mac who can run Safari) to try to reproduce #1967 while we're still thinking about it before we merge, because it will be harder to find the necessary information if it's reported again in a few weeks.

gnestor:
Ok, @rgbkrk or @minrk, could you check this out and test for codemirror single-space bugs (https://github.com/jupyter/notebook/issues/1967) in Safari?

takluyver:
OK, let's merge this and hope that the problem is gone.

PR #2610, opened 2017-06-27
Update link of messaging  in Jupyter

vharsh:
The link given in the docs is for an old version of IPython and it has a link to the newer docs i.e. https://jupyter-client.readthedocs.io


vharsh:
Sends the reader to https://jupyter-client.readthedocs.io/en/latest/messaging.html#messaging instead of http://ipython.org/ipython-doc/dev/development/messaging.html

takluyver:
Thanks!

PR #2615, opened 2017-06-28
Remove references to header cells 

mpacer:
In addition to removing direct references I removed best practices that better matched the existence of single-purpose heading cells. For example, it no longer makes sense to always make headings unique cells and only include other content on later cells (since headings are _just_ markdown headings they should be able to be treated like markdown text).

Context: heading cells were removed in nbformat v4, which has been the standard for almost 2 years.


rgbkrk:
Hurrah!

Issue #2250, opened 2017-03-03
remove IPYTHON_ALLOW_DRAFT_WEBSOCKETS_FOR_PHANTOMJS


minrk:
with updates to phantomjs being based on Qt5 (v2.0, I think), we probably don't need to ship the backport to deprecated implementations of the websockets for the phantomjs tests.

I won't say we should do this for 5.0 since it's close to release, but I think we can drop this bit of testing code.

PR #2628, opened 2017-07-04
Remove old websocket draft76 protocol.

takluyver:
Support for this older protocol should no longer be needed with Phantom JS 2.

Closes gh-2250


takluyver:
@minrk you opened the issue about this: do you see any reason to keep the draft76 code for another release cycle?

minrk:
Should be fine to do this now, since we don't use it anymore and it's purely internal testing code.

PR #2629, opened 2017-07-04
Fix last-modified column of notebook list

takluyver:
This was showing 'seconds ago' for every file, because it was looking for the nonexistant 'modified' field. The correct name is 'last_modified'.

I'm not sure how long this has been broken.


takluyver:
Looks like it was accidentally broken in #2449, so we haven't made a release with the bug yet.

takluyver:
Merging this, because it's a simple fix.

PR #2630, opened 2017-07-04
Cell tags UI: Ensure close button does not wrap

### LINKED ISSUE #2459: UI issues when too many cell tags are entered
To replicate:
1. Show the tags toolbar: View --> Cell Toolbar --> Tags
2. Enter a string with a few dozen spaces in it and click "Add Tag".
3. See the first few tags, then the rest are cut off.  Moreover, the x's to delete tags are mysteriously missing.

I noticed this when doing my from-scratch React.js based implementation of the Jupyter notebook and re-implementing this feature.  On the top is what things should look like (in my humble opinion) and on the bottom is what they do look like in Jupyter now:

screen shot 2017-04-29 at 11 46 04 am

Another UI comments which you may disagree with: The ... to pop up a modal dialog to edit the same list of tags is completely pointless and redundant.  It provides no extra functionality, and is itself quite confusing (since you would assume that it does provide extra functionality).

vidartf:
Add CSS styling to prevent close button from line-wrapping inside the cell tag elements.

Closes #2459 (minimal fix).

### DIFF
--- a/notebook/static/notebook/less/tagbar.less
+++ b/notebook/static/notebook/less/tagbar.less
@@ -62,6 +62,7 @@
 
 .cell-tag {
     background-color: #fff;
+    white-space: nowrap;
 }
 
 .tags-input input[type=text]:focus {


takluyver:
Thanks, I've tested that locally.

Issue #2256, opened 2017-03-06
Copying Multi Line from Terminal Inserts Line Breaks


cancan101:
When I copy text from a jupyter notebook terminal session that spans multiple lines, line breaks are inserted.

takluyver:
Can you describe a bit more the results and what you expected? Thanks.

cancan101:
I run:
```
# python -c 'print("M"*455)'
```
And I get (notice the line breaks when I copy out of the terminal):
```
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
```

takluyver:
I think that's not unusual for terminal emulators, although modern ones (including Gnome terminal, which I use) preserve it without the linebreaks. If you want to chase it up, the terminal component we use is [xterm.js](https://github.com/sourcelair/xterm.js), so that's where it would need to be changed. See maybe sourcelair/xterm.js#404.

cancan101:
I think this is the upstream issue: https://github.com/sourcelair/xterm.js/issues/443.

cancan101:
This has now been fixed in: https://github.com/sourcelair/xterm.js/pull/693

takluyver:
Thanks @cancan101 . We should remember to update our version of xterm when there's a release with that fix.

cancan101:
[...]

PR #2638, opened 2017-07-06
Upgrade xterm.js to newly released 2.8.0

cancan101:
Fixes https://github.com/jupyter/notebook/issues/2256

/CC @gnestor @takluyver


takluyver:
Thanks @cancan101 . Assuming the tests pass, I'm happy to merge this straight away.

takluyver:
In fact, I don't think we test the terminal at all, so even if the tests fail, it's probably spurious. ;-)

PR #2643, opened 2017-07-08
Upgrade xterm.js to newly released 2.8.1

cancan101:
Fixes issue with scroll wrapping in initial viewport: https://github.com/sourcelair/xterm.js/pull/768


takluyver:
Thanks!

cancan101:
What is the plan for the 5.1 release?

takluyver:
My plan is 'do it quite soon'. ;-)

It looks like Min is doing a lot of merges today, so we're moving in the right direction. :-)

Issue #968, opened 2016-01-18
/files/ should use StaticFileHandler when files are local


minrk:
Right now, the ContentsManager is used for requests on `/files/`. This is the most generic and will work for any ContentsManager, but is very inefficient when it's not needed. For the default case of FileContentsManager, we should go back to using AuthenticatedFileHandler, which is now unused (#865).

The simplest way to express this, I think, is a files_handler_class attribute on the ContentsManager. The base class should set the default to what we use now, and FileContentsManager should set it to AuthenticatedFileHandler.

takluyver:
Closed by #2656

PR #2656, opened 2017-07-17
Use StaticFileHandler when files are local

### LINKED ISSUE #1024: Embedding local videos works imperfectly
I have a notebook where I'm generating and then displaying a video file using the following HTML snippet:

``` html
<video controls autoplay>
    <source src="movie.mp4" type="video/mp4"/>
</video>
```

This works to some degree but there seems to be something wrong.
- In Chrome and Opera, the video plays fine, but the timeline remains in "loading" state and it is not possible to jump to any location except back to the start.
- In Safari, the video does not play at all and remains in "loading" state.

The notebook server logs the same thing in all cases:
`[I 18:11:02.417 NotebookApp] 302 GET /notebooks/path/to/movie.mp4 (::1) 1.22ms`

If I instead use a direct `file:///` URL the video works as expected with the ability to jump around.

Embedding images eg. `` works fine.

A search through the web yields several pages that advise using Base64 data instead of an URL, but it seems like this almost works. Is it a known issue with the way local files are served?

agermanidis:
Implements the set of changes described at #968, and fixes #1024. 

In summary, a `files_handler_class` is added as an attribute to `ContentsManager`, specifying a custom request handler for the `/files` requests. For `FileContentsManager`, this is set to `AuthenticatedFilesHandler` (which is a light proxy over Tornado's `StaticFileHandler`).

Using tornado's `StaticFileHandler` class for serving local files is more efficient and enables HTTP Range requests, which are necessary for seekable video embeds (see #1024).

An additional change is made to the `test_contents_manager` test case to allow both `text/plain` 
and `text/plain; charset=UTF-8` as acceptable `content-type`s. The reason being that while `FilesHandler` explicitly specifies the content type ([here](https://github.com/jupyter/notebook/blob/master/notebook/files/handlers.py#L50)), Tornado's static file handler just sets the content type to the mime-type of the file, which doesn't include charset; overwriting that would be awkward.

This is my first time digging through this codebase so any changes/comments/suggestions would be much appreciated.

### DIFF
--- a/notebook/tests/test_files.py
+++ b/notebook/tests/test_files.py
@@ -95,7 +95,7 @@ class FilesTest(NotebookTestBase):
 
         r = self.request('GET', 'files/test.txt')
         self.assertEqual(r.status_code, 200)
-        self.assertEqual('text/plain', r.headers['content-type'])
+        self.assertEqual(r.headers['content-type'], 'text/plain; charset=UTF-8')
         self.assertEqual(r.text, 'foobar')
     
     def test_download(self):
--- a/notebook/tests/test_files.py
+++ b/notebook/tests/test_files.py
@@ -95,7 +95,7 @@ class FilesTest(NotebookTestBase):
 
         r = self.request('GET', 'files/test.txt')
         self.assertEqual(r.status_code, 200)
-        self.assertEqual(r.headers['content-type'], 'text/plain; charset=UTF-8')
+        self.assertEqual('text/plain', r.headers['content-type'])
         self.assertEqual(r.text, 'foobar')
     
     def test_download(self):
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -472,11 +472,7 @@ class AuthenticatedFileHandler(IPythonHandler, web.StaticFileHandler):
 
     @web.authenticated
     def get(self, path):
-        if os.path.splitext(path)[1] == '.ipynb':
-            name = path.rsplit('/', 1)[-1]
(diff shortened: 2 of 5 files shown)


notebook/tests/test_files.py
```diff
@@ -95,7 +95,7 @@ def test_contents_manager(self):
 
         r = self.request('GET', 'files/test.txt')
         self.assertEqual(r.status_code, 200)
-        self.assertEqual(r.headers['content-type'], 'text/plain; charset=UTF-8')
```
takluyver: This PR is effectively reverting #2402, so it will likely reintroduce #2397.

gnestor: It looks like the tests are passing on Travis... It looks good to me. What says @minrk?

minrk: We can define [AuthenticatedFileHandler.get_content_type](https://github.com/tornadoweb/tornado/blob/v4.5.1/tornado/web.py#L2641) to include the charset.

takluyver: Are you requesting that @agermanidis does that in this PR, or suggesting that we do it separately?

agermanidis:
Thanks, I wasn't aware of `get_content_type`. `AuthenticatedFileHandler` should now set the charset properly. I reverted the test as well.

takluyver:
Thanks! @minrk @gnestor feel free to merge this if you're happy.

gnestor:
Thanks @agermanidis!! We appreciate your contribution!

PR #2661, opened 2017-07-19
Remove space before ? in prompt message

adl1995:


takluyver:
Thanks, but that file is automatically generated, so we shouldn't edit it by hand. The message comes from `notebook/static/notebook/js/notebook.js` - can you edit it there? Then we can regenerate this file automatically (see the [i18n README](https://github.com/jupyter/notebook/blob/master/notebook/i18n/README.md)).

adl1995:
@takluyver Yeah, sure.

adl1995:
@takluyver Please check #2665. Closing this now.

PR #2662, opened 2017-07-19
Type error in notebookapp.py

iSmartie:
"though ssh" should be "through ssh"


iSmartie:
By the way, is there a way I can use code auto complete and parenthesis matching when editing .py file (not .ipynb file) in jupyter notebook's web browser's text editor?

minrk:
Thanks!

> By the way, is there a way I can use code auto complete and parenthesis matching when editing .py file (not .ipynb file) in jupyter notebook's web browser's text editor?

Not at the moment. We should enable CodeMirror completions, though. I assumed we had already.

Issue #2094, opened 2017-01-31
Link with token from UI


minrk:
It would probably be convenient for users who switch browsers / toggle private browsing to be able to get a link with the auth token in the URL that can be copied from somewhere in the notebook UI. That way opening a new authenticated browser can be done easily from the already authenticated browser, without needing to get the token from `jupyter notebook list` in a terminal.

Possibly related:

- Should we expose `jupyter notebook list` results in the UI and/or API? For the local case it sounds nice, but I'm thinking it's probably not a good idea as it breaks a notebook's isolation, exposing other notebook servers for the same user on the system (including their auth info).

jasongrout:
> Should we expose jupyter notebook list results in the UI and/or API? For the local case it sounds nice, but I'm thinking it's probably not a good idea as it breaks a notebook's isolation, exposing other notebook servers for the same user on the system (including their auth info).

I think not - I think it's really important to preserve the notebook's isolation.

jasongrout:
> It would probably be convenient for users who switch browsers / toggle private browsing to be able to get a link with the auth token in the URL that can be copied from somewhere in the notebook UI. That way opening a new authenticated browser can be done easily from the already authenticated browser, without needing to get the token from jupyter notebook list in a terminal.

I think it would make sense to make it easy to retrieve a one-time token from the rest api, which could be exposed in the browser, or another program could launch another browser, etc.

psychemedia:
I run notebooks using Docker via Kitematic and get a status message along the lines of:

![image](https://cloud.githubusercontent.com/assets/82988/23316979/8c469900-fac4-11e6-8d0c-15b56979eaeb.png)

Which is sort of true but sort of not - because when running Kitematic against a VM, the ports are forwarded:

![image](https://cloud.githubusercontent.com/assets/82988/23317016/b0725f94-fac4-11e6-8e07-f3c1d99db47a.png)

In this case, it would be more user friendly if the status message just said to post the token - and then give the token - into the initial authentication screen text box and not mention the localhost / port 8888 URL combination.

So rather than:

    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://localhost:8888/?token=NNNNN

maybe say something like:

    Enter the following token at the initial sign-in to the notebook:
        NNNN
    Alternatively, if the server is running on the local host without port forwarding, visit the URL:
        http://localhost:8888/?token=NNNNN

takluyver:
[...]

PR #2666, opened 2017-07-19
Expose URL with auth token in notebook UI

takluyver:
This provides a crude way to copy the link from one browser (right click the Jupyter logo) to get authenticated in another browser.

Closes gh-2094

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -118,7 +118,10 @@ dir="ltr">
 
 <div id="header">
   <div id="header-container" class="container">
-  <div id="ipython_notebook" class="nav navbar-brand pull-left"><a href="{{default_url}}?token={{token}}" title='{% trans %}dashboard{% endtrans %}'>{% block logo %}<img src='{{static_url("base/images/logo.png") }}' alt='Jupyter Notebook'/>{% endblock %}</a></div>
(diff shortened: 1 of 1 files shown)


notebook/templates/page.html
```diff
@@ -118,7 +118,7 @@
 
 <div id="header">
   <div id="header-container" class="container">
-  <div id="ipython_notebook" class="nav navbar-brand pull-left"><a href="{{default_url}}" title='{% trans %}dashboard{% endtrans %}'>{% block logo %}<img src='{{static_url("base/images/logo.png") }}' alt='Jupyter Notebook'/>{% endblock %}</a></div>
+  <div id="ipython_notebook" class="nav navbar-brand pull-left"><a href="{{default_url}}?token={{token}}" title='{% trans %}dashboard{% endtrans %}'>{% block logo %}<img src='{{static_url("base/images/logo.png") }}' alt='Jupyter Notebook'/>{% endblock %}</a></div>
```
minrk: This exposes the login token on the login page, so unauthenticated users can just click this link to login. We should make sure to only include this if the page being served is already authenticated, and only when the token is defined.

takluyver: Good job one of us is thinking about this! ;-)

takluyver:
Should be fixed now. I'm using the same conditions (`logged_in and token`) as we already use for putting  `data-jupyter-api-token` in the body tag.

PR #2671, opened 2017-07-20
avoid modifying settings['headers'] in add_default_headers

minrk:
Use a copy to avoid writing content security policy into settings['headers'], which can be a problem because APIHandlers have a stricter CSP than page handlers, adding `default-src: 'none'`.

If an API request is made before the first page request, pages would fail to load due to CSP violations because the API CSP would be saved in `settings['headers']` and used for all subsequent requests.

cc @tkinz27 @kalvinnchau who reported this on Gitter


takluyver:
Seems straightforward, thanks.

PR #2672, opened 2017-07-20
Render the logout page without auth tokens

takluyver:
The user is technically still logged in when the logout page is produced, but this is not obvious to the user, and they may leave the page open.

This ensures that it is rendered without information such as auth tokens which should only be available to a logged-in user.


minrk:
Thanks!

PR #2674, opened 2017-07-20
ensure "default-src 'none'" CSP is added to APIHandlers

minrk:
even if custom Content-Security-Policy header is specified, which was previously setting the same value for both APIHandlers and page handlers.


takluyver:
I don't feel like I understand this well enough to review it. @rgbkrk , do you have time to look at it?

rgbkrk:
That is much improved.

takluyver:
Super :-)

Issue #2590, opened 2017-06-21
Clicking left-margin to collapse output no longer works


takluyver:
I just noticed that clicking on the area to the left of output no longer collapses the output into a scroll area on master. Was this a deliberate change I missed, or an accidental regression?

gnestor:
It looks like this was introduced by https://github.com/jupyter/notebook/pull/2501.

It looks like `this.collapsed` is returning `false` and therefore `this.prompt_overlay.show()` is not being called: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/outputarea.js#L155

Fixed by https://github.com/jupyter/notebook/issues/2681 👍

PR #2681, opened 2017-07-24
Fix hidden prompt overlay button

gnestor:
Closes https://github.com/jupyter/notebook/issues/2590

It looks like `this.collapsed` is returning `false` and therefore `this.prompt_overlay.show()` is not being called: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/outputarea.js#L155

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -77,7 +77,6 @@ define([
 
     OutputArea.prototype.style = function () {
         this.collapse_button.hide();
-        this.prompt_overlay.hide();
         
         this.wrapper.addClass('output_wrapper');
         this.element.addClass('output');


gnestor:
It looks like this was introduced by https://github.com/jupyter/notebook/pull/2501.

It looks like `this.collapsed` is returning `false` and therefore `this.prompt_overlay.show()` is not being called: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/outputarea.js#L155

@minrk Mind taking a look at this? Just wanna make sure that this doesn't behave unexpectedly in some edge case.

Carreau:
Let's try it to see if it has side effects.

minrk:
I think this is right. I saw this when I did my PR, but then forgot to investigate.

jasongrout:
FYI, this breaks the `prompt_area: false` argument to the output area constructor. We're seeing this in ipywidgets, where we explicitly set that to `false`, but now the prompt area is still shown.

For now, we'll just hardcode some css to hide the prompt area in output widgets. I think it would be enough to check `this.prompt_area` and if it is false, explicitly hide the prompt area as before.

CC https://github.com/jupyter-widgets/ipywidgets/issues/1765

jasongrout:
Proposed fix: https://github.com/jupyter/notebook/pull/2961

Issue #2207, opened 2017-02-20
Undo after merge operation results in loss of cell


adelq:
After merging cells with `shift-m`, the expected behavior for undo (`C-z`) is to split the cells back apart again.

However, the actual displayed behavior is that the newly merged cell's code is removed, with no new cells created. Thus, the code that was merged in is lost.

So, 2 cells are merged into 1 with the code of both, undo results in a single cell that has the code of just 1 cell.

takluyver:
We don't have a great undo mechanism at the moment. Each cell has its own undo history, and this is what's deleting the code that came from the second cell. We also have a simple undo mechanism to restore deleted cells. I think this should be available in the edit menu after you merge cells.

Carreau:
We can clear CM history on merge, it will prevent undo beyond merge point.

takluyver:
@Carreau you milestoned it for 5.0 - would you like to work on it?

Carreau:
Bumping to 5.1, even if Ctrl-Z undo Ctrl-Shift-Z should redo so that's not critical. 
It should be a 1 line fix in the right place (likely the Merge command). But it will take me an afternoon to re-dive in this code.

gnestor:
@Carreau What 1 line are you thinking of? Here is the `merge_cells` function: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/notebook.js#L1740-L1787

Carreau:
Yes, if you call, [`doc.clearHistory()`](https://codemirror.net/doc/manual.html#clearHistory), it should prevent to Ctrl-Z to before the merge to an empty state.

PR #2684, opened 2017-07-25
Reset the target cell's undo history after merge_cells

gnestor:
Closes https://github.com/jupyter/notebook/issues/2207

This will prevent the scenario where a user merges one or more cells, does an undo, types something, and loses the contents of the other cells that were merged into this cell.


Carreau:
+1 thanks !

Issue #2673, opened 2017-07-20
Huge amount of requests to favicon-busy.ico


sasdf:
Kernel switch between idle and busy status very quickly while using interactive plot(`%pylab notebook`). It will generate many http requests to `favicon-busy.ico`.  Even just move the cursor over a interactive plot can generate about 40 requests in a second.  This kind of network traffic may trigger some firewall system because it looks like we're attacking the server.

This behavior was found in current pip version, but not in the version on tmpnb.
I install jupyter by `pip3 install jupyter --user`

takluyver:
Yikes, I had assumed the browser would cache that, but it looks like it makes a new request each time the favicon changes. Even worse, in master we're doing an animation in Javascript, so it repeatedly requests `favicon-busy-1.ico` and `favicon-busy-3.ico` in alternation while the kernel is busy.

takluyver:
#2675 removes the animation part of it. For the main part of it, I see a few possible options:

a. Get rid of the busy favicon change entirely
b. Debounce it, so it only changes to busy if it's been busy for e.g. 1 second.
c. Find some way to change it without causing new HTTP requests, e.g. using data URLs.

takluyver:
#2687 aims to mitigate this by only changing the icon when execution lasts > 1 second.

PR #2687, opened 2017-07-26
Only show busy favicon when execution lasts > 1s

takluyver:
This avoids making lots of HTTP requests as we rapidly change the favicon between the 'notebook' (idle) and 'busy' icons. Now it should be no more than one per second in the most pathological case.

Closes gh-2673


gnestor:
Thanks @takluyver!

PR #2688, opened 2017-07-26
Fix jitter caused by rounding height.

### LINKED ISSUE #2668: Output area height is rounded when fixing height on clear
In https://github.com/jupyter/notebook/blob/5192d72c63369d7167c5472d06f87072d68ecc53/notebook/static/notebook/js/outputarea.js#L969, we use jquery to find and fix the height of an output area in order to prevent jumpiness when clearing an output area. However, the jquery `.height()` function rounds the height to the nearest integer, which causes some jitter. It would be better to use `.getBoundingClientRect().height`, which doesn't round the height.

jasongrout:
Fixes #2668.

Before height rounding fix (notice that the cell below the interact jitters its height):
jitter

After:
jitterfixed

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -966,7 +966,8 @@ define([
             // Fix the output div's height if the clear_output is waiting for
             // new output (it is being used in an animation).
             if (!ignore_clear_queue && this.clear_queued) {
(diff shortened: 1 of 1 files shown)


jasongrout:
CC @gnestor

gnestor:
Thanks @jasongrout!

PR #2689, opened 2017-07-26
Don't assume that application/ mime types are editable

takluyver:
The application media type includes many common binary files like `application/zip` and `application/pdf`. I don't think we can assume it's editable.

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -30,9 +30,11 @@ define([
     var includes_extension = function(filepath, extensionslist) {
       return item_in(extension(filepath), extensionslist);
     };
-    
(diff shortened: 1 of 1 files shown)


rgbkrk:
Except for application/json

takluyver:
We're already whitelisting several relevant extensions, like `.json`, `.jsonld` and `.geojson`. Is that likely to be enough, or do you want to whilelist `application/*+json` as well?

takluyver:
I've included a check for `+json` and `+xml` mimetypes.

rgbkrk:
> Is that likely to be enough, or do you want to whilelist application/*+json as well?

That sounds good, though it looks like you've already added it. 😄

PR #2691, opened 2017-07-26
Restore flat edge of rendered html code block

invisibleroads:
The extra padding causes the first line of each markdown rendered code block to be indented, but does not indent the rest of the code block, resulting in a jagged left edge. Removing this bit of CSS restores a flat left edge.

You can verify this by trying to render the following:

    abc
        def
        ghi

This fix will revert 533aea11fdcb3a358add0ae4cbe41c97dac2cf2c.

screenshot from 2017-07-26 15-39-22

### DIFF
--- a/notebook/static/notebook/less/renderedhtml.less
+++ b/notebook/static/notebook/less/renderedhtml.less
@@ -52,7 +52,6 @@
 
     code {
         background-color: #eff0f1;
-        padding: 1px 5px;
     }
 
     pre code {background-color: @body-bg;}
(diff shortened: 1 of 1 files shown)


gnestor:
Thanks for your contribution @invisibleroads! Can you please check that "Allow edits from maintainers" box in the right column or add this CSS rule to renderedhtml.less?

```
p code {
    padding: 1px 5px;
}
```

This will preserve the padding for inline code elements (and not code block elements).

invisibleroads:
Now I understand that you were trying to add padding to inline code in rendered markdown.

gnestor:
Thanks! Will merge as soon as CI build passes 👍

Issue #1976, opened 2016-12-14
Unhelpful error when selecting unavailable CodeMirror keyMap


takluyver:
I was helping someone configure their CodeMirror keymap, with `~/.jupyter/nbconfig/notebook.json`:

```json
{
  "CodeCell": {"cm_config": {"keyMap": "emacs"}}
}
```

The result was:

![screenshot from 2016-12-14 15-51-13](https://cloud.githubusercontent.com/assets/327925/21189030/3afe875c-c215-11e6-869a-640664f8c175.png)

The error tracks back to [this line](https://github.com/codemirror/CodeMirror/blob/5.21.0/src/edit/options.js#L92). I think we should find some way to check that the keyMap exists before trying to set it.

takluyver:
This error also prevents the notebook from loading - config stuff should generally be more robust against simple bad values.

gnestor:
@takluyver Good catch! Can you take a stab at validating keymaps and throwing a more informative error message if it fails?

PR #2699, opened 2017-07-27
Validate CodeMirror keyMap from frontend config

takluyver:
@gnestor Unfortunately I can't see a general way to validate the config for CodeMirror, so I think we just have to check individual things.

The [CodeMirror docs](https://codemirror.net/doc/manual.html#config) say:

> Options are not checked in any way, so setting bogus option values is bound to lead to odd errors.

Closes gh-1976


gnestor:
Good enough! Thanks!

Issue #2676, opened 2017-07-20
Extra newline between bulleted list sections with links/backquote


cstjean:
I'm a heavy bulleted list user, and they often look like this:

![screen shot 2017-07-20 at 10 10 11](https://user-images.githubusercontent.com/8622776/28421649-c135d1f8-6d33-11e7-8e61-a1403029526f.png)

Is there a work-around to remove the newlines?

takluyver:
It's not a real newline, it's an additional top margin. The relevant rule seems to be from `renderedhtml.less`:

```less
* + ul {margin-top: 1em;}
```

I'm not quite sure what this rule is meant to do; it looks like it could just be removed. @gnestor @minrk any ideas?

gnestor:
@cstjean I can't reproduce. Can you share a snippet of markdown?

The following Markdown:

```
# A bullet list

* test
  * 2nd level
    * 3rd level
* test quotes
  * > a quote
  * not a quote
* test quotes
  * [a link](jupyter.com)
  * not a link
* test code
  * `some code`
```

renders:

![image](https://user-images.githubusercontent.com/512354/28643804-eb81ea28-720b-11e7-9dca-0bcc8968f47a.png)

cstjean:
It's when the markdown is in the "parent" bullet
```
- [A link](a)
 * a link
- `Some code`  
 * some code
- $latex too$
 * a
```

takluyver:
I was able to reproduce it with `inline code` in the parent bullet.

gnestor:
Ok, I see now. https://github.com/jupyter/notebook/pull/2701 fixes this 👍

PR #2701, opened 2017-07-27
Prevent extra newline between nested lists

gnestor:
Closes https://github.com/jupyter/notebook/issues/2676


takluyver:
Seems to work, though I worry that we're playing CSS whack-a-mole.

gnestor:
Gotta love CSS 🤗

This one actually makes sense, I don't think we'll encounter unexpected consequences here...

PR #2704, opened 2017-07-28
Ensure NbserverListApp produces valid json even for multiple/no servers

jcb91:
This PR makes a small adjustment to the NbserverListApp class to ensure valid json output.

Currently, the NbserverListApp outputs a json object for each running server, without wrapping them in a list. This means that if there are no running servers, there is no output, and if multiple servers are running, the response is invalid json.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -389,18 +389,28 @@ class NbserverListApp(JupyterApp):
     description=_("List currently running notebook servers.")
     
     flags = dict(
+        jsonlist=({'NbserverListApp': {'jsonlist': True}},
+              _("Produce machine-readable JSON list output.")),
         json=({'NbserverListApp': {'json': True}},
-              _("Produce machine-readable JSON output.")),
+              _("Produce machine-readable JSON object on each line of output.")),
     )
     
+    jsonlist = Bool(False, config=True,
+          help=_("If True, the output will be a JSON list of objects, one per "
+                 "active notebook server, each with the details from the "
+                 "relevant server info file."))
     json = Bool(False, config=True,
           help=_("If True, each line of output will be a JSON object with the "
-                  "details from the server info file."))
+                  "details from the server info file. For a JSON list output, "
+                  "see the NbserverListApp.jsonlist configuration value"))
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -398,12 +398,12 @@ class NbserverListApp(JupyterApp):
...
     def start(self):
-        if not self.json:
-            print(_("Currently running servers:"))
-        for serverinfo in list_running_servers(self.runtime_dir):
-            if self.json:
-                print(json.dumps(serverinfo))
-            else:
+        serverinfo_list = list(list_running_servers(self.runtime_dir))
+        if self.json:
+            print(json.dumps(serverinfo_list))
```
takluyver: Maybe pass `indent=2` so that the JSON is nicely formatted if a human needs to look at it?

takluyver:
Correction: it's been there since IPython 2.0, released in April 2014. See ipython/ipython#4772. We discussed making it single JSON array then, but I was lazy (and apparently just dumping the whole list as JSON didn't occur to me :confused: ).

I think we should make this behaviour a new option, and then maybe go through a deprecation cycle with `--json`. Even if we think no-one is using the old behaviour, we can't be sure, and I think we should take backwards compatibility a bit more seriously than we do.

jcb91:
Sounds reasonable to me, I'll update this to add a separate option when I get a chance

notebook/notebookapp.py
jcb91: :+1:

takluyver:
Thanks for being so understanding about it! :-)

jcb91:
ok, I've added a (new) separate option `jsonlist`. For now, this is used in preference to the plain `json` flag if both are specified; is that ok, or would you rather they were made mutually exclusive, like the `--sys`, `--user` and `--sys-prefix` flags of `jupyter nbextension` and `jupyter serverextension`?

takluyver:
Thanks, that looks good to me. I'm not worried about the behaviour when both are specified.

I'll let @gnestor have a look - if you're happy, feel free to merge.

PR #2706, opened 2017-07-29
Bug Fix: use the nbconvert.exporters.base instead of deprecated expor…

gabyx:
This is a bug fix, hopefully correct. the variable exporter_map is deprecated and get_exporter should be used, to make availabe also exporters registered by entrypoints.


mpacer:
This looks right to me…I want to give time for others to look this over and I'll merge Tuesday if no one else has issues with it.

Issue #2499, opened 2017-05-17
once-only events fired before custom.js is loaded


mpacer:
Earlier in https://github.com/jupyter/notebook/issues/1730 we had an issue where the point at which `custom.js` was loaded was delayed so that standard triggers wouldn't work. It seems that with notebook 5.0 this bug was reintroduced.

This is a problem for scripts using `notebook_loaded.Notebook` as well as `app_initialized.NotebookApp` (likely the cause of https://github.com/jupyter/notebook/issues/2403). 

I don't know if https://github.com/jupyter/notebook/pull/1732 give an idea for a similar solution — but if it should be a Promise, my guess is it needs a slightly different kind of fix. Happy to attempt a fix if someone wants to point me in the right direction; seems like a good way to learn about Promises.

mpacer:
this is the test `custom.js` that I was using to test stuff. 

```javascript
// file ~/custom/custom.js

console.log("hello1");

require([
	"base/js/namespace",
	"base/js/events"
], 
	function(Jupyter, events){
		console.log("hello2");	
		events.on('notebook_loaded.Notebook', function(){
			console.log("hello3");
		});
		events.on('app_initialized.NotebookApp', function(){
			console.log("hello4");
		});
});

console.log("hello5");
```

rolweber:
You're right. From a brief look, I'd say that of all the changes in #1732, only these are still in place:

* `notebook/templates/page.html`
* `tools/build-main.js`

while these are missing:

* `notebook/static/edit/js/main.js`
* `notebook/static/notebook/js/main.js`
* `notebook/static/terminal/js/main.js`
* `notebook/static/tree/js/main.js`

You could work from the [file diff](https://github.com/jupyter/notebook/pull/1732/files) to re-apply the missing changes. If you have time to spend and/or want to learn git, you could also examine the history to see where they got lost :-)

takluyver:
Looks like they were changed back in #1378, because we don't want a bad custom.js to prevent the notebook from loading.

I think the longer term fix is that rather than fiddling with the loading order, we should have promises for one-time events like `app_initialized.NotebookApp`, so that custom code and extensions can wait for something that may already have occurred.

rolweber:
@mpacer: I suggest to rename this issue. Something like "once-only events fired before custom.js is loaded", maybe.

I'm not firm with Promises either, so the following are really just pointers. I like this introduction: https://developers.google.com/web/fundamentals/getting-started/primers/promises

[...]

PR #2710, opened 2017-07-31
Add a promise for app_initialized event.

takluyver:
See gh-2499

The idea is that custom.js code which was previously listening for `app_initialized.NotebookApp` events can do this instead:

```javascript
define([
        'base/js/namespace',
        'base/js/promises'
     ], function(Jupyter, promises) {
         promises.app_initialized.then(function(appname) {
             if (appname === 'NotebookApp') {
                 do_stuff()
             }
         });
     });
```

I've milestoned this for 5.1 so that we consider it, as I think this is the answer to a long-running annoyance with the event firing before the code listens for it. But I don't mind if we decide to push it back to a later release.


minrk:
Seems like a good solution to me.

gnestor:
Looks good to me. Should we update the custom.js docs?

minrk:
@gnestor good call. I pushed an update to custom.js

takluyver:
Thanks - I was just about to look at it, so you saved me a few minutes. :-)

PR #2715, opened 2017-08-01
Add "all" flag to nbextension and serverextension commands

gnestor:
@takluyver @minrk I'm not sure whether this is the right approach... As opposed to "check every level where the enable config is defined", this is just adding an "all" trait to the top-level class and some logic to the `start` method that will set `user` and `sys_prefix` to `True` if the all flag is enabled. This requires all subclass instances to call `super(CommandInstance, self).start()`.

Closes https://github.com/jupyter/notebook/issues/2149

### DIFF
--- a/notebook/extensions.py
+++ b/notebook/extensions.py
@@ -58,6 +58,7 @@ class BaseExtensionApp(JupyterApp):
 
     user = Bool(False, config=True, help="Whether to do a user install")
     sys_prefix = Bool(False, config=True, help="Use the sys.prefix as the prefix")
+    all = Bool(False, config=True, help="Whether to do both a user and sys.prefix install")
     python = Bool(False, config=True, help="Install from a Python package")
 
     # Remove for 5.0...
@@ -71,6 +72,11 @@ class BaseExtensionApp(JupyterApp):
     def _log_format_default(self):
         """A default format for messages"""
         return "%(message)s"
+        
+    def start(self):
+        if self.all:
+            self.user = True
+            self.sys_prefix = True
 
 def _get_config_dir(user=False, sys_prefix=False):
     """Get the location of config files for the current context
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -671,6 +671,8 @@ class InstallNBExtensionApp(BaseExtensionApp):
 
     def start(self):
         """Perform the App's function as configured"""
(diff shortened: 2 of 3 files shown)


takluyver:
I think it's just going to end up raising [this error](https://github.com/jupyter/notebook/blob/9f5926ec14f504547a711db3df41592740a5501d/notebook/nbextensions.py#L1001) in some cases, and not doing what you expect in others. The machinery is designed to select one location to make a change, not a set of locations.

Rather than trying to add this for all commands, I'd focus on the `disable` command, and maybe `uninstall`. It doesn't make much sense to install or enable an extension in all locations.

There are questions over what it should do if you don't have write permissions to all the directories it wants to change (there are system config locations, so that will be common if you run it as a regular user). Should it always error out in that case, or only if it would change a non-writable file? Should it fail quickly, or make any changes it can before throwing an error?

minrk:
I don't think setting sys_prefix and user both to True will do the right thing, because only one is triggered in the end, unless something changed. I thought this would through a mutually-exclusive flag error.

I think you'll need to handle `all` specially, and I see two options:

1. Follow pip's example, and remove the first one you see in the priority order.
2. go through them all in order.

Option 1. seems like the more logical behavior, to me. It has the potential disadvantage (like pip) that you have to re-run it until it shows nothing installed to be sure that it's truly gone if you have something enabled/installed at multiple levels. But it has the advantage that it only does one action.

Option 2. might better address the specific use case @ellisonbg brings up in #2149, where the real goal is "scrub every trace of this and start from scratch". However, 'disable' is not quite the right action for that use case, because `disable` sets an explicit `False` flag (thus overriding lower-priority config), whereas scrubbing every trace should pop values from config, not set them to False. That makes "remove every last trace of this" a somewhat distinct action from disable.

This ambiguity doesn't affect `uninstall`, though.

It seems like what would be really nice is if uninstalling an extension also removed enabling configuration, but that's getting pretty complicated.

Per @takluyver's comment about errors, I think it should raise any PermissionErrors it encounters.

takluyver:
> I think you'll need to handle all specially, and I see two options:
>
> 1. Follow pip's example, and remove the first one you see in the priority order.

I've had a go at this in #2725 .

> 2. go through them all in order.

And this in #2724 .

gnestor:
Thanks @takluyver! I'm going to close this one...

PR #2718, opened 2017-08-01
Allow notebooks to be trusted without triggering a save

### LINKED ISSUE #195: trust from menu not working for readonly ipy notebook
I use jupyter hub for multi user and open another notebook server for readonly sharing.
there are js code in output cell that isn't executing. I click trust menuitem but not working. 
however, running `ipython trust xyz.ipynb` works.

Madhu94:
Closes #195. 

Trust should now add the notebook's signature without needing to trigger a save. Save will still trust the notebook, after checking and if it has not already been trusted. 

I have not added an automatic refresh after trust, let me know if you think this would be a good idea.

### DIFF
--- a/notebook/services/contents/handlers.py
+++ b/notebook/services/contents/handlers.py
@@ -307,6 +307,17 @@ class NotebooksRedirectHandler(IPythonHandler):
     put = patch = post = delete = get
 
 
+class TrustNotebooksHandler(IPythonHandler):
+    """ Handles trust/signing of notebooks """
+
+    @json_errors
+    @web.authenticated
+    @gen.coroutine
+    def post(self,path=''):
+        cm = self.contents_manager
+        yield gen.maybe_future(cm.trust_notebook(path))
+        self.set_status(201)
+        self.finish()
 #-----------------------------------------------------------------------------
 # URL to handler mappings
 #-----------------------------------------------------------------------------
@@ -318,6 +329,7 @@ default_handlers = [
     (r"/api/contents%s/checkpoints" % path_regex, CheckpointsHandler),
     (r"/api/contents%s/checkpoints/%s" % (path_regex, _checkpoint_id_regex),
         ModifyCheckpointsHandler),
+    (r"/api/contents%s/trust" % path_regex, TrustNotebooksHandler),
     (r"/api/contents%s" % path_regex, ContentsHandler),
     (r"/api/notebooks/?(.*)", NotebooksRedirectHandler),
 ]
--- a/notebook/services/contents/manager.py
+++ b/notebook/services/contents/manager.py
@@ -426,7 +426,7 @@ class ContentsManager(LoggingConfigurable):
         nb = model['content']
         self.log.warning("Trusting notebook %s", path)
         self.notary.mark_cells(nb, True)
-        self.save(model, path)
+        self.check_and_sign(nb, path)
 
     def check_and_sign(self, nb, path=''):
         """Check for trusted cells, and sign the notebook.
@@ -443,7 +443,7 @@ class ContentsManager(LoggingConfigurable):
         if self.notary.check_cells(nb):
             self.notary.sign(nb)
         else:
-            self.log.warning("Saving untrusted notebook %s", path)
+            self.log.warning("Notebook %s is not trusted", path)
 
     def mark_trusted_cells(self, nb, path=''):
         """Mark cells as trusted if the notebook signature matches.
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2876,10 +2876,24 @@ define([
                                 cell.output_area.trusted = true;
                             }
                         }
-                        nb.events.on('notebook_saved.Notebook', function () {
-                            window.location.reload();
+                        // If its write only and dirty, save before 
+                        // trusting
+                        var pr;
+                        if(nb.writable && nb.dirty) {
+                            pr = nb.save_notebook();
+                        }
+                        else {
+                            pr = Promise.resolve();
+                        }
+                        return pr.then(function() {                            
+                            nb.contents.trust(nb.notebook_path)
+                            .then(function(res) {
+                                nb.events.trigger("trust_changed.Notebook", true);
+                                window.location.reload();
+                            }, function(err) {
+                                console.log(err);
+                            });
                         });
-                        nb.save_notebook();
                     }
                 }
             }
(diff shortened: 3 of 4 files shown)


notebook/services/contents/filemanager.py
```diff
@@ -438,7 +438,10 @@ def save(self, model, path=''):
         try:
             if model['type'] == 'notebook':
                 nb = nbformat.from_dict(model['content'])
-                self.check_and_sign(nb, path)
+                # We need this check in case the notebook is 
+                # trusted before save
+                if not self.notary.check_signature(nb):
+                    self.check_and_sign(nb, path)
```
minrk: check_signature shouldn't be called before `check_and_sign`, which consumes trust metadata. I think `.save` probably doesn't need to change at all.

notebook/services/contents/handlers.py
```diff
@@ -307,6 +307,17 @@ def get(self, path):
     put = patch = post = delete = get
 
 
+class TrustNotebooksHandler(IPythonHandler):
+    """ Handles trust/signing of notebooks """
+
+    @json_errors
+    @web.authenticated
+    @gen.coroutine
+    def put(self,path=''):
```
minrk: POST is probably more appropriate for a request with no content.

notebook/services/contents/manager.py
```diff
@@ -443,7 +443,7 @@ def check_and_sign(self, nb, path=''):
         if self.notary.check_cells(nb):
             self.notary.sign(nb)
         else:
-            self.log.warning("Saving untrusted notebook %s", path)
+            self.log.warning("Signing notebook %s", path)
```
minrk: This message is backwards. In this event, the notebook is *not* signed. It is still untrusted.

notebook/static/notebook/js/notebook.js
```diff
@@ -2876,10 +2876,12 @@ define([
                                 cell.output_area.trusted = true;
                             }
                         }
-                        nb.events.on('notebook_saved.Notebook', function () {
-                            window.location.reload();
+                        nb.contents.trust(nb.notebook_path)
+                            .then(function(res) {
+                                nb.events.trigger("trust_changed.Notebook", true)
```
minrk: it should still trigger reload to get the newly trusted notebook without sanitized output.

And `.save` should only be avoided when the notebook is read-only. We should still save & reload for non-read-only notebooks to avoid losing unsaved changes (which is unavoidable for read-only).

Madhu94:
@minrk  Thanks for the feedback!

notebook/services/contents/manager.py
```diff
@@ -443,7 +443,9 @@ def check_and_sign(self, nb, path=''):
         if self.notary.check_cells(nb):
             self.notary.sign(nb)
         else:
-            self.log.warning("Saving untrusted notebook %s", path)
```
minrk: Maybe leave this at "Notebook %s is not trusted."

notebook/static/notebook/js/notebook.js
```diff
@@ -2876,10 +2876,13 @@ define([
...
-                        nb.events.on('notebook_saved.Notebook', function () {
-                            window.location.reload();
+                        nb.contents.trust(nb.notebook_path)
+                            .then(function(res) {
+                                nb.events.trigger("trust_changed.Notebook", true);
+                                window.location.reload();
+                            }, function(err) {
+                                console.log(err);
                         });
-                        nb.save_notebook();
```
minrk: Can you still trigger save here if the notebook is not read-only? That way we can avoid losing unsaved changes when the notebook is editable.

```python
var p;
if (nb.writable and nb.dirty) {
    p = nb.save_notebook();
} else {
    p = Promise.resolve();
}
p.then(function () {
    return nb.contents.trust...
```

Madhu94: So if its read-only, we save it and then trust..Thanks, never noticed the ``dirty`` attribute.

Madhu94:
@minrk  In addition to the changes you requested, I added [this](https://github.com/Madhu94/notebook/blob/trust_without_save/notebook/services/contents/manager.py#L448), I think the check makes sense here.

gnestor:
@minrk Do you think this ready to merge?

notebook/services/contents/manager.py
```diff
@@ -440,10 +440,15 @@ def check_and_sign(self, nb, path=''):
         path : string
             The notebook's path (for logging)
         """
+        # Warning: The check_cells method removes the 
+        # "trusted"" key from the cells' metadata 
         if self.notary.check_cells(nb):
-            self.notary.sign(nb)
+            # We're checking the signature *after* the
+            # trusted keys have been removed
+            if not self.notary.check_signature(nb):
```
minrk: I think this entire method should be left exactly as it was. Adding this check here is not correct because it prevents updating the signatures of trusted notebooks.

minrk:
@gnestor almost. There's one addition in check_and_sign that should be reverted, but the rest should be set.

notebook/services/contents/manager.py
Madhu94: @minrk The way ``store_signature`` for the signature store is implemented, the signatures _don't_ get updated. 
I logged an issue with nbformat, let me know if I'm wrong [here](https://github.com/jupyter/nbformat/issues/99)

Madhu94:
@minrk The way ``store_signature`` for the signature store is implemented, the signatures _don't_ get updated. 
I logged an issue with nbformat, let me know if I'm wrong [here](https://github.com/jupyter/nbformat/issues/99)

Madhu94:
I think this is a bug even on master. I noticed that without the check, the same signature was re-inserted each time the notebook was trusted. 

@minrk, I probably shouldn't check this here - the issue should be fixed in nbformat.  So, I'll revert that change, squash my commits and then we're good to go?

minrk:
@Madhu94 exactly right. This shouldn't be changed here, but fixed in nbformat itself. Thanks for digging in!

Madhu94:
@minrk I rebased and squashed my commits and reverted that check. Let me know if there's anything else left?

takluyver:
@minrk this is, I think, the last remaining PR for 5.1. Is it ready to go?

minrk:
Yes, indeed.

minrk:
Thanks, @Madhu94!

Madhu94:
Can I add this to the 5.1 changelog ?

takluyver:
Yes, please do!

PR #2724, opened 2017-08-04
--all flag for jupyter nbextension disable

takluyver:
See #2149, #2715.

This roughly implements option 2 from [Min's comment](https://github.com/jupyter/notebook/pull/2715#issuecomment-319687449), which is what I'd expect from `--all`.

The alternative is to disable in only the first location where it is found enabled. I might even make that the default behaviour if we do that.

### DIFF
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -844,6 +844,13 @@ class DisableNBExtensionApp(ToggleNBExtensionApp):
     """
     _toggle_value = None
 
+    flags = _base_flags.copy()
+    flags['all'] = ({
+        "DisableNbExtensionApp" : {
+            "all_config_locations" : True,
+        }}, "Go through all config locations to disable loading an extension"
+    )
+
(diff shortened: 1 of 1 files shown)


notebook/nbextensions.py
```diff
@@ -833,6 +844,35 @@ class DisableNBExtensionApp(ToggleNBExtensionApp):
...
+    )
+
+    all_config_locations = Bool(False, config=True,
+        help="Go through all config locations to disable loading an extension",
+    )
+
+    def toggle_nbextension_python(self, module):
+        if self.all_config_locations:
+            for config_dir in jupyter_config_path():
+                disable_nbextension_python(self.section, module,
```
minrk: We should make sure this does:

```python
if enabled at this level:
   set enabled = None # (not False. None pops config, False stores a value)
else:
   do nothing
```

So that it doesn't try to write any configuration files where there wasn't any to start with.

takluyver:
Closing in favour of #2729 .

PR #2725, opened 2017-08-04
Disable nbextension in first location found

takluyver:
Alternative to #2724.

This implements option 1 from [Min's comment](https://github.com/jupyter/notebook/pull/2715#issuecomment-319687449): find the first location where the config enables the extension, and remove it. This makes it the default, and simplifies the options available, on the grounds that 99% of the time, when you're asking to disable something, you don't care which bit of config is enabling it.

It adds some code because it bypasses the inheritance tree of applications that has been built. Looking at this code, I think we've over-abstracted the actual operations a bit, which makes it less clear what's going on. I'd like to flatten it a bit, but not in this PR.

### DIFF
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -243,10 +243,7 @@ def uninstall_nbextension(dest, require=None, user=False, sys_prefix=False, pref
     ----------
     
     dest : str
-        path to file, directory, zip or tarball archive, or URL to install
-        name the nbextension is installed to.  For example, if destination is 'foo', then
-        the source file will be installed to 'nbextensions/foo', regardless of the source name.
-        This cannot be specified if an archive is given as the source.
+        Name of the installed nbextension file or folder.
     require : str [optional]
         require.js path used to load the extension.
         If specified, frontend config loading extension will be removed.
@@ -279,6 +276,25 @@ def uninstall_nbextension(dest, require=None, user=False, sys_prefix=False, pref
     if require:
         for section in NBCONFIG_SECTIONS:
             cm.update(section, {"load_extensions": {require: None}})
+
+def _find_uninstall_nbextension(filename, logger=None):
+    """Remove nbextension files from the first location they are found.
+
+    Returns True if files were removed, False otherwise.
+    """
+    filename = cast_unicode_py2(filename)
+    for nbext in jupyter_path('nbextensions'):
+        path = pjoin(nbext, filename)
+        if os.path.lexists(path):
+            if logger:
+                logger.info("Removing: %s" % path)
+            if os.path.isdir(path) and not os.path.islink(path):
+                shutil.rmtree(path)
+            else:
+                os.remove(path)
+            return True
+
+    return False
 
 
 def uninstall_nbextension_python(module,
@@ -730,8 +746,8 @@ class UninstallNBExtensionApp(BaseExtensionApp):
         """The default config file name."""
         return 'jupyter_notebook_config'
     
-    def uninstall_extensions(self):
-        """Uninstall some nbextensions"""
+    def uninstall_extension(self):
+        """Uninstall an nbextension from a specific location"""
         kwargs = {
             'user': self.user,
             'sys_prefix': self.sys_prefix,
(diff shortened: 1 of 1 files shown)


minrk:
I like this better than #2724. It's what I would expect from disable. I would also expect `uninstall` to do the same.

notebook/nbextensions.py
```diff
@@ -831,7 +849,43 @@ class DisableNBExtensionApp(ToggleNBExtensionApp):
...
     """
-    _toggle_value = None
+    version = __version__
+    flags = {"py" : ({
+        "DisableNBExtensionApp" : {
+            "python" : True,
+        }}, "Disable an extension from a Python package name")
+    }
+    flags['python'] = flags['py']
+    aliases = {'section': 'ToggleNBExtensionApp.section'}
```
minrk: This should still inherit the --system, --user, --sys-prefix flags, etc. from `ToggleNBExtensionApp`. There's still a valid use case:

A system-wide installation enabled an extension, but you want to disable it in your own configuration. Passing `--user` in this case (which should store False, not pop config) makes sense as an override of the system defaults.

takluyver: I've added the `--user` and `--sys.prefix` flags. I don't think `--system` is necessary, since there's nothing for system config to override.

minrk:
I was in the process of writing a comment that it would be nice to not lose the inheritance then saw that you had just updated it to do exactly what I was going to ask for :)

minrk:
Oops. I was confused - the diff just moved because you added the uninstall application.

Restoring comment: It would probably be nice if we can keep the extensionapp inheritance, if only for preserving the consistency of the flags/aliases without having to duplicate it all. All of the enable flags should also be on disable (--system is missing, as is --debug).

notebook/nbextensions.py
```diff
@@ -712,37 +746,56 @@ def _config_file_name_default(self):
...
+            sys.exit("Only one nbextension allowed at a time. "
+                     "Call multiple times to uninstall multiple extensions.")
+        elif self.user or self.sys_prefix or self.prefix or self.nbextensions_dir:
             try:
-                self.uninstall_extensions()
+                self.uninstall_extension()
             except ArgumentConflict as e:
                 sys.exit(str(e))
+        else:
+            self.find_uninstall_extension()
```
minrk: 👍

notebook/nbextensions.py
```diff
@@ -831,7 +884,60 @@ class DisableNBExtensionApp(ToggleNBExtensionApp):
...
+    sys_prefix = Bool(False, config=True,
+                      help="Explicitly disable in sys.prefix config")
+
+    python = Bool(False, config=True, help="Install from a Python package")
+
+    section = Unicode('notebook', config=True,
+                      help="""Which config section to disable the extension in."""
+                      )
+
+    def start(self):
```
minrk: Could we get away with only overriding start, instead of losing the ToggleNBExtensionApp inheritance?

takluyver:
I decided there wasn't enough significant code in common between the enable & disable apps to keep a base class, so I folded the remainder of ToggleNBExtensionApp into EnableNBExtensionApp and simplified it.

gnestor:
I'd like to publish a release candidate by EOD so that people can start testing 5.1 this weekend and then try to officially release on Tuesday. Does this PR close https://github.com/jupyter/notebook/issues/2149 or is it a part of the solution? From my understanding, with or without an `--all` flag, we want to be able to disable an extension from all prefixes with a single command. Does this PR get us there? How much more would it take to get this for serverextensions, too? And do we also want the same for uninstall or does uninstall not work the same?

takluyver:
It doesn't add a `--all` flag, but I think it's a different approach to meeting the needs described in #2149: you can disable or uninstall an extension without working out where it's installed/enabled.

I'd consider saving the `--all` flag to remove/disable all extensions (rather than removing/disabling one in all locations).

It shouldn't be much more work to do it for serverextensions, but there's not much of the day left here now. I'll have a go after dinner, but I might not get it done today.

gnestor:
Ok, no worries. Let's shoot for a rc tomorrow!

takluyver:
I think the find-and-remove strategy makes sense for uninstall, but on reflection, I wonder if the default action for disable should be to set it to False in the user config directory, overriding system or installed config.

In the scenario Min described - extension enabled systemwide, user wants to disable it - running `jupyter nbextension disable foo` is the obvious thing to try, but as this PR stands, it would give you a permission error as it tried to remove the systemwide config.

takluyver:
In fact, what I'm suggesting is what `jupyter nbextension disable` already does before this PR. @minrk how would you feel about me reverting the changes to enable/disable, and just changing uninstall for now?

minrk:
> In fact, what I'm suggesting is what jupyter nbextension disable already does before this PR. @minrk how would you feel about me reverting the changes to enable/disable, and just changing uninstall for now?

That sounds great to me, actually.

takluyver:
#2729 does that. Closing this one.

PR #2729, opened 2017-08-07
Uninstall an nbextension from the first location it is found

takluyver:
Closes gh-2725 (see discussion on that PR)


minrk:
Looks great, thanks!

PR #2731, opened 2017-08-07
Add buffers argument to comm open/close

### LINKED ISSUE #2730: JS services' Comm.open/close should take buffers
Currently `services.kernel.Comm.send()` takes an optional `buffers` argument, but its `open()` and `close()` methods does not. It would be really useful for me if especially the open method accepted buffers, and I cannot see any reason that it should not. Buffers are well defined for general messages, and the ipykernel implementation supports buffers on both reception and sending of comm open messages.

In both the notebook and jupyterlab implementation of Comm.open/close, this seems to be simply a matter of adding the argument to the functions, and passing it along to the inner message send function (which already supports it). I can PR those changes if agreed.

For now I am able to work around this with the following function on the JS side:
```js
var commOpenWithBuffers = function(comm, content, callbacks, metadata, buffers) {
    return comm.kernel.send_shell_message(
        "comm_open", content, callbacks, metadata, buffers);
}
```

vidartf:
Adds an argument `buffers` to `services.kernels.comm.Comm.open()`/`close()`. It is passed through to the `send_shell_message` similarly to what is done in `Comm.send()`.

Fixes #2730.

### DIFF
--- a/notebook/static/services/kernels/comm.js
+++ b/notebook/static/services/kernels/comm.js
@@ -30,8 +30,8 @@ define([
             kernel.register_iopub_handler(msg_type, $.proxy(this[msg_type], this));
         }
     };
-    
-    CommManager.prototype.new_comm = function (target_name, data, callbacks, metadata, comm_id) {
+
+    CommManager.prototype.new_comm = function (target_name, data, callbacks, metadata, comm_id, buffers) {
         /**
          * Create a new Comm, register it, and open its Kernel-side counterpart
          * Mimics the auto-registration in `Comm.__init__` in the Jupyter Comm.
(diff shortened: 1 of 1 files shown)


stonebig:
https://github.com/jupyter/notebook/issues/2730 ?

takluyver:
Thanks. Should `CommManager.prototype.new_comm` also gain a `buffers` parameter and pass it through to `comm.open()`?

takluyver:
@minrk @gnestor I've marked this as 5.1 since it's a fairly simple change that makes sense to me, and seems unlikely to have adverse effects.

vidartf:
@takluyver I added the buffers argument to `new_comm` as well. I added it at the end to retain compatibility with old signatures, even if the grouping of arguments then gets a little different.

takluyver:
Thanks, I think that's the right thing to do.

gnestor:
Thanks @vidartf 1

Issue #2737, opened 2017-08-08
Plain text editor formatting issues


ndrewxie:
Hello,
  When I used the plain text editor (new -> plain text file) and selected the language C++, all C++ code I enter has a weird formatting, like this:
```
#include <iostream>
     #include <string>
```
It's very inconvenient and annoying to have to delete all of those tabs. Could you please make formatting language - specific (at least for the popular ones, like C, C++, Java, Javascript, Python, etc)? Same for the syntax highlighting.

takluyver:
I can reproduce this. It is supposed to be language specific - it should use the Codemirror [C like mode](http://codemirror.net/mode/clike/index.html) for C++, but something seems to be going wrong.

takluyver:
Aha, found it! It was effectively using the C mode, I think, for C++ - we need to use mime types to distinguish them.

ndrewxie:
OK thanks. Waiting for version 5.1 :smile.

PR #2739, opened 2017-08-08
Set CodeMirror mode using mimetype, not mode name

takluyver:
One mode definition file can contain different modes, e.g. clike includes C, C++, C#, Java, and a few others. We need to use the mime type to distinguish which one we want.

Closes gh-2737

@gnestor can we sneak this fix into 5.1?


gnestor:
Sure!

Issue #2740, opened 2017-08-08
Shortcut Key in File Editor to Block Comment


cancan101:
Many text editors (see for example [Sublime](https://stackoverflow.com/a/11443903/2638485)) have a shortcut key that block comments the selected lines of code. Given the file editor has as language selection, this language can be used to properly block comment. ie in Python this would just be prepending each line in the selection with a `#`.

gnestor:
It looks like the file editor defaults to non-Sublime keymaps (which is not the case for the notebook). If you simply switch to Sublime Text keymaps from the Edit menu, then block comment shortcut will work.

cancan101:
Does the choice get persisted across editors?if not is there a way to make the choice persistent? Perhaps keybindings should be made to default the same for notebook and editor?

gnestor:
Yes it does

gnestor:
Closing this for now... 👍

cancan101:
Does it make sense to change the editor default for consistency?

gnestor:
I think it does! Will submit a PR momentarily...

gnestor:
Notebook 5.1.0rc1 is available on PyPI so please give it a try and confirm that this is resolved 👍:

`pip install notebook --pre --upgrade`

PR #2745, opened 2017-08-10
Make "extraKeys" consistent for notebook and editor

gnestor:
Closes https://github.com/jupyter/notebook/issues/2740

### DIFF
--- a/notebook/static/edit/js/editor.js
+++ b/notebook/static/edit/js/editor.js
@@ -78,10 +78,11 @@ function(
         extraKeys: {
             "Tab" :  "indentMore",
         },
+        keyMap: 'sublime',
         indentUnit: 4,
         theme: "ipython",
         lineNumbers: true,
-        lineWrapping: true,
+        lineWrapping: true
     };
     
     Editor.prototype.load = function() {
@@ -170,7 +171,7 @@ function(
             Editor: {
                 file_extension_modes: update_mode_map,
             }
-        })
+        });
     };
 
     Editor.prototype.get_filename = function () {
--- a/notebook/static/edit/js/editor.js
+++ b/notebook/static/edit/js/editor.js
@@ -78,7 +78,6 @@ function(
         extraKeys: {
             "Tab" :  "indentMore",
         },
-        keyMap: 'sublime',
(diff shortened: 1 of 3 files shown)


Carreau:
I have no objections to changing the keymap, though that would be inconsistent with the notebook that default to "default" command amp. Would it be better to be consistent and change notebook as well ?

I also would try to get  a wider input on if we should change as it may break user habits.

gnestor:
> Would it be better to be consistent and change notebook as well ?

It seems like sublime is the default for the notebook (e.g. if I <kbd>CMD + /</kbd> it will comment out a block).

Carreau:
> It seems like sublime is the default for the notebook (e.g. if I CMD + / it will comment out a block).

Well no, the default is "default", the Ctrl-/ shortcut is set [here](https://github.com/jupyter/notebook/blob/f8d4d6ad2dcaaed68851483db1a4efe742b52278/notebook/static/notebook/js/cell.js#L139-L140)

gnestor:
Ahhh! Ok, we don't need to make sublime the default, I just need to make the notebook and editor `extraKeys` consistent 👌

notebook/static/edit/js/editor.js
```diff
@@ -76,12 +76,20 @@ function(
...
     Editor.default_codemirror_options = {
         extraKeys: {
-            "Tab" :  "indentMore",
+            "Cmd-Right": "goLineRight",
+            "End": "goLineRight",
+            "Cmd-Left": "goLineLeft",
+            "Tab": "indentMore",
+            "Shift-Tab" : "indentLess",
+            // "Cmd-Alt-[" : "indentAuto",
+            // "Ctrl-Alt-[" : "indentAuto",
```
takluyver: Should these be in or out? If we don't want them, let's remove them completely rather than commenting them.

gnestor: I'll remove...

takluyver: Looks like you've removed them from cell.js, but not here ;-)

takluyver:
Reopening to rerun tests with master

PR #2752, opened 2017-08-12
schedule IOLoop.stop on the main thread

minrk:
loop.stop() will not be safe from a thread in tornado 5

There has never been a guarantee to this effect, but it has happened to be true so far.

The result is that the stop event will not fire until the next loop iteration (e.g. triggered by an HTTP request). Using add_callback ensures that the main thread wakes and handles the stop event.

cf tornadoweb/tornado#2119

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1292,7 +1292,7 @@ class NotebookApp(JupyterApp):
     
     def _signal_stop(self, sig, frame):
         self.log.critical(_("received signal %s, stopping"), sig)
-        self.io_loop.add_callback(self.io_loop.stop)
+        self.io_loop.add_callback_from_signal(self.io_loop.stop)
 
     def _signal_info(self, sig, frame):
         print(self.notebook_info())
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1279,7 +1279,7 @@ class NotebookApp(JupyterApp):
                 self.log.critical(_("Shutdown confirmed"))
                 # schedule stop on the main thread,
                 # since this might be called from a signal handler
-                self.io_loop.add_callback(self.io_loop.stop)
+                self.io_loop.add_callback_from_signal(self.io_loop.stop)
                 return
         else:
             print(_("No answer for 5s:"), end=' ')
@@ -1288,7 +1288,7 @@ class NotebookApp(JupyterApp):
         # set it back to original SIGINT handler
         # use IOLoop.add_callback because signal.signal must be called
         # from main thread
-        self.io_loop.add_callback(self._restore_sigint_handler)
+        self.io_loop.add_callback_from_signal(self._restore_sigint_handler)
     
     def _signal_stop(self, sig, frame):
         self.log.critical(_("received signal %s, stopping"), sig)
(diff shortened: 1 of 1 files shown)


stonebig:
do you intend to push it in notebook-5.1 final ?

notebook/notebookapp.py
```diff
@@ -1277,7 +1277,9 @@ def _confirm_exit(self):
             line = sys.stdin.readline()
             if line.lower().startswith(yes) and no not in line.lower():
                 self.log.critical(_("Shutdown confirmed"))
-                ioloop.IOLoop.current().stop()
+                # schedule stop on the main thread,
+                # since this might be called from a signal handler
```
takluyver: The docs for `add_callback` say that it shouldn't be used from a signal handler - there's a separate method for that.

http://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.add_callback

minrk: Thanks. Fixed.

takluyver: Sorry, I've just gone and looked at the code, and we actually call this in a thread started by a signal handler. I'm not 100% sure, but I think that should be using the `add_callback()` method after all.

notebook/notebookapp.py
```diff
@@ -1286,11 +1288,11 @@ def _confirm_exit(self):
         # set it back to original SIGINT handler
         # use IOLoop.add_callback because signal.signal must be called
         # from main thread
-        ioloop.IOLoop.current().add_callback(self._restore_sigint_handler)
+        self.io_loop.add_callback_from_signal(self._restore_sigint_handler)
     
     def _signal_stop(self, sig, frame):
         self.log.critical(_("received signal %s, stopping"), sig)
-        ioloop.IOLoop.current().stop()
+        self.io_loop.add_callback(self.io_loop.stop)
```
takluyver: This one should be using the `_from_signal` version of the method, though, as this function is registered directly as a signal handler.

gnestor:
Do we want to include this in 5.1.0?

takluyver:
My leaning would be to leave it for 5.2, but it really depends on how soon tornado 5 is expected - if that's coming soon, then it's worth taking some extra risk to get this in sooner.

gnestor:
What if we change [this line](https://github.com/jupyter/notebook/blob/master/setup.py#L147) to `'tornado>=4', <=5,` and then plan on releasing this as part of a patch or minor release?

takluyver:
@minrk , thoughts? I'd also be fine with pushing it in as a last minute fix, so long as you're confident it's working with tornado 4.

minrk:
This should be fine as-is. The `from_signal` doesn't really do anything other than improve tracebacks if exceptions are raised inside the callback, but I've added it where it ought to be. With asyncio, it's literally `add_callback_from_signal = add_callback`.

takluyver:
OK, in that case let's push it in (once the tests have passed). We can always make 5.1.1 if it turns out there's a problem in some particular case.

takluyver:
There seems to be an issue with building the JS on Travis. I can't reproduce this locally at the moment.

minrk:
Flushed the travis cache and restarted the builds

takluyver:
Thanks, I guess it must have been something wrong in the cache.

minrk:
Thanks! Yeah, it seems that the Travis cache can get in a bit of a weird state once in a while, and clearing it comes to mind when something fails during the js build, especially due to missing files. I suspect this is caused by imprecise "do I need to re-run npm install" logic in setup.py.

Issue #2742, opened 2017-08-09
Download as HTML puts "+" symbols instead of spaces in filename


blokeley:
Using the File > Download > As HTML... command, I would expect the downloaded HTML file to have the same filename as the Jupyter Notebook.

However, every space is replaced by a + symbol.

Versions:
Python 3.6.1
IPython 5.3.0
Jupyter 1.0.0
Windows 10 Pro x64

gnestor:
@takluyver I assume that there is a good reason why spaces are replaced with pluses? @minrk?

minrk:
it's probably url-escaping being applied to the wrong thing.

gnestor:
Notebook 5.1.0rc1 is available on PyPI so please give it a try and confirm that this is resolved 👍:

`pip install notebook --pre --upgrade`

PR #2761, opened 2017-08-15
Don't url escape filenames on download

gnestor:
Closes https://github.com/jupyter/notebook/issues/2742

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -478,7 +478,7 @@ class AuthenticatedFileHandler(IPythonHandler, web.StaticFileHandler):
     def get(self, path):
         if os.path.splitext(path)[1] == '.ipynb' or self.get_argument("download", False):
             name = path.rsplit('/', 1)[-1]
-            self.set_header('Content-Disposition','attachment; filename="%s"' % escape.url_escape(name))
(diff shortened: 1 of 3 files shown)


takluyver:
This looks fine to me, but I'd like @minrk to have a quick look too. I notice you've also marked it for 5.2, so we should release 5.1 before merging it.

gnestor:
We can fit this in 5.1 as long as we're confident that this won't break anything else! 😮

minrk:
This solves a problem for spaces in ascii names but reintroduces a problem for non-ascii names.

Test with a filename like 'üni spåce™' to be sure it's working.

minrk:
Some diagnoses:

1. the `+` was introduced by using the wrong escape function (`tornado.escape.url_escape`, which is for URL query parameters by default, not url components)
2. Putting a name in quotes means it shouldn't be escaped, but it also has to be Latin-1 if you do that.
3. [RFC 6266](https://tools.ietf.org/html/rfc6266) explains how to specify filename encodings properly (url escapes were the right thing to do, we just did them wrong)

#2767 should implement filename encoding correctly. I've tested it on Safari, Chrome, and Firefox.

takluyver:
Closing in favour of #2767.

PR #2767, opened 2017-08-16
use RFC5987 encoding for filenames

### LINKED ISSUE #2742: Download as HTML puts "+" symbols instead of spaces in filename

minrk:
as described in RFC 6266 describing Content-Disposition

This should result in the correct download filename for filenames with spaces and non-latin characters

finishes up #2761
closes #2742

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -279,6 +279,20 @@ class IPythonHandler(AuthenticatedHandler):
         if self.allow_credentials:
             self.set_header("Access-Control-Allow-Credentials", 'true')
     
+    def set_attachment_header(self, filename):
+        """Set Content-Disposition: attachment header
+
+        As a method to ensure handling of filename encoding
+        """
+        escaped_filename = url_escape(filename)
+        self.set_header('Content-Disposition',
+            'attachment;'
+            " filename*=utf-8''{utf8}"
+            .format(
+                utf8=escaped_filename,
+            )
+        )
+
     def get_origin(self):
         # Handle WebSocket Origin naming convention differences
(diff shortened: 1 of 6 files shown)


minrk:
Hold the phone, this doesn't work on Firefox, which apparently doesn't implement RFC 5987

takluyver:
On #2761 you said you've tested this on Firefox. Does that supersede your comment here that it doesn't work?

minrk:
@takluyver yes, sorry. Commenting in two issues. My first attempt didn't work on Firefox, but it does now.

takluyver:
Thanks. Confirmed with a manual test.

I'd be in favour of putting this in 5.1, but if we're not confident enough it's fine to bump it to 5.2.

minrk:
It worked for me, but the tiniest typo while I was figuring this out caused pretty weird failures that different browsers handled differently, so I'd be okay merging right after 5.1.

gnestor:
I tested and I'm still seeing filenames with pluses vs. spaces?

gnestor:
Ok, nevermind. I guess that I was not using a develop version of notebook or something. Looks good to me! 

@minrk What do you think about merging this with 5.1 now? Feel free to mark as 5.2 otherwise.

minrk:
Let's be conservative and call it 5.2

PR #2769, opened 2017-08-16
[Prototype JCon] Mimetype selection in frontend

Carreau:
This is a prototype (mostly to demonstrate tings during Jupytercon), in
particular that the frontend is sending all the mimetypes, and that you
could (in theory) select the mimetype representation after the fact,
even when the kernel has been restarted. This – so far – works only with
the execute result, but there is not reason for it to be limited.

I don't particularly intend to push that forward, it's mostly as a demo, unless someone want to take over during sprints. 

mimeselector

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -466,21 +466,13 @@ define([
         // Notify others of changes.
         this.element.trigger('changed', {output_area: this});
     };
+    
+    OutputArea.prototype._core_insert = function(json, toinsert, type){
+        toinsert.find('.output_subarea').remove()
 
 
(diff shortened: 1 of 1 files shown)


minrk:
Cool!

gnestor:
Nice @Carreau! I've been meaning to prototype this as well. I think it's a useful feature. What sort of feedback have you received about it?

Carreau:
> Nice @Carreau! I've been meaning to prototype this as well. I think it's a useful feature. What sort of feedback have you received about it?

People are excited. And I got some "Oh..." and "AHhh...", "Why couldn't you build that before", "Will it work on nbviewer ?"

Feel free to push changes on this PR and take over if i'm slow to update.

Issue #2757, opened 2017-08-14
Error while launching a Jupyter notebook


Ross-Jo:
I encountered with a strange error(I wasn't able to see the file TREE!!!) while executing a jupyter notebook. And the below ones are the error logs.
--------------------------------------------------------------------------------------------------------------------
```
Server error: Traceback (most recent call last):
File "C:\Users\350U2A\Anaconda3\lib\site-packages\notebook\base\handlers.py", line 516, in wrapper
 result = yield gen.maybe_future(method(self, *args, **kwargs))
File "C:\Users\350U2A\Anaconda3\lib\site-packages\tornado\gen.py", line 1055, in run
 value = future.result()
File "C:\Users\350U2A\Anaconda3\lib\site-packages\tornado\concurrent.py", line 238, in result
 raise_exc_info(self._exc_info)
File "<string>", line 4, in raise_exc_info
File "C:\Users\350U2A\Anaconda3\lib\site-packages\tornado\gen.py", line 307, in wrapper
 yielded = next(result)
File "C:\Users\350U2A\Anaconda3\lib\site-packages\notebook\services\contents\handlers.py", line 124, in get
 path=path, type=type, format=format, content=content,
File "C:\Users\350U2A\Anaconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 382, in get
 model = self._dir_model(path, content=content)
File "C:\Users\350U2A\Anaconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 264, in _dir_model
 model = self._base_model(path)
File "C:\Users\350U2A\Anaconda3\lib\site-packages\notebook\services\contents\filemanager.py", line 230, in _base_model
 created = tz.utcfromtimestamp(info.st_ctime)
File "C:\Users\350U2A\Anaconda3\lib\site-packages\notebook\_tz.py", line 30, in utc_method
 dt = unaware(*args, **kwargs)
OSError: [Errno 22] Invalid argument
```

![error](https://user-images.githubusercontent.com/14027563/29281572-810dbeac-815a-11e7-9739-1e58651fc8e5.JPG)
--------------------------------------------------------------------------------------------------------------------
So I've tried to solve the error with various methods, but, I wasn't able to solve one till now. 
And there are some lists what I've tried but failed to solve the problem
1) I appointed the starting location of jupyter via 'jupyter_notebook_config.py' file like this.
    -> c.NotebookApp.notebook_dir = r"C:/Users/350U2A/Desktop" 
[...]

PR #2770, opened 2017-08-16
Catch OSError when getting file mtime & ctime

takluyver:
Closes gh-2757

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -235,15 +235,18 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         info = os.lstat(os_path)
         try:
             last_modified = tz.utcfromtimestamp(info.st_mtime)
-        except ValueError:
+        except (ValueError, OSError):
             # Files can rarely have an invalid timestamp
             # https://github.com/jupyter/notebook/issues/2539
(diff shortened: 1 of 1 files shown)


notebook/services/contents/filemanager.py
```diff
@@ -235,15 +235,18 @@ def _base_model(self, path):
         info = os.lstat(os_path)
         try:
             last_modified = tz.utcfromtimestamp(info.st_mtime)
-        except ValueError:
+        except (ValueError, OSError):
             # Files can rarely have an invalid timestamp
             # https://github.com/jupyter/notebook/issues/2539
+            # https://github.com/jupyter/notebook/issues/2757
             # Use the Unix epoch as a fallback so we don't crash.
+            self.log.warn('Invalid mtime %s for %s', info.st_mtime, os_path)
```
minrk: log.warning to avoid deprecation warnings (and below)

takluyver: Done. I always get that one wrong.

takluyver:
@gnestor I've optimistically marked this for 5.1, but feel free to bump it to 5.2 if you'd prefer.

gnestor:
I think this should be included in 5.1. Thanks @takluyver!

PR #2774, opened 2017-08-17
Factor out output_prompt_function, as is done with input prompt

iamlemec:
This allows for much cleaner customization of the output prompt in custom.js, as is currently the case with the input prompt.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -467,6 +467,11 @@ define([
         this.element.trigger('changed', {output_area: this});
     };
 
+    OutputArea.output_prompt_classical = function(prompt_value) {
+        return '<bdi>'+i18n.msg.sprintf(i18n.msg._('Out[%d]:'),prompt_value)+'</bdi>';
+    };
+
+    OutputArea.output_prompt_function = OutputArea.output_prompt_classical;
 
     OutputArea.prototype.append_execute_result = function (json) {
         var n = json.execution_count || ' ';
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks!

We're just trying to make a 5.1 release - I've marked this as 5.2 as it's probably not urgent enough to go in at the last minute. Hopefully we can merge it soon after the release.

notebook/static/notebook/js/outputarea.js
```diff
@@ -467,6 +467,11 @@ define([
         this.element.trigger('changed', {output_area: this});
     };
 
+    OutputArea.output_prompt_classical = function(prompt_value) {
+        return '<bdi>'+i18n.msg.sprintf(i18n.msg._('Out[%d]:'),prompt_value)+'</bdi>';
```
takluyver: Can this construct a node like the old code does, rather than assembling an HTML string? While I don't think there's a security concern here, it's good practice to use the safer style throughout.

iamlemec: Ok, just made the change. Btw, the code for input_prompt_classical in codecell.js also uses an HTML string, so it might be worth changing that too.

gnestor:
@iamlemec This looks good to me. I just re-ran the Travis build and it's passing now. @takluyver Is this ready to merge?

takluyver:
Yup, I think so. 

@iamlemec thanks, and thanks for checking the input prompt method. If you want to make a separate PR to update that to create an element, I think we could merge that.

Issue #2775, opened 2017-08-18
opening file with backslash(%5C) fails on linux


j31d0:
In windows there couldn't exist file which have backslash, but linux could.
I tested with version 5.0.0.
test script : 

```
:/test$ echo "foo" > test1
:/test$ echo "bar" > \\test1
```

![image](https://user-images.githubusercontent.com/8012197/29439790-1ebb83ee-83fb-11e7-8afe-8588d6c5c08f.png)
when I enter to \test1 : 
![image](https://user-images.githubusercontent.com/8012197/29439824-54e8330e-83fb-11e7-9694-d811d7fcfcce.png)

takluyver:
Thanks, I think #2777 should fix it.

PR #2777, opened 2017-08-18
URL escape file path in iframe view handler

takluyver:
Closes gh-2775

### DIFF
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -9,5 +9,5 @@ store the current version info of the notebook.
 
 # Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**. 
 
-version_info = (5, 1, 0, 'rc1')
+version_info = (5, 1, 0, 'rc2')
 __version__ = '.'.join(map(str, version_info[:3])) + ''.join(version_info[3:])
--- a/notebook/static/base/js/namespace.js
+++ b/notebook/static/base/js/namespace.js
@@ -73,7 +73,7 @@ define(function(){
     // tree
     jglobal('SessionList','tree/js/sessionlist');
 
-    Jupyter.version = "5.1.0rc1";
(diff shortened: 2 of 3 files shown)


gnestor:
Thanks @takluyver! I can't reproduce this issue because I don't have a Windows environment, but it looks good to me. Shall I merge?

takluyver:
I think it should be reproducible on any filesystem - try making a file like `a?b.html`, then selecting it and clicking the 'view' button.

gnestor:
Hmmm, I'm actually getting a 404 for http://localhost:8888/view/a%3Fb.html 🤔

takluyver:
With this PR or before it? That's expected before this PR.

gnestor:
No I checked out this branch and rebuilt the JS (even though the change is a Python one) and restarted the server

takluyver:
Weird. Were you definitely running with the server you restarted, and not another one you had running at the same time?

For reference, here's a test file in the editor:

screenshot from 2017-08-19 09-39-46

Here's the `/view/` page on master (you can't tell from this, but the 404 is inside the iframe: the `/view/` URL loads correctly, but refers to an invalid `/files/` URL):

screenshot from 2017-08-19 09-40-04

The error logged in the terminal:
```
[W 09:39:27.062 NotebookApp] 404 GET /files/a?b.html (127.0.0.1) 16.96ms referer=http://localhost:8888/view/a%3Fb.html
```

And here's it loading correctly on this branch:

screenshot from 2017-08-19 09-41-13

@minrk @Carreau and anyone else reading this, could you check this?

gnestor:
I tried again on a new computer and alas, it worked! Merging...

Could this be another instance of indeterminate browser caching issue? It looks like you only edited Python so probably not...

takluyver:
I don't think caching could affect this, but who knows!

Issue #2780, opened 2017-08-18
Mathjax file not loading when opening or reloading notebook


amahadevan99:
When I open any jupyter notebook I get a small `loading MathJax` notification in the bottom left corner of the window. For most `.ipynb` files this proceeds, then says `Processing Math: 100%`, then goes away. However there is one particular notebook which behaves strangely. When I open or reload it, this text, instead of going to `Loading Math: 100%`, gives me an error `File failed to load: http://localhost:8888/static/components/MathJax/jax/element/mml/optable/SpacingModLetters.js`. I also get the following error in the Terminal: `404 GET /static/components/MathJax/jax/element/mml/optable/SpacingModLetters.js?rev=2.6.0 (::1) 17.79ms referer=http://localhost:8889/notebooks/data/filename.ipynb`. It is a small thing but still a bit annoying. Furthermore, the notebook seems to run and render fine, despite this error. And this notebook is similar to the other notebooks that do not cause this problem. Any ideas as to why I am getting this warning/error or what the missing file is supposed to do? Looks like this is related to issue #1108 .

takluyver:
Are you willing to share the notebook that you see the error with?

amahadevan99:
I wouldn't mind sharing the markdown cells, if that would be helpful, but I would prefer not to share the entire notebook.

takluyver:
Does the same thing still happen if you make a copy with only the markdown cells?

amahadevan99:
Yes -- In fact I just discovered that the offending cell is 
    Therefore the slope of the $E$ versus $q$ graphs should be the following:
    $\sim xxxx$ eV$\cdot\mathring{A}$ for no spin flip,
    $xxxx$ eV$\cdot\mathring{A}$ for $\mathbf k_p\to-\mathbf k_p$,
    $xxxx$ eV$\cdot\mathring{A}$ for $\mathbf k_s\to-\mathbf k_s$
Not sure what is wrong with it though.

takluyver:
I copied and pasted that, and it seems to be working on my machine. Is this what it's supposed to look like?

![screenshot from 2017-08-21 16-27-58](https://user-images.githubusercontent.com/327925/29526496-d027079a-868d-11e7-8a10-dabf3b1a0113.png)

amahadevan99:
[...]

Issue #1108, opened 2016-02-19
Mathjax js file not loading


moorepants:
I just loaded up a notebook that I last used with IPython < 4.0 and the MathJax math from SymPy output is failing to load. I'm getting this 404:

```
[W 16:47:28.747 NotebookApp] 404 GET /static/components/MathJax/jax/element/mml/optable/BasicLatin.js?rev=2.5.3 (127.0.0.1) 10.63ms referer=http://localhost:8888/notebooks/n01_dynamics_overview.ipynb
```

My previous method for installing MathJax was to do this:

```
In [1]: from IPython.external import mathjax
In [2]: mathjax.install_mathjax()
```

I see from some issues that that may not be correct anymore.

takluyver:
The `install_mathjax()` function is gone, because MathJax should now be bundled with the notebook. How did you install/upgrade?

moorepants:
I am using conda to upgrade IPython.

takluyver:
Odd, that shouldn't have this issue. Can you dig into your installation directory and find out how much of the 404'd path is missing. It's rooted in the notebook package directory, i.e.:

```python
import notebook
os.path.join(os.path.dirname(notebook.__file__), 'static/...')
```

moorepants:
This is all I find in the `jax` dir:

```
moorepants@garuda:eme185-website(master)$ ls /home/moorepants/miniconda3/lib/python3.5/site-packages/notebook/static/components/MathJax/jax/
input  output
```

Here is the complete tree:

```
moorepants@garuda:eme185-website(master)$ tree /home/moorepants/miniconda3/lib/python3.5/site-packages/notebook/static/components/MathJax/jax/
/home/moorepants/miniconda3/lib/python3.5/site-packages/notebook/static/components/MathJax/jax/
├── input
│   └── TeX
│       ├── config.js
│       └── jax.js
└── output
    └── HTML-CSS
        ├── autoload
        │   ├── annotation-xml.js
        │   ├── maction.js
        │   ├── menclose.js
        │   ├── mglyph.js
        │   ├── mmultiscripts.js
        │   ├── ms.js
        │   ├── mtable.js
        │   └── multiline.js
        ├── config.js
        ├── fonts
        │   └── STIX-Web
        │       ├── Alphabets
        │       │   ├── Bold
        │       │   │   └── Main.js
        │       │   ├── BoldItalic
        │       │   │   └── Main.js
        │       │   ├── Italic
        │       │   │   └── Main.js
[...]

Issue #2747, opened 2017-08-11
Notebook failed to load error


fonnesbeck:
After upgrading to the notebook from master ('5.1.0.dev'), I get a popup error saying that the notebook failed to load:

![](https://d.pr/i/Zujuuq+)

the underlying notebook is also truncated at the second cell, when in fact it is quite a large notebook.

If I close the notebook, and re-open it (without restarting it), it renders fine. The only thing unusual I notice in the console is:

    [I 08:37:27.331 NotebookApp] Adapting to protocol v5.1 for kernel 515185bf-a09c-44c5-80c9-53390ff1b98e

gnestor:
Similar: https://github.com/jupyter/notebook/issues/2733

gnestor:
I will investigate that "adapting to protocol 5.1" message...

Was this notebook already open in your browser and you just refreshed after the upgrade? Perhaps some 5.0 JS was being loaded from cache?

What browser are you using? Would you happen to have LastPass extension installed?

fonnesbeck:
I am running on Firefox 55 on macOS. Not using LastPass, but do have 1Password.

If I use a different conda env with an older version of the notebook, the notebook runs fine.

gnestor:
Ok, you're seeing the `Adapting to protocol v5.1` message because `protocol_version != client_protocol_version`, which leads me to believe that the browser is loading 5.0 JS vs. 5.1. Try clearing the cache and/or doing a hard refresh.

fonnesbeck:
Just got this after clearing the browser cache:

![](http://d.pr/i/dlZvUl+)

gnestor:
Hmmm... So when you close the notebook (I assume when you close the tab) and re-open, you don't get the error?

fonnesbeck:
Actually, it sometimes errors again, but in some cases the error goes away. Also, if I use Safari this does not seem to happen. I will try switching back to Safari as my regular browser and see if it ever comes up.

stonebig:
I had the same error once, and only once,  on 5.1.0rc1, on windows10/jupyter-5.0/chrome/very_slow_pc. Just restarting the kernel solved it. 

I would suspect a slowness, or some remains of a previous widgets/whatever versions that are cleared with the relaunch.

fonnesbeck:
I've now seen the issue pop up in Safari:

![](http://d.pr/i/a6lrtE+)

Closing the notebook tab and reopening it loaded the notebook properly.

[...]

PR #2788, opened 2017-08-21
Add Mathjax/jax/element/mml to package_data

takluyver:
Closes gh-2780
Closes gh-1108

One of the comments on #1108 concerns a file `mathchoice.js` which is the `extensions` directory under Mathjax. Do we want to include that as well (350KB total)?


takluyver:
Nevermind about the `extensions` directory, we're actually including that already. Maybe that commenter was using an older notebook release.

takluyver:
It's also conceivable that this is related to the mysterious issue #2747, but I think it unlikely.

Issue #2733, opened 2017-08-07
Testing RC and existing notebook failed to load


willingc:
Hi @gnestor,

Running the notebook RC in a conda env and received a strange pop-up when opening an existing notebook. I don't know  if something else needs updating (extensions?) but had never seen this before.

[pastebin of debug log](https://pastebin.com/bydwV1hs)

<img width="1227" alt="screen shot 2017-08-07 at 3 49 07 pm" src="https://user-images.githubusercontent.com/2680980/29049232-99e78b3c-7b89-11e7-83bb-9576f91587d5.png">
<img width="698" alt="screen shot 2017-08-07 at 3 50 11 pm" src="https://user-images.githubusercontent.com/2680980/29049233-99f6b5e4-7b89-11e7-99d7-e0e77aceb586.png">
<img width="1198" alt="screen shot 2017-08-07 at 3 54 40 pm" src="https://user-images.githubusercontent.com/2680980/29049234-9a08c9fa-7b89-11e7-8858-6d498f3fce66.png">

takluyver:
I had a go at reproducing this, but it appears to be working normally for me, in both Firefox and Chromium.

Are there any error details in the JS console? That error message ("Can't make callback from given data") comes [from Mathjax](https://github.com/mathjax/MathJax/blob/2.6.0/unpacked/MathJax.js#L315), but I don't know how we're calling that code.

willingc:
Thanks @takluyver. The error is happening in Safari Version 10.1.1 (12603.2.4). It is fine in Chrome.

<img width="1320" alt="screen shot 2017-08-08 at 7 44 02 am" src="https://user-images.githubusercontent.com/2680980/29077837-db112030-7c0d-11e7-95bb-0dded3e580df.png">

gnestor:
Hey Carol! Thanks for this! I ran into an issue (albeit different) after installing this release. 

I tried in Safari and I can't reproduce this "Can't make callback from given data" error, but I am seeing the 404 errors in the browser console (which should be fixed). Did you try emptying the cache?

willingc:
@takluyver @gnestor A bit more detail...

- Same behavior after clearing cache in both regular Safari window and private Safari window
- It seems to happen most when opening an existing notebook for the first time. If you then try to open again while it is running, it works.
- If I disable all extensions from the Develop menu in Safari, it seems to work with no issues.

[...]

PR #2792, opened 2017-08-22
Attempt to fix mysterious mathjax callback error

takluyver:
Looking at #2733 again, I noticed that @willingc's JS console did show tracebacks, which I had  somehow overlooked before. Partly that was just me not looking closely enough, but we can also make it stand out more by logging it with `console.error`, which this PR does.

That pointed to the code which was modified in PR #2677 just last month; this fits with people first running into the error when we made an RC.

That change asks Mathjax to call `MathJax.InputJax.TeX.resetEquationNumbers()` after typesetting each thing. That function is defined in a Mathjax extension `extensions/TeX/AMSmath.js`. While I still can't reproduce the error, I guess that looking up the function is in a race condition with the extension loading; if it loses the race, then there's nothing to call, so an error pops out.

This PR:

* Catches errors from enqueueing resetEquationNumbers(), logs them and carries on, so that these (hopefully) cannot cause a failure to load the notebook.
* Based on my speculation, enqueues a Mathjax `Require` call immediately before to ensure that the extension is loaded before we use it.

If this passes review, I suggest that we make an RC3 soon after merging it and ask the people who can reproduce the issue to see if it's fixed.

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -945,7 +945,8 @@ define([
             MathJax.Hub.Queue(["Typeset", MathJax.Hub, this]);
             try {
                 MathJax.Hub.Queue(
-                    ["resetEquationNumbers", MathJax.InputJax.TeX]
+                    ["Require", MathJax.Ajax, "[MathJax]/extensions/TeX/AMSmath.js"],
+                    function() { MathJax.InputJax.TeX.resetEquationNumbers(); }
                 );
             } catch (e) {
                 console.error("Error queueing resetEquationNumbers:", e);
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -940,12 +940,16 @@ define([
         if(!window.MathJax){
             return;
         }
-        return $el.map(function(){
+        $el.map(function(){
             // MathJax takes a DOM node: $.map makes `this` the context
-            return MathJax.Hub.Queue(
-                ["Typeset", MathJax.Hub, this],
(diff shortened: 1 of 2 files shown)


gnestor:
@takluyver Excellent!! Thanks for tracking this down! With your blessing, I will merge this and create a new rc release.

takluyver:
That's fine by me - I think it's the easiest way to get people to test it.

I have run with this locally and it doesn't obviously break anything for me, but I haven't tried using the equation numbering the code in question was meant to enable.

mpacer:
The code doesn't enable equation numbering. It enables assigning labels to
equations without breaking them on further renders.

If you render the same cell twice does it produce a weird box around he raw
text of the internals instead of showing displayed math?

takluyver:
> If you render the same cell twice does it produce a weird box around he raw text of the internals instead of showing displayed math?

Not for me. Rendering mathematical expressions appears to be working as it does in master, at least on my system.

mpacer:
Including if you have a label in the eq? Sorry, I should have been specific.

takluyver:
How do you put a label in the equation?

mpacer:
\label{your_label} should work

takluyver:
screenshot from 2017-08-23 17-18-09
screenshot from 2017-08-23 17-18-16

Seems to be fine. I've flipped it between rendered and unrendered a few times.

willingc:
@takluyver Thank you :D I haven't seen the mathjax error while using this fix. :tada:

Issue #2798, opened 2017-08-23
Cannot build docs: NameError: name '_' is not defined


hroncok:
I'm trying to package 5.1.0rc2 as an RPM for Fedora Rawhide (to be Fedora 28). Part of the process is to build the documentation.

I have the sdist from PyPI unpacked + 81d0e6146d73a716e8f46b3054dd389ef67779be applied as a patch. I'm running CPython 3.6.2.

The error is:

```
Exception occurred:
  File "/notebook-5.1.0rc2/notebook/services/contents/manager.py", line 69, in ContentsManager
    untitled_notebook = Unicode(_("Untitled"), config=True,
NameError: name '_' is not defined
```

<details>
<summary>full command+stdout+stderr</summary>

```console
# sphinx-build-3 docs/source html
Running Sphinx v1.6.3
making output directory...
DEBUG:: sys.path
================
/usr/bin
/usr/lib64/python36.zip
/usr/lib64/python3.6
/usr/lib64/python3.6/lib-dynload
/usr/lib64/python3.6/site-packages
/usr/lib/python3.6/site-packages
repo_root
=====================
/notebook-5.1.0rc2
DEBUG:: Post insert to sys.path
===============================
/notebook-5.1.0rc2
/notebook-5.1.0rc2/docs/sphinxext
/usr/bin
/usr/lib64/python36.zip
/usr/lib64/python3.6
/usr/lib64/python3.6/lib-dynload
/usr/lib64/python3.6/site-packages
/usr/lib/python3.6/site-packages
Initializing GitHub plugin
loading pickled environment... not yet created
loading intersphinx inventory from https://ipython.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://nbconvert.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://nbformat.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://jupyter.readthedocs.io/en/latest/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 32 source files that are out of date
updating environment: 32 added, 0 changed, 0 removed
reading sources... [  3%] changelog
reading sources... [  6%] comms
reading sources... [  9%] config_overview
reading sources... [ 12%] contributing
reading sources... [ 15%] development_faq
reading sources... [ 18%] development_release
reading sources... [ 21%] examples/Notebook/Connecting with the Qt Console
Executing notebook with kernel: python3
[...]

PR #2799, opened 2017-08-23
Have _ defined in notebook.transutils

hroncok:
Fixes https://github.com/jupyter/notebook/issues/2798

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -345,3 +345,6 @@ intersphinx_mapping = {
 
 spelling_lang='en_US'
 spelling_word_list_filename='spelling_wordlist.txt'
+
+# import before any doc is built, so _ is guaranteed to be injected
+import notebook.transutils
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -35,11 +35,7 @@ except ImportError: #PY2
 
 from jinja2 import Environment, FileSystemLoader
 
-# Set up message catalog access
-base_dir = os.path.realpath(os.path.join(__file__, '..', '..'))
-trans = gettext.translation('notebook', localedir=os.path.join(base_dir, 'notebook/i18n'), fallback=True)
-trans.install()
-_ = trans.gettext
+from notebook.transutils import trans, _
 
 # Install the pyzmq ioloop. This has to be done before anything else from
 # tornado is imported.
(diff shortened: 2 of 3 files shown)


hroncok:
This allows to use _ from anywhere.

hroncok:
@takluyver from https://github.com/jupyter/notebook/issues/2798#issuecomment-324465518:

> Yup, that's the i18n machinery - `_` is injected into the builtins to look up translations, but evidently it isn't when the docs build imports the notebook.
>
> Thanks for the report, it's embarrassing that we haven't caught that already. I suspect it can be worked around by modifying `conf.py` - if you get that working, please do send us a PR.

If that's the case, maybe importing it in `notebook/services/contents/manager.py` is not a good idea and injecting it to builtins in `conf.py` would be preferred. Thoughts?

takluyver:
I'd be inclined to import it from `conf.py` - calling `.install()` special cases `_` to be used throughout the code without importing it. Importing it into one file risks that it will spread to another file and we'll be back to problems with the docs.

Adding it to builtins is not very Pythonic, though, so other people might disagree with me.

hroncok:
So, let's keep the transutils module but import `_` from it in `conf.py` instead of `notebook/services/contents/manager.py`?

takluyver:
Sounds OK. I don't think it needs to import `_` in conf, though; the important bit is that `trans.install()` gets called before we try importing other notebook modules.

hroncok:
Well adding the following code to the end of `conf.py` actually makes sure it gets called before everything else:

```python
# import before any doc is built, so _ is guaranteed to be injected
from notebook.transutils import _
```

It's a bit black magik and can be exchanged with `from notebook import transutils`.

takluyver:
Does "import notebook.transutils" work too?

hroncok:
Yes it does. Amended.

takluyver:
Thanks! Could I ask you to add a one-line docstring to the `transutils` module? Other than that, I think this is ready to go.

hroncok:
Done. Also copied the copyright header from `utils.py`

takluyver:
Super, thanks.

hroncok:
AppVeyor fails with bower. Seems not related.

takluyver:
Yup. I clicked rebuild just so it goes green.

I'm not addicted to little green CI check marks. I can quit any time I like.

Issue #2768, opened 2017-08-16
`COPYING.md` to `LICENSE`


jzf2101:
We should change the name of our license file to be consistent across repos at Project Jupyter

rolweber:
Would that imply some re-formatting of the content? It's currently in markdown format, and probably won't be rendered as such by GitHub anymore if the file suffix changes.

Also, there's more in the file than just the license.

danielskatz:
I was checking to see the license for this project, and was surprised not to find a license file in the repo, which led me to this issue.  It would be good if there was a LICENSE.txt file that was just a text version of the license, which doesn't mean there also can't be a COPYING.md file that has some of what is now in the file (except the license) and points to LINCENSE.txt

sammyrTX:
Hi! Would it be okay to just copy the contents of the jupyter/LICENSE file into a jupyter/notebook/LICENSE.txt file and then updating the existing notebook/COPYING.md file to reference the new file (as suggested by @danielskatz)? Or are there any additional requirements for this issue? Thanks!

sammyrTX:
I opened PR #4765 which is based on my previous comment. Review pending.

jtpio:
Closing as the current `LICENSE` is aligned with the other Jupyter projects: https://github.com/jupyter/notebook/blob/main/LICENSE

PR #2813, opened 2017-08-26
revise license [ci skip]

aliandra:
rename COPYING.md to LICENSE and update language. Closes issue #2768


Carreau:
If this is renamed, MANIFEST.in should be updated accordingly. 

Also we should test that in CI, and should not CI-skip it.

Zsailer:
This was addressed [here](https://github.com/jupyter/notebook/commit/13039db1a6f1e4aa0f1790b3472027e477a2df11). Thanks, @aliandra!

PR #2815, opened 2017-08-28
typos fix

sytays:


takluyver:
Thanks - actually heading cells are no longer a thing (you just use markdown cells for headings now). So it should say there are three types of cells.

sytays:
Alright. Fixing it now...

takluyver:
Thanks :-)

Issue #2659, opened 2017-07-18
Enhancement: Option to allow serving from hidden directory


jendas1:
```
[I 21:48:37.098 NotebookApp] Refusing to serve hidden directory, via 404 Error
```
Since iCloud Drive folder on macOS is reachable from user folder but it goes through hidden folder, I would love to have the ability to override the hidden directory restriction so I can view it (if the jupyter server has user folder as root).

rolweber:
There is no configuration setting. The logic is implemented in various places of the FileManager. You could implement and configure your own FileManager to allow access to (some) hidden directories.

artonimous:
I ran into the same issue trying to work on .ipynb's and seemed to be able to work around it with the following two approaches.
**Using Jupyter-notebook:** 
1. $ cd ~/Library/Mobile\ Documents/com~apple~CloudDocs
2. $ jupyter-notebook
OR
**Using the JupyterLab Alpha Preview interface** the somewhat more recursive, solution:
1. From Anaconda Navigator launch a JupyterLab session
2. Within the lab open an terminal session
3. $ cd ~/Library/Mobile\ Documents/com~apple~CloudDocs
4. $ jupyter-lab

Initial test seem fine using either of these approaches but comments/feedback on the limitations or a more effective/stable approach would be appreciated.

artonimous:
Forgot this is markdown - ignore the strike through on the paths in the previous comment

takluyver:
There is now a config option `ContentsManager.allow_hidden` to allow serving hidden files. They're still hidden from the list in the dashboard, but you can access them by URL. Added in https://github.com/jupyter/notebook/pull/2819

wenmin-wu:
starts the notebook with this command: `jupyter notebook --ContentsManager.allow_hidden=True`

PR #2819, opened 2017-08-30
Added a flag to allow access of hidden files

danpf:
I thought this flag would be useful. However, looking back it looks like there was some discussion as to if this should ever be allowed.  So not a big deal if you don't want it in, but since I've been using this for myself for a while I figured I'd at least give it a shot.

The flag '--allow-hidden' will allow Tornado to access hidden files
such as '.images/my_img.jpg'

relevant threads:
issue:#2659
https://github.com/jupyterlab/jupyterlab/issues/2910
https://github.com/ipython/ipython/issues/5157

Sorry for the multiple commits, didn't think to test something other than jupyter notebook originally

- [x] Fix how we aren't checking if a file is under the root directory.
 ~~Reduce option priority.~~ I don't think this makes much sense considering how much had to be changed.


takluyver:
Thanks! I think this makes sense. We're just trying to release notebook 5.1, so we'll review this more closely once we've got that out.

takluyver:
5.1 is now out, so we can get back to this. :-)

@minrk @gnestor @Carreau any thoughts on whether/how this should be done?

notebook/notebookapp.py
```diff
@@ -445,6 +445,11 @@ def start(self):
     _("Allow the notebook to be run from root user.")
 )
 
+flags['allow-hidden']=(
```
minrk: I don't believe this argument is of sufficient priority that it should get this level of promotion. There's a cost to adding noise here. All options are always accessible via lower level config, e.g. `--NotebookApp.allow_hidden=True`.

notebook/notebookapp.py
```diff
@@ -550,6 +555,10 @@ def _default_log_format(self):
         help=_("Whether to allow the user to run the notebook as root.")
     )
 
+    allow_hidden = Bool(False, config=True,
```
minrk: It seems like this would more logically be on `FileContentsManager`, rather than NotebookApp. What do you think?

takluyver: Interestingly, the abstract ContentsManager (which doesn't know about files) defines an `is_hidden()` method, but doesn't appear to use it for anything. It's FileContentsManager which is deciding not to show hidden files.

Carreau:
I'm happy to have that as an option. We have to be careful, I believe some of the "is_hidden" logic does (used to ?) do two things; a file is hidden if either:
  -  somewhere in its path the name starts with `.`
  - its abspath (after resolving links) is outside the server root.

I think we _still_ want the files outside of the root to be unaccessible. In which case this implementation is not totally correct.

danpf:
> I think we still want the files outside of the root to be unaccessible. In which case this implementation is not totally correct.
That's a good point, I will think of a way to change that.

notebook/notebookapp.py
danpf: That's a good option.  The only thing that is weird about it is that it's kind of unintuitive to type `jupyter lab --NotebookApp.allow_hidden=True` when you're actually launching the LabApp. Although the back end is still notebook, it might seem weird for a user to have to type that when launching a different app (or they might not even think twice about it).

Alternatively, a user probably wouldn't have come across this option unless they knew what they were doing, so they might understand why it's a NotebookApp option.

notebook/notebookapp.py
danpf: Hrm I think, for my original intentions, it would be mildly confusing to put it in `FileContentsManager`.

* `FileContentsManager` controls the ability to travel to hidden directories.
* `AuthenticatedFileHandler` controls the ability to access hidden files in markdown (maybe more?)
* `TreeHandler` controls the ability to access hidden files in the filebrowser

For my personal use, this PR was to allow access to hidden files through markdown, not to allow travel to hidden directories/open hidden files. But issue #2659 wanted that ability, so I'm not sure what the best course of action is.

If you think I should include the ability to travel to hidden directories and access hidden files, I can though. Then each class will have their own `allow_hidden` and the option can be accessed separately.  Although `FileContentsManager` and `TreeHandler` should probably be linked somehow (can't think of a reason to separate that logic.)

minrk:
> its abspath (after resolving links) is outside the server root.

I don't think this has ever been true. I believe we've always allowed the use of symlinks to serve files outside root (my default notebooks dir is just a folder full of symlinks to elsewhere).

gnestor:
Is this ready to merge? Or should we mark it as 5.3?

danpf:
I don't know exactly how to proceed. I mentioned this above, but I'll summarize:

For my personal use, this PR was to allow access to hidden files through markdown, not to allow travel to hidden directories/open hidden files in the filebrowser.

In order to add that ability you would have to add an option to the filebrowser to view these, you pretty much only have to edit the function `is_hidden` in utils. 

Editing that is relatively easy and straightforward, but i feel like the filebrowser would be super clogged with tons of dotfiles, and could be annoying unless there's a gui option on the filebrowser for showing hidden files, I think it would have to be a toggle specific to TreeHandler, could be wrong, once I start I'll double check.

Do you think I should go ahead with that instead of my original proposal?

takluyver:
Yeah, there's really two things to control: allowing access to hidden files, and showing them in the list. I think it's fine to add an option to allow access but leave them hidden. Showing them is more complex, because you probably want that to be switchable at runtime, not just from a config file.

It sounds like @gnestor is getting ready to release 5.2 in the next couple of days, so I think this PR may need to wait for 5.3.

gnestor:
Proposal: Add a "Toggle hidden files" item to the View menu?

danpf:
I don't know javascript at all. I tried for a little bit, but I couldn't figure it out. Maybe someone else can help on that part? 

In any case,  the --allow-hidden flag now should let you see hidden files in the filebrowser, and travel to hidden directories.  You also still cannot create hidden files via the filemanager. I don't know how to move variables between JS and python, but it looks relatively simple if you could.

takluyver:
I've had a go at simplifying this:

- Condensed to one attribute `ContentsManager.allow_hidden`. Configurable, default False.
- For now, setting it to True allows access to hidden files but doesn't show them in the listing (so they're still hidden). We can think about showing them in the list as a separate step after this.
- At @minrk's suggestion, I've removed the command line flag. It's still possible to specify it at the command line by using the full name: `--ContentsManager.allow_hidden=True` .

gnestor:
Thanks @takluyver! Should we try to include this in 5.3?

takluyver:
I'd say go for it - it should be low risk.

PR #2840, opened 2017-09-14
create a vdom output type

rgbkrk:
This implements a renderer for declarative "VDOM" style output (HTML as JSON if you will). It's not even _really_ vdom, as you could do this with `document.createElement` and enough patience. Maybe sprinkling some [morph dom](https://github.com/patrick-steele-idem/morphdom) on top. This was first done in nteract and made to be non-specific to React (phosphor should be able to handle it too). Similar PR in nteract: https://github.com/nteract/nteract/pull/1866

## `application/vdom.v1+json` Specification

The top level element must be a `VDOMEl` as described below:

```js
type VDOMEl = {
  tagName: string, // Must be a valid HTMLElement (web components 👌🏻)
  children: VDOMNode,
  attributes: Object,
};

type VDOMNode = VDOMEl | string | Array<VDOMNode> | null;
```

## Goals

The eventual goal is to be able to do display updates that keep DOM state (only changing the nodes that matter):

whoa emoji time

Here's it working in classic:

screen shot 2017-09-14 at 2 49 22 pm

/cc @willingc @gnestor @lheagy @mariusvniekerk @Carreau @jackparmer

### DIFF
--- a/notebook/static/notebook/js/object-to-preact.js
+++ b/notebook/static/notebook/js/object-to-preact.js
@@ -93,8 +93,16 @@ define(function() {
       } else if (typeof item === "string") {
         result.push(item);
       } else if (typeof item === "object") {
-        const keyedItem = item;
-        item.key = i;
+        const keyedItem = {
+          tagName: item.tagName,
+          attributes: item.attributes,
+          children: item.children
+        };
+        if (item.attributes && item.attributes.key) {
+          keyedItem.key = item.attributes.key;
+        } else {
+          keyedItem.key = i;
+        }
         result.push(objectToPreactElement(keyedItem));
       } else {
         console.warn("invalid vdom data passed", item);
--- a/notebook/static/notebook/js/object-to-preact.js
+++ b/notebook/static/notebook/js/object-to-preact.js
@@ -1,0 +1,108 @@
+/**
+ * The original copy of this comes from
+ * https://github.com/remarkablemark/REON/blob/1f126e71c17f96daad518abffdb2c53b66b8b792/lib/object-to-react.js
+ *
+ * MIT License
+ *
+ * Copyright (c) 2016 Menglin "Mark" Xu <mark@remarkablemark.org>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+ * LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+define([], function() {
+  "use strict";
+  // Since apparently this is global rather than part of the UMD loading
+  var preact = window.preact;
+
+  /**
+    * Convert an object to preact element(s).
+    *
+    * @param  {Object}       obj - The element object.
+    * @return {ReactElement}
+    */
+  function objectToPreactElement(obj) {
+    // Pack args for preact.h
+    var args = [];
+
+    if (!obj.tagName || typeof obj.tagName !== "string") {
+      throw new Error("Invalid tagName on ", JSON.stringify(obj, null, 2));
+    }
+    if (!obj.attributes || typeof obj.attributes !== "object") {
+      throw new Error("Attributes must exist on a VDOM Object");
+    }
+
+    // `React.createElement` 1st argument: type
+    args[0] = obj.tagName;
+    args[1] = obj.attributes;
+
+    const children = obj.children;
+
+    if (children) {
+      if (Array.isArray(children)) {
+        // to be safe (although this should never happen)
+        if (args[1] === undefined) {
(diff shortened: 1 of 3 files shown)


rgbkrk:
I've started making a `vdom` library as well which I'd love to collaborate on others with. It allows users to write code that has resulting output like below:

```python
VDOM(
    div([
        h1('Now Incredibly Declarative'),
        p(['Can you believe we wrote ', bold('all this from scratch'), '?']),
        img(src="https://media.giphy.com/media/xUPGcguWZHRC2HyBRS/giphy.gif"),
        p('SO COOL!'),
    ])
)
```
# Now Incredibly Declarative

Can you believe we wrote *all this from scratch*?



SO COOL!

rgbkrk:
Where I'm a bit stuck is in figuring out how I could do update displays without nuking the previous element (thereby losing the actual benefit of the stateful display).

rgbkrk:
I'll go ahead and make the diff manageable, then I'll leave this alone to touch the update display stuff later.

rgbkrk:
```diff
7d6
<     "underscore",
14d12
<     "notebook/js/object-to-preact",
19d16
<     _,
26d22
<     otp,
283d278
<     var MIME_VDOM = "application/vdom.v1+json";
294d288
<       MIME_VDOM,
726d719
<     OutputArea.safe_outputs[MIME_VDOM] = true;
777,790d769
<     var append_vdom = function(vdom, md, element) {
<       var type = MIME_VDOM;
<       var toinsert = this.create_output_subarea(
<         md,
<         "output_html rendered_vdom",
<         type
<       );
<
<       element.append(toinsert);
<       preact.render(otp.objectToPreactElement(_.clone(vdom)), toinsert[0]);
<
<       return toinsert;
<     };
<
1188d1166
<       MIME_VDOM,
1202d1179
<     OutputArea.append_map[MIME_VDOM] = append_vdom;
```

rgbkrk:
By the way, does anyone know if there's a way to get `preact` as a UMD bundle we can work with require.js here rather than the assumed global that is used in other parts of the code here?

rgbkrk:
@minrk can you think of a good strategy for how I could make update less destructive? I was confused here about the difference between `toinsert` and `element`. It looks like `element` is tracked, and I could reuse it for `preact.render`.

takluyver:
Awesome!

minrk:
> I was confused here about the difference between toinsert and element

Generally `toinsert` refers to an element that's specifically not on the page yet. If you have a good idea for how to update elements in-place, that would be cool. The way it is right now, we do it in two steps:

1. create a [new output](https://github.com/jupyter/notebook/blob/5.1.0/notebook/static/notebook/js/outputarea.js#L615)
2. [replace the original output via jQuery](https://github.com/jupyter/notebook/blob/5.1.0/notebook/static/notebook/js/outputarea.js#L451) after everything else is done

I don't see an obvious way to handle the update-in-place, other than to possibly push down the possible replacement logic into each `append_foo` method, rather than letting them assume they always populate a new element, as they do now.

rgbkrk:
> other than to possibly push down the possible replacement logic into each append_foo method, rather than letting them assume they always populate a new element, as they do now

Ok, at first I worried about doing that. Perhaps that's ok though. I could end up making updates for the other mimetypes smoother as a result.

minrk:
I think it's fine if you figure out a way to make it work.

rgbkrk:
As requested by @mpacer, here's a GIF showing what happens when the update happens to the VDOM output.

wipeout

I'd actually love to get this in as is then work on the update semantics later so that I'm not changing any base APIs within the outputs.

rgbkrk:
```
FAIL ".CodeMirror-code" still did not exist in 10000ms
#    type: uncaughtError
#    file: /home/travis/build/jupyter/notebook/notebook/tests/base/utils.js
#    error: ".CodeMirror-code" still did not exist in 10000ms
#    stack: not provided
```

Is this normal?

rgbkrk:
Looks like this is happening on `master` too:

```
 {master %=} kylek@nfml-83G ~/code/src/github.com/jupyter/notebook$ python -m notebook.jstest notebook

Test group: notebook
Running tests with notebook directory '/var/folders/kd/cylz4mhs1_9cpsjh0_c_gzfr0000gn/T/tmp_YfpvE'
Test file: /Users/kylek/code/src/github.com/jupyter/notebook/notebook/tests/notebook/attachments.js
Page Error
  Message:   Error: Script error for "components/moment/locale/en-us"
  Call stack:
    line 140 of static/components/requirejs/require.js in defaultOnError
    line 544 of static/components/requirejs/require.js in onError
    line 1732 of static/components/requirejs/require.js in onScriptError
FAIL ".CodeMirror-code" still did not exist in 10000ms
#    type: uncaughtError
#    file: /Users/kylek/code/src/github.com/jupyter/notebook/notebook/tests/notebook/attachments.js
#    error: ".CodeMirror-code" still did not exist in 10000ms

#    stack: not provided
```

rgbkrk:
This is looking like `components/moment/locale/en-us` is missing:

```
$ ls notebook/static/components/moment/locale/en-*
notebook/static/components/moment/locale/en-au.js notebook/static/components/moment/locale/en-ca.js notebook/static/components/moment/locale/en-gb.js
```

rgbkrk:
Fun fact time! It turns out you don't have to do camelCased property names for CSS. It will accept either of them if they match the CSS name:

screen shot 2017-09-25 at 9 31 23 pm

rgbkrk:
If possible, I'd like to merge this one as is and not worry about the display updates portion of this as it works just not as well as it could.

gnestor:
@rgbkrk Sorry I haven't had a chance to review this yet! If you can resolve the Travis issue, then feel free to merge. Otherwise, I will have a look at it tonight 👍

rgbkrk:
Yeah I have no idea how to resolve the issue on Travis. I'm getting a timeout error for the codemirror cell.

gnestor:
I can't figure out this Travis issue, but I can reproduce it by running the tests locally...

@minrk Any idea what might be going on here? https://travis-ci.org/jupyter/notebook/jobs/279802361#L816

rgbkrk:
I have this sneaky feeling that `preact` isn't defined on the page by the time this runs and it is currently bombing out. Perhaps the precursor to this PR should be to get `react` from bower (and trim out the current preact usage), then I'll have this PR do the right thing.

Madhu94:
I hope this hasn't been dropped. What needs to be done to move this along ? @rgbkrk @gnestor

rgbkrk:
The main blocker is the test suite, which has been a work in progress to migrate off of the old slimerjs / phantomjs setup. cc @takluyver

rgbkrk:
At least to me, I'm dropping this PR for the time being pending classic notebook being brought to the future with npm packages. As it stands, most development is invested in all the newer frontends whether nteract or jupyterlab (or Hydrogen, Neutron, etc.). Totally cool if you bring it forward. 😄 

Looking at the commits, I'd prefer to have this package be based on `@nteract/transform-vdom`.

blink1073:
Closing, since we ended up not switching to npm packages.   Thanks for experimenting with this @rgbkrk!

PR #2849, opened 2017-09-18
Upgrade xterm.js to 2.9.2

cancan101:


rgbkrk:
Does this fix #2850 for you?

takluyver:
It looks like the fix for that issue is still waiting in a PR on xterm.

I've restarted one test job that failed; if it passes next time, I'm happy to merge this.

takluyver:
@Carreau it looks like the notebook tests rely on passing a base64-encoded unicode string to `IPython.display.Image()`, and the new release of IPython has broken that (the test only fails on Python 3, but I suspect it will display an invalid image with Python 2).

Do you think we should change the notebook tests to not rely on that, or put back the ability to pass base64 encoded data into the class in IPython?

rgbkrk:
> Do you think we should change the notebook tests to not rely on that, or put back the ability to pass base64 encoded data into the class in IPython?

Change the notebook test

minrk:
#2851 should update the tests to use bytes instead.

minrk:
If there is a fix to be made on the IPython side, I think it would be to make the `b64encode` conditional of if the input data is `bytes` or `unicode`:

```python
if isinstance(data, unicode):
    # already base64
    b64data = data
else:
    b64data = b2a_base64(data).decode('ascii')
```

takluyver:
Reopening to rerun the tests now that #2851 is merged.

PR #2852, opened 2017-09-20
Test Python 3.6 on Travis, drop 3.3

takluyver:
3.3 reaches its end of upstream support at the end of this month, so I think it's fine to stop testing with it now.


takluyver:
https://en.wikipedia.org/wiki/CPython#Version_history for Python version support dates

minrk:
We should also remember to update `python_requires` in setup.py

PR #2854, opened 2017-09-21
Allow copy operation on modified, read-only notebook

### LINKED ISSUE #2541: "Make a copy" no longer works on read-only notebooks
When editing a read-only jupyter notebook (made read-only via file system) I noticed that I can no longer use the "Make a copy" feature. All that happens is that a new empty tab is opened. 

I played a bit around to find out when this happens exactly. 

1. One can get lucky if one is quick enough to hit the button before Jupyter is finished rendering the notebook. 
 2. I could only reproduce this issue when the Jupyter Notebook is served via a port reroute (using iptables from port 443 or 80 to 8888) via web. On a local installation this does not happen. 
 
As a note this did not use to be an issue like half a year ago were I ran the same set of Jupyter Notebooks with the exact same set up. I can be sure of that since I used the same docker setup.

I can supply a running live server where this can tested out if needed.

Madhu94:
Should close #2541, let me know if something else is needed

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2910,7 +2910,7 @@ define([
         var w = window.open('', IPython._target);
         var parent = utils.url_path_split(this.notebook_path)[0];
         var p;
-        if (this.dirty) {
+        if (this.dirty && this.writable) {
             p = this.save_notebook(true);
         } else {
             p = Promise.resolve();


Madhu94:
There should be a way for contributors to re-run Travis for just their commits...

rgbkrk:
Thanks @madhu94!

PR #2866, opened 2017-09-25
load locale dynamically only when not en-us

rgbkrk:
While trying to investigate failing tests on master, I noticed that there were errors because moment doesn't have a separate locale file for `en` or `en-us`. They keep it as a default per https://github.com/moment/moment/issues/2081.

Since the whitespace was misleading in this file (I thought some of these functions were inside each other until I cleaned it up), I ran `prettier` over it.

### DIFF
--- a/notebook/static/bidi/bidi.js
+++ b/notebook/static/bidi/bidi.js
@@ -1,56 +1,57 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-define([
-    'bidi/numericshaping',
-], function (numericshaping){
-    "use strict";
-    
-    var shaperType="";
-    var textDir="";
-    
-    var _uiLang= function (){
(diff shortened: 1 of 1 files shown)


notebook/static/bidi/bidi.js
```diff
@@ -1,56 +1,57 @@
...
+    return navigator.language.toLowerCase();
+  };
+
+  var _loadLocale = function() {
+    if (_isMirroringEnabled()) {
+      $('body').attr('dir', 'rtl');
+    }
+    var uiLang = _uiLang();
+
+    if (uiLang !== 'en' && uiLang !== 'en-us') {
```
rgbkrk: The primary change here, with comment below

rgbkrk:
Since this is _pretty much_ passing, I'm thinking I need to take a second look at how I'm doing the vdom mimetype in #2840. O____o

PR #2887, opened 2017-10-01
give Page constructor default arguments

jcb91:
Otherwise, any extension using a template which relied on the old no-parameter formation breaks since #2279


rgbkrk:
Thanks @jcb91

PR #2888, opened 2017-10-01
fix notebook require config to match tools/build-main

jcb91:
#2140 altered the dependencies of a bunch of modules, without updating the require config used on page.html template. The updated tools/build-main.js require config ensures that this isn't a problem for pages with built minified js, but the changes currently break any extension template which tries to load affected modules.

This only goes partway to fixing this, since the final thing I haven't figured out how to fix correctly is `nbjs.json`, which unfortunately sits below the server-root `static` directory (d'oh!) so can be loaded by the minifier build, but not by requirejs in an extension-templated page. Not sure what the best approach to fix that would be - I'd go for moving it into the static directory, but don't know enough (anything!) about the i18n mechanics to know whether that might cause other problems. Any thoughts appreciated - @JCEmmons I guess you're the most appropriate to ask about the i18n stuff?


rgbkrk:
@Carreau @gnestor we'll want to make another release soon after because of this one

PR #2899, opened 2017-10-04
[MRG]Fixes #740

### LINKED ISSUE #740: Remove debug statements introduced in #729
Logging of debug statements was added in #729 in order to aid with debugging an issue with strange file change messages outlined in #484.

This is a reminder to remove these debug statements once the issue in #484 has been resolved.

cc/ @ellisonbg @Carreau @jdfreder @fperez

ragnerok:
Removed debug statements for #484

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -75,7 +75,7 @@ define([
      * Contains and manages cells.
      * @class Notebook
      * @param {string}          selector
-     * @param {object}          options - Dictionary of keyword arguments.  
+     * @param {object}          options - Dictionary of keyword arguments.
      * @param {jQuery}          options.events - selector of Events
      * @param {KeyboardManager} options.keyboard_manager
      * @param {Contents}        options.contents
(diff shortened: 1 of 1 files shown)


ragnerok:
Sorry, it seems like my text editor removed indents from empty lines, if it's a problem I can add them back

ragnerok:
Also just wanted to add that according to the [comment](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/notebook.js#L2735) a 0.5 sec delay has been allowed due to the bug, therefore should I remove it and the console warning

takluyver:
This variable is also now used in a console warning if the notebook has been modified on disk. If we're going to remove it, we'll need to remove it from there too. I'm not sure if we should, though - we worked around some problems with the changed-on-disk code, but I wouldn't be surprised if there are more lurking there. Storing a bit of info for debugging costs almost nothing.

ragnerok:
Yeah, makes sense to not remove the console warning and the variable ! Although I think the 0.5 sec delay should be removed as stated in my earlier comment.

takluyver:
We only just put the 0.5s fuzz-factor in there to deal with issues users were reporting! The underlying cause is definitely something to do with inconsistencies on network filesystems, but it's not something we can fix, so the 0.5s is a workaround.

takluyver:
cc @Carreau - do you think it's worth clearing this up, or just leave the debugging stuff in place and close #740?

ragnerok:
I think if the debug statements are to be left, at least the comment `debug 484` should be removed.

ragnerok:
@takluyver so what should I do now ?

takluyver:
Sorry for the delay. I think we can remove the 'debug 484' comments and put in a more general, descriptive comment, saying that it's there for debugging issues with the modified-on-disk check.

If you could do that in a fresh branch with your editor configured not to strip trailing whitespace, that would be good. Thanks!

ragnerok:
Ok I'll close this then

PR #2911, opened 2017-10-09
Improve Edit/View behavior

gnestor:
Follow up from https://github.com/jupyter/notebook/issues/2203#issuecomment-334607372

This makes the default path `edit` for files so that all files can edited. If a file is binary or otherwise uneditable, a user can still edit it and they will see the following error:

```
Error: [FILE] is not UTF-8 encoded
Saving is disabled.
See Console for more details.
```

They're browser won't crash or anything, so no harm, no foul.

If a file's MIME type matches `text/html` or the whitelist of viewable extensions (a list of browser-supported extensions borrowed from https://support.google.com/chromebook/answer/183093?hl=en), then it will use the `view` path by default.

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -721,17 +721,9 @@ define([
             icon = 'running_' + icon;
         }
         var uri_prefix = NotebookList.uri_prefixes[model.type];
-        if (model.type === 'file' && !this._is_editable(model))
-        {
-            uri_prefix = 'files';
-        }
         if (model.type === 'file' && this._is_viewable(model))
         {
             uri_prefix = 'view';
-        }
-        if (model.type === 'file' && this._is_editable(model))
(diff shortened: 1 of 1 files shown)


billgreenwald:
Sorry if this isnt the right thread for this question:

Since upgrading to jupyterhub 5.2.0, the default option for .pdf's switched from view to edit.  Is this intentional?

Separately, where is this "white list" located, so that I could manually tell the hub to view .pdf's instead of edit them.

Thanks!

takluyver:
(i'm guessing you mean jupyter notebook - *jupyterhub* is on version 0.8.1)

The list of viewable filetypes is here:
https://github.com/jupyter/notebook/blob/bbf0ead68f91ef44c5a57f3cdc83587ab06d4c80/notebook/static/tree/js/notebooklist.js#L558-L561

PDF should probably be on it.

PR #2916, opened 2017-10-10
allow disabling offline message buffering

minrk:
escape hatch if the new buffering (#2871) is causing anyone trouble


takluyver:
Makes sense to have a way to turn it off just in case.

gnestor:

rgbkrk:
Smart.

rgbkrk:
Thus far my deployment hasn't run into any trouble, users have been pleasantly surprised.

PR #2917, opened 2017-10-10
register contents_manager.files_handler_class directly

minrk:
rather than trying to call one handler from another, which is unreliable and can cause misbehavior.

closes jupyter/jupyter#292

### DIFF
--- a/notebook/files/handlers.py
+++ b/notebook/files/handlers.py
@@ -12,12 +12,19 @@ except ImportError: #PY2
     from base64 import decodestring as decodebytes
 
 
-from tornado import web, escape
+from tornado import web
 
 from notebook.base.handlers import IPythonHandler
 
+
 class FilesHandler(IPythonHandler):
-    """serve files via ContentsManager"""
+    """serve files via ContentsManager
+
+    Normally used when ContentsManager is not a FileContentsManager.
+
+    FileContentsManager subclasses use AuthenticatedFilesHandler by default,
+    a subclass of StaticFileHandler.
+    """
 
     @web.authenticated
     def head(self, path):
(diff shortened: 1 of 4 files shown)


notebook/files/handlers.py
```diff
@@ -12,12 +12,19 @@
...
 
+
 class FilesHandler(IPythonHandler):
-    """serve files via ContentsManager"""
+    """serve files via ContentsManager
+
+    Normally used when ContentsManager is not a FileContentsManager.
+
+    FileContentsManager subclasses use AuthenticatedFilesManager by default,
+    a subclass of StaticFileManager.
```
takluyver: "AuthenticatedFilesManager" and "StaticFileManager" should be "...Handler", I think?

notebook/notebookapp.py
```diff
@@ -299,7 +299,8 @@ def init_handlers(self, settings):
         handlers.extend(load_handlers('services.kernelspecs.handlers'))
         handlers.extend(load_handlers('services.security.handlers'))
         handlers.extend(load_handlers('services.shutdown'))
-        
+        handlers.extend(settings['contents_manager'].get_extra_handlers())
```
takluyver: This is being added after the default files handler (the one which uses the contents manager). Doesn't that mean that this one (using static files) will never actually be used?

ebraminio:
Thank you @minrk, just for the sake of curiosity, do you know why this could cause more issue on Safari? Thanks

notebook/notebookapp.py
minrk: I removed the global registration of the default files handler [here](https://github.com/jupyter/notebook/pull/2917/files#diff-aa5d3e97b16f9c0253d881fdec65c8c3R78), so this is where it gets registered, whichever class it is. On the base `ContentsManager`, it will be the generic `FilesHandler`, but on `FileContentsManager` subclasses it will be `AuthenticatedFileHandler`.

notebook/files/handlers.py
minrk: I got Managers on the mind. Fixed.

minrk:
@ebraminio I have no idea why I can only reproduce this in Safari. My guess is that it's a coincidence of how it implements fetching and/or caching of requests that makes it more likely to trigger.

notebook/notebookapp.py
minrk: The relevant change [below](https://github.com/jupyter/notebook/pull/2917/files#diff-c604f4531aba170341bbf6e9a155d5daR136) is that the base ContentsManager has `files_handler_class = FilesHandler` by default, so `get_extra_handlers` always returns something unless a ContentsManager explicitly overrides `files_handler_class = None`, in which case the `/files/` endpoint is disabled.

takluyver: Got it, thanks.

ebraminio:
Confirmed the fix, thanks!

PR #2919, opened 2017-10-10
Fix some errors caused by raising 403 in get_current_user

minrk:
get_current_user is called in a few places that really shouldn’t raise

move the raising to `get_login_url`, which is called in `@web.authenticated`, where we want to replace redirect logic with 403.

@gnestor this fixes a regression in the 5.2 rc (introduced in #2853)

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -104,6 +104,9 @@ class AuthenticatedHandler(web.RequestHandler):
         For example: in the default LoginHandler, if a request is token-authenticated,
         origin checking should be skipped.
         """
+        if self.request.method == 'OPTIONS':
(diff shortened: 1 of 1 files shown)


notebook/base/handlers.py
```diff
@@ -507,7 +516,6 @@ def options(self, *args, **kwargs):
                         'accept, content-type, authorization, x-xsrftoken')
         self.set_header('Access-Control-Allow-Methods',
                         'GET, PUT, POST, PATCH, DELETE, OPTIONS')
-        self.finish()
```
minrk: finish is already called on every method, no need to include it explicitly

gnestor:
Thanks @minrk!

PR #2920, opened 2017-10-10
allow token-authenticated requests cross-origin by default

minrk:
we already apply this logic in our server-side checks, but browsers check `Access-Control-Allow-Origin` headers themselves as well, meaning that token-authenticated requests can’t be made cross-origin without CORS headers from browsers, only scripts.

This makes default browser and server-side origin checks consistent

</del>includes #2919 to avoid merge conflicts</del>


minrk:
My conflict-avoidance strategy was not successful, apparently. Rebased.

takluyver:
I'm never entirely sure about these CORS issues, but I think I understand the basic idea, and I'm happy to trust the two of you. Do you want anyone else to review it? Do you want it in for 5.2?

gnestor:
We're going to try to release 5.2.1. Should this be included?

PR #2921, opened 2017-10-10
allow setting token via JUPYTER_TOKEN env

minrk:
more convenient for certain deployments than CLI arguments


takluyver:
If it's set in both the config file and the environment, the config file will win, right? It's probably not a big problem, but I'd generally expect the environment variable to have precedence.

minrk:
Yes. There's a *huge* discussion about supporting env values better over on traitlets, but this is what's available to us for now, without doing some extra shenanigans.

takluyver:
Wrong button

PR #2922, opened 2017-10-11
'Play' icon for run-cell toolbar button

takluyver:
Based on @ellisonbg's research in gh-2914, most tools with similar functionality to run a piece of code use either the 'play' icon, or something custom designed (RStudio and Spyder are in the latter camp).

The analogy with audio and video does suggest that 'step forward' is more similar to running one cell, but it's not clear whether users are actually thinking about it in those terms.

I'm opening this for discussion, not for immediate merge. cc @SylvainCorlay who originally changed it to 'step forward', and @ssanderson and @rgbkrk who discussed that PR (#201). Sylvain did say at the time that he had seen users confused by it.


blink1073:
For visual reference:

The proposal is to change from the current:
<image src="https://user-images.githubusercontent.com/2096628/31440227-c5a1003e-ae54-11e7-9499-b46cbae537ee.png" width=64>

to:
<image src="https://user-images.githubusercontent.com/2096628/31440259-e114060e-ae54-11e7-9530-5766fcd1c614.png" width=64>

blink1073:
Spyder's Run Cell Icon:

image

blink1073:
And RStudio:

image

takluyver:
Thanks @blink1073!

I hadn't noticed that Spyder's icon incorporates the 'step forward'  icon until I saw it enlarged.

ccordoba12:
@SylvainCorlay designed our icons too.

rgbkrk:
Yeah I think we should go back to the play button.

gnestor:
Play button is more simple. If both jupyterlab and nteract are using it, then I think that is enough reason to switch to using it, too. 

There is another discussion about adding a "restart and run all" button to the toolbar at https://github.com/jupyter/notebook/pull/2965 and what the icon for that should be. The step-forward icon does imply "run and stop" whereas the play icon doesn't. Please chime in there 👍

ellisonbg:
I am in favor of the play button with no text. See https://github.com/jupyter/notebook/pull/2965 for related points about toolbar design.

Zsailer:
Circling back here, even if it's 2.5 years later 😆. Looks like everyone is in agreement!

I'll give this a couple of days to see if there is any objections—then, I'll plan to merge. Thanks, all!

rgbkrk:
Thanks for bumping this thread @Zsailer!

Issue #2858, opened 2017-09-23
 .ipy_checkpoints doesn't respect umask


dclong:
I used the following configuration in `/etc/jupyter/jupyter_notebook_config.py`. Newly created notebooks have the right permission (664), however the `.ipy_checkpoints` folders do not.
```
import os
os.umask(0o002)
```

dclong:
And `setfacl` doesn't make `.ipynb_checkpoints` directories have the right permission, either.

dclong:
I quickly checked the code and found the following piece of code which might be the root cause. 
```
./services/contents/filecheckpoints.py:            ensure_dir_exists(cp_dir)
```
The function `ensure_dir_exists` is in ipython_genutils/path.py. It has a parameter `mode` with default value `0o755`. Passing the right mode to `ensure_dir_exists` might resolve the issue.

takluyver:
I suspect that someone - possibly me - saw that the default for os.makedirs() was 0o777 and assumed that that it would actually create a world-writable directory by default. So I implemented a default equivalent to umask 0o022. I think we can probably let the default be determined by the umask.

dclong:
@takluyver 
That would resolve the issue.

takluyver:
See #2924

PR #2924, opened 2017-10-11
Rely on umask for default directory permissions

takluyver:
Closes gh-2858

@minrk this feels somewhat unsatisfactory. The function I want to change is in `ipython_genutils`, but we're not making any changes there. So do we need to extract it to each package that uses it to make similar fixes?

### DIFF
--- a/notebook/nbextensions.py
+++ b/notebook/nbextensions.py
@@ -24,7 +24,7 @@ from jupyter_core.paths import (
     jupyter_data_dir, jupyter_config_path, jupyter_path,
     SYSTEM_JUPYTER_PATH, ENV_JUPYTER_PATH,
 )
-from ipython_genutils.path import ensure_dir_exists
+from notebook.utils import ensure_dir_exists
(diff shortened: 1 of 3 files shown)


blink1073:
This [search](https://github.com/ipython/ipython/search?utf8=%E2%9C%93&q=ipython_genutils&type=) shows that `ipython` itself is no longer using `ipython_genutils` outside of `indent` and `dedent` in the docs generator.  Perhaps moving the functions to `jupyter_core` makes sense?

blink1073:
To be clear: I mean copying the functions currently used in this repo to `jupyter_core`, releasing `jupyter_core`, and removing the dependency of this repo on `ipython_genutils`.

takluyver:
That's a good idea. jupyter_core probably makes reasonable sense as a place for this.

takluyver:
jupyter/jupyter_core#113 would add it to jupyter_core.

takluyver:
Closing in favour of #3002.

PR #2930, opened 2017-10-12
Add 5.2.0 to changelog

gnestor:


rgbkrk:
It seems there are a few other items missing from the changelog, which I'm assuming might be my own fault for not adding the output buffering to `whatsnew.rst`:

https://github.com/jupyter/notebook/compare/5.1.0...5.2.0

gnestor:
The issue here is that those PRs were not marked with the 5.2 milestone. I should've double-checked. I just marked them and I will updated the changelog on Github and blog.

rgbkrk:
Thanks!

PR #2943, opened 2017-10-17
improve swagger API docs

mpacer:
as mentioned in #2413 our swagger docs are invalid. 

This gives an actual fully specified url, fixes typos around finding definitions, removes an invalid reference to a parameter inside of a schema declaration, moves a reference inside of a schema where it belongs.


rgbkrk:
Excellent!

PR #2947, opened 2017-10-18
Whitelist <a knowl="..."> for including knowls

takluyver:
At the workshop in Simula, @rbeezer is interested in using [knowls](https://www.aimath.org/knowlepedia/) in Jupyter. These work with a tag `<a knowl="foo.html">`,  which Javascript turns into a clickable thing to open a knowl.

The `knowl=` attribute is currently getting sanitized in markdown cells. It seems likely that it's safe in general; it doesn't do anything without the Javascript to process it. However, we should consider this carefully - if a Jupyter deployment includes the knowl JS to allow knowls in notebooks, could a maliciously coded knowl execute code on your kernel? It doesn't look like it's using an iframe by default.

### DIFF
--- a/notebook/static/base/js/security.js
+++ b/notebook/static/base/js/security.js
@@ -37,6 +37,7 @@ define([
         // disable the check
         // https://www.owasp.org/index.php/Script_in_IMG_tags
         ATTRIBS['img::src'] = 0;
+        ATTRIBS['a::knowl'] = 0;
         return caja.sanitizeAttribs(tagName, attribs, opt_naiveUriRewriter, opt_nmTokenPolicy, opt_logger);
     };
     


rbeezer:
@davidfarmer is the expert on the knowl JS.

David - Lots of interest here in knowls.  They **almost** work quite easily in a Jupyter notebook.  Right know the `knowl=` attribute gets stripped when used in a text (Markdown) cell.  The code change whitelists that attribute.  But I guess having the content of the knowl in an external file will provide an unanticipated avenue for adding content into a notebook.

@takluyver suggests wrapping the knowl in an `iframe` perhaps?  Not sure if that will degrade performance generally in other uses?

davidfarmer:
I don't think I understand the benefit of wrapping the knowl in
an iframe.  Is it the knowl link or the content that gets wrapped?

By default, the knowl content appears after the block element
containing the knowl link.  Is there some reason to do that differently
in a Jupyter notebook?

takluyver:
My concern is the following scenario:

You are running Jupyter modified to include the knowl Javascript, so that knowl `<a>` elements are automatically set up. I (with my evil hat on) write a knowl containing some Javascript, and put it on a public web-server configured to allow cross-origin requests. I make an interesting notebook which references my knowl, and send it to you. You open it up, start reading, and click on the knowl link. My evil Javascript gets to run in your notebook page, where it can send code to the kernel connected to the notebook. I can now take control of your computer.

I haven't tested this all the way through, but I have tested that Javascript in a knowl runs, and it attempts to load knowls from other domains. Even if it didn't load from other domains, I think there are still ways to exploit it - you're unlikely to look carefully at an HTML file buried in a git repository you've just cloned, for instance.

Putting the content in a sandboxed iframe would mean that the Javascript runs in a separate context from the notebook, so it can't talk to your kernel. This isn't entirely specific to Jupyter notebooks: it could matter on any page where the Javascript has access to some private resources (e.g. a login token) and knowls might come from an untrusted source.

davidfarmer:
I agree that this probem is not specific to Jupyter.

I could modify the knowls javascript to look at the knowl content and
react appropriately if it contained javascript.  Or make the output
be in an iframe.

But when the evil person sends you that interesting notebook/webpage,
how do you know it uses the knowls javascript that I maintain,
and not a nasty version?

rbeezer:
> But when the evil person sends you that interesting notebook/webpage, how do you know it uses the knowls javascript that I maintain, and not a nasty version?

As I am doing it now, and as I understand it (Thomas can correct me), you **must** load the Javascript with a cell that is visible and which the reader must conciously execute.  So you can see the URL and you must elect to use it.  I could send a screenshot later today if that would make it clearer.

takluyver:
Yep, that's right. When a notebook is untrusted (you haven't run the code), we don't allow it to run Javascript on load. The idea is that the user can at least see what the code is before they make a decision whether to execute it.

I'm imagining a scenario where the user or the sysadmin extends the notebook to load the knowls javascript in every notebook view. They have decided to trust the knowls javascript itself to run in the notebook context. But that shouldn't imply that the content of all knowls is also trusted.

(Aside: in the knowls docs, can I suggest that the example should load the Javascript over `https://`? I see that the aimath.org server redirects to https, but code.jquery.com does not, and it's best to specify https explicitly anyway.)

davidfarmer:
So, if the knowls code detects content that contains javascript,
then it should wrap the content in an iframe?  If that is sufficient,
then we just need to make sure it doesn't break existing cases that
have javascript (such as WeBWorK.  Are there others?)

I updated the (very old) instructions to use https.

rbeezer:
WeBWorK uses knowls pervasively to show/hide solutions and hints to problems, so that is one use I thought about overnight.  Not sure if they have a copy or grab AIM's version as needed.

In PreTeXt output, "staged" WeBWorK problems needed to go into an "iFrameResizer" gadget so they would behave inside knowls (iirc).  Not sure if an extra iframe will interfere with that.  Of course, Alex would be the one to consult.  Example is Exercise 2.2 here:

http://mathbook.pugetsound.edu/examples/webwork/sample-chapter/html/section-2.html

takluyver:
The ideal from a security point of view would be to put it in an iframe always; otherwise the code to detect Javascript becomes a security target (evil me wants to include Javascript while tricking the detector into giving it the all-clear).

When a user views an HTML file inside Jupyter, we put it in a sandbox like this:

```html
<iframe id="iframe" sandbox="allow-scripts" src="{{file_url}}"></iframe>
```

That allows Javascript to run inside it, but treats it as being on a separate domain to the notebook server, so it can't do notebook stuff. (`{{file_url}}` is a template variable)

> I updated the (very old) instructions to use https.

Thanks!

takluyver:
Heads up @davidfarmer - the instructions on https://www.aimath.org/knowlepedia/ are still using http (not https) urls to load the Javascript.

davidfarmer:
Thanks.  I had forgotten about that page, since it is not the one we
link to.  But I updated it anyway.

takluyver:
Thanks! What is the one you link to? That was the first result when I searched for knowls.

davidfarmer:
Same content, but copied when we switched to WordPress:

https://aimath.org/knowls-browsing/

takluyver:
Thanks @davidfarmer :-)

Given the security concerns around Javascript and transclusion, I'm going to shelve this PR for now.

PR #2949, opened 2017-10-18
Handle a compound extension in new_untitled

blink1073:
This allows file names with compound extensions (e.g. `.tar.gz`) to be properly incremented when creating a new file.  Previously we would end up with `untitled.tar1.gz`.

cf https://github.com/jupyterlab/jupyterlab/issues/3113

### DIFF
--- a/notebook/services/contents/manager.py
+++ b/notebook/services/contents/manager.py
@@ -4,7 +4,6 @@
 # Distributed under the terms of the Modified BSD License.
 
 from fnmatch import fnmatch
-import gettext
 import itertools
 import json
 import os
@@ -31,6 +30,7 @@ from traitlets import (
 )
 from ipython_genutils.py3compat import string_types
 from notebook.base.handlers import IPythonHandler
+
 
 copy_pat = re.compile(r'\-Copy\d*\.')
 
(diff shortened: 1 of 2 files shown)


notebook/services/contents/manager.py
```diff
@@ -317,22 +317,31 @@ def increment_filename(self, filename, path='', insert=''):
...
         Returns
         -------
         name : unicode
             A filename that is unique, based on the input filename.
         """
-        path = path.strip('/')
-        basename, ext = os.path.splitext(filename)
+        # Extract the full suffix from the filename (e.g. .tar.gz)
+        dirname = os.path.dirname(filename)
+        basename = os.path.basename(filename)
```
takluyver: This can be summarised as `dirname, basename = os.path.split(filename)`.

Also, I think we should leave in the `.strip('/')` that was there before, so that passing in `foo/bar/` gives you `foo/bar1`, not `foo/bar/1`.

notebook/services/contents/manager.py
```diff
@@ -317,22 +317,31 @@ def increment_filename(self, filename, path='', insert=''):
...
+        parts = basename.split('.')
+        basename = os.path.join(dirname, parts[0])
+        suffix = '.' + '.'.join(parts[1:])
+        if suffix == '.':
+            suffix = ''
```
takluyver: How about using [str.partition()](https://docs.python.org/3/library/stdtypes.html#str.partition):

```python
name, dot, ext = basename.partition('.')
basename = os.path.join(dirname, name)
suffix = dot+ext
```

blink1073: Thanks!  Updated.

notebook/services/contents/manager.py
```diff
@@ -317,22 +317,29 @@ def increment_filename(self, filename, path='', insert=''):
...
             if i:
                 insert_i = '{}{}'.format(insert, i)
             else:
                 insert_i = ''
-            name = u'{basename}{insert}{ext}'.format(basename=basename,
-                insert=insert_i, ext=ext)
-            if not self.exists(u'{}/{}'.format(path, name)):
+            name = u'{basename}{insert}{suffix}'.format(basename=basename,
+                insert=insert_i, suffix=suffix)
+            if not self.exists(os.path.join(path, name)):
```
takluyver: Ah, hang on. This function works with an API path, which is always `/` separated, rather than a platform-dependent native path. So using `os.path.split` and `os.path.join` is wrong here.

blink1073: Updated.  Looking again, `filename` should never have a directory in it.

takluyver:
Thanks, merging.

blink1073:
Ha, that was easy, thanks for your patience ;).

gnestor:

PR #2958, opened 2017-10-19
Add 'reason' field to JSON error responses

### LINKED ISSUE #2957: APIHandler.write_error should set `reason` field in JSON error response.
When the `@json_errors` decorator was removed, to be replaced with `APIHandler.write_error` (see #2853), the establishment of the `reason` field was not preserved.   Since `@json_errors` set that field and some notebook clients depend on its presence, it should be set in the compatible replacement.

Please see Issue #2942 for the _meandering details_ of how this conclusion came about.

kevin-bates:
During the deprecation/removal of the `@json_errors` decorator, the
`reason` field was not carried forward into the compatible replacement
method `APIHandler.write_error`.  This broke some client (tests) that
relied on that field's presence.

Fixes #2957.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -461,6 +461,8 @@ class APIHandler(IPythonHandler):
         message = responses.get(status_code, 'Unknown HTTP Error')
         reply = {
             'message': message,
+            # some clients expect 'reason' to exist and equate to status_code text
+            'reason': message,
         }
         exc_info = kwargs.get('exc_info')
         if exc_info:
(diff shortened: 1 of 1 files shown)


takluyver:
@kevin-bates looking back at the old decorator code before #2853, it set reason to `e.reason` when e is an HTTPError, and None otherwise. I'd suggest that we replicate that.

kevin-bates:
@takluyver - yeah, that's probably the best approach.  Change committed.

takluyver:
Thanks. I'll merge tomorrow if no-one says otherwise.

PR #2963, opened 2017-10-21
Config option to shut down server after N seconds with no kernels

takluyver:
This is something I was discussing with @minrk yesterday: allow the notebook server to shut down automatically after some time with no kernels active. This could be useful with nbopen: a notebook server can be started automatically when you double-click a notebook file, and then clean itself up automatically when you've finished using it.

This is not yet ready for merge: I think it should take all activity into account for the timeout, not just kernel activity, so it doesn't shut down while you're editing a file or using a terminal. It will probably be a few days before I have time to finish it off.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1070,6 +1070,14 @@ class NotebookApp(JupyterApp):
     rate_limit_window = Float(3, config=True, help=_("""(sec) Time window used to 
         check the message and data rate limits."""))
 
+    shutdown_no_kernels_timeout = Integer(0, config=True,
+        help=("Use a positive integer, to shut down the server after N "
+             "seconds with no kernels running. This can be used together with "
+             "culling idle kernels (MappingKernelManager.cull_idle_timeout) to "
+              "shutdown the notebook server when it's not in use. "
+              "Do not rely on this being accurately timed.")
+    )
+
     def parse_command_line(self, argv=None):
         super(NotebookApp, self).parse_command_line(argv)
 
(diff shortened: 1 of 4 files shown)


takluyver:
This is now ready for review: it considers kernels, terminals and API activity, so it shouldn't shut the server down if you're still doing anything.

You could configure it like this:

```python
c.MappingKernelManager.cull_idle_timeout = 600
c.NotebookApp.shutdown_no_activity_timeout = 600
```

With those options, if you close a notebook tab leaving the kernel running, it will cull the kernel after 10 minutes. After another 10 minutes, if you haven't done anything else, the server will shut down.

There is currently no option to cull idle terminals, so if you leave a terminal open, the server will never shut down. We could add an option for that separately.

takluyver:
@Carreau @gnestor @minrk ?

minrk:
Makes sense. Only thing I'd like to add is that there's already an API endpoint for retrieving activity info in `services/api/handlers.py`. It would be good to get this updated `last_activity` to the field reported there, as well, so the two stay in sync. I'm not sure what that means wrt where the last_last_activity calculation belongs. At the app level makes sense, but at the same time accessing a method on the app object from a handler doesn't seem right.

gnestor:
@minrk To clarify: Does the `last_activity` provided by `services/api/handlers.py` include kernel and terminal activity? Can the app get the `last_activity` value from there? That would simplify this quite a bit 👍

minrk:
@gnestor it should be the last activity of any kind on the server as a whole i.e. exactly the same value added here in `app.last_activity`.

takluyver:
It doesn't currently include terminals, because I only added terminal activity timestamps in this PR. I'll look for a way to unify them.

takluyver:
I have unified them in a method on the web application object. This seemed a good compromise, as it's easily accessible both from the `NotebookApp` and from the API handler.

gnestor:
@minrk Care to take one last look and merge?

PR #2967, opened 2017-10-23
Fix ANSI inverse

mgeier:
The "inverse" escape sequence was implemented in #2186, but not by actually inverting foreground and background.

@tonycpsu Since you created the original "inverse" implementation, can you please check if that's OK for you?

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -380,6 +380,12 @@ define([
                     case 22:
                         bold = false;
                         break;
+                    case 24:
+                        underline = false;
+                        break;
+                    case 27:
+                        inverse = false;
+                        break;
                     case 30:
                     case 31:
                     case 32:
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -247,8 +247,6 @@ define([
                 fg += 8;  // Bold text uses "intense" colors
             }
             if (inverse) {
-                if (fg.length === 0) { fg = [0, 0, 0]; }
-                if (bg.length === 0) { bg = [255, 255, 255]; }
                 [fg, bg] = [bg, fg];
             }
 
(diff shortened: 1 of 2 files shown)


tonycpsu:
That looks like a very elegant way to "fake" inverse support with CSS!

However, one problem I'm seeing in my output is that it appears to be assuming that the background color (which becomes the foreground of the inverse) is white when it may not be.  Example:

here

I suspect there's no reliable way to deal with that given the need to precompute styles ahead of time and not knowing what the background will be...  But if there must be a default, I would think black foreground would be the most sensible, since the ANSI stuff tends to be used for terminal output, and terminals have traditionally used black as the default background.

mgeier:
@tonycpsu Thanks for checking this out!

It indeed assumes that the default text color is black and the background is white.
I chose this because that's how I thought it is typically used in the Jupyter notebook, which has a white background.
Where is your example screenshot from?

If I'm supposed to make this customizable, I could try to define two new CSS classes like `ansi-default-fg` and `ansi-default-bg` (or probably somebody can come up with better names?) and use those for "inverse" where foreground or background are not defined.
Those CSS classes could then be overwritten whenever white background is not appropriate.

tonycpsu:
>  Where is your example screenshot from?

It's from a notebook (console output from [cdiff](https://pypi.python.org/pypi/cdiff) but I use ```custom.css``` to make the output look more like an "old school" console.

Relevant ```custom.css``` in my case:

```css
.output {
    font-family: Menlo,Consolas,Lucida Console,monospace;
    font-size: 9pt;
}

div.output_area pre {
    font-family: Menlo;
    background: #202020;
    color: #37f14a;
}

div.output_area pre::selection {
    color: #202020;
    background: #37f14a;
}
div.output_area pre::-moz-selection {
    color: #202020;
    background: #37f14a;
}
```

I like your idea of new classes for the default fg/bg that I could just override in ```custom.css```.

mgeier:
@tonycpsu I've added a commit with 2 new overridable CSS classes: ab7b19dd9910f15610f7cc0fd93ee36dc1bf7886

By default this still assumes a white background, but you can override `ansi-default-inverse-fg` and `ansi-default-inverse-bg` in your `custom.css`.

Does this work for you?

tonycpsu:
Looks perfect!  Thanks so much for doing this!

gnestor:
@mgeier Is this ready to merge?

mgeier:
@gnestor Thanks for the review!

Yes, I think this is ready. I've rebased this PR in order to get rid of the AppVeyor error.

The only remaining question is if you are OK with adding two additional CSS classes and if the names `ansi-default-inverse-fg` and `ansi-default-inverse-bg` are OK.

gnestor:
Yes and yes. Thanks @mgeier!!

PR #2969, opened 2017-10-23
Work on loading UI translations 

takluyver:
To test: set `zh-CN` to the most preferred language in browser preferences, then launch the notebook with the LANG environment variable set, i.e. `LANG=zh_CN jupyter notebook`.

screenshot from 2017-10-23 16-01-42

- Only transfers translations for languages the user wants.
- (For strings used in JS) can handle multiple fallbacks - e.g. an Afrikaans speaker could request Afrikaans > Dutch > English, and get each translation in the best available language.

Writing the translation JSON into the template and storing it as a JS global seems a bit ugly, but it ensures that the data is there before any Javascript strings are used, without either bundling data for all languages into the JS bundle, or trying to do clever things with requirejs.

The biggest thing that I'm not happy about is that the language for the JS is determined by the browser, and the language for the template strings by an environment variable in the server. If you only set one, you see half the strings translated, with no obvious reason for the difference. Using the browser header seems better, and it allows for multiple fallbacks, but there isn't an obvious way to use that with `jinja2.ext.i18n`.


takluyver:
The remaining failure on Appveyor appears to be an unrelated problem with installing nbconvert from conda-forge.

JocelynDelalande:
> To test: set zh-CN to the most preferred language in browser preferences, then launch the notebook with the LANG environment variable set, i.e. LANG=zh_CN jupyter notebook.

Just tested, it does work. Thanks :-)

Note : there is no fallback within a same lang (maybe this is normal behavior) : if I set browser lang to `zh`, it will never load the `zh-CN` translation.

takluyver:
Yeah, I think we should probably rename `zh_CN` to just `zh`, and introduce an automatic fallback when loading so that if you specify `zh_CN`, it will use `zh` as well. But I think those can be in future PRs: I just wanted to get the translations working first.

gnestor:
Great work, @takluyver!! I tried testing and I'm seeing the translations that depend on the browser's language (working in both Chrome and Firefox) but I'm not seeing the translations that depend on the environment variables. I'm using `LANG=zh_CN jupyter notebook` to run the notebook server... 🤔

takluyver:
I'm not sure what's going on there - maybe there are other environment variables besides LANG that affect it? I'm inclined to get this merged and keep iterating.

gnestor:
Ok let's do it 👍

hikui:
Hi @takluyver, the translation doesn't work in the release version. I noticed that `.mo` and `nbjs.json` files are gitignored and there's no corresponding compiling process in the setup.py?

I had to manually follow the README and compiled my own version of end files to make the translation work. Perhaps we should keep .mo and nbjs.json files?

takluyver:
There's an issue for someone to integrate the steps to build the translations: https://github.com/jupyter/notebook/issues/3102

PR #2972, opened 2017-10-24
update version to dev version for master to not overwrite editable installs

mpacer:
Return to a dev version number so that it gets precedence when doing `pip install -U ___` for other packages that have the notebook as a dependency.

### DIFF
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -9,5 +9,5 @@ store the current version info of the notebook.
 
 # Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**. 
 
-version_info = (5, 2, 0)
+version_info = (5, 3, 0, ".dev0")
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! @gnestor , note that we typically do this as part of the release process. No harm done, though :-)

gnestor:
Ah yes, I knew that. Thanks @mpacer 👍

mpacer:
makes me think bumpversion might be useful here too :P

takluyver:
Yup, it might be!

I'm also working on a tool for automating release processes (wrapping tools like bumpversion and git). It's still in early stages, but if you're interested in playing around: https://pypi.python.org/pypi/modeltee

takluyver:
(PS requires Python 3.6, because experimental ;-)

gnestor:

gnestor:

PR #2981, opened 2017-10-25
Fix nbconvert handler

### LINKED ISSUE #2974: server error when downloading PDF via LaTeX
I have installed a clear version of Python 3.6 and ran the following `$ pip install notebook` to install jupyter on my Windows machine.

When I try to download my notebook as PDF via LaTeX I receive the following error:

errorjupyter

nbconvert was installed as a dependency with jupyter. Furthermore, I have MikTex installed for LaTeX.

Worth mentioning, I have tried this with different python versions and on my Mac as well. 

Hopefully someone could shed some light on this error.

---- 
    
Originally opened as jupyter/nbconvert#697 by @MustiDarsh, migration requested by @mpacer

mpacer:
closes #2974, 

Other small improvements while I was doing this:

Building the resources dict separately from invocation is more readable. 

`resources['metadata']['name']` now will use the notebook level metadata.title value as passed through to name (if present) and fall bad to stripping the extension using os.path.splitext. This notion of `name` is only used in the html and slides exporters to create a `<title/>` tag.

### DIFF
--- a/notebook/nbconvert/handlers.py
+++ b/notebook/nbconvert/handlers.py
@@ -103,7 +103,7 @@ class NbconvertFileHandler(IPythonHandler):
 
         # create resources dictionary
         mod_date = model['last_modified'].strftime(text.date_format)
-        nb_title = nb.metadata.get("title","") or os.path.splitext(name)[0]
+        nb_title = os.path.splitext(name)[0]
 
         resource_dict = {
             "metadata": {
--- a/notebook/nbconvert/handlers.py
+++ b/notebook/nbconvert/handlers.py
@@ -97,20 +97,29 @@ class NbconvertFileHandler(IPythonHandler):
             # not a notebook, redirect to files
             return FilesRedirectHandler.redirect_to_files(self, path)
 
+        nb = model['content']
+
         self.set_header('Last-Modified', model['last_modified'])
+
+        # create resources dictionary
+        mod_date = model['last_modified'].strftime(text.date_format)
+        nb_title = nb.metadata.get("title","") or os.path.splitext(name)[0]
+
+        resource_dict = {
+            "metadata": {
+                "name": nb_title,
+                "modified_date": mod_date
+            },
+            "config_dir": self.application.settings['config_dir']
+        }
+
+        if ext_resources_dir:
+            resource_dict['metadata']['path'] = ext_resources_dir
 
         try:
             output, resources = exporter.from_notebook_node(
-                model['content'],
-                resources={
-                    "metadata": {
-                        "name": name[:name.rfind('.')],
-                        "modified_date": (model['last_modified']
-                            .strftime(text.date_format)),
-                        "path" : os_path
-                    },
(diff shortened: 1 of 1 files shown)


mpacer:
I can't restart the Appveyor build… it looks like that error might just go away since i just verified that the conda-forge package does exist…

But for some reason I can't log in to appveyor,

mpacer:
@gnestor Pdf exports were broken with 5.2 (#2974), so this should be prioritised in a patch release.

mpacer:
Can someone please add me to the list of those with permissions to restart appveyor builds? I don't understand why it doesn't just use github permissions, but it seems to not do that.

gnestor:
@Carreau Do you have those powers?

mpacer:
@gnestor I really don't think this should wait until 5.3… instead aiming at 5.2.1 or are you using 5.3 to also organise patch releases for 5.2.x?

Carreau:
> @Carreau Do you have those powers?

Not sure. I tried something, @mpacer you should receive an invite I guess from AppVeyor, if it does not work I think we need to as Kyle.

gnestor:
We don't have a milestone for 5.2.1 so I marked it as 5.3. I'll create one now and shoot for releasing 5.2.1 tomorrow.

notebook/nbconvert/handlers.py
```diff
@@ -76,40 +76,50 @@ def get_exporter(format, **kwargs):
...
             # not a notebook, redirect to files
             return FilesRedirectHandler.redirect_to_files(self, path)
 
+        nb = model['content']
+
         self.set_header('Last-Modified', model['last_modified'])
 
+        # create resources dictionary
+        mod_date = model['last_modified'].strftime(text.date_format)
+        nb_title = nb.metadata.get("title","") or os.path.splitext(name)[0]
```
takluyver: Not convinced about the title from notebook metadata taking precedence over the filename used to identify it. The notebook metadata is already available in the template, it doesn't need to be added to resources.

gnestor:
The 5.2.1 branch is ready for release so we're just waiting on this PR 👍

notebook/nbconvert/handlers.py
mpacer: Fine, this was a fix that I had added on the nbconvert side for the LaTeX interpreter and was just trying to keep it consistent here. But I'm not wedded to it.

mpacer:
fixed!

gnestor:
I tested and export to PDF is working 👍 

@mpacer Shall I merge?

minrk:
@gnestor 🚢

gnestor:

gnestor:
I backported this manually to 5.2.1 👍

mpacer:
Apologies! didn't see you ask me if it could be merged — I had intended it to be done, yes. Any thoughts as to why the meseeksdev backport didn't work?

gnestor:
No prob! No I don't have access to the meseeksdev logs but there must've been a merge conflict

PR #2993, opened 2017-10-29
Prevent uneditable cells from having their type changed

### LINKED ISSUE #2992: Non-editable cells can have their cell type changed
It seems that even when cells have the `"editable": false` in their metadata, their cell type can still be changed (even though the source cannot). This seems counterintuitive to me and causes problems with nbgrader as well (see https://github.com/jupyter/nbgrader/issues/878 ). I will make a PR shortly to address the issue.

jhamrick:
Fixes #2992 

This both prevents cells from having their types changed, and also disables the cell type selector for uneditable cells.

### DIFF
--- a/notebook/static/notebook/js/maintoolbar.js
+++ b/notebook/static/notebook/js/maintoolbar.js
@@ -80,11 +80,17 @@ define([
             .append(multiselect);
         this.notebook.keyboard_manager.register_events(sel);
         this.events.on('selected_cell_type_changed.Notebook', function (event, data) {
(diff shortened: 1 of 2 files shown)


jhamrick:
Thanks for reviewing! I do, restarting it again... it failed before too (different error though) and I already restarted it once. The errors seem unrelated though. Maybe appveyor is a bit wonky today.

blink1073:
They were wonky on Friday as well.

jhamrick:
Still failing 😒

minrk:
Dealt with the appveyor failure in #2999

Issue #2937, opened 2017-10-14
Jupyter notebook stop - permission denied


ann-qin-lu:
Hi, 
I just upgraded notebook to 5.2.0 and tried `_jupyter notebook stop 8888_`, but got:     
os.kill(server['pid'], signal.SIGTERM)
PermissionError: [WinError 5] Access is denied

My system is win10
I tried to run the cmd in admi mode, but still not worked.

Appreciate any helps.
Thanks!

takluyver:
I think that feature may be written in a way that doesn't work at all on Windows. Sorry,  most of the core developers use Linux & Mac systems.

We should change it to try requesting a shutdown using the shutdown HTTP endpoint first, and only resort to signals if it hasn't shut down in a few seconds.

PR #3000, opened 2017-10-31
Send HTTP shutdown request on 'stop' subcommand

takluyver:
Signals don't work on Windows. This tries the HTTP shutdown request first. On Unix, it will try SIGTERM after 5 seconds, and SIGKILL after another 5, if the server didn't already exit.

Closes gh-2937

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -24,6 +24,7 @@ import signal
 import socket
 import sys
 import threading
+import time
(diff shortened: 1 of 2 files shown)


xiaosuyu1997:
image
Meet a problem when trying to stop jupyter notebook on a server. It seems the httpRequest sent has some permission issue as I'm not root on this server?

Issue #1530, opened 2016-06-11
Shutdown/Restart Server menu item: Feature Request


GadgetSteve:
Many times when running Jupyter notebooks I find myself wishing for a quick clean mechanism to cleanly shutdown &/or restart the server without having to go to the command window that it is running in and hit control-C and without the risk of not having saved and shutdown all kernals that may be running at the time.

Giving it some thought this feature needs a mechanism for disabling &/or hiding it on Jupyter servers that the user should not be able to shutdown or as a configuration item that on a specific machine this is enabled or not so that on a server the configuration would, _normally_, be **no** shutdown/restart entries - _possibly with the command line option of starting a server with them enabled for testing_ - but on a personal machine the option could be set to enable them, and possibly to hide the command window.

Carreau:
I **think** we can consider that covered by jupyterhub:

<img width="761" alt="screen shot 2016-06-16 at 16 32 54" src="https://cloud.githubusercontent.com/assets/335567/16136519/043b06a0-33e0-11e6-9655-d1fd82df8c36.png">

Which is reasonable to use even on a local machine. 

Though it will **not** warn you if you have multiple notebook/kernels running.

GadgetSteve:
@Carreau I suspect that many users would find running jupyterhub a little intimidating - certainly at work I only use Jupyter a part of the time so would not wish to have the hub running all of the time.

takluyver:
I do think it makes sense to have a way to shut down notebook servers besides interrupting them in a terminal. [Nbmanager](https://github.com/takluyver/nbmanager) is one tool I wrote to do this. It probably needs some updates by now to work with Jupyter, but that should be easy enough if anyone's interested.

JamiesHQ:
@GadgetSteve : We're doing a little housekeeping on our issue log and noticed this thread from 2016. Did you try [nbmanager](https://github.com/takluyver/nbmanager) and were you able to find a solution to your issue? Please let us know! We'll close this issue if you found what you need. thanks!

GadgetSteve:
@JamiesHQ : Thanks for pointing out nbmanager but I would have expected server shutdown to be a part of the core functionality of Jupyter rather than an external utility, or using Control-C in the shell/command window.
@takluyver : It is helpful that nbmanager shows the users currently running notebooks - I wonder if it wouldn't be possible to hook it into a button next to the jupyter logout, (with it defaulting to the current server selected).

takluyver:
[...]

PR #3004, opened 2017-10-31
Add shutdown button option

gnestor:
Closes https://github.com/jupyter/notebook/issues/1530

Introduces a new option `--NotebookApp.shutdown_button` which will add a button to the dashboard header that when clicked will shutdown the notebook server.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1025,8 +1025,8 @@ class NotebookApp(JupyterApp):
     def _update_mathjax_config(self, change):
         self.log.info(_("Using MathJax configuration file: %s"), change['new'])
         
-    shutdown_button = Bool(False, config=True,
-        help="""Whether or not to display a shutdown server button in the dashboard"""
+    shutdown_button = Bool(True, config=True,
+        help="""If True, display a button in the dashboard to shutdown the server"""
     )
 
     contents_manager_class = Type(
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -273,6 +273,7 @@ class NotebookWebApplication(web.Application):
             websocket_url=jupyter_app.websocket_url,
             mathjax_url=jupyter_app.mathjax_url,
             mathjax_config=jupyter_app.mathjax_config,
+            shutdown_button=jupyter_app.shutdown_button,
             config=jupyter_app.config,
             config_dir=jupyter_app.config_dir,
             allow_password_change=jupyter_app.allow_password_change,
@@ -1023,6 +1024,10 @@ class NotebookApp(JupyterApp):
     @observe('mathjax_config')
     def _update_mathjax_config(self, change):
         self.log.info(_("Using MathJax configuration file: %s"), change['new'])
+        
+    shutdown_button = Bool(False, config=True,
+        help="""Whether or not to display a shutdown server button in the dashboard"""
+    )
 
     contents_manager_class = Type(
         default_value=LargeFileManager,
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -273,7 +273,7 @@ class NotebookWebApplication(web.Application):
             websocket_url=jupyter_app.websocket_url,
             mathjax_url=jupyter_app.mathjax_url,
             mathjax_config=jupyter_app.mathjax_config,
-            shutdown_button=jupyter_app.shutdown_button,
+            shutdown_button=jupyter_app.quit_button,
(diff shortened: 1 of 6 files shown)


takluyver:
Aha, you beat me to it, thanks :-)

- I'd say it should be enabled by default, otherwise basically no-one will ever know about it.
- I don't entirely like 'shutdown' as a user-facing name, because that sounds like shutting down the computer. But I don't know how we make it clear that this is something more than closing the browser tab it's in.
- As shutting down is a destructive action, should we give the user a chance to change their mind? I don't like 'are you sure' dialogs, but it could turn into a countdown: 'Shutting down in 10 seconds... *Cancel*'
- How would this interact with the shutdown buttons in Jupyterhub? @minrk ?

This also reminds me that I don't think we should be showing a 'logout' button when the user is authenticated by a token, since from their perspective they're not 'logged in'.

minrk:
> How would this interact with the shutdown buttons in Jupyterhub?

From JupyterHub's perspective, this would mean that the server died and the Hub would take a finite (generally short) period of time to notice. It would probably be best in the future for SingleUserServer to override the method to trigger shutdown via the Hub API, so if it's easy to override that would be ideal.

gnestor:
> I don't entirely like 'shutdown' as a user-facing name, because that sounds like shutting down the computer. But I don't know how we make it clear that this is something more than closing the browser tab it's in.

How about "Disconnect"?

> It would probably be best in the future for SingleUserServer to override the method to trigger shutdown via the Hub API, so if it's easy to override that would be ideal.

Seems like it should be simple to override. Here is the shutdown handler: https://github.com/jupyter/notebook/blob/master/notebook/services/shutdown.py

takluyver:
I like disconnect even less ;-). Disconnect implies something will remain to which you can reconnect later.

GadgetSteve:
Since this is not shutting down the machine and is not disconnecting how about "HALT JUPYTER" or "HALT SERVICE" as a label? Either should make it clear what is going on.

Regarding default enabling or not if the option is added to the configuration file options and included in the default generated by `jupyter notebook --generate-config` _plus mentioned in the documentation_ then the user/maintainer of the installation can choose to show or hide it according to their preferences.

takluyver:
Returning to this, I think we need to:

- Enable it by default
- Make it configurable for Jupyterhub (@minrk what would this involve? pointing it to a different URL for the POST request?)
- Show some indication on the page that the server has been shutdown, so it's clear that it worked.
- Leave the text as 'Shutdown' for now. I still don't love it as a name, but so far I like all the proposed alternatives less.

@gnestor do you have time to come back to this?

GadgetSteve:
Just a suggestion how about "Terminate" possibly with a hover of "Shutdown Jupyter Server Backend" this rather than "Shutdown" - it rightly sounds more drastic than close/exit/etc. but does not imply that the machine will shutdown. I like the idea of a countdown, _(especially if the length of the countdown is a configurable item)_, but would suggest that it would be a good idea to have a "**Now**" button as well as a "**Cancel**" - everybody on Windows swears regularly at the "Applying Updates: Do not switch off...." message when they are in a hurry to get their machine shutdown so that they can go home or catch a bus/train/flight and I would like to avoid being in the same category.

takluyver:
'Terminate' could work, though it sounds a bit violent ;-) Or maybe 'Quit'?

If we're trying to get this into 5.5, I'd probably leave the countdown for a future version to keep things simple.

takluyver:
I've rebased this, enabled the button by default, changed the label to 'Quit', made sure all the human readable text is marked for translation, and made it display a dialog on successful shutdown. The dialog looks pretty ugly, but it's better than not showing anything. We can improve on it later.

Min and I agreed that Jupyterhub can simply disable this button and show its own buttons. Shutdown via the hub involves a HTTP DELETE request, so it would have to have config options for both the URL and the HTTP verb.

gnestor:
@takluyver Thanks for doing this. I was just about to dig into it. I'm not sure if "Quit" is better than "Shutdown"... We are using "shutdown_button" as the name for the module, so I guess that we should rename that as well if we go with "quit." You're call. Feel free to merge 👍

takluyver:
My rationale is that 'shutdown' is clearer for people who understand the idea of a local HTTP server, and 'quit' is clearer for people who don't. But on reflection,  I think the config name should match what the button says. I'll update it.

takluyver:
I've renamed the config option to `quit_button`, but left it as shutdown in the code, because the API endpoint is called shutdown too.

PR #3006, opened 2017-11-01
adding options to run notebook with kernel gateway

### LINKED ISSUE #3005: Add option to run Notebook with Kernel Gateway
We propose to make 2 handlers in Jupyter Notebook configurable. Currently, Jupyter Notebook supports a configurable handler for Login. We want to make the following settings configurable at run-time:
	1. 'services.kernels.handlers'
	2. 'services.kernelspecs.handlers'
Additionally, if the Jupyter Community wishes, we could also add the files from the nb2kg demo to the Jupyter Notebook project so that others can use the integration just by changing the configuration. 

These changes will allow users to use Kernel Gateway with Notebook as shown in the nb2kg demo.

shiti-saxena:
resolves #3005

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -144,13 +144,14 @@ class NotebookWebApplication(web.Application):
 
     def __init__(self, jupyter_app, kernel_manager, contents_manager,
                  session_manager, kernel_spec_manager,
-                 config_manager, log,
+                 config_manager, kernel_handlers, kernelspecs_handlers, log,
                  base_url, default_url, settings_overrides, jinja_env_options):
 
 
         settings = self.init_settings(
             jupyter_app, kernel_manager, contents_manager,
-            session_manager, kernel_spec_manager, config_manager, log,
+            session_manager, kernel_spec_manager, config_manager,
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -977,6 +982,18 @@ def _update_mathjax_config(self, change):
 
     info_file = Unicode()
 
+    kernel_handlers = Unicode(
+        default_value='services.kernels.handlers',
+        config=True,
+        help=_('The kernel handlers to use.'),
+    )
+
+    kernelspecs_handlers = Unicode(
```
minrk: nit: singluar 'kernelspec_handlers' to match 'kernel_handlers'

shiti-saxena:
@minrk It is possible to modify the `load_handlers` method to allow overriding any handler as currently it adds `notebook.` prefix. But I wasn't sure if we would want to expose that to users. Should I update the PR to do that as well?

shiti-saxena:
@minrk @parente I've created another PR https://github.com/jupyter/notebook/pull/3158 based on the feedback and comments, what do you think of it?

takluyver:
Closing for now in favour of #3158

PR #3008, opened 2017-11-02
When login-in via token, let a chance for user to set the password

Carreau:
When token is enabled, the login page will present a form to the user
asking them if they want to set a password at the same time. This is
almost equivalent to running `jupyter notebook password` on the command
line.

The experience can likely be better, but just submitting that as a POC
for feedback

### DIFF
--- a/docs/source/public_server.rst
+++ b/docs/source/public_server.rst
@@ -63,15 +63,28 @@ using the following command::
 
   $ jupyter notebook --generate-config
 
-.. _hashed-pw:
 
-Preparing a hashed password
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Automatic Password setup
+~~~~~~~~~~~~~~~~~~~~~~~~
 
-As of notebook version 5.0, you can enter and store a password for your
-notebook server with a single command.
-:command:`jupyter notebook password` will prompt you for your password
-and record the hashed password in your :file:`jupyter_notebook_config.json`.
+As of notebook 5.3, the first time you log-in using a token, the notebook server
+should give you the opportunity to setup a password from the user interface.
+
+You will be presented with a form asking for the current _token_, as well as
+your _new_ _password_ ; enter both and click on ``Login and setup new password``.
+
+Next time you need to log in you'll be able to use the new password instead of
+the login token, otherwise follow the procedure to set a password from the
+command line.
+
+The ability to change the password at first login time may be disabled by
+integrations by setting the ``--NotebookApp.allow_password_change=True``
+
+
+Starting at notebook version 5.0, you can enter and store a password for your
+notebook server with a single command. :command:`jupyter notebook password` will
+prompt you for your password and record the hashed password in your
+:file:`jupyter_notebook_config.json`.
 
 .. code-block:: bash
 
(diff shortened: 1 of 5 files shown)


notebook/auth/login.py
```diff
@@ -72,16 +73,26 @@ def passwd_check(self, a, b):
...
+
+
+        
         if self.get_login_available(self.settings):
-            if self.passwd_check(self.hashed_password, typed_password):
+            if self.passwd_check(self.hashed_password, typed_password) and not new_password:
                 self.set_login_cookie(self, uuid.uuid4().hex)
             elif self.token and self.token == typed_password:
                 self.set_login_cookie(self, uuid.uuid4().hex)
+                if self.new_password:
```
minrk: extra `self.`

notebook/templates/login.html
```diff
@@ -85,6 +85,22 @@
       <p>
         Cookies are required for authenticated access to notebooks.
       </p>
+      <h3>{% trans %}Setup a Password{% endtrans %}</h3>
```
minrk: This whole block should be conditional on password not being set already. And there should probably be a `set_password_available` for explicit disabling, as well.

Carreau:
Yep I thought to most of those. I'm even considering redirecting users to
the login screen immediately if they set a password to be sure it works.

I'll work on that this week-end or next week. Thanks !

Carreau:
Updated w/ documentation and options to disable.

notebook/notebookapp.py
```diff
@@ -750,6 +751,18 @@ def _token_changed(self, change):
                       """
     )
 
+    allow_password_change = Bool(True, config=True, 
+                    help="""Allow password to be changed at login for the notebook server. 
+
+                    While login-in with a token, the notebook server UI will give the opportunity to
```
takluyver: logging in

notebook/templates/login.html
```diff
@@ -85,6 +85,24 @@
...
+        <p> You can also setup a password by entering your token and a new password
+        on the fields below:</p>
+        <form action="{{base_url}}login?next={{next}}" method="post" class="">
+                {{ xsrf_form_html() | safe }}
+          <div class="form-group">
+            <input type="password" name="password" id="password_input" class="form-control" placeholder="Token">
+          </div>
+          <div class="form-group">
+            <input type="password" name="new_password" id="new_password_input"
+            class="form-control" placeholder="New password" required>
```
takluyver: Should we get the user to type the new password twice, and check that it's the same?

takluyver:
Can you show a screenshot of the new login page?

notebook/templates/login.html
Carreau: I don't think it is necessary, they can still issue `jupyter notebook password` to reset.

Carreau:
In the end I think we want a change-password page that may or may not be available, and potentially pre-fill the token if it is given in the URL (and not redirect). 

Though I do not want to spend too much time on that as most user will not even see this change password field.

screen shot 2017-11-13 at 07 55 02

takluyver:
Thanks. I'm happy to merge this and see how it goes, but I'll give it a while for other people to have a look.

docs/source/public_server.rst
```diff
@@ -63,15 +63,28 @@ using the following command::
...
+
+You will be presented with a form asking for the current _token_, as well as
+your _new_ _password_ ; enter both and click on ``Login and setup new password``.
+
+Next time you need to log in you'll be able to use the new password instead of
+the login token, otherwise follow the procedure to set a password from the
+command line.
+
+The ability to change the password at first login time may be disabled by
+integrations by setting the ``--NotebookApp.allow_password_change=True``
```
minrk: =False

Carreau:
> Minor comment that I think the flag for disabling password change is True where it should be False, but 👍 to merge with that typo fixed (or my understanding corrected).

Oops. Should be fixed.

PR #3009, opened 2017-11-02
Hash cookie secret with user hashed password.

Carreau:
Currently changing the password does not revoke current session:

  - jupyter notebook password <password1>
  - jupyter notebook
  - Logging in
  - Kill server
  - jupyter notebook password <other password>
  - jupyter notebook
  - Oh ! I'm still logged in.

With this, as the "effective" secret depends on the (hashed) password,
changing it void any existing session (which I believe is the goal of
most password change)

cc @minrk, @rgbkrk, and @yuvipanda for review.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -27,6 +27,7 @@ import threading
 import time
 import warnings
 import webbrowser
+import hmac
 
 try: #PY3
     from base64 import encodebytes
@@ -674,11 +675,16 @@ class NotebookApp(JupyterApp):
     def _default_cookie_secret(self):
         if os.path.exists(self.cookie_secret_file):
             with io.open(self.cookie_secret_file, 'rb') as f:
-                return f.read()
+                key =  f.read()
         else:
-            secret = encodebytes(os.urandom(1024))
-            self._write_cookie_secret_file(secret)
-            return secret
+            key = encodebytes(os.urandom(1024))
+            self._write_cookie_secret_file(key)
+        h = hmac.HMAC(key)
+        h.digest_size = len(key)
+        h.update(self.password.encode())
+        return h.digest()
+
+
     
     def _write_cookie_secret_file(self, secret):
         """write my secret to my secret_file"""


notebook/notebookapp.py
```diff
@@ -674,11 +676,16 @@ def _default_cookie_secret_file(self):
...
             with io.open(self.cookie_secret_file, 'rb') as f:
-                return f.read()
+                key =  f.read()
         else:
-            secret = encodebytes(os.urandom(1024))
-            self._write_cookie_secret_file(secret)
-            return secret
+            key = encodebytes(os.urandom(1024))
+            self._write_cookie_secret_file(key)
+        h = sha256(digest_size=len(key))
```
rgbkrk: I was going to make a comment here about what happens if the cookie file happened to be empty (leading to a 0 length), then I noticed that at least on my systems I can't set this keyword argument:

```ipython
In [1]: from hashlib import sha256

In [2]: sha256(digest_size=0)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-94e5dda9a83d> in <module>()
----> 1 sha256(digest_size=0)

TypeError: openssl_sha256() takes no keyword arguments
```

notebook/notebookapp.py
```diff
@@ -674,11 +676,16 @@ def _default_cookie_secret_file(self):
...
+        h.update(key)
+        h.update(self.password.encode())
```
yuvipanda: IMO this should be a [HMAC](https://docs.python.org/3/library/hmac.html), with the secret key as the HMAC key & the password as the message. HMACs are built for this exact specific use case, and supported in the stdlib.

notebook/notebookapp.py
Carreau: hum, weird.

notebook/notebookapp.py
Carreau: I disagree that HMAC are built for that (no code is going to check hmac(key, pw) equal the value here). The result of this function will be used by HMAC as the key later on (in tornado), so we're basically doing `HMAC(HMAC(key, pwd_sha), cookie_value)`; Though I agree that the returned value with HMAC will still be (pseudo-)random enough.

Carreau:
Should we also (or as an alternative) have the the `jupyter notebook password` command to regenerate the cookie secret ? Because If I change my password to an old one, then the previous sessions will become valid again. Or is this a "feature" ?

PR #3015, opened 2017-11-04
typo fix

PHaeJin:
Fix some broken links

### DIFF
--- a/docs/source/links.txt
+++ b/docs/source/links.txt
@@ -25,8 +25,8 @@
 .. _notebook_p1: https://nbviewer.jupyter.org/github/jupyter/notebook/blob/master/docs/source/examples/Notebook/Running%20Code.ipynb
 
 .. Other python projects
-.. _matplotlib: http://matplotlib.org
-.. _nbviewer: http://nbviewer.jupyter.org
+.. _matplotlib: http://matplotlib.org/
+.. _nbviewer: http://nbviewer.jupyter.org/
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -72,7 +72,7 @@
 explanatory text. The Notebook has support for multiple programming
 languages, sharing, and interactive widgets.
 
-Read `the documentation <https://jupyter-notebook.readthedocs.io>`_
+Read `the documentation <https://jupyter-notebook.readthedocs.io/en/stable/>`_
```
gnestor: What if we support other languages in the future? I think readthedocs can potentially redirect to the appropriate labguage version based on locale. I think we would be better off sticking with the original.

docs/source/links.txt
```diff
@@ -25,8 +25,8 @@
 .. _notebook_p1: https://nbviewer.jupyter.org/github/jupyter/notebook/blob/master/docs/source/examples/Notebook/Running%20Code.ipynb
 
 .. Other python projects
-.. _matplotlib: http://matplotlib.org
-.. _nbviewer: http://nbviewer.jupyter.org
+.. _matplotlib: http://matplotlib.org/
+.. _nbviewer: http://nbviewer.jupyter.org/
```
gnestor: 🤷‍♂️

setup.py
PHaeJin: Alright. Fixing it now!

gnestor:
Thanks @PHaeJin!

PR #3020, opened 2017-11-05
Add link to Korean Version of Installation

ChungJooHo:
I add my repo link on README so that people who wants to read it
translated version of Korean.

As I heared your idea, I made new repo. And I will continue to translate such docs as soon as I can.
The link I added is a link to my repo, so people who wants to read it can see Korean version.

I will update my repo steadily. Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ If you are interested in contributing to the project, see [`CONTRIBUTING.rst`](C
 - [Project Jupyter website](https://jupyter.org)
 - [Online Demo at try.jupyter.org](https://try.jupyter.org)
 - [Documentation for Jupyter notebook](https://jupyter-notebook.readthedocs.io/en/latest/) [[PDF](https://media.readthedocs.org/pdf/jupyter-notebook/latest/jupyter-notebook.pdf)]
+- [Korean Version of Installation](https://github.com/ChungJooHo/Jupyter_Kor_doc/)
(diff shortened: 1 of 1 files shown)


rgbkrk:
We could also add a `docs-translations` folder, similar to [how Electron does it](https://github.com/electron/electron/blob/master/docs-translations/ko-KR/README.md). I think this is fine for now though. Thank you for taking the time to do translation.

ChungJooHo:
Oh that's a idea. I will make new folder on Jupyter project similiar to Electron!!  Thank you for suggesting such idea!!

willingc:
For simplicity, I would use a `docs-translations` folder unless @jzf2101 has a different desired approach.

Internationalization of docs is a big effort. Django is one of the best examples. Here are some links:
- http://www.sphinx-doc.org/en/stable/intl.html
- https://groups.google.com/forum/#!forum/django-i18n

takluyver:
RTD has a feature where you can set up another RTD project and then cross-link it as the docs for another language. This was what I was thinking of when I suggested @ChungJooHo translated in a separate repository.

https://docs.readthedocs.io/en/latest/localization.html#project-with-multiple-translations

rgbkrk:
You can still use the same repository with RTD localization. My main preference is that we keep it in the Jupyter org at the very least and help people steward their localization efforts.

takluyver:
Sounds good.

PR #3022, opened 2017-11-06
Add Folder for document translation

ChungJooHo:
As you suggested I first add translation folder. And I added my translation 'Readme.md' in Korean.
Thank you for your suggestion. I will steadily update docs.

### DIFF
--- a/docs-translations/ko-KR/README.md
+++ b/docs-translations/ko-KR/README.md
@@ -1,0 +1,55 @@
+# Jupyter Notebook
+
+[![Google Group](https://img.shields.io/badge/-Google%20Group-lightgrey.svg)](https://groups.google.com/forum/#!forum/jupyter)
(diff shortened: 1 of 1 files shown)


willingc:
Thanks!

I'll leave for @jzf2101 to review and merge :-)

gnestor:
Another options is to use a file name convention like `README.ko-KR.md` 👍

rgbkrk:
> Another options is to use a file name convention like README.ko-KR.md

Oh that's true for the base. Then we would just make that folder for the `docs/` translations.

ChungJooHo:
Okay!! Then I will just add folder for translations and rename readme to readme.ko-KR.md :)

ChungJooHo:
But I think making each language folders like docs/transation/ko-KR is more comfortable to access docs .

jzf2101:
@ChungJooHo this is great thanks!

PR #3031, opened 2017-11-09
Use `/files` prefix for PDF-like files

gnestor:
Fixes https://github.com/jupyter/notebook/pull/2911#issuecomment-343301587

This adds one more conditional for PDF-like files (currently only PDF) that can be viewed (opened) in the browser but don't work with:

```html
<iframe id="iframe" sandbox="allow-scripts" src="{{file_url}}"></iframe>`
```

Other file extensions that are like this (off the top of my head) are office files (`.doc`, `.xls`, `.docx`, `.xls`, `.xlsx`, `.ppt`) (at least using Chrome).


takluyver:
Makes sense. Let's merge and iterate in master.

PR #3034, opened 2017-11-10
tornado 5: PeriodicCallback loop arg will be removed

minrk:
PCs are always run with the current eventloop, which is what the loop we have been passing explicitly already is.

Also fixed a case where we were closing ZMQStream and socket separately,
which may have been safe in the past, but can raise errors due to trying to unregister a closed socket with zmq.


notebook/services/kernels/handlers.py
```diff
@@ -452,7 +452,6 @@ def on_close(self):
                 # close the socket directly, don't wait for the stream
```
takluyver: This comment should probably be removed as misleading. And the assignment below is now unnecessary.

minrk: Good atch, fixed.

PR #3037, opened 2017-11-11
Fix typo

PHaeJin:


Carreau:
Hi there ! Thanks for your pull request. 

It looks like you may have struggled a bit with git; did you managed to do what you were looking for, or do you need guidance to fix a couple of things ? There seem to be a couple of changes that disappeared in your final version.

Git can be quite confusing and it may take a couple of month to be fluent with it; we'll be happy to offer advice.

takluyver:
I think you're reusing the master branch for a series of pull requests, merging our master branch into it each time.

We can resolve that here, but it's neater to make a new branch for each change you want to submit. Something like this:

```shell
git clone https://github.com/jupyter/notebook.git
cd notebook
git remote add PHaeJin https://github.com/PHaeJin/notebook.git

# Make a new branch
git checkout -b typo1

# Make your changes, then commit
git commit -m "Fix a typo"

git push -u PHaeJin typo1
```

That way you won't have an ever-increasing number of commits in each pull request.

PHaeJin:
thank you for your advice!! 
i will fix it right away.

takluyver:
This was merged as #3043 , so I'll close this one.

PR #3045, opened 2017-11-15
Arrange quotation marks

bacboc:


minrk:
Thanks for the PR! But please don't make style change pull requests like this. If such a pass would make significant changes to code, it's a good indicator that it is not likely to be a desirable change. If you would like to re-issue a PR that only fixes the link you found instead of changing the quote style, that would be very welcome.

PR #3048, opened 2017-11-16
 Fix moment locale loading in bidi support

jcb91:
This is throwing errors like

```
Uncaught Error: Script error for "components/moment/locale/en-gb"
    at makeError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:165)
    at HTMLScriptElement.onScriptError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1732)
makeError @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:165
onScriptError @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1732
```
because of the way it's been implemented (issuing a require call to a presumed locale js file, which may not exist, especially since packaged versions of notebook are shipping without separate locale files). This PR does it in the correct way (getting moment to load it from its own internal set, if possible, also falling back to less-specific versions as necessary).

### DIFF
--- a/notebook/static/bidi/bidi.js
+++ b/notebook/static/bidi/bidi.js
@@ -5,7 +5,6 @@ define(['bidi/numericshaping'], function(numericshaping) {
   'use strict';
 
   var shaperType = '';
-  var textDir = '';
 
   var _uiLang = function() {
     return navigator.language.toLowerCase();
@@ -13,7 +12,7 @@ define(['bidi/numericshaping'], function(numericshaping) {
 
   var _loadLocale = function() {
     if (_isMirroringEnabled()) {
-      $('body').attr('dir', 'rtl');
+      document.body.dir = 'rtl';
     }
 
     require(['moment'], function (moment) {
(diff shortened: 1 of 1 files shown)


jcb91:
Also, while looking at this, I've noticed that

* there are quite a few country-specific arabic locales (e.g. `ar-sa`), but the current check for whether to use the numericshaping at [static/bidi/bidi.js#L29](https://github.com/jupyter/notebook/blob/74fbc5b5784d5dbadcc66d7a78cedf029d1bb7d9/notebook/static/bidi/bidi.js#L29) is checking for the exact string `ar`, which means it'll decide not to map them. This might be intentional, but I suspect it isn't?
* moment's arabic locales already provide functions for transforming numbers in strings (see, e.g. [moment/src/locale/ar-sa.js#L85-L89](https://github.com/moment/moment/blob/328d51e7235b7a2d443bfdf69c1844939c115660/src/locale/ar-sa.js#L85-L89)). Perhaps it might make more sense to use these for the numericshaping functionality than rolling new ones in notebook? This might make the above check redundant too, we could instead do something like

  ```javascript
  var shapeNumerals = moment.localeData().postformat;
  var applyBidi = function (value) {
      return shapeNumerals(value);
  };
  ```

takluyver:
Thanks, I don't know much about this, but this looks like an improvement to me.

cc @samarsultan and @rgbkrk (you've touched this file)

takluyver:
I think you're right about the exact `ar` check, as well - it should be checking if it starts with that.

takluyver:
I've done a brief test with an arabic locale set. I made and installed an sdist to check for packaging problems. It seemed to be working, whereas 5.2.1 installed through conda still showed me 'ago' times in English, and had the error about loading the locale.

Issue #2953, opened 2017-10-18
404 GET error


smartinsightsfromdata:
I have:
python 3.6; jupyterlab 0.28.3; jupyter-nbextensions-configurator 0.2.8; ipython 6.2.1; ipywidgets 7.0.1

This is what I get when I launch `jupyter notebook`:

```
jupyter notebook
[I 21:55:58.313 NotebookApp] Loading IPython parallel extension
[I 21:55:58.354 NotebookApp] [jupyter_nbextensions_configurator] enabled 0.2.8
[I 21:55:58.384 NotebookApp] JupyterLab alpha preview extension loaded from /Users/e/anaconda3/lib/python3.6/site-packages/jupyterlab
[I 21:55:58.392 NotebookApp] Serving notebooks from local directory: /Users/e/Dropbox/devlib/notebooks
[I 21:55:58.392 NotebookApp] 0 active kernels
[I 21:55:58.392 NotebookApp] The Jupyter Notebook is running at:
[I 21:55:58.392 NotebookApp] http://localhost:8484/?token=ff4e7fc9d15c5a81949d5fec9bb32c3a32b49307dbb6ae2c
[I 21:55:58.392 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 21:55:58.394 NotebookApp] 
    
    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://localhost:8484/?token=ff4e7fc9d15c5a81949d5fec9bb32c3a32b49307dbb6ae2c
[I 21:55:59.287 NotebookApp] Accepting one-time-token-authenticated connection from ::1
[W 21:55:59.763 NotebookApp] 404 GET /static/components/moment/locale/en-gb.js?v=20171018215557 (::1) 10.01ms referer=http://localhost:8484/tree
```

I haven't found any new problem related to this, but it doesn't seem right.

How can I get rid of the `404` ?

pavlin99th:
I had the exact same error on my Mac. Why did it request `en-gb` locale?
It turned out the code uses `navigator.language` property and in my chrome console it indeed showed `en-GB`. In my case I expected `en-US`.
So, if you are in US locale, not in UK, and on a Mac, the solution is: go to System Preferences -> Language & Region, there replace `English` with `English (U.S.)` in Preferred languages.
This solved the issue for me.

chrisramsay:
[...]

PR #3051, opened 2017-11-16
Include moment locales in package_data

takluyver:
Closes gh-2953

I'm not quite sure how this interacts with minification, but it appears that in at least some circumstances it expects these files.


takluyver:
Ah, actually, I think #3048 is probably a better fix to the same issue.

takluyver:
Closed in favour of #3048.

PR #3055, opened 2017-11-17
Compare non-specific language code when choosing to use Arabic numerals

jcb91:
As discussed in https://github.com/jupyter/notebook/pull/3048#issuecomment-344905231,

Currently, country-specific locales like `ar-sa` (Arabic, Saudi Arabia), end up using the default numerals. This PR alters the comparison to only use the part of the locale before any `-`.


gnestor:
Thanks @jcb91!

Issue #3044, opened 2017-11-15
Wrong size for some interface buttons in the tree view with Firefox


ferdas:
![screen](https://user-images.githubusercontent.com/8852116/32812029-3abf8af2-c9a3-11e7-9c77-e5785d35f42f.png)

Using Firefox, there are a lot of small sizing problem across the whole UI.
The ones highlighted in the screenshot above are the among the most annoying since they appear often in the daily usage.

I'll send a pull request later on to fix them :)

takluyver:
I also notice this, so I'm looking forward to your PR :-)

ferdas:
I was busier than I expected but finally I made it :)

I hope give more contribution in the future!

Thank you

ferdas:
Thanks :)

PR #3058, opened 2017-11-18
Fixes some UI bugs in Firefox #3044

ferdas:
Here's my PR to solve the reported problems :)

I've cleaned a bit some CSS and I've refined the usability of the "Upload" button": now the clickable area is exactly equal to the button area.
I've tested the changes into Chromium, Firefox (Quantum, too) and Edge and it behaves as expected!

Let me know if you have questions or requests!

Thank you for your work!

### DIFF
--- a/notebook/static/tree/less/altuploadform.less
+++ b/notebook/static/tree/less/altuploadform.less
@@ -13,16 +13,24 @@
 
     input.fileinput
     {
-        text-align: center;
-        vertical-align: middle;
-        display: inline;
+        position: absolute;
+
+        display: block;
+        width: 100%;
+        height: 100%;
+
+        cursor: pointer;
+
         opacity: 0;
         z-index: 2;
-        width: 12ex;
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks - I've just tested this, and it looks good to me.

Oddly, when I move my mouse over it in Chromium, the pointer (hand) cursor appears momentarily, then it goes back to the default (arrow) cursor. In Firefox it uses the hand cursor like the other buttons. I wouldn't hold up the improvements over this, though.

ferdas:
Yes, I've noticed it: it's due to the fact that Chrome/Chromium doesn't like to have the cursor changed on some kind of object..! Let me see if I can find a clean way to force it :)

ferdas:
I had to add a browser-specific rule but I think it's worth it since it's very straightforward :)
Now it behaves uniformly.

takluyver:
Thanks, nice one :-)

PR #3061, opened 2017-11-19
Fix broken link

bacboc:


docs/source/public_server.rst
```diff
@@ -36,9 +36,9 @@ This document describes how you can
 
 
 
-.. _ZeroMQ: http://zeromq.org
+.. _ZeroMQ: https://zeromq.org
 
-.. _Tornado: http://www.tornadoweb.org
+.. _Tornado: https://www.tornadoweb.org
```
minrk: These two links cannot be accessed via https

bacboc: Oh. I'm sorry, I'll revise it and send you a PR.

takluyver: It looks like you've worked this out already, but there's no need to make a new PR - when you push on the same branch, this PR updates automatically.

bacboc:
It is now a step in the beginning and is immature. So now I know. I will do that in the future. Thanks for letting me know.

takluyver:
Thanks!

PR #3071, opened 2017-11-22
Add a copy capability to jupyter notebooks

agrippa:
This is a small change to add the ability to copy files in the Notebook UI, in addition to moving them. Because of my infamiliarity with the notebook code base, I tried to minimize the amount of changes necessary to achieve this. Mostly, this appropriates the same code paths as Move but adds a keep_old flag that indicates whether the source file should be cleared out (move) or not (copy).

This is my first pull request on the repo, so apologies if I skipped/mangled a whole bunch of steps in creating this pull request (as I imagine I may have).

### DIFF
--- a/notebook/static/base/js/namespace.js
+++ b/notebook/static/base/js/namespace.js
@@ -73,7 +73,7 @@ define(function(){
     // tree
     jglobal('SessionList','tree/js/sessionlist');
 
-    Jupyter.version = "5.2.1";
+    Jupyter.version = "5.3.0.dev0";
     Jupyter._target = '_blank';
     return Jupyter;
 });
--- a/notebook/static/base/js/namespace.js
+++ b/notebook/static/base/js/namespace.js
@@ -73,7 +73,7 @@ define(function(){
     // tree
     jglobal('SessionList','tree/js/sessionlist');
 
-    Jupyter.version = "5.3.0.dev0";
+    Jupyter.version = "5.2.1";
     Jupyter._target = '_blank';
     return Jupyter;
 });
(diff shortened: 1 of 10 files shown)


takluyver:
Thanks! It will probably be a few days before I can look at this properly, but someone else might be able to have a look sooner.

It looks like the CI tests have failed - can you have a look and see if that's related to the change you made?

There is already a method `ContentsManager.copy()` in the API - maybe it's possible to use that?

notebook/static/base/js/namespace.js
```diff
@@ -73,7 +73,7 @@ define(function(){
     // tree
     jglobal('SessionList','tree/js/sessionlist');
 
-    Jupyter.version = "5.2.1";
+    Jupyter.version = "5.3.0.dev0";
```
dibyadas: @agrippa I think version changes automatically when start developing the code base. And should be changed when we are making a PR, if I'm not mistaken.
I'm currently new to the codebase so I might be wrong in this.  😄

takluyver:
`ContentsManager` already has a `copy()` method - can we try using that rather than adding a parameter to `rename()`? Or is there some reason that won't work?

agrippa:
@takluyver Thanks for the feedback! I looked at just using copy() and did a bit of refactoring, is that along the lines  of what you were thinking?

takluyver:
That's a start, but I don't think we need the `keep_old` parameter anywhere - the contents manager class already has a `copy()` method separate from `rename()`. Maybe we need to expose it to the notebook UI.

Zsailer:
Thanks, @agrippa, for this work!

This has been inactive for awhile, so I'm going to close this PR for now. 

I think it's a good feature to add at some point, but it looks like developer time is limited here. If anyone has time to devote to it, feel free reopen or resubmit a new PR. 

Thanks!

PR #3087, opened 2017-11-27
Add support for terminals on Windows

blink1073:


rgbkrk:
Sweet

notebook/terminal/__init__.py
```diff
@@ -6,14 +6,20 @@
...
 from .handlers import TerminalHandler, TermSocket
 from . import api_handlers
 
 def initialize(webapp, notebook_dir, connection_url, settings):
-    shell = settings.get('shell_command', [os.environ.get('SHELL') or 'sh'])
+    default_shell = which('sh')
+    if not default_shell and os.name == 'nt':
+        default_shell = 'cmd.exe'
+    shell = settings.get('shell_command',
+        [os.environ.get('SHELL', default_shell)]
```
takluyver: I tend to do defaults for environment variables like `os.environ.get('SHELL') or default_shell`, because setting an environment variable to the empty string is equivalent by convention to unsetting it.

blink1073:
I'm going to hold off on this pending support for Python 2.7 on Windows.  cf https://github.com/spyder-ide/pywinpty/issues/75

takluyver:
Up to you, but I'm happy for it to be a Python-3-only feature. Next year we might be looking at doing Notebook 6.0 and requiring Python 3, in any case.

blink1073:
Already mostly done: cf https://github.com/jupyter/terminado/pull/50

blink1073:
Good to go

blink1073:
Updated to use `powershell` on Windows if `sh` is not available, which is what VS Code uses.

gnestor:
Thanks @blink1073!

stonebig:
hi, I think I'm hitting this with Jupyterlab-0.30.5 / terminado-0.8.1 / Pywinpty-0.5.1... Could you do a release soon ?

liuzhousi:
Sorry to bother.. I think I may encounter the same problem now, as I still could not open the terminal in the latest win10 64bit system. Besides,  my environment situation lists below: Python3.6.4, jupyter notebook5.7.8, terminado0.8.2, and pywinpty0.5.5. The detailed problem is: Terminals not available (error was DLL load failed: %This is not a valid Win32 Application.

PR #3088, opened 2017-11-27
Allow programmatic copy to clipboard

### LINKED ISSUE #2975: "Programmatic" copy to clipboard don't work in Jupyter 5
Because of #1286, any extension that relies on ``document.execCommand('copy')`` for copying to clipboard don't work, as the notebook captures _all_ copy events triggered. 

For instance, if a notebook extension used clipboard.js - an invisible textarea is created whose value is set to whatever is supposed to go on the clipboard.  In that case [window.getSelection().isCollapsed](https://github.com/jdfreder/notebook/blob/72eeb99ad6f234467d561e2ebfc79343c35aa2b2/notebook/static/notebook/js/clipboard.js#L38) is set to True and the contents of the cell (not the value of the invisible textarea) is stored in the system clipboard.  

In this case, ``Ctrl+C`` would never be pressed and it is surprising for the extension users to click a button on the toolbar or access a menu-item and find that the current cell's contents are copied over.

A workaround might be to disable ``Jupyter.keyboard_manager`` and re-enable it once the extension's work is done.

Do you think the notebook can somehow try to detect explicit requests to copy (when the user hits ``Ctrl+C``) as opposed to capturing all copy events that are fired ?  

Thanks!

Madhu94:
Trying to close #2975.

@takluyver I did not go with ``document.execCommand`` because it does not allow you to set a mime-types like html, json, etc. (or I couldn't find a way to do that anyway) and I see you need that.

Let me know if these changes are okay.

### DIFF
--- a/notebook/static/notebook/js/clipboard.js
+++ b/notebook/static/notebook/js/clipboard.js
@@ -34,10 +34,18 @@ function load_json(clipboard) {
   return JSON.parse(s.slice(pix + jcbprefix.length, six));
 }
 
+function isProgrammaticCopy(event) {
+  return (typeof(event.target.selectionStart) !== 'undefined'
+    && typeof(event.target.selectionEnd) !== 'undefined'
+    && ((event.target.selectionEnd - event.target.selectionStart) > 0));
+}
+
 function copy(event) {
   if ((Jupyter.notebook.mode !== 'command') ||
         // window.getSelection checks if text is selected, e.g. in output
-        !window.getSelection().isCollapsed) {
+        !window.getSelection().isCollapsed ||
+        // Allow programmatic copy
+        isProgrammaticCopy(event)) {
     return;
   }
   var selecn = Jupyter.notebook.get_selected_cells().map(


Madhu94:
@takluyver Here is an example you can paste into a notebook cell to reproduce this bug and see what this fix does.
```
class copytoclipboard(object):
    def _repr_html_(self):
        return """
            <button class="btn" data-clipboard-text="Just because you can">Copy to clipboard</button>
            <script>
                function copyToClipboard(copytext) {
                    var ta = document.createElement('textarea');
                    ta.value = copytext;
                    ta.style.position = 'absolute';
                    ta.style.left = '-9999px';
                    document.body.append(ta);
                    ta.select()
                    ta.focus();
                    console.log("Attempting to copy" + copytext);
                    var success=document.execCommand('copy');
                    console.log(success);
                    
                }
                $('.btn').click(function(){
                  copyToClipboard("Ho ho");
                })
            </script>
        """
copytoclipboard()
```

gnestor:
I tested and it works! Looks good to me. 

@takluyver Care to merge?

takluyver:
Thanks, I've also tested this, and it seems to be fine.

For my own understanding, why does the existing `window.getSelection().isCollapsed` check not do the same? Is the invisible text field ignored when finding `window.getSelection()`?

Madhu94:
@takluyver ``window.getSelection().isCollapsed`` would not return ``True`` if, in the above example, you were to set ``ta.innerHTML`` to the copyText and not ``ta.value``.  

I don't find it intuitive either.

takluyver:
Fair enough. Browsers are strange.

takluyver:
By the way, thanks for providing the example so we could easily test what you meant!

PR #3100, opened 2017-12-01
Fixes #2760. Allows deletion of non empty directories.

### LINKED ISSUE #2760: Allow deleting nonempty directorys
Hello,
  It is frequently useful to be able to delete non empty dirs, especially if there are many subfolders. Could you please add a feature for that (deleting nonempty dirs, maybe with a confirmation)?

kirit93:
Submitting this PR from the PyData NYC Sprint session.
I've commented out the original function and added the modified function right above it in ```notebook/services/contents/filemanager.py```

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -489,16 +489,8 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         path = path.strip('/')
         os_path = self._get_os_path(path)
         rm = os.unlink
-        if os.path.isdir(os_path):
-            listing = os.listdir(os_path)
-            # Don't delete non-empty directories.
-            # A directory containing only leftover checkpoints is
-            # considered empty.
-            cp_dir = getattr(self.checkpoints, 'checkpoint_dir', None)
-            for entry in listing:
-                if entry != cp_dir:
-                    raise web.HTTPError(400, u'Directory %s not empty' % os_path)
-        elif not os.path.isfile(os_path):
+        
+        if not os.path.isfile(os_path):
             raise web.HTTPError(404, u'File does not exist: %s' % os_path)
 
         if self.delete_to_trash:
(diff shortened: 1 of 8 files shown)


takluyver:
Feel free to delete the old version - there's no need to comment it out when we've got version control. ;-)

takluyver:
I think we should probably still refuse to delete non-empty directories when `delete_to_trash` is False (i.e. deleting really deletes things). The check could be moved a bit further down the function.

kirit93:
@takluyver ```test_delete_dirs``` fails with this error message ```requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://localhost:12341/a%40b/api/contents/unicod%C3%A9``` 
I can't seem to figure out why this is happening. I create a set out the the ```self.dirs``` variable so that when a directory is deleted, the sub directories in it are not included in the list because they will be deleted along with the top level directory due to the change I made to the delete function. Could you take a look at this and let me know what you think the issue could be? Thanks!

notebook/services/contents/filemanager.py
```diff
@@ -489,16 +489,8 @@ def delete_file(self, path):
...
-            # Don't delete non-empty directories.
-            # A directory containing only leftover checkpoints is
-            # considered empty.
-            cp_dir = getattr(self.checkpoints, 'checkpoint_dir', None)
-            for entry in listing:
-                if entry != cp_dir:
-                    raise web.HTTPError(400, u'Directory %s not empty' % os_path)
-        elif not os.path.isfile(os_path):
+        
+        if not os.path.isfile(os_path):
```
takluyver: Previously this was checking (in pseudocode):

```python
if isdir():
    ...
elif isfile():
    error
```

Now it only checks `isfile()`, so directories hit the error case.

takluyver: sorry, the second check is `elif not isfile()`.

i.e. the error should occur if the path is neither a file nor a directory. With your changes, it occurs if it's not a file.

kirit93: Okay, thanks. I'll make the changes and send in a new PR.

PR #3108, opened 2017-12-04
Allowing non empty dirs to be deleted

### LINKED ISSUE #2760: Allow deleting nonempty directorys

kirit93:
Fixes #2760

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -80,7 +80,9 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
     save_script = Bool(False, config=True, help='DEPRECATED, use post_save_hook. Will be removed in Notebook 5.0')
 
     @observe('save_script')
-    def _update_save_script(self):
+    def _update_save_script(self, change):
+        if not change['new']:
+            return
         self.log.warning("""
         `--script` is deprecated and will be removed in notebook 5.0.
 
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -489,17 +489,8 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         path = path.strip('/')
         os_path = self._get_os_path(path)
         rm = os.unlink
-        if os.path.isdir(os_path):
-            listing = os.listdir(os_path)
-            # Don't delete non-empty directories.
-            # A directory containing only leftover checkpoints is
-            # considered empty.
-            cp_dir = getattr(self.checkpoints, 'checkpoint_dir', None)
-            for entry in listing:
-                if entry != cp_dir:
-                    raise web.HTTPError(400, u'Directory %s not empty' % os_path)
-        elif not os.path.isfile(os_path):
-            raise web.HTTPError(404, u'File does not exist: %s' % os_path)
+        if not os.path.exists(os_path):
+            raise web.HTTPError(404, u'File or directory does not exist: %s' % os_path)
 
         if self.delete_to_trash:
             self.log.debug("Sending %s to trash", os_path)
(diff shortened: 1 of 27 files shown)


kirit93:
@takluyver It seems that the delete function is unable to delete the unicode directory from the test case. However the only changes I made to the delete function are those that allow directories to be deleted so I am unsure why unicode isn't being deleted.

notebook/services/contents/filemanager.py
```diff
@@ -489,19 +489,19 @@ def delete_file(self, path):
...
-                    raise web.HTTPError(400, u'Directory %s not empty' % os_path)
-        elif not os.path.isfile(os_path):
-            raise web.HTTPError(404, u'File does not exist: %s' % os_path)
+        if not os.path.exists(os_path):
+            raise web.HTTPError(404, u'File or directory does not exist: %s' % os_path)
 
         if self.delete_to_trash:
+            if os.path.isdir(os_path):
+                listing = os.listdir(os_path)
+                # Don't send non-empty directories to trash.
```
takluyver: I think you've got this backwards. The idea is that if we're using trash (if `self.delete_to_trash` is true), then we *can* delete non-empty directories, because the user can get them back from trash. If we're *not* using trash, then deleting is permanent, so we should still refuse to delete non-empty directories.

notebook/services/contents/tests/test_contents_api.py
```diff
@@ -518,7 +518,8 @@ def test_delete_dirs(self):
         for name in sorted(self.dirs + ['/'], key=len, reverse=True):
             listing = self.api.list(name).json()['content']
             for model in listing:
-                self.api.delete(model['path'])
+                if os.path.exists(model['path']):
```
takluyver: API paths shouldn't be used as filesystem paths - you can't pass them to functions like `os.path.exists()`, because they probably don't start from the same directory where the tests are running. This check shouldn't be needed in any case.

kirit93:
@takluyver the ```GROUP=js/services test``` fails with ```The command "pip install --upgrade setuptools wheel nose coverage codecov" failed and exited with 2 during .```
Any ideas what this could be due to?

takluyver:
Looks like it's passed now. Sometimes there's a random connection failure while installing a package.

notebook/services/contents/tests/test_contents_api.py
```diff
@@ -522,11 +522,6 @@ def test_delete_dirs(self):
         listing = self.api.list('/').json()['content']
         self.assertEqual(listing, [])
 
-    def test_delete_non_empty_dir(self):
```
takluyver: Can we add a test that deleting a non-empty directory does work?

notebook/services/contents/tests/test_contents_api.py
```diff
@@ -522,6 +522,12 @@ def test_delete_dirs(self):
         listing = self.api.list('/').json()['content']
         self.assertEqual(listing, [])
 
+    def test_delete_non_empty_dir(self):
+        # Test that non empty directory can be deleted
+        self.api.delete(u'å b')
+        # Assertion will pass only if self.api.delete does not throw and error
```
takluyver: A test doesn't have to have an assert - you can leave this off, and it will still fail if there's an error. When a test just checks that something runs without an error, that's a 'smoketest' (as in 'does this make smoke come out').

However, it would be good to check that it has actually been removed as well. Maybe copy some code from `test_list_nonexistant_dir`.

kirit93: Oh okay. Sure, I'll make these changes!

kirit93:
@takluyver I've added the required tests. Is the PR okay to be merged?

takluyver:
Yup, thanks!

Dmitry1987:
Non-empty folders can't be deleted in version 6 of the notebook, was it changed intentionally back to old behavior?  https://github.com/jupyter/notebook/issues/4916

PR #3115, opened 2017-12-07
[WIP] Use DOM history API for navigating directories

gnestor:
Closes https://github.com/jupyter/notebook/issues/307

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -366,7 +366,7 @@ define([
             var path = '';
             window.history.pushState({
                 path: path
-            }, 'Home', '/tree');
+            }, 'Home', utils.url_path_join(that.base_url, 'tree'));
             that.update_location(path);
             return false;
         });
@@ -376,6 +376,7 @@ define([
             path_parts.push(path_part)
             var path = path_parts.join('/')
             var url = utils.url_path_join(
+                that.base_url,
                 '/tree',
                 utils.encode_uri_components(path)
             );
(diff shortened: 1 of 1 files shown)


gnestor:
This passes my basic user testing but I'm pretty sure there are some edge cases that may break this so it's marked as WIP.

rgbkrk:
Awesome thanks for going after this!

One edge case worth trying — do a right click and “open in new tab”

blink1073:
This is great @gnestor!  We'll be cribbing this for use with `lab/tree`.  cf https://github.com/jupyterlab/jupyterlab/issues/2532

blink1073:
cc @afshin

afshin:
Thanks for the heads up, @blink1073. Nice one, @gnestor!

takluyver:
Nice, this feels super snappy. :-) I've tried opening in a new tab as @rgbkrk suggests, and it seems to work correctly.

However, something is going wrong with the breadcrumbs - when I click on them to navigate back up, the URL bar does not update. And if I navigate down three directories and then use the breadcrumbs to go up one, it requests the wrong thing (e.g. if I navigate to `Code/astcheck/dist`, and then click *astcheck* in the breadcrumbs, it requests `astcheck` instead of `Code/astcheck`).

takluyver:
I pushed a couple of commits to fix things for me and make sure we're using `base_url` where necessary.

The test failures appear to be unrelated - #3127

takluyver:
@gnestor is there more you want to do with this, or shall we merge it so that people can try it out in master?

takluyver:
:-) As this is a user-visible change, it would be good to write a little bit about it for the release notes.

gnestor:
Is this the same test error that you saw before: https://travis-ci.org/jupyter/notebook/builds/316101183?utm_source=email&utm_medium=notification

Ok, I'll update release notes now...

takluyver:
No, that's a different failure. It seems to be that the linkcheck is failing because nbviewer is down.

takluyver:
https://github.com/jupyter/nbviewer/issues/746

PR #3128, opened 2017-12-13
Pin to ipykernel 4.6.1 on Python 2 to avoid kernelspec issue

takluyver:
https://github.com/ipython/ipykernel/issues/288


takluyver:
Merging this so we have useful CI again. I'll open an issue to unpin it when there's a new release of ipykernel.

blink1073:
Nice catch!

Issue #3135, opened 2017-12-15
Jupyter Kernel Spec + N times refreshing for N Kernels + Slow start up


prabhu1984:
Hi Dev,

I found some strange scenario on Notebook. I'm running `jupyter notebook 4.1.0` version on `Python 3.5.2 :: Anaconda 4.1.1 (64-bit)` platform.

Issue: If I enable N kernels, when I refresh Jupyter home page, then it's getting kernel spec N times and hence page refresh is too slow. Say for example: If I enable 10 custom kernels, when I refresh jupyter home page, it's taking 10+ seconds. When I see log, most of the time is taking in refreshing kernels again and again for 10 times as "Found kernel <> in /usr/local/share/jupyter/kernels". In logs, it show 100 times that kernel as found.

Version:
```
[22:48]:[root@myhost:~]# python -V
Python 3.5.2 :: Anaconda 4.1.1 (64-bit)
[22:48]:[root@myhost:~]# jupyter --version
4.1.0
```
Starting Jupyter with default Python Kernel:
```
[22:48]:[root@myhost:~]# jupyter notebook --no-browser --ip 0.0.0.0 --port 9000 --debug
```
Log: With only one Kernel by default:
```
[D 22:40:45.377 NotebookApp] 200 GET /api/sessions?_=1513320006484 (10.196.177.146) 0.85ms
[D 22:40:45.378 NotebookApp] 200 GET /api/terminals?_=1513320006485 (10.196.177.146) 0.42ms
[D 22:40:45.459 NotebookApp] 200 GET /api/contents?type=directory&_=1513320006486 (10.196.177.146) 1.18ms
[D 22:40:49.684 NotebookApp] Using contents: services/contents
[D 22:40:49.685 NotebookApp] 304 GET /tree (10.196.177.146) 1.46ms
[D 22:40:49.789 NotebookApp] 304 GET /custom/custom.css (10.196.177.146) 0.52ms
[D 22:40:50.111 NotebookApp] 304 GET /custom/custom.js?v=20171214222020 (10.196.177.146) 0.84ms
[D 22:40:50.112 NotebookApp] 200 GET /static/base/images/favicon.ico?v=30780f272ab4aac64aa073a841546240 (10.196.177.146) 0.64ms
[D 22:40:50.221 NotebookApp] 200 GET /api/config/common?_=1513320049961 (10.196.177.146) 2.28ms
[...]

PR #3136, opened 2017-12-15
Find available kernelspecs more efficiently

takluyver:
Using `find_kernel_specs()` followed by `get_kernel_spec()` has quadratic performance with the number of available kernels, because it effectively rescans for each available kernel. Hands up, I designed this, and I can't think why I did it this way.

I thought this was going to require some ugly workaround, but thankfully @captainsafia was on the case two years ago, adding the `get_all_specs()` method (https://github.com/jupyter/jupyter_client/pull/93). So all I've done here is switch the notebook code to use it.

This should improve performance loading pages for anyone with several kernels installed.

Closes gh-3135

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -148,7 +148,7 @@ install_requires = [
     'ipython_genutils',
     'traitlets>=4.2.1',
     'jupyter_core>=4.4.0',
-    'jupyter_client>=4.2.0',
+    'jupyter_client>=5.2.0',
     'nbformat',
     'nbconvert',
     'ipykernel', # bless IPython kernel for now
--- a/setup.py
+++ b/setup.py
@@ -148,7 +148,7 @@ install_requires = [
     'ipython_genutils',
     'traitlets>=4.2.1',
     'jupyter_core>=4.4.0',
-    'jupyter_client',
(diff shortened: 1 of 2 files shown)


takluyver:
It's failing the test for listing kernelspecs when one is invalid, and I don't think that can easily be fixed here; I'll need to change it in jupyter_client.

takluyver:
Reopening to test with jupyter_client 5.2, which was just released.

blink1073:
Looks like appveyor will fail until the juoyter_client conda package makes its way through the wickets.

takluyver:
Reopening to test with jupyter_client 5.2, which was just released.

setup.py
```diff
@@ -148,7 +148,7 @@
     'ipython_genutils',
     'traitlets>=4.2.1',
     'jupyter_core>=4.4.0',
-    'jupyter_client',
+    'jupyter_client>=4.2.0',
```
minrk: 5.2?

takluyver: Yeah, it probably should be 5.2.

blink1073:
Thanks!

Issue #3151, opened 2017-12-20
filemanager shows broken symlinks


steverweber:
https://github.com/jupyter/notebook/blob/ae011a179742a7f4348476beb27faf375b1884fe/notebook/services/contents/filemanager.py#L311

is using: `os.lstat` 
Equivalent to stat(path, follow_symlinks=False).

the comment following that block of code says:
`# skip over broken symlinks in listing`

perhaps in the past the comment was true... however as it is now the file listing shows broken links.
The broken links are an issue for our university because we have symlinks to extra home dirs
that wont resolve... when users click them they get a `404 error page`

can this be changed to use `os.stat` so broken symlinks are hidden, or could a config option be added to hide broken symlinks... or perhaps color code them red.

thanks

scanon:
We are seeing this issue on our deployments and it is causing pretty serious usability problems especially with the labs interface.  If a user navigates up the tree from their home directory they can quickly get to a point where they can't navigate back.  I will submit a PR based on Steve's change which seems to work.

PR #3152, opened 2017-12-20
skip broken symlinks

steverweber:
skip broken symlinks

https://github.com/jupyter/notebook/issues/3151

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -308,7 +308,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                     continue
 
                 try:
-                    st = os.lstat(os_path)
+                    st = os.stat(os_path)
                 except OSError as e:
                     # skip over broken symlinks in listing
                     if e.errno == errno.ENOENT:


takluyver:
Using git blame, it appears this was changed the other way in #2574.

Pinging @minrk - you OKed the behaviour change in that PR.

minrk:
Following #2574, I think hiding broken symlinks was always the wrong thing to do. Those files do exist in the directories being listed and clicking on them 404s, as it should. Coloring them red would be nice, but I'm not sure if there's a place for that info in the API.

At the very least, the comment should be updated to reflect the fact that we are preserving rather than ignoring broken symlinks.

@steverweber where are you running into symlinks that are broken and should be hidden rather than fixed or removed from the filesystem?

steverweber:
our university has multiple facilities and each one has storage for the users.
we are deploying jupyterhub as a somewhat campus wide solution and want to allow users access to all their home dirs... symlinks in home dirs point to automounts of cifs mounts that require kerberos tickets.

its working well.

it might be possible for us to remove the symlinks when we first create them... but not an idea solution because generation/regeneration of the links would need to be done everytime the pam session or auth stack is hit for the user... besides its nicer for the user if at runtime their extra home dir would become active without requiring a logout and login.

anyways i can workaround it... just my solutions are not going to be as simple as removing one byte in a python file.

if bad symlinks being visible is intended, that's fine.. I'll just hack it in for our jupyter deployment.
```
sed -i 's|st = os.lstat(os_path)|st = os.stat(os_path)|' /usr/local/lib/python3.5/dist-packages/notebook/services/contents/filemanager.py
```

perhaps can set an icon as a red 'X' for the file icon. I'll do some more investigation.

steverweber:
well i tried... dont think i'll get the icon idea done in 30min so i give up :)

feel free to close

scanon:
I added a comment to #3151.  You can easily see the issue.  Just create a link in a directory to a directory you don't have permissions to and you will see the problem.

On my mac I can do...

ln -s /etc/cups/certs .
jupyter notebook

We have users hitting this issue and it is causing some usability issues.  Steve's patch seems to fix it but let me see if I can come up with another PR that addresses your concern.

coolingozone:
This issue had been open and not updated for quite some time. Do we want to close it?

Zsailer:
Closing, since this has been inactive for awhile. Happy to review a new PR if someone has time to devote to this.

PR #3160, opened 2017-12-21
warn_bad_name should not use global name

Sukneet:
warn_bad_name in get_name should use the local variable name_or_data not name


takluyver:
Thanks, good catch!

PR #3173, opened 2017-12-31
Alignment issue [FIXED]

### LINKED ISSUE #3109: out of alignment in chrome

Axle7XStriker:
To fix the alignment issue which was found out to occur in most of the browsers including Chrome and Firefox because of the scroll bar becoming visible and invisible whichever is required.

The main problem this was being caused was because of 'margin-left' property being set to 'auto' as well as the 'overflow' property being set to 'auto'.

So, to fix this issue I updated the 'margin-left' property to align it with the header of the page whenever changes like page reload, resize, etc occurs.

closes #3109

### DIFF
--- a/notebook/static/base/js/page.js
+++ b/notebook/static/base/js/page.js
@@ -74,7 +74,21 @@ define([
         if (!(e && e.target && e.target.tagName)) {
             $('div#site').height($(window).height() - $('#header').height());
         }
+        this._align_header_site();
     };
+
+
+
+    Page.prototype._align_header_site = function(e) {
+        /**
+         * Align the site and header divs
+         */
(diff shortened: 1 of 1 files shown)


Axle7XStriker:
I am not able to figure out why the Travis CI build failed. Can anyone please tell me the reason behind it.

takluyver:
Probably a random failure. I've restarted the job that failed.

I'm not that keen on javascript to fix CSS layout issues. Is it possible to do something with CSS calc() instead?

Axle7XStriker:
I don't think we can get the same flexibility with CSS calc() as we get from using the 'auto' value in margin.

Zsailer:
I think we need to revert this change. It's causing new alignment issues discovered in #5687 that are slightly worse than before.

Issue #2591, opened 2017-06-21
Auto-indent is only available when editing cells


edgimar:
The auto-indent feature introduced in PR #963 only applies when editing cells in a notebook.  It would be nice if this could also apply to the file-editor interface (e.g. when viewing / editing an entire python file).

gnestor:
The `indentAuto` shortcut was removed because it was causing issues for French/German/etc keyboards (see https://github.com/jupyter/notebook/pull/2535). It would be nice to add it back with a different keymap. 

@Carreau Apparently, keymaps like `Ctrl-Alt-[` and `Cmd-Alt-[` disable French and German keyboards from typing `[`. What about `Cmd-Shift-[`? Would that be ok? What about`Cmd-\`?

gnestor:
`Cmd-Shift-[` is the shortcut for switching tabs in Chrome...

Carreau:
The typical reasonig for shortcut is the following:

No shortcut should use Alt (or Shift but in a lesser measure) by default. None. Ever. 

Many non-english keyboard need Alt and or Shift to type characters. So Shift and Alt are not available.

Typically, french keyboard use Shift for Numbers. So `Cmd-Shift-[` is actually "`Cmd-5`" on a french keyboard... which is taken, and on French keyboard typing `[` actually require pressing `Alt-Shift-5`... so I'm not even sure how that will conflict resolve, or if this could trigger ghost shortcuts.

Hence I would prefer _not_ to do it. What we can (and maybe should) do, is provide a set of "nice but not enabled by default" keymap that are a click away, and provide a "Querty-layout" extra, "Azerty layout extra", "Cyrillic extra", where we can go as crazy as we like.

PR #3175, opened 2018-01-02
Add Jupyter action for CodeMirror indentAuto command

gnestor:
Closes https://github.com/jupyter/notebook/issues/2591

Adds a shortcut for CodeMirror’s `indentAuto` command that doesn’t collide with special characters on French/German keyboards (https://github.com/jupyter/notebook/pull/2535) and also adds it to both the cell editor and text/file editor.

### DIFF
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -684,6 +684,16 @@ define([
                 }
             }
         },
+        'auto-indent': {
+            cmd: i18n.msg._('automatically indent selection'),
+            help : i18n.msg._('automatically indent selection'),
+            handler : function(env) {
+              // Get selected cell
+              var selected_cell = env.notebook.get_selected_cell();
+              // Execute a CM command
+              selected_cell.code_mirror.execCommand('indentAuto');
+            }
+        }
     };
 
     /**


takluyver:
How confident are we that this shortcut is not already used for anything in some browser? Generally, whenever we add a Ctrl- shortcut, we get complaints that it overrides some other shortcut that people like.

gnestor:
Similar sentiment expressed by @Carreau at https://github.com/jupyter/notebook/issues/2591#issuecomment-354978107:

> The typical reasonig for shortcut is the following:
>
> No shortcut should use Alt (or Shift but in a lesser measure) by default. None. Ever.
>
> Many non-english keyboard need Alt and or Shift to type characters. So Shift and Alt are not available.
>
> Typically, french keyboard use Shift for Numbers. So Cmd-Shift-[ is actually "Cmd-5" on a french keyboard... which is taken, and on French keyboard typing [ actually require pressing Alt-Shift-5... so I'm not even sure how that will conflict resolve, or if this could trigger ghost shortcuts.
>
> Hence I would prefer not to do it. What we can (and maybe should) do, is provide a set of "nice but not enabled by default" keymap that are a click away, and provide a "Querty-layout" extra, "Azerty layout extra", "Cyrillic extra", where we can go as crazy as we like.

gnestor:
Would it be possible to add some of these "nice to have but not necessary" CodeMirror commands to the notebook actions, allowing people to access `indentAuto` from the command palette as well as set a custom keyboard shortcut for it? In this instance, it could be as simple as:

```js
// Get selected cell
var selected_cell = notebook.get_selected_cell();
// Execute a CM command
selected_cell.code_mirror.execCommand('indentAuto');
```

takluyver:
I think that makes sense, yep. :-)

notebook/static/edit/js/editor.js
```diff
@@ -82,10 +82,10 @@ function(
             "Tab": "indentMore",
             "Shift-Tab" : "indentLess",
             "Cmd-/" : "toggleComment",
-            "Ctrl-/" : "toggleComment",
+            "Ctrl-/" : "toggleComment"
         },
         indentUnit: 4,
-        theme: "ipython",
+        theme: "default",
```
takluyver: Not sure why this change is part of this PR?

takluyver:
@gnestor if you want to get rid of the editor theme change, I think it's fine to merge this PR for 5.3.

notebook/static/edit/js/editor.js
gnestor: Initially, the intent was to make the CodeMirror config (shortcuts, theme, etc.) consistent between code cells in the notebook and the file editor. The code cells use the "default" theme and the editor uses the "ipython" theme. Is this intentional?

takluyver: I don't know. I don't have any particular objection if you think it's the right thing to do, but I think it would have been better to do it as a separate PR so it could be discussed separately.

Git blame shows @minrk set it to use the IPython theme three years ago in 0efd335553afe8d84d4a16aed9fc37db643390f2 . Min, do you have any recollection whether there was a need for that?

minrk: Oof, that's asking a lot of my foggy memory. Maybe it had to do with how magics were highlighted? Try highlighting some IPython-specific syntax and see if it works the same. If it looks fine, I've no objection.

gnestor:
Ok, I'm moving the syntax stuff to another PR. This is ready to merge 👍

takluyver:
Thanks :-)

Issue #2850, opened 2017-09-18
Copy from Terminal Selecting Wrong Value


cancan101:
I have the following on my terminal:
```
jjia@node002-jupyter-20170918-141251:/usr/src/app$ date
Mon Sep 18 15:27:19 EDT 2017
jjia@node002-jupyter-20170918-141251:/usr/src/app$
```
but when I try to copy and then paste the first two, I end up getting:
```
Mon Sep 18 15:27:19 EDT 2017
jjia@node002-jupyter-20170918-141251
```

cailiang9:
https://github.com/sourcelair/xterm.js/pull/910

parente:
Quick experimentation for a hacky workaround while things get fixed upstream. Replacing this rule:

https://github.com/jupyter/notebook/blob/master/notebook/static/terminal/less/terminal.less#L24-26

with:

```
.xterm-rows {
  top: 10px;
  left: 10px;
}
```

and adding rule:

```
.xterm-selection {
  top: 10px;
  left: 10px;
}
```

seems to work around the problem with xterm 2.8.1 included in the notebook 5.1 release.

cancan101:
@parente do you want to pr that change?

takluyver:
Is working around it going to cause any problems when it's fixed in xterm? I don't think we have a great way to ensure that the workaround is removed when we upgrade it.

oeway:
@cancan101 Could you please fix this issue?  sourcelair/xterm.js#910 seems ready. It's a bit annoying for copy and paste from the current terminal. Thank you.

cancan101:
V3 hasn't been released yet. Should be any day

cancan101:
New version of xterm (v3.0.1) has been released. That being said, they have [dropped support for bower](https://xtermjs.org/docs/guides/download/#no-more-bower-support). Any chance to pull in dependencies from npm / yarn?

PR #3189, opened 2018-01-08
Upgrade xterm.js to 3.1.0

cancan101:
Closes https://github.com/jupyter/notebook/issues/2850

### DIFF
--- a/notebook/static/terminal/js/main.js
+++ b/notebook/static/terminal/js/main.js
@@ -40,7 +40,6 @@ requirejs([
     
     page.show_header();
     
-    // var size = calculate_size();
     var terminal = terminado.make_terminal($("#terminado-container")[0], ws_url);
     
     page.show_site();
--- a/notebook/static/terminal/js/main.js
+++ b/notebook/static/terminal/js/main.js
@@ -32,9 +32,9 @@ requirejs([
     var login_widget = new loginwidget.LoginWidget('span#login_widget', common_options);
 
     // Test size: 25x80
-    var termRowHeight = function(){ return 1.00 * $("#dummy-screen")[0].offsetHeight / 25;};
+    // var termRowHeight = function(){ return 1.00 * $("#dummy-screen")[0].offsetHeight / 25;};
         // 1.02 here arrived at by trial and error to make the spacing look right
-    var termColWidth =  function() { return 1.02 * $("#dummy-screen-rows")[0].offsetWidth / 80;};
+    // var termColWidth =  function() { return 1.02 * $("#dummy-screen-rows")[0].offsetWidth / 80;};
 
     var base_url = utils.get_body_data('baseUrl').replace(/\/?$/, '/');
     var ws_path = utils.get_body_data('wsPath');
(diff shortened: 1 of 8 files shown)


blink1073:
I'm not very familiar with bower, but we would need to get `https://unpkg.com/xterm@~3.0.1/dist/xterm.css` as well.

taion:
You may be able to get away with inventing a fake "xtermjs-css" package and using that.

blink1073:
This suffers from the same problem that is holding up JupyterLab: we can't specify the padding yet.  You could work around it here by adding an extra wrapper div so that the terminal host div is already inset from the wrapper.

cf https://github.com/jupyterlab/jupyterlab/pull/3553, https://github.com/xtermjs/xterm.js/issues/946


image

blink1073:
For reference, on master:

image

Tyriar:
@blink1073 couldn't you just wrap the element you open xterm in inside a container with some padding? Does this not work? `<div style="padding: 10px"><div id="term"></div></div>`

blink1073:
Yep, that's what I recommended above :wink:  We didn't do that in JupyterLab because we were about to have a major release and there were a few other things that were not quite working yet in the changeover.

Tyriar:
@blink1073 oh ok, I thought that was a blocker for you

cancan101:
Ok, so what to do about this PR?

blink1073:
@cancan101, I'd say wrap this element as @Tyriar recommended above:

https://github.com/cancan101/notebook/blob/131502de3be2e6d4b12da2d93303e2cfb25409ea/notebook/templates/terminal.html#L26

cancan101:
I'm not sure that fix works. Maybe we have to wait for https://github.com/xtermjs/xterm.js/pull/1123?

blink1073:
@cancan101, why wouldn't wrapping the cell work?  It might also take updating some CSS/LESS.

cancan101:
image

blink1073:
Ah, I see what you mean.  It is more invasive than I'd hoped.  Yeah, I think we should wait for the upstream fix.

takluyver:
https://github.com/xtermjs/xterm.js/pull/1208 has been merged, so I guess we're just waiting for xtermjs 3.1.0 to be released

cancan101:
I bumped the version pin. Now I just need to set the padding.

cancan101:
Just pulling in 3.1.0 but without making any CSS changes, I get:
image

The left / top look okay, but something is overflowing on the right.

Tyriar:
Yes resizing is a little more involved now, please report if the fit addon doesn't meet your needs.

cancan101:
Ok, I think it's working now:
image

notebook/static/terminal/js/main.js
```diff
@@ -29,13 +29,6 @@ requirejs([
     var common_config = new configmod.ConfigSection('common', common_options);
     common_config.load();
 
-    var login_widget = new loginwidget.LoginWidget('span#login_widget', common_options);
```
takluyver: Please leave this in place - it makes the 'logout' button in the top right work.

notebook/static/terminal/js/main.js
```diff
@@ -45,32 +38,17 @@ requirejs([
...
     page.show_site();
     
     utils.load_extensions_from_config(config);
     utils.load_extensions_from_config(common_config);
     
-    window.onresize = function() { 
-      var geom = calculate_size();
-      terminal.term.resize(geom.cols, geom.rows);
-      terminal.socket.send(JSON.stringify(["set_size", geom.rows, geom.cols,
-                                    $(window).height(), $(window).width()]));
```
takluyver: We still need to send the new size to the server so that it can trigger a resize in the running process. Try running a full-screen program like `vim` or `aptitude`, and resize the terminal while it's running.

notebook/static/terminal/js/terminado.js
```diff
@@ -1,14 +1,10 @@
...
         var ws = new WebSocket(ws_url);
-        var term = new Terminal({
-          cols: size.cols,
-          rows: size.rows
-        });
+        Terminal.applyAddon(fit);
+        var term = new Terminal();
         ws.onopen = function(event) {
-            ws.send(JSON.stringify(["set_size", size.rows, size.cols,
-                                        window.innerHeight, window.innerWidth]));
```
takluyver: We'll still need to send this as well, so it knows how big the initial window is.

cancan101:
@takluyver ok reverted those and added some comments.

cancan101:
@Tyriar ran into this issue: https://github.com/xtermjs/xterm.js/issues/1283 . worked around it for the time being.

cancan101:
Looks like tests failed since "PhantomJS has crashed".

takluyver:
Thanks! The tests passed after I restarted the relevant job, and I've tried it manually.

PR #3195, opened 2018-01-09
Limit length of GUI autocomplete to 1000

areading314:
If an object's dir() call returns a long list (greater than a few thousand), the UI locks up while it is creating this long list of <option> elements. This just limits the maximum length to 1000 elements.

### DIFF
--- a/notebook/static/notebook/js/completer.js
+++ b/notebook/static/notebook/js/completer.js
@@ -306,7 +306,8 @@ define([
     };
 
     Completer.prototype.build_gui_list = function (completions) {
-        for (var i = 0; i < completions.length; ++i) {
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks. We also introduced a limit in IPython - see https://github.com/ipython/ipython/pull/10743 - but I think the new Jedi completions might not be affected by that.

takluyver:
I've marked this as for 5.4, as we're just about to release notebook 5.3.

areading314:
@takluyver is there anything else I need to do to get this merged?

takluyver:
No, I think this makes sense as is. Thanks!

Issue #3196, opened 2018-01-09
Problem signing out of Jupyter Notebooks if c.NotebookApp.base_url is set (cookie not cleared)


leontp587:
This is a bug report.

Versions tested:
- Notebook 5.2.2 on python 2.7 on RedHat Enterprise Linux 6.9 and Windows 7
- Notebook 5.2.1 on python 2.7 on Windows 7

Reproduce steps:
1) Install Jupyter notebooks with "pip installing jupyter"
2) Generate a config file with "jupyter notebook --generate-config"
3) Set a password in c.NotebookApp.password in .jupyter/jupyter_notebook_config.py
4) Set a base URL in c.NotebookApp.base_url in .jupyter/jupyter_notebook_config.py
5) Run "jupyter --notebook"
6) Login with password, and click logout, then login again. You'll see the user isn't logged out. Closing the browser doesn't help. The cookie is still present.

If I don't set a base_url (step 4 above), then the logout and cookie clearing works normally.

Can someone confirm this behavior? I tried on both Windows and Linux in my environment and hit this bug.

leontp587:
After looking at the code: the problem is in notebook/base/handlers.py

def clear_login_cookie(self):
        cookie_options = self.settings.get('cookie_options', {})
        path = cookie_options.setdefault('path', self.base_url)
        **self.clear_cookie(self.cookie_name, path=path) // first clear_cookie**
        if path and path != '/':
            # also clear cookie on / to ensure old cookies
            # are cleared after the change in path behavior.
            **self.clear_cookie(self.cookie_name) // second clear_cookie**

The problem is in the last line where I've marked //second clear_cookie. Commenting this out fixes the problem.

The issue is that clear_cookie is implemented by the Tornado backend on the server side by marking the cookie as expired. Cookies don’t actually expire on the server though, they are implemented by the client browser on the http response. Cookies have paths, the first call to clear_cookie on the line where I marked // first clear_cookie has a path set correctly, and the Tornado backend sets the cookie for expiration. The second call to clear_cookie inside the if block has no path set, so the Tornado backend gets confused and messes up the cookie expiration from the first call. So when the browser gets the response, it doesn’t clear the cookie.

The reason this doesn’t happen if Jupyter doesn’t have a base_url is because in that case path=’’, so the two clear_cookie calls are equivalent.

[...]

PR #3207, opened 2018-01-12
Fix clearing two cookies with the same name

takluyver:
This is an ugly workaround, and it might go wrong other code tries to manipulate the same cookies for the same response. But at least in my testing, it fixes the issue reported in #3196 .

I think a properly robust fix would involve replacing `http.cookies.BaseCookie` with something that can handle multiple cookies with the same name. But Tornado appears to consider it a corner case that's not worth supporting (https://github.com/tornadoweb/tornado/issues/1248), and it seems like a lot of work for us to build that just to tackle this one bug. Then again, it is a security bug, so maybe we should be taking that time.

cc @minrk .

Closes gh-3196

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -122,7 +122,7 @@ class AuthenticatedHandler(web.RequestHandler):
     def clear_login_cookie(self):
         cookie_options = self.settings.get('cookie_options', {})
         path = cookie_options.setdefault('path', self.base_url)
-        self.force_clear_cookie(self.cookie_name, path=path)
+        self.clear_cookie(self.cookie_name, path=path)
         if path and path != '/':
             # also clear cookie on / to ensure old cookies
             # are cleared after the change in path behavior.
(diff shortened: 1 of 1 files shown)


notebook/base/handlers.py
```diff
@@ -91,14 +94,39 @@ def set_default_headers(self):
...
     def clear_login_cookie(self):
         cookie_options = self.settings.get('cookie_options', {})
         path = cookie_options.setdefault('path', self.base_url)
-        self.clear_cookie(self.cookie_name, path=path)
+        self.force_clear_cookie(self.cookie_name, path=path)
         if path and path != '/':
             # also clear cookie on / to ensure old cookies
             # are cleared after the change in path behavior.
-            self.clear_cookie(self.cookie_name)
+            self.force_clear_cookie(self.cookie_name)
```
minrk: Maybe just call the force_clear_cookie here so we are using the normal clear_cookie most of the time? And add a FIXME noting that this is a backward-compatibily bit added in 5.3?

takluyver: Thanks, done.

takluyver:
I'm going to slip this into 5.3.

Issue #840, opened 2015-12-11
Color printing should be an option (or should be the default?)


aroberge:
Buried in the css files are various `@media print` configurations which prevent color printing.  They can be edited by hand if one has access to the installation directories, but there should be a way to have color printing be done without this type of editing.

willingc:
@aroberge Thanks for the suggestion. Personally, I never noticed that printing defaults to B&W whether from the Print Preview menu option or the browser print menu (at least on Firefox).

@carreau @ellisonbg Is there a setting that can be easily changed to enable color printing? If so, it would seem reasonable to document here and the user docs.

ellisonbg:
Hmm, I don't think we set anything that makes it BW. We will have to look
into this.

On Sat, Dec 12, 2015 at 3:54 AM, Carol Willing <notifications@github.com>
wrote:

> @aroberge <https://github.com/aroberge> Thanks for the suggestion.
> Personally, I never noticed that printing defaults to B&W whether from the
> Print Preview menu option or the browser print menu (at least on Firefox).
>
> @carreau <https://github.com/carreau> @ellisonbg
> <https://github.com/ellisonbg> Is there a setting that can be easily
> changed to enable color printing? If so, it would seem reasonable to
> document here and the user docs.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jupyter/notebook/issues/840#issuecomment-164142268>.
>



-- 
Brian E. Granger
Associate Professor of Physics and Data Science
Cal Poly State University, San Luis Obispo
@ellisonbg on Twitter and GitHub
bgranger@calpoly.edu and ellisonbg@gmail.com

aroberge:
I've installed Jupyter + IPython using the Anaconda Python distribution.  In my system, the relevant css file is (if I am not mistaken):
Anaconda3/pkgs/notebook-4.06-py34_0/Lib/site-packages/notebook/static/style/style.min.css

Carreau:
This is a bootstrap thing that include a `print.less` with `!important`. 
It's not easy to get rid of, but we can try.

mjbright:
Any progress on this?
Let me bump this one - it's really important to me.

I'm using Jupyter (+bash_kernel) to create some labs (Docker) where I use Jupyter to create the lab and then create a PDF to distribute to the students-*.

Those documents would be much more readable in colour.

Thanks for a great tool !

[...]

PR #3212, opened 2018-01-15
Patch out bootstrap rule forcing black and white printing

takluyver:
A thoroughly ugly hack. Bootstrap forces all text to be black when printing, resulting in a long running issue (#840). I tried to override it with `color: inherit !important`, but that didn't produce the right results. I can't find any way to tell LESS to remove a rule or act as if it didn't exist. So the only thing I can think to do is to patch the source we get from bootstrap to get rid of the rule entirely.

Frustratingly, it looks like bootstrap is fixing this (getting rid of the rule) for v4, but it's also moving from LESS to SCSS, so we can't easily switch to the new version.

Closes gh-840

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -398,6 +398,28 @@ class Bower(Command):
         update_package_data(self.distribution)
 
 
+def patch_out_bootstrap_bw_print():
+    """Hack! Manually patch out the bootstrap rule that forces printing in B&W.
+
+    We haven't found a way to override this rule with another one.
+    """
+    print_less = pjoin(static, 'components', 'bootstrap', 'less', 'print.less')
+    with open(print_less) as f:
+        lines = f.readlines()
+
+    for ix, line in enumerate(lines):
+        if 'Black prints faster' in line:
+            break
+    else:
+        return  # Already patched out, nothing to do.
+
+    rmed = lines.pop(ix)
+    print("Removed line", ix, "from bootstrap print.less:")
+    print("-", rmed)
+    print()
+    with open(print_less, 'w') as f:
+        f.writelines(lines)
+
 class CompileCSS(Command):
     """Recompile Notebook CSS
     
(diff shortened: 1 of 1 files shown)


setupbase.py
```diff
@@ -398,6 +398,28 @@ def run(self):
...
+    for ix, line in enumerate(lines):
+        if 'Black prints faster' in line:
+            break
+    else:
+        return  # Already patched out, nothing to do.
+
+    rmed = lines.pop(ix)
+    print("Removed line", ix, "from bootstrap print.less:")
+    print("-", rmed)
+    print()
```
damianavila: why this empty `print` here?

setupbase.py
```diff
@@ -398,6 +398,28 @@ def run(self):
...
+    """Hack! Manually patch out the bootstrap rule that forces printing in B&W.
+
+    We haven't found a way to override this rule with another one.
+    """
+    print_less = pjoin(static, 'components', 'bootstrap', 'less', 'print.less')
+    with open(print_less) as f:
+        lines = f.readlines()
+
+    for ix, line in enumerate(lines):
+        if 'Black prints faster' in line:
```
damianavila: I guess this is enough specific to patch the proper line, right?

setupbase.py
takluyver: It makes a blank line, so the text above is slightly easier to pick out.

setupbase.py
takluyver: Should be. The original file is not very long:

https://github.com/twbs/bootstrap/blob/v3.3.7/less/print.less

setupbase.py
damianavila: Fair enough 👍

Carreau:
Thanks both !

damianavila:
@mpacer @takluyver, we need to update the css in the CDN so nbconvert can catch this change.
Will open an issue to track this need.

Issue #3215, opened 2018-01-16
Favicon symlink causing Windows problems


takluyver:
`notebook/static/favicon.ico` is a symlink in the repository. Under at least some conditions, this causes issues on Windows:

```
    File "C:\Users\appveyor\AppData\Local\Temp\1\pip-le56htf7-build\setupbase.py", line 229, in check_package_data
      assert os.path.exists(path), "Missing package data: %s" % path
  AssertionError: Missing package data: notebook\static\favicon.ico
```

I'm not yet sure how common these conditions are. I'm trying to work around it on conda-forge.

PR #3220, opened 2018-01-16
Replace symlinked favicon with a copy

takluyver:
This is the simplest way to fix it for now. I may need to do a quick 5.3.1 release.

Closes gh-3215


Carreau:
Thanks for the quickfix , 

may I suggest a unit-test in travis that fail if we introduce symlink in the repo ?

takluyver:
You may, but I might suggest a PR ;-)

Hope you're getting over your jet lag OK.

PR #3225, opened 2018-01-17
Require Python >= 3.4

takluyver:
We discussed this in a recent dev meeting, and people seemed to agree that it's time to start doing this.

* We already did the same with IPython 6.0, and there hasn't been too much pushback - partly because of improvements to PyPI and pip ensuring they do the right thing for Py2 users.
* The upcoming Tornado 5 uses asyncio where available. This means there will be more difference between Python 2 and 3, and it will be easier for us to take advantage of Py3-only features.
* It will still be possible to run Python 2 code inside notebooks, by installing a Python 2 kernel. This isn't completely orthogonal to what the server runs, because it's easiest to install the server and the kernel on the same version of Python, but we're not leaving Python 2 users in the cold.

Like IPython, I imagine that we will continue to backport some fixes to 5.x and make some releases, but I think we'll probably apply a higher bar for what is backported, since people have the option of running the notebook server on Python 3 and using a Python 2 kernel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ cache:
         - $HOME/.cache/pip
 python:
     - 3.6
-    - 2.7
 
 sudo: required
 
@@ -70,9 +69,6 @@ matrix:
           env: GROUP=python
         - python: 3.5
           env: GROUP=python
-    exclude:
-        - python: 2.7
-          env: GROUP=docs
 
 after_success:
     - codecov
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,8 +6,8 @@ environment:
   matrix:
     - CONDA_PY: 35
       CONDA_INSTALL_LOCN: "C:\\Miniconda35-x64"
-    - CONDA_PY: 27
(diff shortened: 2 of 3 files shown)


rgbkrk:
Needs a rebase now, I'm happy to merge after that.

takluyver:
@mpacer is working towards releasing 5.6, so let's save this for after that's done.

If people want to help progress towards 6.0, I'd like to get more eyes on the API and applications of the new modules jupyter_kernel_mgmt and jupyter_protocol. I've recently had a go at updating jupyter_kernel_test to use these, for instance: https://github.com/jupyter/jupyter_kernel_test/pull/41

One specific design question: for request/reply messages, should an 'error' reply be automatically translated into an exception, or returned to let the application code deal with the different reply statuses? The API currently raises an `ErrorInKernel` exception, but in a couple of places now I've caught the exception to turn it back into a return value, so I wonder if that should be how it works.

blink1073:
Superseded by https://github.com/jupyter/notebook/pull/3926, thanks!

Issue #3235, opened 2018-01-18
Notebook 5.2.x produces sprintf error when trying to open a Notebook


rwmajor2:
We have a sandbox environment for sample Notebooks and when we try to open an existing sample (http://notebooks.esri.com/notebooks/samples/03_org_administrators/move_existing_user_content_to_a_new_user.ipynb), it generates the following error:

![image](https://user-images.githubusercontent.com/3799165/35109803-617abcfa-fc45-11e7-99c5-6b5e07302a5e.png)

It appears to be having a problem parsing parts of the Notebook file.  However, we do not experience this issue when working with Notebook 5.0.x.

Web browser console reports the following:

![image](https://user-images.githubusercontent.com/3799165/35109926-a9b48442-fc45-11e7-8b88-64008fd5a7ea.png)

```javascript
 /**
     * Success callback for loading a notebook from the server.
     * 
     * Load notebook data from the JSON response.
     * 
     * @param {object} data JSON representation of a notebook
     */
    Notebook.prototype.load_notebook_success = function (data) {
        var failed, msg;
        try {
            this.fromJSON(data);
        } catch (e) {
            failed = e;
            console.error("Notebook failed to load from JSON:", e);
        }
```

Any thoughts?

takluyver:
It looks like the error is coming from the translation machinery (Jed). I think it's trying to produce a warning and that's going wrong. Are you willing to share a copy of the notebook that causes the problem?

takluyver:
#3236 should fix this.

To work around it in the meantime, replace `null` with `1` (or any integer) on line 101.

rwmajor2:
@takluyver Man, I am having a tough time manually applying this fix.  If I  edit the file on my local Windows PC (C:\Anaconda3\Lib\site-packages\notebook\static\notebook\js\outputarea.js) and then start jupyter notebook, the JS still has the old value in Developer Tools (error still presents itself):

```javascript
    OutputArea.output_prompt_classical = function(prompt_value) {
        return $('<bdi>').text(i18n.msg.sprintf(i18n.msg._('Out[%d]:'),prompt_value));
    };
```

Where do I need to edit this file so it "sticks"?  I have cleared my browser and done everything I can think of.

Below is what my JS file on disk now looks like:
```javascript
   OutputArea.output_prompt_classical = function(prompt_value) {
        return $('<bdi>').text(i18n.msg.sprintf(i18n.msg._('Out[%s]:'),prompt_value));
    };
```

[...]

PR #3236, opened 2018-01-18
Fix output prompt when execution_count missing.

takluyver:
Closes gh-3235


takluyver:
Self merging, because it's a one character change.

Issue #3263, opened 2018-01-25
creating new notebook files not respecting current directory location unless browser refresh


steverweber:
```
# ubuntu 17.10
python3 -m pip install --upgrade notebook
jupyter-notebook --version
5.3.1
```

after login im in my home directory '/tree'

- use [new] > folder  "foo"
- click on folder "foo"
- click [new] > 'notebook' lets use python.. any notebook would work for this example

the new untitled notebook opens .. but discover it's not in the folder "foo" its in the root of my home directory.

If i repeat this process but use [new] > 'text file' the correct directory is used,

To workaround the issue
- use [new] > folder  "foobar"
- click on folder "foobar"
- refresh the browser F5
- creating a new notebook file now goes into the correct directory.

takluyver:
I can reproduce this, working on a PR.

gnestor:
Similar: https://github.com/jupyter/notebook/issues/3248

sargense:
Ticket should be reopened. Still experiencing this issue on 5.3.0

Linlinzhao:
I am having the same issue on 5.3.0 on Manjaro

gnestor:
Please upgrade to notebook 5.6.0

Issue #3248, opened 2018-01-23
v5.3.1 keeps creating notebooks at toplevel dir (notebook_dir) regardless of location.


draperjames:
I've seen this problem manifest on Linux Mint 18.3, Windows 7 and Windows 10 running FireFox 57.0.4 (64-bit). 

When I start Jupyter Notebook my browser launches to http://localhost:8888/tree then I navigate to a subdir let's call it `my_notebooks` when I the `New` button and select the type of notebook I want the notebook is created back at the toplevel dir http://localhost:8888/tree. 

This might not be connected but I noticed that the [changelog](https://github.com/jupyter/notebook/blob/4179d60e4912f6badd4ac34927edc7d47bb862aa/docs/source/changelog.rst#id10) 
 has this line:
> - Use DOM History API for navigating between directories in the file browser (:ghpull:`3115`).

I have Firefox's privacy settings set to never remember history. Would that break the new navigation method?

sarmoho:
I have the same problem in chrome with Windows10  since `conda update notebook --update-dependencies` from defaul notebook version of Anaconda3(notebook v5.0.0)

draperjames:
@sarmoho I find that down-grading to v5.2.1 fixes the problem right away. I think that I use the following commands from a administrator prompt;

    conda uninstall notebook
    pip install notebook==5.2.1

takluyver:
You were quite right about PR #3115 being the culprit, but the issue is there even with the default browser settings. #3263 should fix it.

PR #3264, opened 2018-01-25
Change the path to create notebooks in when navigating directories

takluyver:
Closes gh-3263
Closes gh-3248

There's still an issue where the dashboard goes back to showing the root directory when you create a notebook. I'll look into that later.

### DIFF
--- a/notebook/static/tree/js/newnotebook.js
+++ b/notebook/static/tree/js/newnotebook.js
@@ -13,7 +13,6 @@ define([
     var NewNotebookWidget = function (selector, options) {
         this.selector = selector;
         this.base_url = options.base_url;
-        this.notebook_path = options.notebook_path;
         this.contents = options.contents;
         this.events = options.events;
         this.default_kernel = null;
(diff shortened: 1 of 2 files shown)


takluyver:
I haven't fully understood the issue with the list showing the previous directory yet, but what I have found is that:

- I can trigger it by creating a new text file or a directory, as well as a new notebook.
- The `onpopstate()` handler we add is getting called, for some reason.

takluyver:
Educated guesswork led me to use `event.preventDefault()` on the links that make new files and folders, which seems to fix the issue with the files list jumping back to a previous state.

takluyver:
@gnestor @Carreau if you have a chance to review this, I think the issue is worth doing a 5.3.2 bugfix release for. I can make the release tomorrow if this gets merged.

Carreau:
Haven't looked at notebook codebase in a while but this look good to me.
Happy if we make a 5.3.2.

gnestor:
Thanks for fixing this @takluyver! I think that releasing a patch for this is a good idea as it's a regression that was reported by several people. Let me know if I can help with that 👍

takluyver:
I've released 5.4 to PyPI with the fix (and a few other changes). Now to update conda-forge.

PR #3273, opened 2018-01-28
Make buffer time between last modified on disk and last modified on last save configurable

gnestor:
Suggested by https://github.com/jupyter/notebook/issues/484#issuecomment-359508990

This will look in `notebook.json` in `~/.jupyter/nbconfig/notebook.json` for a `last_modified_buffer` value. For example:

```json
{
  "Notebook": {
    "Header": false,
    "Toolbar": true
  },
  "Cell": {
    "cm_config": {
      "lineNumbers": false
    }
  },
  "last_modified_buffer": 600
}
```

Can anyone think of better name than `last_modified_buffer`?

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2738,11 +2738,11 @@ define([
             return this.contents.get(this.notebook_path, {content: false}).then(
                 function (data) {
                     var last_modified = new Date(data.last_modified);
-                    var difference_buffer = that.config.data['last_modified_buffer'] || 500; // 500 ms
+                    var last_modified_check_margin = (that.config.data['last_modified_check_margin'] || 0.5) * 1000; // 500 ms
                     // We want to check last_modified (disk) > that.last_modified (our last save)
                     // In some cases the filesystem reports an inconsistent time,
                     // so we allow 0.5 seconds (configurable in nbconfig/notebook.json as `last_modified_buffer`) difference before complaining.
-                    if ((last_modified.getTime() - that.last_modified.getTime()) > difference_buffer) {  
+                    if ((last_modified.getTime() - that.last_modified.getTime()) > last_modified_check_margin) {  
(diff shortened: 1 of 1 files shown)


takluyver:
I'm trying to think of a better word than 'buffer', because that suggests a temporary storage area for data. Maybe `last_modified_check_margin`?

Also, I might make the config option in seconds rather than milliseconds; I think we generally try to keep intervals in seconds throughout our API for consistency (cc @minrk )

notebook/static/notebook/js/notebook.js
```diff
@@ -2738,10 +2738,11 @@ define([
             return this.contents.get(this.notebook_path, {content: false}).then(
                 function (data) {
                     var last_modified = new Date(data.last_modified);
+                    var last_modified_check_margin = (that.config.data['last_modified_check_margin'] || 0.5) * 1000; // 500 ms
                     // We want to check last_modified (disk) > that.last_modified (our last save)
                     // In some cases the filesystem reports an inconsistent time,
-                    // so we allow 0.5 seconds difference before complaining.
-                    if ((last_modified.getTime() - that.last_modified.getTime()) > 500) {  // 500 ms
+                    // so we allow 0.5 seconds (configurable in nbconfig/notebook.json as `last_modified_buffer`) difference before complaining.
```
takluyver: Don't forget to update the name in the comment too.

brettbrocato:
"last_modified_buffer": 2 or so caused all my notebooks to fail on opening with a 500 internal server error.  I set the buffer at 2 because I assumed the team had settled on seconds.  I've been trying to avoid the "overwrite?" dialog.  I'm on Win10 and notebook 5.7

philastrophist:
Adding the line `"last_modified_check_margin": 0.5` does nothing for me. This issue is not closed

ckbhatt:
Is there any plan to have this property configurable in Jupyter lab as well?

blink1073:
@ckbhatt I opened https://github.com/jupyterlab/jupyterlab/issues/8556 to track the feature in JupyterLab.

PR #3282, opened 2018-01-31
Fix clicking with modifiers, page title updates

### LINKED ISSUE #3276: Cmd/Ctrl clicking folders no longer opens in new tabs in 5.3
Instead, the file browser navigates to the folder in the current browser pane.

Probably related to the new history navigation feature. JS might be eating the hotkey modifier preventing the default browser behavior.

### LINKED ISSUE #3277: Folder Name does not change HTML title until refresh of folder's page (in 5.3)
HTML title is only updated to be the folder name when the page is refreshed, not when originally navigating to the folder's page.

After clicking on a folder:  
image

Then, refreshing that page:
image

parente:
* Allow the default browser behavior when the user is holding any modifier key while clicking on a folder or breadcrumb in the fileview.
* Update the page title when navigating via the History API

Fixes #3276, fixes #3277

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -412,6 +412,10 @@ define([
         $('body').attr('data-notebook-path', path);
         // Update the file tree list without reloading the page
         this.load_list();
+        // Update the page title so the browser tab reflects it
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks!

PR #3284, opened 2018-01-31
Improve kernel restart logic to reestablish ZMQ connection when a kernel is restarted with new ports

skukhtichev:
This proposal is based on https://github.com/jupyter/jupyter_client/pull/348.

This proposal defines the logic for creating new ZMQ connection without closing existing websocket connection if a kernel was restarted with new ports (e.g. when the kernel is not responding during initial startup).
The following changes were added to the `on_kernel_restarted` function:
- `kwargs` support
- `newports` parameter from `kwargs` used if it is defined. If the parameter is not in `kwargs` set it to `False`
- If `newports == True` then existing ZMQ connection to the kernel is closed and new stream with updated ports is opened.
- If `newports == False` then client is notified that the kernel was restarted
This proposal solves the issue described in https://github.com/jupyter/jupyter_client/issues/347

### DIFF
--- a/notebook/services/kernels/handlers.py
+++ b/notebook/services/kernels/handlers.py
@@ -461,13 +461,32 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
         msg['channel'] = 'iopub'
         self.write_message(json.dumps(msg, default=date_default))
 
-    def on_kernel_restarted(self):
+    def on_kernel_restarted(self, **kwargs):
+        if 'newports' in kwargs:
+            newports = kwargs['newports']
+        else:
+            self.log.warning('newports parameter is not defined, setting default to False')
+            newports = False
+        logging.info("Restarting kernel with new ports: {}".format(newports))
+        if newports:
+            km = self.kernel_manager
+            km.remove_restart_callback(
+                self.kernel_id, self.on_kernel_restarted,
+            )
+            km.remove_restart_callback(
+                self.kernel_id, self.on_restart_failed, 'dead',
+            )
+            km.stop_buffering(self.kernel_id)
+            km.notify_disconnect(self.kernel_id)
+            self.open(self.kernel_id)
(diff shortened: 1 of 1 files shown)


notebook/services/kernels/handlers.py
```diff
@@ -461,7 +461,32 @@ def _send_status_message(self, status):
...
         self.write_message(json.dumps(msg, default=date_default))
 
-    def on_kernel_restarted(self):
+    def on_kernel_restarted(self, **kwargs):
+        if 'newports' in kwargs:
+            self.log.debug("Kernel restarted with new ports")
+            newports = kwargs['newports']
+        else:
+            self.log.debug('newports parameter is not defined, setting default to False')
+            newports = False
```
takluyver: I think this can be simplified:

```python
def on_kernel_restarted(self, newports=False):
   ...
```

notebook/services/kernels/handlers.py
```diff
@@ -461,7 +461,32 @@ def _send_status_message(self, status):
...
+        if newports:
+            # Stream should be closed before reconnecting to the new ports
+            for channel, stream in self.channels.items():
+                if stream is not None and not stream.closed():
+                    stream.on_recv(None)
+                    stream.close()
+            # Connect to the stream with new ports
+            try:
+                self.create_stream()
+            except web.HTTPError as e:
```
takluyver: I can see this is copied from the `open()` method, but I don't think `create_stream` can actually raise `HTTPError`. Maybe it could at some point in the past. @minrk any ideas?

minrk: I think you're right. My guess is there used to be code in there where we were retrieving the kernel, which could raise 404, etc. Or it was just a mistake and this should have been Exception all along, which is what it should be now.

blink1073:
Closing this one as out of date, since a lot has changed in `juptyer_client` since.

PR #3288, opened 2018-02-01
Minor fixes

kant:
(proposal)


takluyver:
Maybe use `*asterisks for italics*` rather than backticks. The meaning of backticks is configurable in rst, and in some of our docs we configure it to mean inline code.

kant:
Understood!

takluyver:
Ping? If you've got a moment to switch it to asterisks, I'm happy to merge this.

takluyver:
I made the change myself to tidy this up.

PR #3289, opened 2018-02-01
Add save as menu option

### LINKED ISSUE #2816: feature request: "save as..." under the file menu
It would be really helpful if there was a "save as..." item in the file menu of Jupyter notebook.  The current workaround is rather cumbersome:

- "Make a Copy..."
- "Rename..."
- switch back to the old notebook
- "Close and Halt"

Having "Save as..." would be a way of getting most of the benefit of having templates (https://github.com/jupyter/notebook/issues/332) but with much less effort.

Madhu94:
Trying to close #2816

I'd like some quick, initial feedback on this. 

I am checking the tests. The UI needs some work too (maybe a directory picker instead of a text box?) 

Thanks.

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2899,7 +2899,7 @@ define([
                                    $('.save-message').html(`<span style='color:red;'>${msg}</span>`);
                                 });
                         }
-                        that.contents.get(nb_path, {type: 'notebook'}).then(function(data) {
+                        that.contents.get(nb_path, {type: 'notebook', content: false}).then(function(data) {
                             var warning_body = $('<div/>').append(
                                 $("<p/>").text(i18n.msg._('Notebook with that name exists.')));
                             dialog.modal({
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2885,7 +2885,7 @@ define([
                                 'type': 'notebook',
                                 'content': that.toJSON(),
                                 'name': nb_name
-                             };
+                            };
                             return that.contents.save(nb_path, model)
                                 .then(function(data) {
                                     d.modal('hide');
@@ -2893,12 +2893,15 @@ define([
                                     that.notebook_path = data.path;
                                     that.session.rename_notebook(data.path);
                                     that.events.trigger('notebook_renamed.Notebook', data);
-
-                                },function(error) {
-                                   const msg = i18n.msg._(error.message || 'Unknown error saving notebook');
-                                   $('.save-message').html(`<span style='color:red;'>${msg}</span>`);
+                                }, function(error) {
+                                    var msg = i18n.msg._(error.message || 'Unknown error saving notebook');
+                                    $(".save-message").html(
+                                        $("<span>")
+                                            .attr("style", "color:red;")
+                                            .text(msg)
+                                    );
                                 });
-                        }
+                        };
                         that.contents.get(nb_path, {type: 'notebook', content: false}).then(function(data) {
                             var warning_body = $('<div/>').append(
                                 $("<p/>").text(i18n.msg._('Notebook with that name exists.')));
@@ -2916,7 +2919,7 @@ define([
                             });
                         }, function(err) {
                             return save_thunk();
-                        })
+                        });
                         return false;
                     }
                 },
@@ -2931,7 +2934,7 @@ define([
                 d.find('input[type="text"]').val(current_dir).focus();
              }
          });
-    }
+    };
 
     /**
      * Update the autosave interval based on the duration of the last save.
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2853,14 +2853,21 @@ define([
             this.save_notebook();
         }
         var that = this;
+        var current_dir = $('body').attr('data-notebook-path').split('/').slice(0, -1).join("/");
+        current_dir = current_dir? currend_dir + "/": "";
         var dialog_body = $('<div/>').append(
-            $("<p/>").addClass("save-message")
+            $('<p/>').addClass('save-message')
                 .text(i18n.msg._('Enter a notebook path relative to notebook dir'))
         ).append(
-            $("<br/>")
+            $('<br/>')
+        ).append($('<label />').attr('for', 'save-as-dialog')
+            .html('<i class="fa fa-home"></i>')
         ).append(
             $('<input/>').attr('type','text').attr('size','25')
-            .addClass('form-control').attr('placeholder', that.notebook_name)
+            .attr('value', current_dir)
+            .css({'display': 'inline-block', 'margin-left': '0.5em'})
+            .attr('id', 'save-as-dialog')
+            .addClass('form-inline')
         );
 
         var d = dialog.modal({
@@ -2889,7 +2896,7 @@ define([
                              };
                             return that.contents.save(nb_path, model)
                                 .then(function(data) {
-                                    window.open(data.path, '_self');
+                                    window.open(data.path, '_blank');
                                 },function(error) {
                                    console.error(i18n.msg._(error.message || 'Unknown error saving notebook'));
                                 });
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2854,7 +2854,7 @@ define([
         }
         var that = this;
         var current_dir = $('body').attr('data-notebook-path').split('/').slice(0, -1).join("/");
-        current_dir = current_dir? currend_dir + "/": "";
+        current_dir = current_dir? current_dir + "/": "";
         var dialog_body = $('<div/>').append(
             $('<p/>').addClass('save-message')
                 .text(i18n.msg._('Enter a notebook path relative to notebook dir'))
@@ -2896,7 +2896,12 @@ define([
                              };
                             return that.contents.save(nb_path, model)
                                 .then(function(data) {
-                                    window.open(data.path, '_blank');
+                                    d.modal('hide');
+                                    that.notebook_name = data.name;
+                                    that.notebook_path = data.path;
+                                    that.session.rename_notebook(data.path);
+                                    that.events.trigger('notebook_renamed.Notebook', data);
+
                                 },function(error) {
                                    console.error(i18n.msg._(error.message || 'Unknown error saving notebook'));
                                 });
@@ -2924,6 +2929,12 @@ define([
                 },
             },
             open : function () {
+                d.find('input[type="text"]').keydown(function (event) {
+                    if (event.which === keyboard.keycodes.enter) {
+                        d.find('.btn-primary').first().click();
+                        return false;
+                    }
+                });
                 d.find('input[type="text"]').focus().select();
              }
          });
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2858,6 +2858,7 @@ define([
             $('<br/>')
         ).append(
             $('<input/>').attr('type','text').attr('size','25')
+            .attr('data-testid', 'save-as')
             .addClass('form-control')
         );
 
@@ -2894,7 +2895,8 @@ define([
                                     that.events.trigger('notebook_renamed.Notebook', data);
 
                                 },function(error) {
-                                   console.error(i18n.msg._(error.message || 'Unknown error saving notebook'));
+                                   const msg = i18n.msg._(error.message || 'Unknown error saving notebook');
+                                   $('.save-message').html(`<span style='color:red;'>${msg}</span>`);
                                 });
                         }
                         that.contents.get(nb_path, {type: 'notebook'}).then(function(data) {
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2848,10 +2848,6 @@ define([
     };
 
     Notebook.prototype.save_notebook_as = function() {
-        // If current notebook has some changes, save them, or the copied notebook won't have them.
-        if (this.writable && this.dirty) {
-            this.save_notebook();
-        }
         var that = this;
         var current_dir = $('body').attr('data-notebook-path').split('/').slice(0, -1).join("/");
         current_dir = current_dir? current_dir + "/": "";
@@ -2860,14 +2856,9 @@ define([
                 .text(i18n.msg._('Enter a notebook path relative to notebook dir'))
         ).append(
             $('<br/>')
-        ).append($('<label />').attr('for', 'save-as-dialog')
-            .html('<i class="fa fa-home"></i>')
         ).append(
             $('<input/>').attr('type','text').attr('size','25')
-            .attr('value', current_dir)
-            .css({'display': 'inline-block', 'margin-left': '0.5em'})
-            .attr('id', 'save-as-dialog')
-            .addClass('form-inline')
+            .addClass('form-control')
         );
 
         var d = dialog.modal({
@@ -2935,7 +2926,7 @@ define([
                         return false;
                     }
                 });
-                d.find('input[type="text"]').focus().select();
+                d.find('input[type="text"]').val(current_dir).focus();
              }
          });
     }
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2846,7 +2846,82 @@ define([
             this._checkpoint_after_save = false;
         }
     };
-    
+
+    Notebook.prototype.save_notebook_as = function() {
+        // If current notebook has some changes, save them, or the copied notebook won't have them.
+        if (this.writable && this.dirty) {
+            this.save_notebook();
+        }
+        var that = this;
+        var dialog_body = $('<div/>').append(
+            $("<p/>").addClass("save-message")
+                .text(i18n.msg._('Enter a notebook path relative to notebook dir'))
+        ).append(
+            $("<br/>")
+        ).append(
+            $('<input/>').attr('type','text').attr('size','25')
+            .addClass('form-control').attr('placeholder', that.notebook_name)
+        );
+
+        var d = dialog.modal({
+            title: 'Save As',
+            body: dialog_body,
+            keyboard_manager: this.keyboard_manager,
+            notebook: this,
+            buttons: {
+                Cancel: {},
+                Save: {
+                    class: 'btn-primary',
+                    click: function() {
+                        var nb_path = d.find('input').val();
+                        var nb_name = nb_path.split('/').slice(-1).pop();
+                        // If notebook name does not contain extension '.ipynb' add it
+                        var ext = utils.splitext(nb_name)[1];
+                        if (ext === '') {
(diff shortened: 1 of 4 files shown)


takluyver:
I would expect save-as to work without needing any changes to the REST API or the server. It should update the path for the notebook file in the frontend, and then do a regular save (i.e. a PUT request to the URL of the new file).

This is in line with how 'save as' works in most applications: it changes the file path you're working on, so if you open A and then 'save as' B, subsequent saves will write B until you use 'save as' again.

Some applications have a separate 'save a copy' action, which lets you write file B but keep working on A. I don't think that's a priority to add to the notebook - not many other apps have it.

Madhu94:
That's how I started but I think [this line](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/handlers.py#L219) stopped me.

takluyver:
Yup, I'd just send the whole contents again (for now, every save does that). The notebook may have changed since it was last saved, so copying the file we saved before may not be right.

Madhu94:
@takluyver Don't you think that if "Save As" behaves like "Save",  the user could accidentally overwrite a notebook, if they are not careful?

I guess "save as" should do an additional check and warn the user that the file exists..

takluyver:
Yes, ideally if the user tries to 'save as' over an existing file, they should be warned about that.

Madhu94:
@takluyver Now the user is asked to confirm if it is okay to overwrite. Let me know if these changes are okay.

takluyver:
I think the frontend stuff (the dialog, updating the URL, etc.) should be able to mostly use the code we already have in `static/notebook/js/savewidget.js` to rename a notebook. On this side, I would do what you started out doing on the server side: have an option like `copy_file: true` to switch the behaviour from moving to copying.

Madhu94:
> On this side, I would do what you started out doing on the server side: have an option like copy_file: true to switch the behaviour from moving to copying.  

Did you mean for reusing the dialog, I should pass a flag ?

takluyver:
Yep, so it would be called like this:

```javascript
notebook.save_widget.rename_notebook({notebook: notebook, copy_file: true});
```

Madhu94:
I've factored out the common portions for "save as" and rename. I know this
isn't what you had in mind, but using a flag brought a lot of "if -else"
clauses within the click handler of rename. I'll do that if you feel
strongly about it though.

Re the feature itself, do you think it should take paths relative to
current directory and not the notebook root ? That's how creating a new
notebook works..

takluyver:
From a user perspective, I think the starting point should be the directory the notebook is currently saved in. The REST API should always work with paths from the notebook root, though.

notebook/static/notebook/js/notebook.js
```diff
@@ -2845,6 +2845,57 @@ define([
         }
     };
     
+       
+    Notebook.prototype.save_notebook_as = function() {
```
takluyver: I'd probably put this method into `savewidget.js` because of the symmetry with rename.

notebook/static/notebook/js/notebook.js
```diff
@@ -2845,6 +2845,57 @@ define([
...
+            // If current notebook has some changes, save them
+            if (this.writable && this.dirty) {
+                this.save_notebook();
```
takluyver: I don't think I'd expect 'save as' to save over the old name as well, though it's a bit complicated by autosave. @minrk @gnestor @Carreau ?

notebook/static/notebook/js/notebook.js
```diff
@@ -2845,6 +2845,57 @@ define([
...
+            }
+            var that = this;
+            var d = that.save_widget._nb_name_dialog({
+                dialog_title: i18n.msg._('Save As'),
+                message: i18n.msg._('Enter a notebook path relative to notebook dir'),
+                message_classname: 'save-message',
+                button_name: 'Save',
```
takluyver: This should probably be translatable.

notebook/static/notebook/js/notebook.js
```diff
@@ -2845,6 +2845,57 @@ define([
...
+                        nb_name = nb_name + '.ipynb';
+                        nb_path = nb_path + '.ipynb';
+                    }
+                    var save_thunk = function() {
+                      var model = {'type': 'notebook','content': that.toJSON(),
+                                   'name': nb_name};
+                      return that.contents.save(nb_path, model)
+                          .then(function(data) {
+                              var dir_path = utils.get_body_data("baseUrl");
+                              window.open(utils.url_path_join(dir_path, "notebooks", data.path), '_self');
```
takluyver: There's already code to deal with updating the address bar and other details when the notebook is renamed, based on the `notebook_renamed.Notebook` event. Instead of calling window.open(), we should make sure that event fires, as on this line:

https://github.com/jupyter/notebook/blob/5fdf8c23061607824cfab19f6840de692a916d1a/notebook/static/notebook/js/notebook.js#L3036

We should also ensure that the session is updated - that links a kernel to a filename. See the line above the one I just linked to.

notebook/static/notebook/js/savewidget.js
```diff
@@ -65,6 +65,61 @@ define([
...
+    
+    /** A modal dialog that is used for the specific purpose of
+   * getting a notebook name, validating it, and passing a reference 
+   * of it to the button click handler.
+   * @param options Expected to have dialog message, dialog title,
+   * a primary button and a click handler associated with them.
+   */
+   SaveWidget.prototype._nb_name_dialog = function(options) {
+       var that = this;
+       var dialog_template = _.template(`<p class="<%= classname %>"><%= msg %></p>
```
takluyver: Neat! I assume the `_.template()` makes it translatable?

notebook/static/notebook/js/notebook.js
Madhu94: Will do!

notebook/static/notebook/js/notebook.js
Madhu94: This is to save the current notebook before navigating away. We did agree that "Save as" would open the newly created notebook (Right? Did I misunderstand?), so I thought we should try to save the current notebook before moving away. I think "make a copy" and "trust" do this too.

notebook/static/notebook/js/notebook.js
Madhu94: I'm not sure if triggering event will cause the new notebook to open up...I'll check and get back.

notebook/static/notebook/js/savewidget.js
Madhu94: No, I'm sorry. Underscore templates are just plain html templates.  This was just an attempt to share some of the ui code for "save as" and "rename".

Madhu94:
I'm really, really sorry, but I'm not able to understand why the above changes are requested, hence a lot of delay. I'll come out and place my concerns here - 
I don't get how save as and rename are symmetrical.
Rename
- patches an existing notebook 
- Updates the url and
- changes the kernel session
"Save As" 
- creates a new notebook
- checks that a notebook of the same name does not exist
- Opens the newly created notebook (hence no need to change the session in the kernel ??)
The only shared code I could find is that we need a very similar dialog box - we could use underscore templates here.

Please let me know why you think "save as" and "rename" should be implemented the same.

Thanks for your patience.

takluyver:
Let me try to put those as I see them:

**Rename**

- Pick a new name that doesn't already exist
- Ask the server to move the notebook file to the new name
- Update the session so that our kernel is associated with the new name
- Update the frontend (URL, notebook name in the page, possibly other things)

**Save as**


- Pick a new name that doesn't already exist
- Ask the server to copy the notebook file to the new name
- Update the session so that our kernel is associated with the new name
- Update the frontend (URL, notebook name in the page, possibly other things)

Does that make it clearer?

I think maybe the disconnect is that you're talking about opening the new notebook after 'save as' - are you thinking it would open a second tab, so you would have both notebooks open? It's an interesting idea, but what I describe is the way lots of familiar programs do 'save as' - you have just one editor open afterwards, and it's associated with the new file.

Madhu94:
I see. The disconnect was because I implemented "save as" to work with any path relative to the notebook's root_dir, and I don't think that can not be implemented with a rename.

I guess that isn't relevant now as we later agreed to implement save as _only in the current directory_. 

Thanks, I will implement "Save As" as rename + save.

notebook/static/notebook/js/notebook.js
Carreau: > This is to save the current notebook before navigating away. We did agree that "Save as" would open the newly created notebook (Right? Did I misunderstand?), so I thought we should try to save the current notebook before moving away. I think "make a copy" and "trust" do this too.

If you do open the new notebook, then this will likely be almost identical to "Make a Copy", Should we have "Make a copy" take a parameter which is the new name/path ?

notebook/templates/notebook.html
```diff
@@ -32,6 +32,7 @@
 data-ws-url="{{ws_url | urlencode}}"
 data-notebook-name="{{notebook_name | urlencode}}"
 data-notebook-path="{{notebook_path | urlencode}}"
+data-server-root="{{server_root}}"
```
Carreau: I would like to try that without exposing `server_root` it is potentially sensitive informations, and we tried out best to not expose it anywhere in the UI. The root is usually represented as the Home icon. 

As far as I can tell having this information is only for display purpose, so I would just stay with the home icon.

Carreau:
I agree with your plan, and think that with this approach we have something really close to what "Make a copy" does, and it would be nice to get both functionality closer. I think that "Save as..." may be more familiar than "Make a copy", and we could try to see if "Make a copy" could be removed in the longer term.

Madhu94:
I'm a little confused now

I think it would be easier to decide what to do if we fix on how "Save As"
should behave - save a new notebook in the current directory or anywhere
relative to root_dir ?

If we decide that it is the current directory, I agree with @takluyver that
this feature is a lot like rename

If not, rename may not be identical to Save As.

takluyver:
>  I implemented "save as" to work with any path relative to the notebook's root_dir, and I don't think that can be implemented with a rename.

Ah, my bad. Rename and move are the same operation in the REST API, and I thought we'd made the GUI so that rename could move it to a different folder, but I was wrong about that.

>  I think it would be easier to decide what to do if we fix on how "Save As" should behave - save a new notebook in the current directory or anywhere relative to root_dir ?

I think it should be able to save anywhere under the server root directory, but in terms of the UI, it should work relative to the directory the notebook is already in. So if I start the notebook server at `/home/takluyver` and open a notebook `/home/takluyver/foo/ThisNotebook.ipynb`, then 'Save as' should assume that I want to stay in the `foo` folder unless I tell it otherwise.

notebook/templates/notebook.html
takluyver: I actually disagree. I don't think `server_root` is really that sensitive, and it's probably not hard to work out from the list of files, which is already accessible. The last time we thought about this was before we enabled authentication by default; now that we've done that, I think we can be slightly more relaxed about showing this kind of information.

We've been using the home icon for lack of anything better, but I don't think it's great UI:

- It's easy to confuse with your home directory, especially as the two are sometimes but not always the same. File managers frequently use a very similar icon for your home directory (e.g. in Gnome: 
screenshot from 2018-04-21 08-39-43
- It obscures where the files are actually stored, if you want to go and work with them outside the notebook. At one point Anaconda's GUI launcher would create a directory like 'Notebooks' to run the server in, which is hard to discover from inside.

We should make sure it comes from the contents manager and works if it's empty, to allow for non-filesystem contents managers. But I think the benefits of showing it outweigh the disadvantages.

Madhu94:
> unless I tell it otherwise.  

So, if the user types ``bar/copynotebook.ipynb`` inside ``foo``, we  save under ``/home/takluyver/foo/bar/copynotebook.ipynb``?

Madhu94:
I've changed it back to the original implementation which behaved a little like "Make a copy". Even deciding to make it relative to current dir should be as easy as appending ``$('body').attr('data-notebook-path')`` to the notebook path, so I'm removing the "WIP" from the subject

takluyver:
> So, if the user types bar/copynotebook.ipynb inside foo, we save under /home/takluyver/foo/bar/copynotebook.ipynb?

Yes... but if possible, could we make it so that when the dialog appears, `foo/` is already filled in, and the cursor is at the end? So you could type a name to save in the same folder, or backspace to take it to another folder.

notebook/static/notebook/js/notebook.js
```diff
@@ -2844,7 +2844,82 @@ define([
...
+                        }
+                        var save_thunk = function() {
+                            var model = {
+                                'type': 'notebook',
+                                'content': that.toJSON(),
+                                'name': nb_name
+                             };
+                            return that.contents.save(nb_path, model)
+                                .then(function(data) {
+                                    window.open(data.path, '_self');
```
takluyver: Here I think we should use the same functions that rename uses to update the session and the page:

https://github.com/jupyter/notebook/blob/2aca6f98b0e46c9a3d635faf0c03abf113ff4089/notebook/static/notebook/js/notebook.js#L2984-L2985

Madhu94:
I think I got the behavior you wanted now. (Although the save as method has become a callback+promise hell...)

notebook/static/notebook/js/notebook.js
```diff
@@ -2844,7 +2844,100 @@ define([
             this._checkpoint_after_save = false;
         }
     };
-    
+
+    Notebook.prototype.save_notebook_as = function() {
+        // If current notebook has some changes, save them, or the copied notebook won't have them.
```
takluyver: Is this still needed? We're sending a new copy of the notebook to the new name, so I don't think we need to save it at the old name again first.

notebook/static/notebook/js/notebook.js
```diff
@@ -2844,7 +2844,100 @@ define([
...
+        current_dir = current_dir? current_dir + "/": "";
+        var dialog_body = $('<div/>').append(
+            $('<p/>').addClass('save-message')
+                .text(i18n.msg._('Enter a notebook path relative to notebook dir'))
+        ).append(
+            $('<br/>')
+        ).append($('<label />').attr('for', 'save-as-dialog')
+            .html('<i class="fa fa-home"></i>')
+        ).append(
+            $('<input/>').attr('type','text').attr('size','25')
```
takluyver: This input area doesn't show up well for me. I think you're missing the `form-control` class that the rename dialog uses:

https://github.com/jupyter/notebook/blob/2aca6f98b0e46c9a3d635faf0c03abf113ff4089/notebook/static/notebook/js/savewidget.js#L83

Also, I hope we can get rid of the inline styling a couple of lines below.

notebook/static/notebook/js/notebook.js
```diff
@@ -2844,7 +2844,100 @@ define([
...
+                },
+            },
+            open : function () {
+                d.find('input[type="text"]').keydown(function (event) {
+                    if (event.which === keyboard.keycodes.enter) {
+                        d.find('.btn-primary').first().click();
+                        return false;
+                    }
+                });
+                d.find('input[type="text"]').focus().select();
```
takluyver: This currently selects the path pre-filled in the input, which is not ideal. Can we make it put the cursor at the end of the input already there?

notebook/templates/notebook.html
```diff
@@ -89,6 +89,9 @@
                         <li id="copy_notebook"
                             title="{% trans %}Open a copy of this notebook's contents and start a new kernel{% endtrans %}">
                             <a href="#">{% trans %}Make a Copy...{% endtrans %}</a></li>
+                        <li id="save_notebook_as"
+                            title="{% trans %}Save a copy of the notebook's contents and start a new kernel{% endtrans %}">
+                            <a href="#">{% trans %}Save as{% endtrans %}</a></li>
```
takluyver: This label should end `...` to indicate that it brings up a dialog rather than acting immediately.

Madhu94:
> I'd also like to have a basic Selenium test for this - see this folder: https://github.com/jupyter/notebook/tree/master/notebook/tests/selenium

I added one. Also took care of the css issues.

Madhu94:
Hi @takluyver @Carreau, any updates here ?

notebook/static/notebook/js/notebook.js
```diff
@@ -2846,7 +2846,93 @@ define([
...
+                                    that.notebook_path = data.path;
+                                    that.session.rename_notebook(data.path);
+                                    that.events.trigger('notebook_renamed.Notebook', data);
+
+                                },function(error) {
+                                   const msg = i18n.msg._(error.message || 'Unknown error saving notebook');
+                                   $('.save-message').html(`<span style='color:red;'>${msg}</span>`);
+                                });
+                        }
+                        that.contents.get(nb_path, {type: 'notebook'}).then(function(data) {
```
takluyver: This should at least use `content: false` to avoid getting the full content of the other notebook if it does exist. It's also not great in that it assumes any error getting a model means that the file is not there.

The right way to do this is probably to add a parameter when saving, to ask the server to throw an error if the filename already exists. Maybe this way of doing it is good enough for now, though. I'd like to get some other opinions on it.

Madhu94: > This should at least use content: false to avoid getting the full content of the other notebook if it does exist.  

This is done, thank you.

> It's also not great in that it assumes any error getting a model means that the file is not there.
I will try to think of a better way.  Maybe we go back to something like my original implementation which used a flag to toggle between copy/save on the server..

Madhu94:
What do you think of adding a HEAD HTTP method to the contents handler? This would simply call contents_manager.file_exists or contents_manager.dir_exists.  (I'm not sure if things like this are "done")

It might be better than adding a flag while saving, I don't like how that looks - 

```
if exists:
                if model.get('overwrite', True):
                    yield gen.maybe_future(self._save(model.get("model"), path))
                else:
                    raise web.HTTPError(400, "File with that name exists")
```

takluyver:
That's roughly what `content: false` already does, but with the metadata in the JSON model rather than HTTP headers.

@minrk @Carreau @mpacer I'd like to get another pair of eyes on this PR if possible. Thanks!

takluyver:
@Madhu94 It seems to be failing the tests, if you could take a look.

Madhu94:
> @Madhu94 It seems to be failing the tests, if you could take a look.

".CodeMirror-code" still did not exist" => I thought this was some transient Travis bug, will check

takluyver:
I think that error can mean it's failing to load the notebook page - it waits for a codemirror block to exist to check that the notebook is loading correctly. The fact that it's doing that on all frontend tests suggests that there's a real problem somewhere (though it might be that something else changed at the same time).

minrk:
I tracked down the test failure to some ES6 syntax in the new javascript. Removing that should let the tests pass.

takluyver:
Indeed it does :-). Thanks Min, and thanks @Madhu94 for your patience with this. Merging now.

takluyver:
Heads up @mpacer that this is another change for 5.6.

Madhu94:
Thanks @minrk, I was having a lot of trouble with that.  Thanks for all
your help, @taklyuver

rachelgshaffer:
@Madhu94 @takluyver quick question... I am curious about the implementation here. The tooltip indicates that it starts a new session but seems the intent was to associate the new name with the same kernel session. I wanted to double check on the actual behavior: does it keep the same kernel or start a new one?

PR #3291, opened 2018-02-01
Latex and markdown instructions for learners

layik:
Notebook server: 5.4.0, python 2.7.12

### DIFF
--- a/docs/source/examples/Notebook/Working With Markdown Cells.ipynb
+++ b/docs/source/examples/Notebook/Working With Markdown Cells.ipynb
@@ -161,6 +161,7 @@
     "Courtesy of MathJax, you can include mathematical expressions both inline: \n",
     "$e^{i\\pi} + 1 = 0$  and displayed:\n",
     "\n",
+    "Make sure you call %%latex before the statement \n",
     "$$e^x=\\sum_{i=0}^\\infty \\frac{1}{i!}x^i$$\n",
     "\n",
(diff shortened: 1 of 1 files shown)


docs/source/examples/Notebook/Working With Markdown Cells.ipynb
```diff
@@ -161,6 +161,7 @@
     "Courtesy of MathJax, you can include mathematical expressions both inline: \n",
     "$e^{i\\pi} + 1 = 0$  and displayed:\n",
     "\n",
+    "Make sure you call %%latex before the statement \n",
```
takluyver: This whole notebook is about markdown cells. The `%%latex` cell magic is a shortcut to make latex output from *code* cells, which is a separate topic. In Markdown cells, you can write latex surrounded by `$` signs, as the examples already show.

docs/source/examples/Notebook/Working With Markdown Cells.ipynb
```diff
@@ -189,6 +190,7 @@
    "source": [
     "The Notebook webapp supports Github flavored markdown meaning that you can use triple backticks for code blocks:\n",
     "\n",
+    "Make sure change the cell to markdown from Cell menu. \n",
```
takluyver: If this needs clarification, it should probably be at the start of the notebook, because the whole document is talking about markdown cells, not just this bit.

layik:
I will double check but current instruction is not accurate and made me realise something was not quite clear. Will get back to it later.

Issue #3249, opened 2018-01-23
Deleting notebooks is broken


jdemeyer:
With Jupyter notebook 5.3.1, I can no longer delete notebook files:
```
[E 11:16:58.653 NotebookApp] Uncaught exception DELETE /api/contents/Untitled2.ipynb (::1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='DELETE', uri='/api/contents/Untitled2.ipynb', version='HTTP/1.1', remote_ip='::1', headers={'Accept-Language': 'nl,en;q=0.5', 'Accept-Encoding': 'gzip, deflate', 'X-Xsrftoken': '2|53abba32|97dda7043bc73dc31a6c8b157cfbc04a|1515155636', 'Host': 'localhost:8888', 'Accept': 'application/json, text/javascript, */*; q=0.01', 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64; rv:45.0) Gecko/20100101 Firefox/45.0', 'Connection': 'keep-alive', 'X-Requested-With': 'XMLHttpRequest', 'Referer': 'http://localhost:8888/tree', 'Cookie': 'username-localhost-8888="2|1:0|10:1516702258|23:username-localhost-8888|44:MjY1ODA5Njc1NzVlNDZmYmFjYTZlMjkxNDI4ZDM4Yzg=|6b2a4bb9ff2376377b714716929ddda9943e239553d9ebbb381de8de9c60162e"; _xsrf=2|53abba32|97dda7043bc73dc31a6c8b157cfbc04a|1515155636; username-localhost-8889="2|1:0|10:1516109384|23:username-localhost-8889|44:ODFmNjM5YjBkYmQ5NDFiZmE3NWUyYmMzZjUxOTM4MTM=|c76b2846927ce1ac607b0d7da30f7a6c254cbc251d82c43cb316c26622c8f765"'})
    Traceback (most recent call last):
      File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/tornado/web.py", line 1511, in _execute
        result = yield result
      File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
        value = future.result()
      File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
[...]

PR #3304, opened 2018-02-05
Don't trash files on different device to home dir on Linux

takluyver:
This is a bit of a crude check, but I imagine that most users will be working either in their home directory, or on the same device, so it's not worth trying to do a more complex check. I'm going to ask for better exceptions from send2trash so we can distinguish these cases from others.

Closes gh-3249

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -491,13 +491,28 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         if not os.path.exists(os_path):
             raise web.HTTPError(404, u'File or directory does not exist: %s' % os_path)
 
+        def _check_trash(os_path):
+            if sys.platform in {'win32', 'darwin'}:
+                return True
+
+            # It's a bit more nuanced than this, but until we can better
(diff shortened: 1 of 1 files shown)


takluyver:
I've asked on send2trash about having a more specific exception type that we could catch: https://github.com/hsoft/send2trash/issues/20

notebook/services/contents/filemanager.py
```diff
@@ -491,13 +491,28 @@ def delete_file(self, path):
...
+        def _check_trash(os_path):
+            if sys.platform in {'win32', 'darwin'}:
+                return True
+
+            # It's a bit more nuanced than this, but until we can better
+            # distinguish errors from send2trash, assume that we can only trash
+            # files on the same partition as the home directory.
+            file_dev = os.stat(os_path).st_dev
+            home_dev = os.stat(os.path.expanduser('~')).st_dev
+            return file_dev == home_dev
```
michaelkarlcoleman: Unfortunately, this test doesn't do what you'd hope on GPFS (under Linux).  If the directories are part of different filesets, you'll still get an `EXDEV` if you try to rename a file/dir from one to the other.

Issue #1097, opened 2016-02-16
Keyboard shortcut to "Close and Halt"


AGS-Knight:
I'm trying to create a keyboard shortcut to invoke the  `Close and Halt` dropdown button in an open notebook.  I've tried using the following without success:

```
%%javascript

Jupyter.keyboard_manager.command_shortcuts.add_shortcut('ctrl-q', {
    help : 'Halt and Close',
    help_index : 'zz',
    handler : function (event) {
        IPython.notebook.kill_and_exit();
        return false;
    }}
);
```

If I use another method, say `clear_all_output` in place of `kill_and_exit`, the shortcut works as expected.  Am I using the wrong method here?

Carreau:
IIRC this is a browser limitation, you cannot trigger a window closing too long after an user action, which often need to be a click, or someone else. The definition of "Too long" is variable, and I would expect we do too many indirection and the request to close is cancelled. 

We can try to investigate.

AGS-Knight:
Thanks @Carreau , this might not be a crucial feature, but would help workflow, especially without the ability to preview notebooks before opening them.

jaypeedevlin:
+1 on this, it would assist workflow for when you're making changes to lots of notebooks in sequence.

tnlin:
+1, if no one doing this, I can do it

Carreau:
5.0 (now in rc2) have the ability to create custom shortcuts from the help menu. So there is no need for patch, for the shortcut, though it is not exposed as an action. It should not be too hard to do .

takluyver:
I think the active part is to expose the function behind the 'close and halt' menu entry:

https://github.com/jupyter/notebook/blob/bdd62b3c58051cd9c12cd77f27c289b711ed0b29/notebook/static/notebook/js/menubar.js#L230-L240

As an [action](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/actions.js), so that users can easily assign it a keyboard shortcut in the UI.

takluyver:
Fixed by #3314 - it doesn't add a shortcut by default, but it makes it an action so you can easily configure a shortcut in the notebook interface.

PR #3314, opened 2018-02-07
#1097 Add close and halt to shortcut menu

Shels1909:
Hi this is my first pull request for the jupyter notebook project! This is a fix for issue #1097. I appreciate any feedback.

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -232,15 +232,7 @@ define([
         });
 
         this.element.find('#kill_and_exit').click(function () {
-            var close_window = function () {
-                /**
-                 * allow closing of new tabs in Chromium, impossible in FF
-                 */
-                window.open('', '_self', '');
-                window.close();
-            };
-            // finish with close on success or failure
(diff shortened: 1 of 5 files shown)


notebook/static/notebook/js/menubar.js
```diff
@@ -232,15 +232,7 @@ define([
...
-            var close_window = function () {
-                /**
-                 * allow closing of new tabs in Chromium, impossible in FF
-                 */
-                window.open('', '_self', '');
-                window.close();
-            };
-            // finish with close on success or failure
-            that.notebook.session.delete(close_window, close_window);
+            that.notebook.close_and_halt();
```
takluyver: Now that there's an action for this, it should be possible to remove the handler function, and connect it via `id_actions_dict` below.

Shels1909: Ok Thanks! Would it be helpful to change the id from 'kill_and_exit' to 'close_and_halt' in the html for consistency? or vice versa?

takluyver:
There are some tests failing because they look for the `kill_and_exit` ID - could you update the test so it looks for the new ID?

PR #3316, opened 2018-02-08
add missing import _

ehengao:


ehengao:
Hope this can merge to the latest version cos code breaked not doing this.

takluyver:
This isn't actually broken - the `_` gets installed to builtins, so it's available everywhere without needing to import it. But I now think we're better off putting the imports in than forever explaining how it's meant to work.

PR #3326, opened 2018-02-11
Fix a missing import of _

SimonBiggs:
Fixes https://github.com/jupyter/notebook/issues/3324

PR #3330, opened 2018-02-12
Don't install gettext into builtins

takluyver:
With the merge of #3326, I think this is no longer needed. Having `_` in builtins seems convenient, but it turns out that most of the code using it is in a few files, and breaking the normal namespacing rules has caused various problems and confusions.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -51,7 +51,7 @@ help:
 
 clean:
 	rm -rf $(BUILDDIR)/*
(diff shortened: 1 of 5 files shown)


takluyver:
This also unpins nbsphinx for the docs build. Completely unrelated, but it's needed to make the tests pass, as sphinx 1.7 just got released.

takluyver:
Merging so that the tests pass again.

PR #3339, opened 2018-02-14
Move login_widget to a separate file.

Siecje:
I want to remove the Logout button.

Now I can replace `login_widget.html` with an empty file.

Related: https://github.com/jupyterhub/jupyterhub/pull/1663


minrk:
Same as that PR, you can override the login_widget block without needing to put it in a new file:

```python
{% extends "templates/page.html" %}
# empty block to remove login_widget:
{% block login_widget %}
{% endblock %}
```

PR #3341, opened 2018-02-15
Use CSP header to treat served files as belonging to a separate origin

takluyver:
We already contain these files using the `/view/` handler, which displays them inside an iframe with the same sandboxing. This PR ensures that if a user ends up viewing an untrusted HTML/SVG file at a `/files/` URL (instead of `/view/`), it will still be sandboxed, using the [Content-Security Policy: sandbox](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox) header.

The browser then treats the page as having a unique origin, so cross-origin protection prevents any Javascript from communicating with the notebook server.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -601,6 +601,13 @@ class Template404(IPythonHandler):
 class AuthenticatedFileHandler(IPythonHandler, web.StaticFileHandler):
     """static files should only be accessible when logged in"""
 
+    @property
+    def content_security_policy(self):
+        # In case we're serving HTML/SVG, confine any Javascript to a unique
+        # origin so it can't interact with the notebook server.
+        return super(AuthenticatedFileHandler, self).content_security_policy + \
+                "; sandbox allow-scripts"
+
     @web.authenticated
     def get(self, path):
         if os.path.splitext(path)[1] == '.ipynb' or self.get_argument("download", False):
@@ -629,10 +636,6 @@ class AuthenticatedFileHandler(IPythonHandler, web.StaticFileHandler):
         # disable browser caching, rely on 304 replies for savings
         if "v" not in self.request.arguments:
             self.add_header("Cache-Control", "no-cache")
-
-        # In case we're serving HTML/SVG, confine any Javascript to a unique
-        # origin so it can't interact with the notebook server.
-        self.set_header('Content-Security-Policy', 'sandbox allow-scripts')
     
     def compute_etag(self):
         return None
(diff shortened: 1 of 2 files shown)


notebook/base/handlers.py
```diff
@@ -629,6 +629,10 @@ def set_headers(self):
         # disable browser caching, rely on 304 replies for savings
         if "v" not in self.request.arguments:
             self.add_header("Cache-Control", "no-cache")
+
+        # In case we're serving HTML/SVG, confine any Javascript to a unique
+        # origin so it can't interact with the notebook server.
+        self.set_header('Content-Security-Policy', 'sandbox allow-scripts')
```
rgbkrk: This also overrides any currently set `Content-Security-Policy` -- should we merge with the operator's settings?

Note: I refer to operator to mean whoever administers this deployment, whether a local server user or a jupyterhub administrator.

takluyver: Yes, good point. Do you think it's sufficient to use `add_header()` to do this, or should we get into checking any existing `Content-Security-Policy` headers?

minrk: Can multiple CSP headers be set and do what we want? If not, we might want to override the inherited [content_security_policy property](https://github.com/jupyter/notebook/blob/master/notebook/base/handlers.py#L66) instead.

takluyver: [From MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy):

> You can use the Content-Security-Policy header more than once... Adding additional policies can only further restrict the capabilities of the protected resource

I think that means this should work, but maybe the c_s_p property is a neater way to achieve the same thing?

rgbkrk: Interesting, I didn't realize we could add multiple headers. How does `set_header` work when there are multiple headers with the same key from `add_header`?

takluyver: I think `set_header` blows away anything that was set before.

rgbkrk: My question comes down to "what's the set of headers after running the following"

```python
self.add_header("X", "y")
self.add_header("X", "z")
self.set_header("X", "x")
```

takluyver: AIUI, only `X: x` would remain.

rgbkrk: I'm thinking using only one header is good. For now I think the value you've set here is a very sane default and further enhancements on merging content security policies can be separate work.

takluyver:
OK, I've redone this to achieve the same thing by overriding the `content_security_policy` property, so it shouldn't interfere with other policies we've set.

rgbkrk:
I just re-kicked the appveyor build and somehow it went faster than I expected. 
¯\\\_(ツ)\_/¯

At any rate, I certainly approve again. 😄

PR #3356, opened 2018-02-21
Display hostname when running remotely

evandam:
My team often runs Jupyter notebooks on a remote server, but 'localhost' is always displayed when running with an option like `--ip '*'` or `--ip 0.0.0.0`, meaning the link isn't copy-and-pasteable when trying to connect from another machine.

This change checks if you are running locally (either the current hostname, `localhost`, or `127.0.0.1`) and displays `localhost` as expected, but otherwise provides the server name.

Hopefully this is the right way to go about proposing a change!

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1323,7 +1323,6 @@ class NotebookApp(JupyterApp):
         else:
             ip = hostname
         url = self._url(ip)
-        self.log.info(ip, url)
         if self.token:
             # Don't log full token if it came from config
             token = self.token if self._token_generated else '...'
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1317,8 +1317,13 @@ class NotebookApp(JupyterApp):
     
     @property
     def display_url(self):
-        ip = self.ip if self.ip else _('[all ip addresses on your system]')
+        hostname = socket.gethostname()
+        if self.ip in ('localhost', '127.0.0.1', hostname):
+            ip = self.ip
+        else:
+            ip = hostname
         url = self._url(ip)
+        self.log.info(ip, url)
         if self.token:
             # Don't log full token if it came from config
             token = self.token if self._token_generated else '...'
@@ -1625,7 +1630,7 @@ class NotebookApp(JupyterApp):
                 '\n',
                 'Copy/paste this URL into your browser when you connect for the first time,',
                 'to login with a token:',
-                '    %s' % url_concat(self.connection_url, {'token': self.token}),
+                '    %s' % url_concat(self.display_url, {'token': self.token}),
             ]))
 
         self.io_loop = ioloop.IOLoop.current()


minrk:
This is a solid improvement, thanks!

cyberwillis:
Any chance to make this optional in command line ?

*like this: *
```
--no-translate-hostname
```

evandam:
I'm happy to give this a shot if it makes sense to do so. Maybe even something like `--display-name localhost` where you can explicitly set how the URL will be formed?

However, this is really just a convenience change. It really shouldn't cause any functional difference, unless I'm not understanding how your configuration works?

cyberwillis:
Hi...
Sure I can try to explain...

When I build a container for LXC it reserves a IP for the container and also register the specific IP  number into a alias inside the container so I can call it easly. 

**The alias has this pattern:**
``` 
export JUPYTER_IP=$(ip route get 8.8.8.8 | awk '{print $NF; exit}')
alias jlab='jupyter lab --port=$JUPYTER_PORT --ip=$JUPYTER_IP --no-browser --NotebookApp.password='sha1:$JUPYTER_SECRET' --NotebookApp.notebook_dir=/projects --allow-root'
```

Until notebook==5.4.1, the access bu IP was ok. 
Just to mention tough, in case I had omitted  the "--ip" on alias, the nootbook would show up with an address like that  **http://locahost:8888** or **http://0.0.0.0:8888**, that could not access at all because "localhost" cannot be binded between host and container in LXC. In docker it can, or at least does not exist a easy way to do so. There are many threads about that on the internet.

When notebook==5.5.0rc1 showed up, translating ip into domain names / machine names, every time notebook become online i get this address **http://container:8888/** that I cannot access by name unless I start to managing it into my /etc/hosts or remembering the original IP number from before.

What do you propose ?

cyberwillis:
I just found that the new version of LXD 3.0 released abril 2, supports the binding by creating a proxy for that. I will setup to see.

cyberwillis:
Hi again,
I could confirm it. 

Now I can bind to my HOST localhost on LXd 3.0 so independent of been Hostname or IP I can use localhost now.

takluyver:
So the issue is/was that a container had a different host name internally and externally, so it was showing a name that wasn't useful externally? Maybe instead of special casing `localhost` and `127.0.0.1` in one direction, it would be more robust to special case `*` and `0.0.0.0` in the other direction? I.e. look up the hostname in those cases and use `self.ip` otherwise.

@cyberwillis I'm a bit confused by your last message; is this still a problem for you, or is it OK for you with the new version of LXD? Thanks for testing the RC!

cyberwillis:
Hi, again

It becames a problem when the new rc1 started to translate from IP to machine name, because I would have to manage of DNS internal on my host to translate it back.

Anyway my problem was solved. I just ripped off the --ip argument and started to use the new version of my container runtime that can forward IPs and Ports for my localhost.

But what I was tryied to explain earllier is that, is preferable to inform an IP like 192.168.1.30 and be accessible by it instead of translating into names. Thats whay  I sugested a parameter that let me choice if I need or not translation of the IP into machine name.

takluyver:
Right, and we can add an option for it if we have to, but that's only helpful for people who've figured out what's gone wrong and discovered the option. If this is going to be a problem for a lot of people, I'd like to understand what's going on now and work out if we can make it work for more people without requiring them to set an option.

PR #3360, opened 2018-02-21
Do not execute special notebooks with nbsphinx

hroncok:
They can hang in headless environment.

See also db90e16d17acc778efdfff6dca69767b47b60cbd


takluyver:
Makes sense,  thanks.

PR #3367, opened 2018-02-23
fix error in i18n/README.md

forbxy:
my computer working on Ubuntu，so I just test it on Ubuntu，I can not confirm if it is necessary to set environment variable LANGUAGE on other linux system

### DIFF
--- a/notebook/i18n/README.md
+++ b/notebook/i18n/README.md
@@ -15,6 +15,7 @@ However…
 run in).
 
 Hint: if running Windows, you can set it in PowerShell with `${Env:LANG} = "xx_XX"`.
+      if running Ubuntu, you should set environment variable `LANGUAGE="xx_XX"`.
 
 2. The preferred language for web pages in your browser settings (`xx`) is
    also used. At the moment, it has to be first in the list.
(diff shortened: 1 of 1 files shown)


notebook/i18n/README.md
```diff
@@ -15,6 +15,7 @@ However…
 run in).
 
 Hint: if running Windows, you can set it in PowerShell with `${Env:LANG} = "xx_XX"`.
+      if running Ubuntu, you should set environment variable `LANGUAGE="xx_XX"`.
```
takluyver: At least on my system, there's a variable called `LANG` rather than `LANGUAGE`. I think both are used, but I'm not sure which one has priority.

forbxy: on my system(ubuntu14.04)，LANG don't work

willingc: @forbxy If you are ssh'ing into Ubuntu 14.04, you will need to make sure that the following are set:
```
export LANG="en_US.utf8"
export LANGUAGE="en_US.utf8"
export LC_ALL="en_US.utf8"
```

forbxy:
finally，I test it on ubuntu 16,var LANGUAGE is  needless

PR #3368, opened 2018-02-23
API spec improvements, API handler improvements

mpacer:
This changes the base-path to be / instead of /api so that in the future
other APIs that return files and do not redirect you to a new address
(e.g., see `NbconvertConfigHandler` in #2413).

This also improves the way that the api spec handler responds to a
request, by actually displaying the spec with the appropriate
content-type header (text/x-yaml) instead of triggering a download with
the StaticFileHandler's default application/octet-stream mimetype.


minrk:
Nice!

PR #3372, opened 2018-02-25
Use latest codemirror

vaibhavsagar:


takluyver:
Thanks! :cookie:

vaibhavsagar:
My pleasure!

PR #3373, opened 2018-02-25
fix:when upload large file,we cant stop

forbxy:
the cancel button' action  only remove its html item, don't stop upload event before

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -1220,12 +1220,14 @@ define([
     NotebookList.prototype.add_large_file_upload_button = function (file) {
         var that = this;
         var item = that.new_item(0, true);
+        var stop_singnal = false;
         item.addClass('new-file');
         that.add_name_input(file.name, item, 'file');
         var cancel_button = $('<button/>').text("Cancel")
             .addClass("btn btn-default btn-xs")
             .click(function (e) {
                 item.remove();
+                stop_singnal = true;
                 return false;
             });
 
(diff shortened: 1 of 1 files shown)


notebook/static/tree/js/notebooklist.js
```diff
@@ -1220,12 +1220,14 @@ define([
     NotebookList.prototype.add_large_file_upload_button = function (file) {
         var that = this;
         var item = that.new_item(0, true);
+        var stop_singnal = false;
```
takluyver: `singnal` -> `signal`

notebook/static/tree/js/notebooklist.js
```diff
@@ -1276,6 +1278,9 @@ define([
                         var upload_file = null;
                         
                         var large_reader_onload = function (event) {
+                            if (stop_singnal === true) {
+                                return;
```
takluyver: Should this also abort the file reader? https://developer.mozilla.org/en-US/docs/Web/API/FileReader/abort

forbxy: file reader is a local variable，one chunk one file reader var，we stopped before new file reader create (after last chunk uploaded)，

notebook/static/tree/js/notebooklist.js
takluyver: Ping @forbxy - can you make this spelling change?

notebook/static/tree/js/notebooklist.js
forbxy: should I make a commit to fix the word(singnal)?or it is already  changed after you review?

takluyver: You'll need to make a commit to fix it - my comment was just pointing it out.

takluyver:
Thanks!

PR #3377, opened 2018-02-26
Add explanation on how to change the type of a cell to Markdown.

xuhdev:
I was quite confused when I first read the documentation on this page: I knew what Markdown is, but couldn't figure out how a cell type can be Markdown for sometime. This PR should make it more clarified.


minrk:
Thanks!

PR #3379, opened 2018-02-27
Use powershell on Windows

dhirschfeld:
sh can be installed and on the path if e.g. git and the unix tools are installed
Even in this case the user would probably prefer to use powershell

Addresses: https://github.com/jupyterlab/jupyterlab/issues/4034


minrk:
Thanks!

Issue #3365, opened 2018-02-22
Automatically getting logged out in Firefox


SteliosStefanatos:
Hi, 

My issue is as follows. I work with Jupyter notebook in Firefox (updated Anaconda installation; Ubuntu). Since about two weeks, I keep getting logged out automatically (I don't know if I use the correct term here). What happens is that the automatic save feature will fail (the "Forbidden" sign pops up on the notebook upper right corner) and I cannot save "manually" as well. When I attempt to go the dashboard I get prompted to provide the session token. 

The weird thing is that this issue happens randomly during a session. It may appear after hours from starting Jupyter/firefox or even right from the start. Also, when I do provide the token, it often happens that I keep getting logged out right the next second and have to do several attempts till I can have it working. 

This issue does not happen when I work on a different browser (chromium).

P.S.: I see similar reports for Firefox [here](https://github.com/jupyter/notebook/issues/2396)

takluyver:
That's interesting. I've seen something similar running Selenium tests with Firefox, but I haven't seen it in actual use yet. With the tests, it only seemed to affect Firefox 58,  so I told it to install v57 as a workaround. Are you using 58 as well?

Authentication uses cookies, so it could be that something is deleting our authentication cookie. Next time you see this happen, can you open the developer tools, go to the storage tab and check what cookies you see?

BobbyTables42:
I am also using Firefox, v58.0.2 on Fedora 27. When doing as you suggest, when I first open a jupyter notebook from the terminal, I see 2 cookies. The first is named, _xsrf, domain of localhost, expires on session, a value string, httponly of false; the second has a name of username-localhost-8888, domain of localhost, expires in a month, a different value string, and httponly set to true. When I then open any link from the tree after I login (I use a password), the second cookie always disappears. Ideas?

takluyver:
`username-localhost-8888` is our login cookie, so if that disappears, it makes sense that Jupyter would say you're unauthenticated.

We did make some changes to a method that's used to clear the login cookie (#3207). It sounds like you can reproduce the error readily, so could you try commenting out these lines in your local installation, starting a new notebook server and seeing if that makes any difference?

takluyver:
Sorry, forgot the link. These lines:

https://github.com/jupyter/notebook/blob/2aac713937c0487dd2193c85a69f02e9556421fe/notebook/base/handlers.py#L126-L131

takluyver:
[...]

Issue #2396, opened 2017-04-11
How to solve “forbidden” error in jupyter notebook (conda tab)


CorneelDragon:
I installed anaconda and succeeded in making the conda-tab work thanks to the last suggestion in this entry: http://stackoverflow.com/a/41825208/6375668

However, when I try to create an environment or install an additional package I receive a "forbidden" error (see below). When I open anaconda-navigator I am perfectly able to create environments and install packages. In addition, I changed permissions to the anaconda-folders on my Mac and tried starting jupyter with sudo, but that didn't help.

What could be wrong? And how to fix this? Thanks!

!["forbidden error"](http://i.imgur.com/1yo2kld.jpg)

gnestor:
Are you seeing any notebook server errors in the terminal? What version of the notebook and nb-conda are you running?

Here's a similar issue on nb-conda: https://github.com/Anaconda-Platform/nb_conda/issues/44

scaramouche-00:
Since updating my packages in anaconda (conda update --all) and installing doxygen, I have the same issue. Jupyter will throw the error
403 GET

faizankshaikh:
Got the same error. Solved it by opening jupyter in another browser (was using firefox, then used chrome). I think its a "cache" related issue. Maybe someone can explore this futher?

andreagrandi:
@faizankshaikh hi, I came to this page because I was having problems saving a notbook while using Firefox. If I use Chrome I don't have any problem, while if I use Firefox I keep getting the "Forbidden" error when I try to save a notebook. I've also tried to disable any extension but it didn't help.

takluyver:
#3365 may be related.

takluyver:
I've fixed one such issue in #3380, if people want to try that. But there may be more than one problem involved; the precise case I hit could only have come up after PR #2959, which was well after the initial report.

scaramouche-00:
Hello, thanks for this! How can we access the patch? I have anaconda and I have tried updating jupyter (conda update jupyter), but it didn't work (still 403 GET).

takluyver:
It will need to be merged and released before you can get it through Anaconda. For a small change like this, it's probably easiest to [look at what I've changed](https://github.com/jupyter/notebook/pull/3380/files), find the matching lines in your installation, and apply the change manually.

Otherwise, you could use git to get it:

```
git clone https://github.com/takluyver/notebook.git
cd notebook
git checkout i3365
pip install .
```

[...]

PR #3380, opened 2018-02-27
Don't clear login cookie on requests without cookie

takluyver:
Clearing the secure cookie makes sense when it is invalid or expired, but clearing it when it's not sent with a request can unexpectedly log the user out.

I ran into this with misconstructed URLs which didn't incorporate base_url. That was itself a bug, but the result of suddenly being unauthenticated makes it harder to track down.

Closes gh-3365 (hopefully)
Possibly also relevant to gh-2396

I'm not sure why this seems to affect Firefox more than Chrome.

### DIFF
--- a/notebook/auth/login.py
+++ b/notebook/auth/login.py
@@ -175,8 +175,12 @@ class LoginHandler(IPythonHandler):
             # Used in is_token_authenticated above.
             handler._token_authenticated = True
         if user_id is None:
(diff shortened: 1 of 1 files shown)


takluyver:
One user on #3365 has confirmed that this fixed their issue.

takluyver:
cc @minrk @gnestor

gnestor:
LGTM!

Issue #2707, opened 2017-07-29
Trusted button should have a description


fulldecent:
On the top-right there is a button labeled "Trusted".  I don't know what it means. If you click it then it goes away or it doesn't, I forget.

Mousing over this button currently does not provide a description of what this means. All the other buttons do provide a description (`title` attribute) when you mouse over.

Please add a description to this button.

![screen shot 2017-07-29 at 7 06 23 pm](https://user-images.githubusercontent.com/382183/28749325-149e6c56-7491-11e7-8335-04fb8b133f7d.png)

P.S. The "edit mode" button does have a description. But it is useless, I did not learn what it means or why I should click it by reading the description.

takluyver:
Those bits on the right are indicators, rather than buttons - they're not usually meant to be clicked.

'Trusted' means that the notebook has been created or run on your computer, and is trusted to display rich output containing Javascript. If you download a notebook and open it, it will initially be *untrusted*, so that it can't start doing things on your computer until you deliberately run code cells. I'm not sure how to explain that in a tooltip, though. ;-)

fulldecent:
Ah thank you! Recommendation:

> Trusted -- Javascript enabled for notebook display

> Untrusted -- Javascript disabled for notebook display

jzf2101:
Would this be sprint-friendly?

delallea:
I agree it's confusing, because clicking on it makes it disappear, as if it had an action. It should be just a non-interactive label, instead of a clickable button.

ashleytqy:
A non-interactive label would definitely be clearer to me too as the Button UI right now is unintuitive.

If possible, can I work on this issue? 😊

jzf2101:
I don't see why not?

dimaqq:
If it's an indicator, how come it goes away when clicked?

PR #3386, opened 2018-03-01
Adding description for 'Trusted' widgets

ashleytqy:
This PR fixes https://github.com/jupyter/notebook/issues/2707

- Trusted and Untrusted widgets have `title` attributes which describe what they mean
- Trusted widgets now have their clicks disabled

screen shot 2018-02-28 at 10 13 38 pm


minrk:
Nice!

Issue #1883, opened 2016-11-10
Can the look of jupyter notebook be customized in any way according to kernel status?


BoPeng:
I have developed a [Script of Scripts](https://github.com/BoPeng/SOS) kernel that allows switching between different kernels using magics such as `%use R`, and `%with R`. It is working reasonably well but there is a small problem with the current implementation, namely there is no hint that which kernel is currently active. Is it possible to, for example, change the background of notebook according to current kernel? 

In addition, it would be useful if I can mark each cell with the kernel used so that I can get a notebook with cells in different colors. For example

```
[white SoS cell]
    import pandas as pd
    data = pd.read_excel('data.xls')
[purple R cell]
    %with R -i data -o m
    m <- apply(data, 1, median)
[white SoS cell]
    print(m)
```

I have a feeling that this could be done in javascript, but I currently have no idea how the browser (javascript) interacts with the kernel. Any suggestion or code example would be greatly appreciated.

BoPeng:
I have tried but could not find any answer on how the jupyter client and server (kernel) work.

1. How to manipulate cell `metadata`? I can see that some nbextensions can set metadata and the metadata could be saved in `.ipynb`, but I am not sure how the kernel can access cell meta information.
2. How does `user_expressions` work? I mean, how the frontend determines which `user_expressions` to evaluate? The `user_expressions` looks to me the only way for the front and backend to exchange meta information, but I am not sure how to control the frontend.

In the end, what I would like to do, in the context of [Script of Scripts](http://vatlab.github.io/SOS) is to **store and use kernel information in each cell**. More specifically,

1. The frontend reads `metadata` of a cell, and send the kernel information to `do_execute` (via `user_expression`). If this is possible, the frontend can also use different color to display cells created by different kernels.
2. The kernel decides which subkernel to use to evaluate the `code` (current active kernel for new cells and creating kernel for existing cells), and send result, as well as `kernel` information to frontend (through `user_expression`)?
3. The frontend saves kernel info in `metadata` of a cell. The information will be used when the cell is executed again.

I would appreciate it if anyone can let me know if these are at all possible and point me to relevant documentation or source code. Thanks.

takluyver:
> How to manipulate cell metadata? I can see that some nbextensions can set metadata and the metadata could be saved in .ipynb, but I am not sure how the kernel can access cell meta information.

[...]

PR #3391, opened 2018-03-04
Pass cell tags and selected cell metadata to kernel

quinot:
Include tags and kernel_metadata code cell metadata in the execute_request message sent to kernels.

This allows notebook authors to provide additional information to the kernel without making it directly user visible in the contents of the code cell. For example this can be used to provide information about a specific context in which the code cell must be evaluated.

Fixes jupyter/help#223
Also related to https://github.com/jupyter/notebook/issues/1883#issuecomment-269836430


quinot:
Notes:

- The choice of what metadata to pass or not is pretty arbitrary. Here in the initial proposal for this PR we're passing the tags and a specific metadata attribute "kernel_metadata", but we could just as well decide to pass *all* cell metadata to the kernel;
- Technically this can be achieved without changing the Notebook by [monkey patching CodeCell.execute](https://gist.github.com/quinot/e3801b09f754efb0f39ccfbf0b50eb40), but this is a bit ugly and probably more fragile than the proposed Notebook change.

minrk:
Rather than try to pick which metadata is sent, perhaps send the whole `cell.metadata` structure in `metadata.cell_metadata` on the request? And only include it if it's not empty?

quinot:
@minrk right, that's the alternative option I mentioned in the first comment above. I'll be happy to update the PR accordingly if you think that's a preferrable solution.

minrk:
I think it is, since it heads off infinite follow-up "add my metadata, too" PRs.

BoPeng:
I think a bigger question is how kernels handle the passed meta data? According to [the doc](http://jupyter-client.readthedocs.io/en/latest/wrapperkernels.html), the prototype for `do_execute` is

```
do_execute(code, silent, store_history=True, user_expressions=None, allow_stdin=False)
```
which leaves no room to accept metadata.

quinot:
@BoPeng kernels can peek at the passed metadata in the init_metadata method. See for instance
https://github.com/quinot/sparkdisc_kernel/blob/master/binder/sparkdisc_kernel/kernel.py

BoPeng:
Thanks. I never knew the existence of [such a mechanism](https://github.com/ipython/ipykernel/blob/479932d5fe94e431eae22f3b6c6e2bc17940ad84/ipykernel/kernelbase.py#L390). The patch could solve most, but not all of my needs though, because SoS has some special "global" features (e.g. a magic to compose a workflow from multiple cells in the notebook) so I will need some "callback" mechanism to prepare metadata in the frontend. Right now I am using the same monkey patching method as you [described](https://gist.github.com/quinot/e3801b09f754efb0f39ccfbf0b50eb40).

PR #3399, opened 2018-03-05
add missing digestmod arg to HMAC

minrk:
Noticed in test output that creating HMAC without digestmod arg is deprecated.

While there, use proper length of 32 bytes for default tornado cookie_secret. There’s no benefit to using a cookie secret that's longer that the cookie digest size, which is 32 bytes.


minrk:
Based on #3398 since tests won't pass until that lands

PR #3401, opened 2018-03-06
add settings['activity_sources']

minrk:
for extensions to register activity that should prevent shutdown due to inactivity

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -265,7 +265,8 @@ class NotebookWebApplication(web.Application):
             # Jupyter stuff
             started=now,
             # place for extensions to register activity
-            activity_sources={},
+            # so that they can prevent idle-shutdown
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -264,6 +264,8 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
 
             # Jupyter stuff
             started=now,
+            # place for extensions to register activity
+            activity_sources={},
```
takluyver: 'sources' suggests to me that it contains something like functions you call to get activity timestamps. Maybe call it something like `last_activity_times`?

takluyver: Ping @minrk . I can do the name change if you're OK with it.

minrk: Sorry, just got back to it

PR #3409, opened 2018-03-08
Javascript lint in notebooklist.js

takluyver:
Like #3294, this cleans up some simple Javascript mistakes, so that we stand a better chance of noticing when the linter sees things that might actually matter.


takluyver:
I'll merge this tomorrow unless anyone wants more time to look.

PR #3411, opened 2018-03-09
Fix going back to root directory with history in notebook list

takluyver:
Bug found during pair programming with @mpacer


takluyver:
To elaborate a bit, the bug manifests when you click into a subdirectory from the tree view, and then use the back button to return to the first view that was loaded. The history entry for this has an empty state object, but that's still truthy in Javascript, so it was taking `.path` from it, which was undefined.

While fixing this, we realised that the fallback was also to the server root directory instead of the first directory loaded - they're normally the same, but don't have to be. So that is also fixed.

takluyver:
I'll merge this tomorrow unless anyone wants more time to review it.

PR #3412, opened 2018-03-09
Update selenium tests

mpacer:
This PR comes from the same pair programming session with @takluyver that #3411 came from.

This also depends upon #3411.

### DIFF
--- a/notebook/tests/selenium/test_dashboard_nav.py
+++ b/notebook/tests/selenium/test_dashboard_nav.py
@@ -3,13 +3,13 @@ import os
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.support import expected_conditions as EC
-
 pjoin = os.path.join
 
 def get_list_items(browser):
     return [{
         'link': a.get_attribute('href'),
         'label': a.find_element_by_class_name('item_name').text,
+        'element': a,
     } for a in browser.find_elements_by_class_name('item_link')]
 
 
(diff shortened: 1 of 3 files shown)


notebook/tests/selenium/test_dashboard_nav.py
```diff
@@ -3,43 +3,42 @@
...
+            assert authenticated_browser.current_url == url
+        else:
+            break
+    # Going back up the tree while we still have unvisited links
+    while visited_dict:
+        wait_for_selector(authenticated_browser, '.item_link')
+        current_items = get_list_items(authenticated_browser)
+        current_items_links = [item["link"] for item in current_items]
+        stored_items = visited_dict.pop(authenticated_browser.current_url)
+        stored_items_links = [item["link"] for item in stored_items]
```
takluyver: Neat, I like the comparison of what's here now vs what was here before.

notebook/tests/selenium/test_dashboard_nav.py
```diff
@@ -3,43 +3,42 @@
...
+def test_items(authenticated_browser):
+    tree_root_url = authenticated_browser.current_url
+    visited_dict = {}
+    # Going down the tree to collect links
+    while True:
+        wait_for_selector(authenticated_browser, '.item_link')
+        items = get_list_items(authenticated_browser)
+        visited_dict[authenticated_browser.current_url] = items
+        print(authenticated_browser.current_url, len(items))
+        if len(items)>1:
```
takluyver: Why does it not work with 1 item? Does it catch the link to go up a directory?

I think it would be good to filter out which items are directories, so that this doesn't accidentally load a notebook or another file.

mpacer: For the first, I was using that to filter out `..` but I see now that that actually introduces weird behaviour at the root.

For the second, I gather that I can distinguish that based on the link target being on the `/tree` vs. `/notebook` or `/file` endpoints?

takluyver: Yep, that's right.

mpacer: Oops! i forgot about this simple fix and have been focused on getting the notebook navigation stuff working on the markdown stuff. I'll fix this quickly

notebook/tests/selenium/test_dashboard_nav.py
```diff
@@ -13,32 +13,43 @@ def get_list_items(browser):
...
 
 
+def only_dir_links(browser):
+    try:
+        assert 'tree' in browser.current_url
+    except AssertionError:
+        raise("You currently ")
+    wait_for_selector(browser, '.item_link')
+    items = get_list_items(browser)
+    return [i for i in items if 'tree' in i['link'] and i['label'] != '..']
```
takluyver: This will go wrong for any file that has `tree` in the name. Maybe that's unlikely, but I think it's worth doing it a bit more thoroughly.

notebook/tests/selenium/test_dashboard_nav.py
```diff
@@ -3,43 +3,70 @@
...
+        
+
 def get_list_items(browser):
+    """Gets list items from a directory listing page
+    
+    Raises PageError if not in directory listing page (url has tree in it)
+    """
+    try:
+        assert 'tree' in browser.current_url
+    except PageError:
```
takluyver: This is a convoluted way of writing `if 'tree' not in browser.current_url:` , even if it was catching the right error. Let's stick to the simple if statement rather than using assert to achieve the same thing.

mpacer: fair enough — changed.

notebook/tests/selenium/test_dashboard_nav.py
mpacer: ok addressed by using an explicit check for whether urls point to something in the tree (with  `url_in_tree`)

takluyver:
:tada:

PR #3424, opened 2018-03-14
Fix leak of iopub object in activity monitoring

kevin-bates:
After analyzing various leaked items when running either Notebook or
Jupyter Kernel Gateway, one item that recurred across each kernel
startup and shutdown sequence was the a zmq socket instance associated
with the iopub port.  The leaked instance occurs because the activity
monitor and normal port creation logic both call `create_iopub`.

Although this ends up creating a _wrapper_ around the same port, the
object (i.e., wrapper) created by the activity monitor is leaked.  By
setting the kernel manager's `_activity_stream` member to `None` when
the kernel is shutdown, cleanup of the iopub wrapper object will take
place.

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -261,6 +261,7 @@ class MappingKernelManager(MultiKernelManager):
         self._check_kernel_id(kernel_id)
         kernel = self._kernels[kernel_id]
         kernel._activity_stream.close()
+        kernel._activity_stream = None
         self.stop_buffering(kernel_id)
         self._kernel_connections.pop(kernel_id, None)
         self.last_kernel_activity = utcnow()


takluyver:
I don't immediately see why this would result in a leak; the kernel object itself is deleted shortly afterwards (the `super(...).shutdown_kernel()` call calls `.remove_kernel()`, which removes it from the `{id: kernel}` dictionary. The kernel object (which is a KernelManager instance) should be garbage collected, and its `_activity_stream` should be cleaned up.

What am I missing?

kevin-bates:
@takluyver - thanks for the response.  I agree that its quite non-obvious.  However, w/o this statement, the iopub object *wrapper* created [here](https://github.com/jupyter/notebook/blob/master/notebook/services/kernels/kernelmanager.py#L364) will not be garbage collected - with a new (leaked) instance persisted across each subsequent kernel cycle.

I found this and the two PRs ([360](https://github.com/jupyter/jupyter_client/pull/360), [361](https://github.com/jupyter/jupyter_client/pull/361)) submitted to jupyter_client via use of `gc.collect`, `gc.garbage` and `gc.get_referrers` calls in the `start_kernel()` and `shutdown_kernel()` methods of jupyter_client, notebook, kernel gateway, and enterprise gateway along with `resource.getrusage(resource.RUSAGE_SELF).ru_maxrss` to monitor growth.

However, since this leak was found prior to the resolution to the IOLoopKernelManager leak (via PR 361) then I suppose the change may not be required - although, in my opinion, is good practice.  I didn't go and back out this change after breaking the circular references that were preventing the leak in 361.

kevin-bates:
Hmm - I just commented out this change (and running with all other fixes in place) and see the iopub object **and** the kernel manager instance leaks return.  I don't see an obvious circular reference introduced by `_activity_stream`, but given the km instance returns indicates this change is required.

kevin-bates:
@takluyver - is there anything else you need for this?

takluyver:
No, I don't think so. I'll close and reopen it to rerun the tests.

kevin-bates:
Thank you Thomas!

PR #3430, opened 2018-03-15
Trying to fix Appveyor build errors

takluyver:


takluyver:
I thought it was going to be trickier than that. Merging to get the tests passing again; we can always refine it later.

PR #3458, opened 2018-03-22
Selenium utils + markdown rendering tests

mpacer:
This PR introduces a `Notebook` class as a helper for running selenium functions. It still needs work to document it more completely, but… it works and makes it a lot easier to create the relevant tests inside notebooks.

I'm interested to hear people's reviews.

I think @takluyver @Carreau @minrk and @rgbkrk would be particularly interested in this.

### DIFF
--- a/notebook/tests/selenium/quick_selenium.py
+++ b/notebook/tests/selenium/quick_selenium.py
@@ -1,0 +1,16 @@
+from selenium.webdriver import Firefox
+from notebook.notebookapp import list_running_servers
+
+
+def quick_driver():
+    """Quickly create a selenium driver pointing at an active noteboook server.
+
+    """
+    try:
+        server = list(list_running_servers())[0]
+    except IndexError as e:
+        e.message = 'You need a server running before you can run this command'
+    driver = Firefox()
+    auth_url = f'{server["url"]}?token={server["token"]}'
+    driver.get(auth_url)
+    return driver
--- a/notebook/tests/selenium/quick_selenium.py
+++ b/notebook/tests/selenium/quick_selenium.py
@@ -17,6 +17,6 @@ def quick_driver():
     except IndexError as e:
         e.message = 'You need a server running before you can run this command'
     driver = Firefox()
-    auth_url = f'{server["url"]}?token={server["token"]}'
+    auth_url = '{url}?token={token}'.format(**server)
     driver.get(auth_url)
     return driver
--- a/notebook/tests/selenium/quick_selenium.py
+++ b/notebook/tests/selenium/quick_selenium.py
@@ -1,3 +1,8 @@
+"""Utilities for driving Selenium interactively to develop tests.
+
+These are not used in the tests themselves - rather, the developer writing tests
+can use them to experiment with Selenium.
+"""
 from selenium.webdriver import Firefox
 
 from notebook.tests.selenium.utils import Notebook
--- a/notebook/tests/selenium/quick_selenium.py
+++ b/notebook/tests/selenium/quick_selenium.py
@@ -1,4 +1,6 @@
 from selenium.webdriver import Firefox
+
+from notebook.tests.selenium.utils import Notebook
 from notebook.notebookapp import list_running_servers
 
 
@@ -10,6 +12,7 @@ def quick_driver(lab=False):
         from notebook.tests.selenium.quick_selenium import quick_driver
         driver = quick_driver
         
+    Note: you need to manually close the driver that opens with driver.quit()
     """
     try:
         server = list(list_running_servers())[0]
@@ -25,3 +28,16 @@ def quick_driver(lab=False):
         driver.get(driver.current_url.rstrip('lab')+'tree')
     return driver
 
+
+def quick_notebook():
+    """Quickly create a new classic notebook in a selenium driver
+
+
+    Usage example:
+    
+        from notebook.tests.selenium.quick_selenium import quick_notebook
+        nb = quick_notebook()
+
+    Note: you need to manually close the driver that opens with nb.browser.quit()
+    """
+    return Notebook.new_notebook(quick_driver())
--- a/notebook/tests/selenium/quick_selenium.py
+++ b/notebook/tests/selenium/quick_selenium.py
@@ -3,6 +3,10 @@ from selenium.webdriver import Firefox
 from notebook.tests.selenium.utils import Notebook
 from notebook.notebookapp import list_running_servers
 
+class NoServerError(Exception):
+
+    def __init__(self, message):
+        self.message = message
 
 def quick_driver(lab=False):
     """Quickly create a selenium driver pointing at an active noteboook server.
@@ -17,7 +21,8 @@ def quick_driver(lab=False):
     try:
         server = list(list_running_servers())[0]
     except IndexError as e:
-        e.message = 'You need a server running before you can run this command'
+        raise NoServerError('You need a server running before you can run '
+                            'this command')
     driver = Firefox()
     auth_url = '{url}?token={token}'.format(**server)
     driver.get(auth_url)
--- a/notebook/tests/selenium/quick_selenium.py
+++ b/notebook/tests/selenium/quick_selenium.py
@@ -2,15 +2,14 @@ from selenium.webdriver import Firefox
 from notebook.notebookapp import list_running_servers
 
 
-def quick_driver():
+def quick_driver(lab=False):
     """Quickly create a selenium driver pointing at an active noteboook server.
 
-    Usage example
+    Usage example:
     
-    from inside the selenium test directory:
-    
-        import quick_selenium, test_markdown, utils
-        nb = utils.Notebook(test_markdown.notebook(quick_selenium.quick_driver()))
+        from notebook.tests.selenium.quick_selenium import quick_driver
+        driver = quick_driver
+        
     """
     try:
         server = list(list_running_servers())[0]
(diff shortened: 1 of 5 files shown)


rgbkrk:
I'm happy these at least make it easy for Python people to write tests for the JS. 😅

mpacer:
This is now rebased on #3412 to take into account the issues that arise when you create a new notebook in the first place.

notebook/tests/selenium/quick_selenium.py
```diff
@@ -0,0 +1,22 @@
...
+    
+        import quick_selenium, test_markdown, utils
+        nb = utils.Notebook(test_markdown.notebook(quick_selenium.quick_driver()))
+    """
+    try:
+        server = list(list_running_servers())[0]
+    except IndexError as e:
+        e.message = 'You need a server running before you can run this command'
+    driver = Firefox()
+    auth_url = f'{server["url"]}?token={server["token"]}'
```
takluyver: f strings are great, but we're still supporting Python 3.4 and 3.5 for now, so we'll need to stick with the explicit `.format()` method for now.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    @property
+    def body(self):
+        return self.browser.find_element_by_tag_name("body")
+
+    @property
+    def cells(self):
+        """Gets all cells once they are visible.
+        
+        """
+        wait_for_selector(self.browser, ".cell")
```
takluyver: I don't like having a wait inside a property - the rule of thumb for properties is that anything they need to do should be near-instant. If it needs to do something that's not, let's make it a regular method - the call gives people reading the code a hint that there's something going on underneath.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    new_button = wait_for_selector(browser, "#new-buttons", single=True)
+    new_button.click()
+    kernel_selector = '#{} a'.format(kernel_name)
+    kernel = wait_for_selector(browser, kernel_selector, single=True)
+    kernel.click()
+
+@contextmanager
+def new_window(browser, selector=None):
+    """Creates new window, switches you to that window, waits for selector if set.
```
takluyver: Neat! The docstring needs some work, though - this doesn't create a new window, rather it expects that code called in its context will create a new window, and then it switches to that on leaving the context.

This is worth explaining clearly, because most context managers do the interesting stuff when you create/enter them, and exiting just cleans something up. Here the interesting stuff is on exit.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,163 @@
...
+        self.add_cell(index, cell_type="markdown")
+        self.edit_cell(index=index, content=content, render=render)
+
+
+    @classmethod
+    def new_notebook(cls, browser, kernel_name='kernel-python3'):
+        # initial_window_handles = browser.window_handles
+        with new_window(browser, selector=".cell"):
+            select_kernel(browser, kernel_name=kernel_name)
+        browser.execute_script("Jupyter.notebook.set_autosave_interval(0)")
```
takluyver: Is it OK that this is in `new_notebook()` but `remove_safety_check()` is called from `__init__`? I'd expect they'd go together.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,163 @@
...
+        new_index = index + 1 if index >= 0 else index
+        self.convert_cell_type(index=new_index, cell_type=cell_type)
+
+    def add_markdown_cell(self, index=-1, content="", render=True):
+        self.add_cell(index, cell_type="markdown")
+        self.edit_cell(index=index, content=content, render=render)
+
+
+    @classmethod
+    def new_notebook(cls, browser, kernel_name='kernel-python3'):
```
takluyver: Docstring! It should mention that this expects to be called with the browser on the dashboard page. You could use your other utilities to verify that and throw an error if it's not.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        wait_for_selector(self.browser, ".cell")
+        return self.browser.find_elements_by_class_name("cell")
+
+
+    @property
+    def current_cell_index(self):
+        return self.cells.index(self.current_cell)
+
+    def remove_safety_check(self):
```
takluyver: 'safety' is rather vague. `remove_unsaved_check`, maybe? Or base it on the hook name: `remove_onbeforeunload`.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    if single:
+        if visible:
+            conditional = EC.visibility_of_element_located
+        else:
+            conditional = EC.presence_of_element_located
+    else:
+        if visible:
+            conditional = EC.visibility_of_all_elements_located
+        else:
+            conditional = EC.presence_of_all_elements_located
```
takluyver: This is steadily heading towards replicating the complexity of the underlying Selenium API. It's fine for now, but we should keep in mind that it may be better to use the underlying APIs directly in some cases.

notebook/tests/selenium/quick_selenium.py
mpacer: ✓

notebook/tests/selenium/utils.py
mpacer: yea… I think you're right on that account.

mpacer:
I think this now stands on its own. It's not fully handling everything that the markdown tests are doing, but it handles all the basic tests of our markdown rendering. 

The mathjax rendering is a different story — however, I think we might be able to test that even more directly in pure javascript code as it's not really about what's rendered on the page but is entirely about what is parsed on the javascript side.

notebook/tests/selenium/utils.py
mpacer: Agreed — however, I think this covers all the cases where we wanted it and it hides a tonne of the selenium boilerplate so I think it's still ok.

notebook/tests/selenium/utils.py
mpacer: I think I took care of this one.

notebook/tests/selenium/utils.py
mpacer: I was wrong in my belief that it was needed here. One of the reasons I really didn't like it was that it couldn't handle notebooks with no cells; so this always seemed wrong. So it's gone!

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,234 @@
...
+    kernel_selector = '#{} a'.format(kernel_name)
+    kernel = wait_for_selector(browser, kernel_selector, single=True)
+    kernel.click()
+
+@contextmanager
+def new_window(browser, selector=None):
+    """Contextmanager for switching to & waiting for a window created. 
+    
+    This context manager gives you the ability to create a new window inside 
+    the created context and it will switch you to that new window.abs
```
takluyver: `abs`?

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,234 @@
...
+        return self.browser.find_elements_by_class_name("cell")
+
+    @property
+    def current_index(self):
+        return self.index(self.current_cell)
+    
+    def index(self, cell):
+        return self.cells.index(cell)
+
+    def remove_safety_check(self):
```
takluyver: Let's find a more representative name for this.

notebook/tests/selenium/test_markdown.py
```diff
@@ -0,0 +1,44 @@
...
+                 '```aaaa\nx = 1\n```',
+                 ]
+    expected_contents = ['<h1 id="Foo">Foo<a class="anchor-link" href="#Foo">¶</a></h1>',
+                         '<p><strong>Bar</strong></p>',
+                         '<p><em>Baz</em></p>', 
+                         '<pre><code>x = 1\n</code></pre>',
+                         '<pre><code class="cm-s-ipython language-aaaa">x = 1\n</code></pre>'
+                         ]
+    for i, cell in enumerate(nb):
+        nb.append(*cell_text, cell_type="markdown")
```
takluyver: I don't think this needs to be in a loop now - the `.append()` call adds all the cells in one go.

notebook/tests/selenium/quick_selenium.py
```diff
@@ -0,0 +1,43 @@
+from selenium.webdriver import Firefox
```
takluyver: Let's have a module level docstring to explain that these utilities are not used by the tests, but they're intended for interactive exploration while writing tests.

takluyver:
The last few things on this are all docstring and naming points. If you've got other things to do, let me know and I can tidy them up.

takluyver:
@mpacer sorry to take over the PR, but I wanted to get it merged in quickly.

rgbkrk:
Thanks @mpacer and @takluyver!

notebook/tests/selenium/quick_selenium.py
mpacer: I had intended to add this.

notebook/tests/selenium/utils.py
```diff
@@ -0,0 +1,222 @@
...
+        new_index = index + 1 if index >= 0 else index
+        if content:
+            self.edit_cell(index=index, content=content)
+        self.convert_cell_type(index=new_index, cell_type=cell_type)
+
+    def add_markdown_cell(self, index=-1, content="", render=True):
+        self.add_cell(index, cell_type="markdown")
+        self.edit_cell(index=index, content=content, render=render)
+    
+    def append(self, *values, cell_type="code"):
```
hroncok: This is invalid syntax on Python 2. Are those files intended to be supported on all the Pythons as notebook itself?

rgbkrk: Notebook development is Python3 only now. Your runtime / kernel will always be allowed to be in Python2.

hroncok: I'm planning to get python2-notebook out of Fedora, however I was wondering what's the upstream status on this. The docs say Python 2.7 is supported. And except the selenium tests it seems to be.

PR #3459, opened 2018-03-22
Undo patches in teardown before attempting to delete files

takluyver:
Deleting the directory can fail on windows, which leaves the patches in place for other tests which run afterwards. Unfortunately we can't use `addCleanup()` with class-level setup/teardown, AFAIK.

This will probably still fail at first (unless the original issue is random), but it should see one test failure on Appveyor rather than 5.

### DIFF
--- a/notebook/tests/launchnotebook.py
+++ b/notebook/tests/launchnotebook.py
@@ -178,9 +178,9 @@ class NotebookTestBase(TestCase):
     def teardown_class(cls):
         cls.notebook.stop()
         cls.wait_until_dead()
-        cls.tmp_dir.cleanup()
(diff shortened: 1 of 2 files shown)


takluyver:
As expected, that now has only one failure. Now to figure out what causes that.

takluyver:
Hmm, it seems that that fixed it. I was kind of hoping that it wouldn't, because I think that means that something isn't getting cleaned up following a request for a static file, so the file is left 'open'. Tornado devs have told us previously that they don't officially support Windows, so we may have to work around this. But for now this makes tests pass on Appveyor.

mpacer:
Looks good to me, I'm going to merge.

PR #3462, opened 2018-03-23
Tweak the vertical margin of #select-all to fix its vertical position

mrkn:
I want to fix the vertical position of the select-all checkbox.

| Before | After|
| --- | --- |
| screenshot 2018-03-23 22 24 37 | screenshot 2018-03-23 22 24 22 |

### DIFF
--- a/notebook/static/tree/less/tree.less
+++ b/notebook/static/tree/less/tree.less
@@ -257,7 +257,8 @@ ul.breadcrumb {
 #select-all {
     margin-left: @dashboard_lr_pad;
     margin-right: 2px;
-    margin-top: 2px;
+    margin-top: 5px;
+    margin-bottom: -3px;
 
     height: 16px;
 }


minrk:
We need to be careful here. Only Chrome is rendering this box wrong: 

screen shot 2018-03-23 at 15 54 25

And with this change, only Chrome is right (or closer, at least):

screen shot 2018-03-23 at 15 58 26


So we need to figure out why Chrome is giving it the wrong height and apply a fix that won't make the other browsers wrong in exchange.

rgbkrk:
In Firefox and Safari, there's a default margin on `<input />`. Normalize.css (not that we're using it) [sets all inputs to `margin: 0`](https://github.com/necolas/normalize.css/blob/7369f566898b8cefc2649b6a66e112857de4c46f/normalize.css#L147-L161) (among other semi-opinionated things).

blink1073:
Closing as out of date, please feel free to re-open if you'd like to pick this back up @mrkn.

Issue #3335, opened 2018-02-13
Convert JS tests to Selenium


takluyver:
We have a number of Javascript tests:

* https://github.com/jupyter/notebook/tree/master/notebook/tests/base
* https://github.com/jupyter/notebook/tree/master/notebook/tests/notebook
* https://github.com/jupyter/notebook/tree/master/notebook/tests

We would like to convert these to use [Selenium](https://pypi.python.org/pypi/selenium), in [this folder](https://github.com/jupyter/notebook/tree/master/notebook/tests/selenium). To run selenium tests, you will need the Selenium Python package and a 'driver' (a backend for it to talk to your browser). Once you have those, you can run:

```
py.test -v notebook/tests/selenium
```

This is quite a big task, and it probably won't be done in one pull request, but rather several smaller ones converting individual bits of the tests.

I'm marking this 'good first issue', though it's a bit borderline. It requires a decent understanding of Javascript and Python, but you don't need to know much about Jupyter to translate the tests. It should get easier as more of the tests are translated, because there will be more examples to refer to.

parkerduckworth:
I would like to take this on.  Getting started on it now

parkerduckworth:
Hmm.  It appears that the Selenium IDE is no longer being supported in Firefox:
https://addons.mozilla.org/en-US/firefox/addon/selenium-ide/

There is a chrome extension, but it is still under development and has a lot of issues. Have we considered Endtest? Are there any other testing frameworks we would consider using?

takluyver:
The Selenium IDE isn't important to us - I haven't used it at all. We're using [Selenium Webdriver](http://www.seleniumhq.org/projects/webdriver/), and the [Firefox backend](https://github.com/mozilla/geckodriver/) is still supported.

parkerduckworth:
Ah, ok! Can I be assigned this issue?

varshneydevansh:
I want to contribute to this good first issue.
I am proficient in Python, C++ and Good at JS but need help to get started.

takluyver:
Github only lets us assign issues to people with push access to the repo, so just commenting that you plan to work on it is sufficient - especially since this is one a large task which multiple people can work on.

[...]

PR #3465, opened 2018-03-24
Convert jstests to selenium: deletecell.js

Sheshtawy:
This PR is related to #3335. It's mean to convert the file `deletecell.js` into a selenium test.


notebook/tests/selenium/test_deletecell.py
```diff
@@ -0,0 +1,74 @@
...
+    JS = 'return Jupyter.notebook.get_cell({}).is_deletable();'.format(index)
+    return nb.browser.execute_script(JS)
+
+def delete_cell(notebook, index):
+    notebook.focus_cell(index)
+    notebook.to_command_mode
+    notebook.current_cell.send_keys('dd')
+
+def test_delete_cells(notebook):
+    print('testing deleteable cells')
```
takluyver: We shouldn't need this print() - pytest should tell us which test is which when it runs them.

takluyver:
Thanks, it looks like the tests are passing now. Are you satisfied that you've converted the whole of `deletecell.js`? If so, you can delete that file as part of this PR - the aim is to get rid of the Phantom tests as well as creating Selenium ones.

notebook/tests/selenium/test_deletecell.py
```diff
@@ -1,10 +1,5 @@
 import os
 import pytest
-from .utils import Notebook
-
-@pytest.fixture
-def notebook(authenticated_browser):
-    return Notebook.new_notebook(authenticated_browser)
 
 def get_cells_contents(nb):
```
Sheshtawy: @takluyver  Thanks for taking the time to review my PR. I made some changes to address your comments. I have only one more idea that I'm not sure about and I would like your input. I was thinking of moving this function to the `Notebook` class since I think it might be needed/repeated in several tests. I'm not sure about other helper functions like `delete_cell`, `cell_is_deletable` and `set_cell_metadata`, do you think they also should be moved to the `Notebook` class too?

takluyver: I'd say that the `get_cells_contents` and `set_cell_metadata` can move to the notebook class, but let's leave `delete_cell` and `cell_is_deletable` here for now. They can always be moved in another PR if we change our minds. :-)

Sheshtawy: I made the changes. I believe the PR is ready to be merged if you don't have any more comments :)

takluyver:
Thanks!

Issue #3251, opened 2018-01-23
"Download As..." Reusing old pdf


waio1990:
Hi, i've been tweaking my notebook settings and got a custom nbconvert template working with the "download as..." menu via the config file, however subsequent downloads do not recompile the xelatex code and almost instantly give me the report i generated this morning.

Is there a way to force a recompile of the latex code? I figured if i deleted the pdf it would have to do it but cannot find the location its being stored on.

Thanks

takluyver:
I don't think we store it anywhere, but maybe it's cached by the browser.

waio1990:
Bingo! Restarting firefox did the trick, thanks!

now i need to figure out how to disable the download cache on firefox :P

takluyver:
I'm going to reopen this, because we can probably do something better with our HTTP responses to control cache behaviour. At a minimum, we should ensure we set a short cache expiry period on nbconvert URLs, but we can probably do something clever with [ETags](https://en.wikipedia.org/wiki/HTTP_ETag).

juhlpatrick:
Confirmed issue here with download to pdf. +1 for fixing this issue

ashleytqy:
Hi, can I try to work on this issue?

takluyver:
Please do! Do you know about HTTP headers to affect browser caches?

takluyver:
And can you reproduce the issue?

ashleytqy:
@takluyve I can't seem to reproduce the issue on Chrome and Firefox. I downloaded a simple latex file, and then changed the file and then downloaded that changed file. the latex seems to load correctly. I didn't clear the cache for both Firefox and Chrome when I downloaded the two files.

May I know how to reproduce this error correctly? @juhlpatrick / @waio1990

takluyver:
> The behaviour is pretty inconsistent, I'm not totally sure why, because when I tried to download a changed copy of the PDF a few minutes later, it updates correctly.

That does sound like browser caching. The browser can keep a copy of data it gets from the server, to save time if it needs to get the same data again later. But then you need ways to decide when the data has changed, so that the browser can get the updated data. The simple way to do that is to have a time limit. A cleverer way is [ETags](https://en.wikipedia.org/wiki/HTTP_ETag), which let the browser ask for 'give me the latest version if it's not this one that I've already got'.

ashleytqy:
I'll look into ETags, thanks @takluyver !

larsasp:
Rename the notebook and try again

PR #3484, opened 2018-03-30
Disable cache when downloading file

ashleytqy:
Fixes https://github.com/jupyter/notebook/issues/3251

The route handler (`NbconvertFileHandler`) that processes a download request did not have cache-control header so this adds the header so ensure that the downloaded pdf / markdown etc. file will always be the most recent (not cached).

### DIFF
--- a/notebook/nbconvert/handlers.py
+++ b/notebook/nbconvert/handlers.py
@@ -141,9 +141,6 @@ class NbconvertFileHandler(IPythonHandler):
         self.set_header('Cache-Control', 'no-store, no-cache, must-revalidate, max-age=0')
         self.finish(output)
 
-    def compute_etag(self):
-        return None
-
 class NbconvertPostHandler(IPythonHandler):
     SUPPORTED_METHODS = ('POST',)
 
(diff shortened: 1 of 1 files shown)


takluyver:
Was it actually computing the etags wrong, or was it the Cache-Control change that fixed it? Looking through tornado's code, it looks like it computes the etag by hashing the content when we call `self.finish()`, which I think should work.

takluyver:
(I'd assumed before that tornado wasn't setting etags by default except for static files, and that we could try setting it from our code)

ashleytqy:
I think the Cache-Control fixed it – I'll remove the `compute_etag` method!

ashleytqy:
hi @takluyver ! does this look good to you?

takluyver:
Oh, sorry, it dropped off my radar. It does look good, thanks. I assume you've tested that it fixes the issue?

I think we might also need the same in the `respond_zip()` function; that's used when nbconvert produces multiple files, so we assemble them into a zip file to be downloaded.

ashleytqy:
@takluyver yup! i’ll also add the change to respond_zip() !

takluyver:
Thanks :-)

PR #3485, opened 2018-03-30
ConfigManager should not write out default values found in the .d directory

maartenbreddels:
This is a followup of #3116.
When the config manager reads the `.json` + `.d/*.json` files, all the data gets merged into one dict. When writted out again, everything ends on in the root `.json` file. Here I remove all entries that are already present in the `.d/*.json` files.

For instance, when installing ipyvolume, you end up with the file `{prefix}/etc/jupyter/nbconfig/notebook.d/ipyvolume.json`. If you now run `jupyter nbextension ...` for anything, all the data from the `ipyvolume.json` file ends up in `{prefix}/etc/jupyter/nbconfig/notebook.d`. Uninstalling ipyvolume will never clean this up, so it will always stay around. This is now avoided by this PR.

### DIFF
--- a/notebook/config_manager.py
+++ b/notebook/config_manager.py
@@ -36,10 +36,11 @@ def recursive_update(target, new):
         else:
             target[k] = v
 
+
 def remove_defaults(data, defaults):
     """Recursively remove items from dict that are already in defaults"""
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks

Issue #2460, opened 2017-04-30
Remove In[*] from cells when notebook is restarted


dpsanders:
I often restart a notebook when a cell hangs.
If several cells were executed, they are all left with `In[*]` when the notebook is restarted, which can be confusing.

I suggest simply removing all occurrences of `In[*]` when the notebook is restarted.

takluyver:
That sounds reasonable to me, and should hopefully be easy to implement. I think you could just add a handler for the `kernel_restarting.Kernel` event.

danagilliann:
Hey! Can I try to tackle this issue?

takluyver:
Please do! Can you see what you might need to do? You'll probably want to search for the `kernel_restarting.Kernel` event in the Javascript code to see where it's fired and how other code uses it.

danagilliann:
Sweet! I'll probably get to this on Friday

danagilliann:
And I got to this much earlier than Friday. I'm looking into doing something [here](https://github.com/jupyter/notebook/blob/9f578dcec041aa1ad724a27afd2e26b2d79ff099/notebook/static/services/kernels/kernel.js#L311-L317). Am I looking in the right place? If I am, does anyone have any suggestions on removing the `In[*]` instances?

danagilliann:
I'm thinking of adding an if statement to determine if the notebook has been restarted somewhere [here](https://github.com/jupyter/notebook/blob/f81fb46ff68323f9652d638a325e9ccc46f62e83/notebook/static/notebook/js/codecell.js#L339) my only issue though is i'm not sure how to determine if the notebook had been restarted. I'm not sure if this is the way I should be going about this since i'm not sure if there is any data to tell me anything about  that.

takluyver:
That code runs when the user executes a cell - I think the issue here is with restarting the kernel after that has happened.

I think the way to do this will be to handle the kernel_restarting event in codecell.js , in the `bind_events` method - i.e. add some code like this:

https://github.com/jupyter/notebook/blob/faa0cab302bb86f0329a512a4ece1f772b29b4c7/notebook/static/notebook/js/notificationarea.js#L117-L120

But inside the handler, if the prompt number is `'*'` set it back to nothing.

danagilliann:
[...]

PR #3494, opened 2018-04-04
solved waiting asterisk

danagilliann:
At the moment, the PR only solves #2460 for an executed cell, meaning if there are other cells running then they will not be "unhanged." Also, the asterisk just gets stripped out meaning the input would look like `[]` as opposed to `[ ]`. If anyone has any ideas as to how to solve these issues, please let me know :)


danagilliann:
Got it! Ah so I thought that if I ran multiple cells at once then the other cells wouldn't be affected, but it turns out they are so it's all good!

takluyver:
Thanks @danagilliann, this looks good to me.

Issue #2518, opened 2017-05-25
When using PageUp/PageDown in Completer, codemirror mistakenly uses goPageUp/goPageDown.


michael-wise:
When viewing code hints (dropdown list of available methods/attributes) for an object, if I navigate the list using page down (instead of arrow keys), it navigates the list as desired, but has the undesired side effect of  also moving the text cursor down the page (as if you'd pressed page down while not using code hints). Resultingly, after you press enter to confirm your attribute, the cursor is one page below the line you were just working on. 

Here is the commit that added the page-up and page-down functionality to the completer dropdown. [https://github.com/jupyter/notebook/commit/fe37c14bdf2c0507a2aaf861c6d2be855632b85b](https://github.com/jupyter/notebook/commit/fe37c14bdf2c0507a2aaf861c6d2be855632b85b)

> Python 3.5.2 |Anaconda 4.2.0 (64-bit)| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)]
> IPython 5.1.0 -- An enhanced Interactive Python.
> Jupyter Notebook Server 4.2.3.

takluyver:
@gnestor have you got time to look into this?

gnestor:
@takluyver Sorry for the delay, but sure. 

@michael-wise What do you propose? Would it be better if page-up/page-down functionality were removed?

takluyver:
If feasible, I think the ideal would be for page-up/page-down to move the text cursor when not completing, and navigate through completions without affecting the text cursor when the completer is open.

If that's tricky, then let's get a simpler fix in for 5.1 (e.g. disable page-up/page-down in the completer.)

takluyver:
@gnestor I've bumped this yet again so I can start moving towards a 5.5 release, but if you do get time to have a look at it, we can un-bump it again.

gnestor:
@takluyver I just submitted a potential fix. It seems to work for me using <kbd>FN+UP</kbd> and <kbd>FN+DOWN</kbd> on a Mac.

PR #3500, opened 2018-04-05
Prevent default on pageup/pagedown when completer is active

gnestor:
Closes https://github.com/jupyter/notebook/issues/2518


takluyver:
Just got round to testing this, and it works :-)

Issue #2337, opened 2017-03-24
Accepting cookies is required but no useful warning is given


Wrzlprmft:
I tried to open a Jupyter Notebook today and got the error message “Error loading notebook” without any further details. After some digging and guesswork, I found out that the problem was that I did not accept cookies from `localhost`. I suggest that the Notebook checks for this and sends a reasonable warning, if cookies are not accepted.

gnestor:
@Carreau Thoughts?

Carreau:
Hum, I'm not even sure how to check for this. I agree that we should try to have a better error message if we can detect it. Or at least having it in the FAQ.

rosaswaby:
Hi @ckilcrease and I were hoping to contribute this issue but we are not able to re-create it. We are both using the chrome browser and we have disabled the cookies in settings and we are both getting the same result which you can see below but it is not giving us the error that you received initially. Would you be able to give any more details on how to re-create the error message. 

<img width="1255" alt="screen shot 2018-03-27 at 11 45 15 am" src="https://user-images.githubusercontent.com/9499838/37978586-6159c794-31b4-11e8-93f7-23a0e049b111.png">

<img width="855" alt="screen shot 2018-03-27 at 11 36 28 am" src="https://user-images.githubusercontent.com/9499838/37978647-78611438-31b4-11e8-8cb2-a2217ec1d6c6.png">

takluyver:
Educated guesswork here, but if you've got an existing notebook file, try running `jupyter notebook my_notebook.ipynb` to launch Jupyter and open the file.

ckilcrease:
Thanks, we were able to recreate it when running with an existing notebook file. We were thinking one approach might be checking if the user has cookies enabled, and if not, adding a warning message describing this. Does this sound like a good approach?

takluyver:
Yep, that sounds reasonable.

rosaswaby:
Thank you for your help! We (me and @ckilcrease) were working on the issue and we managed to detect if the cookies and enabled and produce the correct error message. Do you have any preference for how we should format the error message in the modal?

takluyver:
If you've got something that works, feel free to open a pull request - that's a good format for discussing the specific details, because we can point to particular bits of code.

takluyver:
Closed by #3511.

PR #3511, opened 2018-04-06
Displays Warning Message if Cookies Not Enabled

ckilcrease:
This addresses #2337 

Adds a check for cookies using navigator.cookiesEnabled (tested in Chrome 64.0.3282.186, works both when cookies are blocked altogether and when only blocked for localhost:8888). If the user doesn't have cookies enabled, a request to enable cookies appears in body of modal as follows:

screen shot 2018-04-06 at 12 34 51 pm


takluyver:
Thanks, this is looking good. A couple of things:

- Could we ask you to do this with your editor configured not to strip trailing spaces? If you look at [the changes](https://github.com/jupyter/notebook/pull/3511/files), you'll see that >100 lines are changed because of automatically removed whitespace. This can cause unnecessary conflict, and it makes `git blame` less useful.
- Having two buttons seems redundant; the user doesn't really have two options. Maybe get rid of the 'OK' button in this case?
- I think it would be good for the message to mention that Jupyter needs cookies to work. Also, if we can find a good website explaining what cookies are and how to make sure they're enabled on different browsers, we could link to it.

ckilcrease:
Made most of the changes discussed, but I wanted to check in re: websites on cookies before choosing one; I think [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) does a good job of explaining them concisely, although parts of the page go into details that might not really concern users (such as creating cookies, scope, etc.) - if this would be an issue, I can find one that only presents the overview/relevant details.

ckilcrease:
Made most of the changes discussed, but I wanted to check in re: websites on cookies before choosing one; I think [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) does a good job of explaining them concisely, although parts of the page go into details that might not really concern users (such as creating cookies, scope, etc.) - if this would be an issue, I can find one that only presents the overview/relevant details.

takluyver:
Thanks!

I like MDN, but I think it's too technical for this. If people want help enabling cookies, they're probably going to be put off by all the detail. I was thinking more of pages like [this one](https://www.whatismybrowser.com/guides/how-to-enable-cookies/) or [this one](https://www.wikihow.com/Enable-Cookies-in-Your-Internet-Web-Browser#Using_Firefox_on_Desktop_sub). (Those are just the first ones I found with a quick search; there may well be better ones out there).

ckilcrease:
That makes sense - I looked at a few other articles as well, but it seems like the WikiHow one is the most accurate (many seem to be outdated) and easy to follow, so at the moment I have that one linked as follows:

screen shot 2018-04-10 at 11 45 31 am

takluyver:
Thanks @ckilcrease and @rosaswaby :-)

Issue #3218, opened 2018-01-16
Accessibility color contrast issues in style.min.css


snidersd:
Also see: [jupyter/notebook#1801](https://github.com/jupyter/notebook/issues/1801)
After running the Dynamic Assessment Plugin for Chrome the following high contrast issues were found:
1. From localhost/tree/docs the contrast between the word docs against the gray background fails. Change the color on line 1094 from 337ab7 to 336db7 or darker passes.
2. Create a new notebook and go back to localhost/tree/docs. The word "running" appears next to the file name of the new notebook, but the color green used for the word "running" fails color contrast. Changing the color on line 9812 from 5cb85c to 5c7f5c or darker passes.
3. From localhost/tree#running (a new notebook is running) the contrast for line 3278 kernel-name #foad4e needs to be darker.
4. From localhost/tree#running (a new notebook is running) the contrast for line 9814 btn-warning #5bc0de needs to be darker.
5. From a new notebook (http://localhost/notebooks/docs/Untitled.ipynb) the contrast for line 11978 kernel_indicator #777 needs to be darker.

takluyver:
I've marked this 'good first issue' as there are specific concrete things to do be done.

takluyver:
But for anyone tackling it: note that `style.min.css` is a generated file. You'll need to identify the source files things come from (probably .less files) and fix the issues there.

var7:
Hello! Can I try my hand at this?

var7:
I'm new to this so please correct me if I am wrong. 😄 
1. Identify where the colour values for the particular elements described are present -> look in .less files. (This is where I am currently stuck, any pointers appreciated 😅)
2. For each of the above mentioned points change the colour values to improve contrasts. 
3. Commit these changes with a description. 
4. Submit a pull request.

takluyver:
Yup, that sounds right.

The less files are scattered around the `notebook/static` directory in the repository. You can try searching for the relevant values with an IDE or a tool like `grep`. Or you can [install the notebook from source](http://jupyter-notebook.readthedocs.io/en/stable/contributing.html#setting-up-a-development-environment) and use your browser's inspector to help find the relevant files (it uses source maps to point back to the less files, even though the browser has loaded the minified css).

var7:
I tried using `grep` and the `find in all files` feature in Sublime text to look in the repository for the colour values but could not find any. 

[...]

PR #3514, opened 2018-04-08
Color contrast improvements

nevanscott:
Fixes https://github.com/jupyter/notebook/issues/3218

- [x] From localhost/tree/docs the contrast between the word docs against the gray background fails. Change the color on line 1094 from 337ab7 to 336db7 or darker passes.
- [x] Create a new notebook and go back to localhost/tree/docs. The word "running" appears next to the file name of the new notebook, but the color green used for the word "running" fails color contrast. Changing the color on line 9812 from 5cb85c to 5c7f5c or darker passes.
- [x] From localhost/tree#running (a new notebook is running) the contrast for line 3278 kernel-name #foad4e needs to be darker.
- [x] From localhost/tree#running (a new notebook is running) the contrast for line 9814 btn-warning #5bc0de needs to be darker.
- [x] From a new notebook (http://localhost/notebooks/docs/Untitled.ipynb) the contrast for line 11978 kernel_indicator `#777` needs to be darker.

### Notes

- I've chosen colors which pass at least [WCAG AA](https://www.w3.org/TR/WCAG20/#visual-audio-contrast7) standards for color contrast.
- The Shutdown and Delete buttons were being used at very small sizes, and contrast was a big issue with both of these. Prior to this change, these were being called out with orange and red, respectively. I'm recommending in some cases to swap the colors for icons instead, so that contrast can remain high, with clear indicators for all users.

## Before

- The orange Shutdown and red Delete buttons have poor contrast, especially at such a small size
- Trash can may be difficult to parse, especially at this size
- Blue links and green icons/labels could have improved contrast (less severe than the buttons)

screenshot 2018-04-14 11 49 12

screenshot 2018-04-14 11 49 27

## After

- Shutdown and Delete buttons use icons to stand out. (They are also adjacent to keep high-stakes actions together)
- Delete button includes a text label for clarity
- Blue and green have been tweaked for better contrast

screenshot 2018-04-14 11 52 23

screenshot 2018-04-14 11 52 45

### DIFF
--- a/notebook/static/base/less/variables.less
+++ b/notebook/static/base/less/variables.less
@@ -8,6 +8,7 @@
 @navbar-height: 30px;
 @brand-success: #006500;
 @brand-info: #165E73;
+@brand-warning: #923D00;
 @breadcrumb-color: darken(@border_color, 30%);
 @blockquote-font-size: inherit;
 @modal-inner-padding: 15px;
--- a/notebook/static/base/less/variables.less
+++ b/notebook/static/base/less/variables.less
@@ -7,6 +7,7 @@
 @link-color: #033EAF;
 @navbar-height: 30px;
 @brand-success: #006500;
+@brand-info: #165E73;
 @breadcrumb-color: darken(@border_color, 30%);
 @blockquote-font-size: inherit;
 @modal-inner-padding: 15px;
--- a/notebook/static/base/less/variables.less
+++ b/notebook/static/base/less/variables.less
@@ -4,6 +4,7 @@
 @text-color: @black;
 @font-size-base: 13px;
 @font-family-monospace: monospace;  // to allow user to customize their fonts
+@link-color: #033EAF;
 @navbar-height: 30px;
 @breadcrumb-color: darken(@border_color, 30%);
 @blockquote-font-size: inherit;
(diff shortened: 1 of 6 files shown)


nevanscott:
One downside to the AAA contrast guideline, is that the colors have to be so dark against a white background that they lose a significant amount of their saturation.

Depending on what anyone else thinks, it may be better to aim only for WCAG AA compliance on this. (This would probably also be a good benchmark for accessibility in general.)

takluyver:
Thanks for looking into this. I'm not familiar with the different WCAG levels - I can see the required contrast ratios, but I don't have a feel for what they mean in practice. Could you post before/after screenshots for this PR (and, if you've got the time, variants that are only AA-compliant as well)? I think we could consider complying to level AA where we think that the colours are valuable.

nevanscott:
@takluyver Thanks for the feedback. I made a couple of additional changes, as noted above.

Since the overarching issue is accessibility, in some cases I'm proposing using icons to differentiate buttons instead of relying on color.

nevanscott:
@gnestor Looking at some of the other UI-related PRs, seemed like it might be worth asking your opinion on this one. 😊

blink1073:
Hi @nevanscott, apologies for the delay in reviewing this.  The PR has picked up some substantial merge conflicts due to other changes to the `tree.html` page.  I am closing this as stale, but please feel free to pick it back up if you're available.

PR #3518, opened 2018-04-10
Create selenium interrupt test

lmcdon14:
I had some trouble with accessing the interrupt message to create an assert statement in the test. I worked around this by using an infinite loop and testing to see if the function executes within an appropriate timeframe. We know that the interrupt commands are working properly if this test returns within the allotted time. I know that this isn't as elegant as the original test, but if anything it will be a good starting place once I figure out how to access the interrupt message.


takluyver:
Thanks!

I don't think the infinite loop thing will actually work as you expect. The loop you send is running in the kernel, and the test suite can carry right on even if the kernel is still in that loop.

I think the thing we're missing to make it work is the equivalent of the Javascript function `wait_for_output()`. When you send the shortcut to interrupt the kernel, there's a small delay before the kernel notices this and sends back the `KeyboardInterrupt` message. So if you look for it straight away, there's no output there. Here's the Javascript function:

https://github.com/jupyter/notebook/blob/ca3e7a3668f6e11270e5c07e15dac480f30bcf60/notebook/tests/util.js#L159-L181

I hope there are more efficient ways of doing that in Selenium with selectors than repeatedly sending the browser some Javascript to execute. But I don't know exactly what they are!

lmcdon14:
So if I am able to create this wait_for_output() function, I should be able to access the interrupt message in the cell's "outputs" field?

takluyver:
Yup

takluyver:
Ping @lmcdon14 - do you think you'll have any more time to work on this any time soon?

takluyver:
I've rebased and updated this to try to get it finished.

Issue #3517, opened 2018-04-10
Custom front-end extensions Issue


dabuside:
Following the [official guide](https://jupyter-notebook.readthedocs.io/en/latest/extending/frontend_extensions.html) ,extension only get loaded when notebook page but not in dashboard(file broswer)
add --section='common' to the guide

takluyver:
Do you want to make a PR? The source of that page is here: https://github.com/jupyter/notebook/blob/master/docs/source/extending/frontend_extensions.rst

takluyver:
Closed by #3525

PR #3525, opened 2018-04-11
add tips

dabuside:
add the ``--section='common'`` argument infomation
fix issue https://github.com/jupyter/notebook/issues/3517#issuecomment-380205926

### DIFF
--- a/docs/source/extending/frontend_extensions.rst
+++ b/docs/source/extending/frontend_extensions.rst
@@ -219,11 +219,11 @@ rather than copying it, so there's no need to reinstall after changes.
 To use your extension, you'll also need to **enable** it, which tells the
 notebook interface to load it. You can do that with another command::
 
-    jupyter nbextension enable my_extension/main [--sys-prefix]
+    jupyter nbextension enable my_extension/main [--sys-prefix|--section='common']
 
 The argument refers to the Javascript module containing your
 ``load_ipython_extension`` function, which is ``my_extension/main.js`` in this
-example. There is a corresponding ``disable`` command to stop using an
+example. The ``--section='common'`` argument will affect all pages. There is a corresponding ``disable`` command to stop using an
 extension without uninstalling it.
 
 .. versionchanged:: 4.2
--- a/docs/source/extending/frontend_extensions.rst
+++ b/docs/source/extending/frontend_extensions.rst
@@ -219,11 +219,11 @@ rather than copying it, so there's no need to reinstall after changes.
 To use your extension, you'll also need to **enable** it, which tells the
 notebook interface to load it. You can do that with another command::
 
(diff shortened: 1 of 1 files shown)


docs/source/extending/frontend_extensions.rst
```diff
@@ -219,11 +219,11 @@ rather than copying it, so there's no need to reinstall after changes.
 To use your extension, you'll also need to **enable** it, which tells the
 notebook interface to load it. You can do that with another command::
 
-    jupyter nbextension enable my_extension/main [--sys-prefix]
+    jupyter nbextension enable my_extension/main [--sys-prefix|--section='common']
```
takluyver: It's not mutually exclusive from the `--sys-prefix` option, so the normal way is to put each one in its own `[square brackets]`.

docs/source/extending/frontend_extensions.rst
```diff
@@ -219,11 +219,11 @@ rather than copying it, so there's no need to reinstall after changes.
...
 
 The argument refers to the Javascript module containing your
 ``load_ipython_extension`` function, which is ``my_extension/main.js`` in this
-example. There is a corresponding ``disable`` command to stop using an
+example. The ``--section='common'`` argument will affect all pages. There is a corresponding ``disable`` command to stop using an
```
takluyver: I think this should mention that the default is for the extension only to be loaded on the notebook view - that is probably what most extensions want.

docs/source/extending/frontend_extensions.rst
```diff
@@ -219,11 +219,11 @@ rather than copying it, so there's no need to reinstall after changes.
 To use your extension, you'll also need to **enable** it, which tells the
 notebook interface to load it. You can do that with another command::
 
-    jupyter nbextension enable my_extension/main [--sys-prefix|--section='common']
+    jupyter nbextension enable my_extension/main [--sys-prefix][--section='common']
 
 The argument refers to the Javascript module containing your
 ``load_ipython_extension`` function, which is ``my_extension/main.js`` in this
-example. The ``--section='common'`` argument will affect all pages. There is a corresponding ``disable`` command to stop using an
+example. The ``--section='common'`` argument will affect all pages, by default it will be loaded on the notebook view only. There is a corresponding ``disable`` command to stop using an
```
takluyver: Can you wrap this so the lines are max 80 columns? Thanks.

dabuside:
have any idea to make extension only loaded in the dashboard(file browser) page? [--section='dashboard']arguments will affect notebook view as well

takluyver:
`tree` should be the name (like in the URL). The list of accepted values is here:

https://github.com/jupyter/notebook/blob/f69ce9fcadd46c282450ff57bf4d855b833f9cae/notebook/nbextensions.py#L37

PR #3527, opened 2018-04-11
jQuery attribute selector value MUST be surrounded by quotes

elgalu:
Same problem as https://github.com/ipython/ipyparallel/pull/321

### DIFF
--- a/notebook/static/tree/js/main.js
+++ b/notebook/static/tree/js/main.js
@@ -209,7 +209,7 @@ requirejs([
     
     // load tab if url hash
     if (window.location.hash) {
-        $("#tabs").find("a[href=" + window.location.hash + "]").click();
+        $("#tabs").find("a[href='" + window.location.hash + "']").click();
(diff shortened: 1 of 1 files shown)


notebook/static/tree/js/main.js
```diff
@@ -209,7 +209,7 @@ requirejs([
     
     // load tab if url hash
     if (window.location.hash) {
-        $("#tabs").find("a[href=" + window.location.hash + "]").click();
```
takluyver: Thanks! It looks like it's possible to use `'single quotes'` ((jQuery docs)[https://api.jquery.com/attribute-equals-selector/]) rather than escaping double quotes, which would be a bit easier to read, IMO. It's not a big deal, though.

elgalu: Good point, much more readable. Fixed!

takluyver:
Thanks!

PR #3534, opened 2018-04-15
set the best monospaced font for users in advance

chuling:
I think that under normal circumstances, **most users will not actively set the font in the bowser**, so we should **set the best font for them in advance.**

I don't like css's default monospace font (usually the Courier New font), so I would like to set the best fonts under different systems (specifically, macOS' Menlo and Monaco, Ubuntu's Ubuntu Mono and Windows' Consolas ) to the default monospaced font for Jupyter notebook. The following is the display    on macOS after beautification.

screen shot 2018-04-16 at 1 29 19 am
screen shot 2018-04-16 at 1 29 53 am
screen shot 2018-04-16 at 1 30 21 am

### DIFF
--- a/notebook/static/base/less/variables.less
+++ b/notebook/static/base/less/variables.less
@@ -3,7 +3,7 @@
 @black: #000;
 @text-color: @black;
 @font-size-base: 13px;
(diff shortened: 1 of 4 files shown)


blink1073:
Closing this as stale, since we're not making substantive UI changes to classic notebook.

PR #3437, opened 2018-03-17
[WIP] [1009] Making the prompt clickable

arovit:
Made changes to make the prompt clickable. Only the portion between "[ ]" on the prompt becomes clickable. Also, cursor turns into a pointer to hint user that the portion is clickable.

Issue link: https://github.com/jupyter/notebook/issues/1009

Please provide further guidance on how can we improve the changes.

screen shot 2018-03-16 at 5 27 28 pm

PR #3535, opened 2018-04-16
[WIP] Executing cell when clicking its prompt

### LINKED ISSUE #1009: add a execute button near "In [ ]" so it can be like the refresh button in dashboard page

lucasoshiro:
Closes #1009. Based on https://github.com/jupyter/notebook/pull/3437. When clicking on a prompt, executes the corresponding cell.

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -164,7 +164,7 @@ define([
         var run_this_cell = $('<div></div>').addClass('run_this_cell');
         run_this_cell.prop('title', 'Run this cell');
         run_this_cell.append('<i class="fa-step-forward fa"></i>');
-        run_this_cell.mouseup(function (event) {
+        run_this_cell.click(function (event) {
             event.stopImmediatePropagation();
             that.execute();
         });
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -505,8 +505,17 @@ define([
         }
         this.input_prompt_number = number;
         var prompt_html = CodeCell.input_prompt_function(this.input_prompt_number, nline);
+
+        var that = this;
+        var input_prompt = this.element.find('div.input_prompt');
+
         // This HTML call is okay because the user contents are escaped.
-        this.element.find('div.input_prompt').html(prompt_html);
+        input_prompt.html(prompt_html);
+        input_prompt.addClass('clickable_prompt');
+        input_prompt.mouseup(function (event) {
+            event.stopImmediatePropagation();
+            that.execute();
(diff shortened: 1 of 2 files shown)


takluyver:
(Repeating my comment from #1009)

Interesting idea! I have some concern that people might be used to clicking on the prompt to select a cell without going into edit mode, and if that suddenly executes the cells, it could be an unpleasant surprise - especially for cells which take a long time to run.

I'll try to think about how we might test it out.

lucasoshiro:
Hello again. Now, when the mouse is over a cell, a small run icon appears on the left of the prompt:

button_mouse_over

takluyver:
I like this idea for mouse-driven use, though it doesn't help so much on a touchscreen. Maybe we can use some CSS media queries to make these buttons always visible on touch devices?

There's a test failure where a test checks the HTML of the input prompt.

lucasoshiro:
I have searched a media query for touchscreen devices. I am using any-pointer: coarse. It worked on Chrome, but didn't worked on Firefox, maybe with some javascript it can be solved... I have also fixed that test failure, now the button is outside the prompt div.

lucasoshiro:
@takluyver Sadly, `any-pointer` media query is not supported by some web browsers (https://caniuse.com/#search=any-pointer)... Do you think it's a good idea show the buttons always on those browsers?

takluyver:
It looks like it's mainly Firefox that doesn't support it, but there's a non-standard `-moz-touch-enabled` that works there. I found this suggested use on a bug report:


```css
@media (-moz-touch-enabled: 1), (pointer: coarse) {
  // Rules for touch screens
}
```

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-moz-touch-enabled

lucasoshiro:
That worked here, on a touchscreen computer. Thanks!

notebook/static/notebook/js/codecell.js
```diff
@@ -160,6 +160,15 @@ define([
 
         var input = $('<div></div>').addClass('input');
         this.input = input;
+
+        var run_this_cell = $('<div></div>').addClass('run_this_cell');
+        run_this_cell.prop('title', 'Run this cell');
+        run_this_cell.append('<i class="fa-step-forward fa"></i>');
+        run_this_cell.mouseup(function (event) {
```
takluyver: Is there a reason to do this on the mouseup event rather than the click event? I think we usually use click.

lucasoshiro: Ok, I'll replace it with click event

takluyver:
Thanks, let's see how this works out. I may ping you if there are issues with it.

Issue #3521, opened 2018-04-10
see file size in file browser 


ggrrll:
Hi, 

I was wondering if there is a way (or if it's a planned feature ) to see the files size in the file browser, next to the modification time, for instance.

Thanks.

takluyver:
I don't think it's on the roadmap, but I don't see a problem with adding it, if someone wants to work on it.

The frontend and the API should treat it as optional information, because some storage backends may not have an easy way to calculate a size.

ckilcrease:
Sounds like a cool feature - could @ashleytqy, @rosaswaby, @danagilliann and I work on this together?

takluyver:
Absolutely! In general, if no-one has indicated that they're working on an issue, there's no need to ask permission to tackle it. :-)

Can you see what bits of code you'll need to look at?

takluyver:
Yep, that looks right. There are three kinds of model that the server can send to the frontend, for directories, notebooks, and files (i.e. files other than notebooks). It probably makes sense for notebooks and files to have a size field.

PR #3539, opened 2018-04-17
Adding file size to views

ashleytqy:
This PR addresses: https://github.com/jupyter/notebook/issues/3521
- Adds file size to files + notebooks, but not for directories (it's `undefined` for directories)
- Displays it in the view, with sorting functionality
- Filesize is humanized (originally in bytes)

@ckilcrease 

Sort by ascending
---
screen shot 2018-04-18 at 1 09 37 pm


Carreau:
Thanks, 

I haven't touched this part of the code base in a while so i'll defer to someone else for complete review. 

You added looking at filesize in each model when doing a Get request. Was that necessary to get that in directory listing ? Or is the server doing a os.stat on every file when listing a directory ? If is is there might be a perf bottle neck we need to be careful there.


You seem to have 16 commits and a collaborator (good job to both of you for handling Git). Do you wish to reduce your commit number (rebase), to make history cleaner. 

From your screenshot I can see that some files are 2B and below, and actually fraction of a Bytes.... Are you sure this is right ? Is there not a factor of 1000 somewhere ? (feel free to use an [external library](https://www.npmjs.com/package/pretty-bytes), assuming it is small enough, the license is right... etc). )

I'll do further inline comments.

notebook/services/api/api.yaml
```diff
@@ -814,6 +814,10 @@ definitions:
         type: string
         description: Last modified timestamp
         format: dateTime
+      size:
+        type: integer
```
Carreau: Integer or None ? If size can't be found, probably say that in the description at least.

notebook/static/tree/js/notebooklist.js
```diff
@@ -520,6 +545,12 @@ define([
             .addClass("item_name")
             .appendTo(link);
 
+        $("<span/>")
+            .addClass("file_size")
+            .addClass("pull-right")
+            .css("width", "65px")
```
Carreau: You likely want to avoid inline css.

notebook/static/tree/js/notebooklist.js
```diff
@@ -835,6 +866,7 @@ define([
         // Add in the date that the file was last modified
         item.find(".item_modified").text(utils.format_datetime(model.last_modified));
         item.find(".item_modified").attr("title", moment(model.last_modified).format("YYYY-MM-DD HH:mm"));
+        item.find(".file_size").html(utils.format_filesize(model.size) || "&nbsp;");
```
Carreau: the use of `.html()` should be a red flag. From a security standpoint it would not pass an audit. We had security issues before because of HTML usage. 

If somewhere where to replace you API to return `cannot find <size> for file <file>` the you get scrip injection in your page, and your client is compromised. 

If you've seen use of `.html()` in other place of the javascript where `.text()` should/could be used, please open an issue.

ashleytqy:
@Carreau I tried to install and use `pretty-bytes` with bower but the library doesn't seem compatible with bower, so i copied over the code with credits if thats okay with you

notebook/static/tree/js/notebooklist.js
ashleytqy: do you know where the css should go? from what i've seen, most of the CSS is defined like this:

https://github.com/jupyter/notebook/blob/8034d714298e212eaef43b8e43f04d55c87bed20/notebook/static/tree/js/notebooklist.js#L1250-L1263

Carreau: Likely in `notebook/static/tree/less/tree.less` there are case when we do inline css, there might be a reason for progress-bar, maybe not, if not that's likely technical dept we need to fix at some point.

You'll  likely need to recompile thee css with `python setup.py css` after modification of the less file.

notebook/static/tree/js/notebooklist.js
Carreau: Good question, there are several way to approach that. A simple one would be to use `.html()` just for `&nbsp;`:

```
if (cfoo){
	$(...).text(foo); 
} else {
    $(...).html("&nbsp");
}
```

That reduce the use of html so it's one less risk. 

Thinking about it more, `&nbsp;` is also just an escape sequence for unicode character [Non Breaking Space](http://www.fileformat.info/info/unicode/char/00a0/index.htm), which you can enter with your keyboard using some clever keyboard combinaison `| |` (nbsp), not like `| |`(space), check the source, on mac it's Alt Space. That's 1) annoying to type 2) annoying to see when reading code. That's not optimal. After thinking a few minutes more you can realize that hopefully most programming language allow you to have escape sequence, knowing that Non-Breaking space is unicode [`U+00A0`](http://www.fileformat.info/info/unicode/char/00a0/index.htm) you can go with : 
```
.text(checked===0 ? '\xA0' : checked); 
```

If you want to even avoid using jQuery you [can use](https://caniuse.com/#feat=innertext) the `innerText` attribute and assign to it.

With your gained knowledge you can even open an issue for `$('#counter-select-all').html(checked===0 ? '&nbsp;' : checked);` to be fixed. 

As an exercise – if you like the fun of hacking – get in a shoes of an attacker and try to modify the server side to not return the size as an integer but a string, and attempt to have a script injection that display a message in the frontend.

notebook/services/api/api.yaml
```diff
@@ -814,6 +814,10 @@ definitions:
...
+        description: "The size of the file or directory."
+        format: bytes
```
takluyver: I think this should just go in the description. Swagger specifies a fixed list of formats for things like `dateTime`. From the spec's point of view, this is just a number - I don't think it has a way of declaring a unit.

notebook/services/contents/filemanager.py
```diff
@@ -183,6 +183,14 @@ def is_hidden(self, path):
         os_path = self._get_os_path(path=path)
         return is_hidden(os_path, self.root_dir)
 
+    def _get_file_size(self, path):
+        try:
+            size = os.path.getsize(path)
```
takluyver: This will do a new stat on the file, which is best avoided for performance reasons. We already get a stat object in `_base_model()` - perhaps we can use the size from that, and remove it again for directories. It's a bit of an ugly trick, but for a big directory, it could make a difference.

notebook/static/base/js/utils.js
```diff
@@ -1038,7 +1038,34 @@ define([
             return (order == 1) ? 1 : -1;
         }
     };
+    
+    // source: https://github.com/sindresorhus/pretty-bytes
```
takluyver: I'm OK with this since it's one fairly small function, and it's unlikely to change. But we should include a copy of the license in the comment.

https://github.com/sindresorhus/pretty-bytes/blob/master/license

notebook/static/tree/js/notebooklist.js
takluyver: Stepping back a bit from unicode discussions: do we actually need a non-breaking space at all? I think the layout should still work if the element has no text.

takluyver:
Looking at the screenshots, I think there should be a bit more whitespace between the last modified time and the size, so it's clear that they're separate fields. Maybe right-align the sizes?

notebook/static/tree/js/notebooklist.js
ashleytqy: I fixed it with using `\xA0`. Without text, the `span` element wrapping it doesn't seem to adhere to the width it's given.

notebook/services/contents/filemanager.py
ckilcrease: @takluyver thank you -- I've updated so that it uses file size from `st_size` (retrieved from stat) in `_base_model()`, but let me know if this isn't what you meant

takluyver:
Thanks, that was what I meant about using the existing stat info. :-)

I think this is looking good now, though I'm too tired to review it closely. But there's some problem with the JS tests on Travis - can you investigate?

Carreau:
Travis seem happy, Appveyor was not but this seem like an unrelated error. I restarted he build.

notebook/services/contents/filemanager.py
```diff
@@ -245,6 +245,14 @@ def _base_model(self, path):
         """Build the common base of a contents model"""
         os_path = self._get_os_path(path)
         info = os.lstat(os_path)
+        
+        try:
+            # size of file 
+            size = info.st_size
+        except (ValueError, OSError):
```
takluyver: I don't think there can be an error here - if `lstat()` returns, then we should have a valid stat object from which we can get `st_size`. But it shouldn't be a problem to be defensive in case I'm wrong.

notebook/services/api/api.yaml
```diff
@@ -814,6 +814,9 @@ definitions:
         type: string
         description: Last modified timestamp
         format: dateTime
+      size:
+        type: integer
+        description: "The size of the file or notebook in bytes. If no size is provided, defaults to None."
```
takluyver: Since this is describing a JSON API, it's probably better to refer to *null*, which is the JSON/Javascript name, rather than *None*, which is from Python. This isn't worth holding the PR up for, so I'll change it myself and then merge.

takluyver:
Thanks!

PR #3541, opened 2018-04-17
basic __version__ reporting for extensions

ivanov:
@mpacer and @ian-r-rose  were just working on jupyterlab-latex stuff and said
"Wouldn't it be nice if `jupyter serverextension list` showed the version?"

With this patch, we now do just that.

```
$ jupyter serverextension list
config dir: /Users/pi/.virtualenvs/nb/bin/../etc/jupyter
    jupyterlab  enabled
    - Validating
      jupyterlab 0.32.0 OK  
```


ivanov:
Here's a screenshot:

screen shot 2018-04-17 at 4 47 30 pm

Carreau:
❤️

PR #3542, opened 2018-04-17
Lay base for structured Logging.

Carreau:
This is an attempt at modifying the log calls in order to prepare for
Structured logging. Structured logging should push us to not format the
string at log time but to actually be able to send logs for example as
JSON.

This is necessary – or at least highly useful – in the contex of
GDPR/Hippa... etc where logs can contain sensitive informations and
should be treated adequately (that is to say anonymisation,
pseudonymisation, encrypted storage with acces control... etc).

It is hard to do with unstructured logging.

By forcing the logs to be structured, with `{name}` formatting and
passing explicit keywords, this will allow a potential structured logger
to be further process and be triggered on specific key names.

Consistency in key names is critical to know what is "sensitive".
Example of sensitive info are user-id, user-name, can be contained in
url, and filepath, and may be sensitive. Error in general can contain
snippets of code, so even more.

That will affect translation, as right now it seem to heavily use %x
syntax that should be avoided to match specific keys.

-- 

cc @yuvipanda thoughts ?

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -384,7 +384,7 @@ class NotebookPasswordApp(JupyterApp):
     def start(self):
         from .auth.security import set_password
         set_password(config_file=self.config_file)
-        self.log.info("Wrote hashed password to %s" % self.config_file)
+        self.log.info("Wrote hashed password to {path}", path=self.config_file)
 
 def shutdown_server(server_info, timeout=5, log=None):
     """Shutdown a notebook server in a separate process.
@@ -653,7 +653,7 @@ class NotebookApp(JupyterApp):
         try:
             s.bind(('localhost', 0))
         except socket.error as e:
-            self.log.warning(_("Cannot bind to localhost, using 127.0.0.1 as default ip\n%s"), e)
+            self.log.warning(_("Cannot bind to localhost, using 127.0.0.1 as default ip\nerror"), error=e)
             return '127.0.0.1'
         else:
             s.close()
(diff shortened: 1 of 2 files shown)


notebook/notebookapp.py
```diff
@@ -653,7 +653,7 @@ def _default_ip(self):
         try:
             s.bind(('localhost', 0))
         except socket.error as e:
-            self.log.warning(_("Cannot bind to localhost, using 127.0.0.1 as default ip\n%s"), e)
+            self.log.warning(_("Cannot bind to localhost, using 127.0.0.1 as default ip\nerror"), error=e)
```
Carreau: missing `{}`

yuvipanda:
Thanks for getting this started, @Carreau!

I think for usable structured logging you need something like http://www.structlog.org/en/stable/ (or something similar) and some way of specifying schemas. This could co-exist with or (ideally) replace the current logging set up.

Can you explain a bit more what you mean by:

> By forcing the logs to be structured, with {name} formatting and
> passing explicit keywords, this will allow a potential structured logger
> to be further process and be triggered on specific key names.

Carreau:
Most of the logging goes through Traitlets' `LoggingConfigurable`, so there is not many places to change. I'm not sure how to enforce a schema.

> By forcing the logs to be structured, with {name} formatting and
> passing explicit keywords, this will allow a potential structured logger
> to be further process and be triggered on specific key names.

means move from:
```
log("Lorem ipsum '%s' dolor sit amet=%d, consectetur" % (a, b))
```

to

```
log("Lorem ipsum '{ssn}' dolor sit amet={duration}, consectetur", ssn=ssn, duration=duration)
```

In the first case the log method get the formatted message – which is hard to extract. We could use positional formatting, but then we loose information as to which fields are sensitive: `(45612789, 12)` is not helpful to filter, `{ ssn:45612789, duration:12}`

yuvipanda:
I started doing something like this (but different) in https://github.com/jupyterhub/binderhub/pull/679

Issue #3557, opened 2018-04-23
Adding a "output_updated.OutputArea" event


philippjfr:
The ``display`` function provides a very convenient way of updating output using the ``display_id`` and the resultant display handle. The updates take a very different codepath to regular ``display_data`` and ``execute_result`` messages however, and therefore do not provide a way to subscribe to a general event which makes the whole mime bundle and the OutputArea available.

In JupyterLab this is possible because the ``renderModel`` method of any registered custom mime renderers is given the whole mime bundle. It would be quite convenient to have an equivalent mechanism, which could easily be implemented by adding an ``output_updated``  event for ``OutputArea``. If this idea sounds reasonable I'd be happy to put together a PR.

takluyver:
Sounds reasonable :-)

takluyver:
Closed by #3560

PR #3560, opened 2018-04-24
Add a output_updated.OutputArea event

philippjfr:
As outlined in https://github.com/jupyter/notebook/issues/3557 this PR adds a ``output_updated.OutputArea`` event, triggered by ``update_display_data`` messages, to match the existing ``output_added.OutputArea`` event which is triggered when ``execute_result`` and ``display_data`` messages are received. Like the ``output_added`` event this allows registering callbacks which are given the entire mime bundle and the OutputArea, matching the capability of JupyterLab MIME renderers which are also given the entire mime bundle when ``update_display_data`` messages are received.

Let me know if there is some place to document this new event and/or add some kind of test for it (couldn't find either in a cursory search). As an aside, very happy about my first ever contribution to the notebook.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -631,6 +631,10 @@ define([
             (json.data[MIME_MARKDOWN] !== undefined)) {
             this.typeset();
         }
+        this.events.trigger('output_updated.OutputArea', {
+            output: json,
+            output_area: this,
+        });
     };
 
     OutputArea.prototype._record_display_id = function (json, element) {


takluyver:
Thanks, looks good to me. I'll give @rgbkrk a chance to look at it before I merge, since he has an interest in the rich display machinery.

rgbkrk:
Sweet, love to have a nice small PR introducing a single event.

takluyver:
Tada! Thank you @philippjfr for your contribution. :cookie:

Issue #3549, opened 2018-04-20
question about extension-defined actions 


parmentelat:
Hey folks

I’m facing an issue about an extension that defines actions
(all this is with the classic notebook and the RISE extension)

What is happening is that:

* users can see those actions in the ‘Edit Keyboard Shortcuts’ UI
* from this UI they can also bind the actions to keystrokes
* and that works fine for the current session
* it is also properly saved in nbconfig/notebook.json
* but then if they come back later on in a new notebook session, this error shows up in the console:

```
Uncaught (in promise) Error: does not know how to deal with : RISE:toggle-note
    at ShortcutManager.add_shortcut (keyboard.js:457)
    at ShortcutManager.add_shortcuts (keyboard.js:475)
    at keyboardmanager.js:80
``` 

which of course suggests that keybinding takes place before the extension is loaded

It seems like a pretty standard situation, so I wonder if I did something wrong in the way I defined these actions; 

where should I start looking to see how folks handle that in other extensions ?

takluyver:
It is a reasonable thing to do, and other people have run into the same roadblock (#2954). It's waiting for someone to dig in and figure out a fix. Maybe a way to register shortcuts without checking their names, and then trigger a check of all names after extensions have loaded?

takluyver:
If you've got time to test, can you check that #3561 fixes it? There are instructions on installing from source here: http://jupyter-notebook.readthedocs.io/en/stable/contributing.html#setting-up-a-development-environment

parmentelat:
as an aside, I see this attached to the 5.5 milestone, is there any known indication about when this would be released, even approximative ? 
Just to get a sense of the best pace for addressing this..

takluyver:
The best way to get a sense is [how many other things are in the milestone](https://github.com/jupyter/notebook/milestone/27) - at the moment there are just a handful, and that's because I'd like to get 5.5 ready in the next few days. :-)

parmentelat:
I struggled a bit to get in position, but as far as I can tell, using takluyver:i3549 indeed solves the issue, the registered action does not trigger any message upon reloading, and the shortcut triggers as expected, so we should be fine

Thanks a lot!

-----
PS. I'd like to add that on the mac I had to explicitly run

```
brew install bower
```

in addition to what was mentioned in the contributing doc that you quoted; this was required before I could run 

```
pip3 install -e .
```

from my local notebook repo, that attempted to run `bower` which was not installed

takluyver:
[...]

Issue #2954, opened 2017-10-19
Can't add keyboard shortcuts to actions of extension


shingo78:
We are developing https://github.com/NII-cloud-operation/Jupyter-multi_outputs

I can't add a keyboard shortcut to actions of this extension using _Edit keyboard shortcuts_.
The added keyboard shortcuts are enabled before reloading the notebook, but disabled with the following error after reloading.

```
Uncaught (in promise) Error: does not know how to deal with : MultiOutputs:pin_output
    at ShortcutManager.add_shortcut (keyboard.js:457)
    at ShortcutManager.add_shortcuts (keyboard.js:475)
    at keyboardmanager.js:80
    at <anonymous>
```

Reproduction procedure:
1. Select the _Edit keyboard Shortcuts_ menu item
2. Add a shortcut to the action of _multi\_outputs_ extension, click OK button
3. Reload the notebook

This `MultiOutputs:pin_output` action is registered by the multi_outputs extension.
https://github.com/NII-cloud-operation/Jupyter-multi_outputs/blob/master/lc_multi_outputs/nbextension/main.js#L604-L608

Could you give me some advice?

takluyver:
I guess there's a race condition, that it's trying to add the shortcut before the extension has loaded and defined the action. It's a tricky one; it should be possible to bind a shortcut to an action name that hasn't been registered *yet*, but it's also nice to be able to catch incorrect action names.

Maybe `add_shortcut` could have an extra parameter to allow it to bind to actions which aren't registered yet. @Carreau @gnestor ?

Carreau:
> Maybe add_shortcut could have an extra parameter to allow it to bind to actions which aren't registered yet. @Carreau @gnestor ?

I though it was already the case by default; I'll have to re-check. 
Though I think that might be because of **Edit** shortcuts, I believe *Edit* is more complicated (as it may interfere with Codemirror); so was not completely implemented.

gnestor:
The issue is that the [`actions` and `keyboard_manager` objects are created](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/main.js#L112-L118) before [extensions are loaded](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/main.js#L218-L221). This is for good reason because otherwise the notebook could take a long time to load. 

[...]

PR #3561, opened 2018-04-24
Allow defining keyboard shortcuts for missing actions

takluyver:
When users bind custom shortcuts to actions coming from extensions, the shortcuts can be loaded before the extensions, so we need to allow defining a shortcut for an action that doesn't exist yet.

Closes gh-3549
Closes gh-2954


takluyver:
@parmentelat has confirmed that this fixes the issue they saw.

PR #3567, opened 2018-04-25
remove es6 syntax

forbxy:
It is necessary to be compatible with a lower version of the browser

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -247,7 +247,9 @@ define([
                 fg += 8;  // Bold text uses "intense" colors
             }
             if (inverse) {
-                [fg, bg] = [bg, fg];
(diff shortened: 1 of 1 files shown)


takluyver:
How low a version are you using? We don't want to be stuck with old syntax forever.

forbxy:
chrome 47.x

takluyver:
That's [over two years old](https://en.wikipedia.org/wiki/Google_Chrome_version_history). Given how security critical browsers are now, I'd strongly recommend you let it upgrade to a current version.

takluyver:
Closing since I think this syntax is [sufficiently well supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Browser_compatibility) that we can rely on it.

Issue #3579, opened 2018-04-27
pip does not know of required pyzmq


solstag:
On an up to date Fedora, using python3, pip install does not grab the required version of pyzmq, leading notebook to try the system version and fail.

```
$ pip install notebook
(...works fine...)

$ jupyter-notebook 
Traceback (most recent call last):
  File "/home/narf/.local/bin/jupyter-notebook", line 7, in <module>
    from notebook.notebookapp import main
  File "/home/narf/.local/lib/python3.6/site-packages/notebook/notebookapp.py", line 45, in <module>
    ioloop.install()
  File "/usr/lib64/python3.6/site-packages/zmq/eventloop/ioloop.py", line 210, in install
    assert (not ioloop.IOLoop.initialized()) or \
AttributeError: type object 'IOLoop' has no attribute 'initialized'

$ pip install --user --upgrade pyzmq

$ jupyter-notebook 
(...works fine...)
```

solstag:
In that case, this is a bug in Tornado?
Pip was happy to install Tornado 5 without requiring an updated pyzmq.
If Tornado 5 does not require pyzmq, but if using it then requires that it be not < 17, perhaps it should check and give a more informative error at loading time?
Or perhaps, since it's Notebook that requires both, it could check the versions when started and tell the user what are his options (upgrade pyzmq/downgrade tornado)?
Thanks!

takluyver:
It's definitely not a bug in tornado. I'm coming round to the idea that we should do something about this, though, because I think one of the changes in pip 10 makes it more likely that people will hit it. Specifically, `pip install --upgrade` now only upgrades dependencies if it needs to satisfy a version requirement, whereas previously it upgraded all dependencies.

@minrk would you be OK with notebook 5.5 requiring pyzmq >= 17? I know that it doesn't really, but it makes it less likely that people will hit this?

minrk:
I'm not a big fan, but I think it's an okay solution if this is going to keep coming up.

takluyver:
OK, #3586 does that.

solstag:
[...]

PR #3586, opened 2018-04-30
"Require" pyzmq>=17

takluyver:
Not really a requirement, but explaining about the pyzmq/tornado incompatibility gets old fast.

Closes gh-3579

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,9 @@ for more information.
     install_requires = [
         'jinja2',
         'tornado>=4',
+        # pyzmq>=17 is not technically necessary,
+        # but hopefully avoids incompatibilities with Tornado 5. April 2018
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thomas, quick question: should this be done for qtconsole too?

takluyver:
Possibly. Have you had many people raising issues due to this incompatibility? I just got tired of explaining it repeatedly, so this should get more people upgraded.

ccordoba12:
> Have you had many people raising issues due to this incompatibility?

Yep, we've had our share of issues about it. I'll open a PR i qtconsole to address it then.

takluyver:
Min already agreed to the idea of doing this, so I'm merging it.

Issue #2793, opened 2017-08-22
After closing the overwrite modal, reload button no longer works


unnamedplay-r:
**How to replicate**:
Save the notebook, change the notebook in an editor then try to save. The notebook will trigger the modal. Once open, click cancel then save again to open the modal. The 'Reload' button no longer works.

**Issue**:
The event handler binded to it originally was not binded again after the modal is removed.

I can tackle this, I just wanted to create an issue to reference when I submit a pull request.

takluyver:
I can still reproduce this on master. @unnamedplay-r are you still interested in making a pull request?

You can find the code which displays this dialog here:
https://github.com/jupyter/notebook/blob/faa0cab302bb86f0329a512a4ece1f772b29b4c7/notebook/static/notebook/js/notebook.js#L2761-L2791

unnamedplay-r:
@takluyver, I unfortunately won't be able to tackle this. It's a relatively simple fix that has a solution demonstrated in the editor.js file of my pull request. It might be a nice way to introduce someone to the codebase.

takluyver:
Thanks. I've marked this 'good first issue', but if no-one new tackles it, one of us can do it before releasing 5.0.

sunilhari:
@unnamedplay-r  Is this issue fixed?

Can you explain me how to reproduce it.I was not able to reproduce.I might be doing something wrong.

Kindly guide me through it

takluyver:
Can you see how to get the "notebook changed on disk" message to appear? If so, it should be possible to reproduce the problem by clicking 'cancel' on that dialog, then getting it to appear again on the same page, and then clicking reload.

sunilhari:
@takluyver I have tried modifying the notebook from an external editor.However i did not get the modal you specified above.

unnamedplay-r:
@sunilhari Here's some detailed instructions:
1. Open up the same notebook twice from Jupyter notebook (NB<sub>1</sub>, NB<sub>2</sub>).
2. In NB<sub>2</sub> edit a cell, then save it.
3. In NB<sub>1</sub>, save immediately and you will trigger the _Notebook changed_ error.
4. In this modal, press **Cancel**.
5. In NB<sub>1</sub>, save again, the _Notebook changed_ error will appear again.
6. In the Modal, press **Reload**. It will close the modal and not reload the page.

This is the unintended behavior.

unnamedplay-r:
Here's the relevant code in Jupyter notebook's editor that solves this issue.

[...]

PR #3589, opened 2018-05-01
Fix for reload button no longer works Issue

sunilhari:
Fix for the issue "Issue #2793:After closing the overwrite modal, reload button no longer works".
This PR fixes the above mentioned issue.


takluyver:
Thanks @sunilhari , that looks like the right fix, and I've tested that it works for me.

Issue #2935, opened 2017-10-14
OSError: [Errno 99] Cannot assign requested address


orobsonpires:
Hello guys,

I am running Jupyter on Ubuntu. I got the following issue:

jupyter@jupyter-VirtualBox:~$ sudo su
[sudo] password for jupyter: 
root@jupyter-VirtualBox:/home/jupyter# cd jupyterNoteBook/
root@jupyter-VirtualBox:/home/jupyter/jupyterNoteBook# clear
root@jupyter-VirtualBox:/home/jupyter/jupyterNoteBook# jupyter notebook
[W 14:25:42.244 NotebookApp] Config option `show_new_folder` not recognized by `NotebookApp`.
[W 14:25:42.245 NotebookApp] Config option `show_new_textfile` not recognized by `NotebookApp`.
[W 14:25:42.246 NotebookApp] Config option `show_terminals` not recognized by `NotebookApp`.
[W 14:25:42.247 NotebookApp] Config option `show_button_upload` not recognized by `NotebookApp`.
[I 14:25:43.490 NotebookApp] [nb_conda_kernels] enabled, 2 kernels found
[W 14:25:44.254 NotebookApp] WARNING: The notebook server is listening on all IP addresses and not using encryption. This is not recommended.
[W 14:25:44.255 NotebookApp] WARNING: The notebook server is listening on all IP addresses and not using authentication. This is highly insecure and not recommended.
[I 14:25:44.275 NotebookApp] [nb_conda] enabled
[I 14:25:44.406 NotebookApp] [nb_anacondacloud] enabled
[I 14:25:44.505 NotebookApp] ✓ nbpresent HTML export ENABLED
[W 14:25:44.505 NotebookApp] ✗ nbpresent PDF export DISABLED: No module named 'nbbrowserpdf'
[I 14:25:44.509 NotebookApp] Serving notebooks from local directory: /home/jupyter/jupyterNoteBook
[I 14:25:44.509 NotebookApp] 0 active kernels 
[I 14:25:44.509 NotebookApp] The Jupyter Notebook is running at: http://[all ip addresses on your system]:950/
[I 14:25:44.510 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[I 14:25:45.386 NotebookApp] Lexcorp: Killed all process jupyter kernel gateway.
[...]

PR #3593, opened 2018-05-04
Prefer IP address over hostname when available

takluyver:
Following discussion on #3356.

@evandam @cyberwillis please let me know if you think this is a good idea. I haven't encountered the same issues as either of you, so I don't want to merge this without some confirmation that it's actually an improvement!

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -30,7 +30,7 @@ New features:
 
 - The files list now shows file sizes (:ghpull:`3539`)
 - Add a quit button in the dashboard (:ghpull:`3004`)
-- Display hostname in the terminal when running remotely (:ghpull:`3356`)
+- Display hostname in the terminal when running remotely (:ghpull:`3356`, :ghpull:`3593`)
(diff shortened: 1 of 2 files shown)


takluyver:
And are there any other values for `self.ip` like '0.0.0.0', which can't be directly used?

cyberwillis:
To me its working fine with this code. 
I replicate your changes locally here and tested:

```
0.0.0.0 => container_name
' ' => localhost 
127.0.0.1  => 127.0.0.1 
xxx.xxx.xxx.xxx => xxx.xxx.xxx.xxx
```
Thanks

evandam:
Looks great to me! Does this handle when you do something like `--ip='*'`? If I recall it's the same thing as passing `0.0.0.0`, too.

cyberwillis:
Yes does I have testes earlier.

cyberwillis:
:)

takluyver:
Thanks both! In it goes.

cyberwillis:
I started to get this problem in my way too

[2935](https://github.com/jupyter/notebook/issues/2935)

PR #3622, opened 2018-05-18
Add Dockerfile

n3uz:
Add Dockerfile easy to run a notebook  in container


takluyver:
Thanks, but we've got so many docker files we have a whole separate repository for them:

https://github.com/jupyter/docker-stacks

PR #3624, opened 2018-05-18
stat symlink to determine access

MathisRosenhauer:
A symlink pointing to a directory where you have no access permissions will result in a 404 for the entire directory containing the symlink.
This could also be fixed in FileContentsManager._dir_model() but I don't know what the implications on  Windows would be.

### DIFF
--- a/notebook/utils.py
+++ b/notebook/utils.py
@@ -140,7 +140,7 @@ def is_file_hidden_posix(abs_path, stat_res=None):
     if os.path.basename(abs_path).startswith('.'):
         return True
 
-    if stat_res is None:
+    if stat_res is None or stat.S_ISLNK(stat_res.st_mode):
         try:
             stat_res = os.stat(abs_path)
         except OSError as e:


notebook/utils.py
```diff
@@ -140,7 +140,7 @@ def is_file_hidden_posix(abs_path, stat_res=None):
     if os.path.basename(abs_path).startswith('.'):
         return True
 
-    if stat_res is None:
+    if stat_res is None or stat.S_ISLNK(stat_res.st_mode):
```
minrk: This should already be the `stat_res` for this path, at least according to the docstring. Do you mean to check ISLNK at another location, perhaps after calling `realpath`?

MathisRosenhauer: In this case the docstring is wrong because _dir_model() passes the result of an lstat() call. To check if we are allowed to read the directory, we have to follow the symlink, though. That's why I called stat() again on the path.
There seems to be a good reason for using lstat() in _dir_model() (#2574).

PR #3630, opened 2018-05-21
[WIP: 838] Add find replace test

arovit:
Adding a selenium based test for finding and replacing in selected cell applied on all cells


notebook/tests/selenium/utils.py
```diff
@@ -258,3 +272,9 @@ def ctrl(browser, k):
     """Send key combination Ctrl+(k)"""
     ActionChains(browser)\
         .key_down(Keys.CONTROL).send_keys(k).key_up(Keys.CONTROL).perform()
+
+def esc(browser, k):
+    """Send key combination esc+(k)"""
+    ActionChains(browser)\
+        .key_down(Keys.ESCAPE).send_keys(k).key_up(Keys.ESCAPE).perform()
```
takluyver: This isn't necessary - `Esc` isn't used as a modifier key that you have to hold down while pressing other keys. Just call `self.to_command_mode()` and then send the other keys as normal.

notebook/tests/selenium/utils.py
```diff
@@ -100,6 +100,20 @@ def focus_cell(self, index=0):
         self.to_command_mode()
         self.current_cell = cell
 
+    def find_and_replace(self, index=0, find_txt='', replace_txt='', replace_all=False):
+        self.focus_cell(index)
+        esc(self.browser, 'F')
+        time.sleep(1)    # TODO: find better way to fill the find and replace form
```
takluyver: Are there any specific elements on the form? We can wait for an element to appear.

notebook/tests/selenium/utils.py
```diff
@@ -100,6 +100,20 @@ def focus_cell(self, index=0):
...
+        self.browser.find_elements_by_css_selector("button.btn.btn-default.btn-sm")[2].click()
+        JS = "document.getElementsByClassName('form-control input-sm')[0].setAttribute('value', '%s')"%find_txt
```
takluyver: We shouldn't need to construct JS just to set the value of a form field like this, I think.

notebook/tests/selenium/utils.py
```diff
@@ -100,6 +100,20 @@ def focus_cell(self, index=0):
         self.to_command_mode()
         self.current_cell = cell
 
+    def find_and_replace(self, index=0, find_txt='', replace_txt='', replace_all=False):
+        self.focus_cell(index)
+        esc(self.browser, 'F')
+        time.sleep(1)    # TODO: find better way to fill the find and replace form
+        self.browser.find_elements_by_css_selector("button.btn.btn-default.btn-sm")[2].click()
```
takluyver: What's this doing?

notebook/tests/selenium/utils.py
arovit: I see okay. Makes sense, will call self.to_command_mode()

notebook/tests/selenium/utils.py
arovit: Hmm, I don't want to comment now before trying again (I rem trying the wait logic while working on this, but couldn't get the wait working because I couldn't find any specific element, but let me retry)

notebook/tests/selenium/utils.py
arovit: right, let me see how I can make it better.

notebook/tests/selenium/utils.py
arovit: This code is basically for clicking double "upside down" arrow (third button) on the find and replace form to enable global replace. It's kinda hacky way of doing it, since the buttons don't have any specific ids on them which made it harder to fetch them directly. But let me will revisit this commit to see if we can make it more robust. Thanks !

takluyver: We could add an ID, so long as that's not going to cause problems. If you do, try closing and reopening the dialog a few times so it's created several successive buttons with the same ID, and check that the button still works.

arovit:
Thanks, added 'ids' in the find_and_replace form. Code looks much clearer now.

notebook/static/notebook/js/searchandreplace.js
```diff
@@ -163,6 +163,7 @@ define([
 
       var allCellsButton = $('<button/>')
       .append($('<i/>').addClass('fa fa-arrows-v'))
+      .attr('id', 'allcells_id')
```
takluyver: IDs don't need to have a `_id` suffix, but they do need to be unique on the whole page - that includes IDs that might come from extensions or from output produced in a notebook. So let's make them a bit more specific, like `findreplace_allcells_btn` - and likewise for the other ones added here.

notebook/tests/selenium/utils.py
```diff
@@ -100,6 +100,16 @@ def focus_cell(self, index=0):
         self.to_command_mode()
         self.current_cell = cell
 
+    def find_and_replace(self, index=0, find_txt='', replace_txt='', replace_all=False):
```
takluyver: Looks like the `replace_all` parameter doesn't do anything. It's fine to drop it for now - it can always be added back when there's a test that wants to use it.

notebook/tests/selenium/test_find_and_replace_apply_all.py
```diff
@@ -0,0 +1,25 @@
+import os
```
takluyver: Let's call the file `test_find_and_replace.py` - we'll probably later want to add tests that aren't doing apply all.

takluyver:
It is looking much clearer  :-)

arovit:
Thanks for reviewing, renamed the ids, removed the replace_all parameter and changed the name of testcase.

takluyver:
Thanks!

PR #3636, opened 2018-05-25
[WIP] Highlight todo comments with overlay

### LINKED ISSUE #1092: Hilight TODO comments
Hi,

Many text editors by default hilight the "TODO" comments. This is not the case for jupyter-notebook which display a `# TODO: Something` comment just the same color as a normal comment (including the "TODO" string).

I think it would be really useful to hilight such comments (at least the "TODO" string).

nicolanrizzo:
Close #1092: Highlight todo comments.

I have added a CodeMirror overlay directly to codemirror-ipython.js. I think that there are better ways to do that, could you please help me cleaning the code? Thanks

### DIFF
--- a/notebook/static/notebook/js/codemirror-ipython.js
+++ b/notebook/static/notebook/js/codemirror-ipython.js
@@ -31,7 +31,22 @@
         } else if (pythonConf.version === 2) {
             pythonConf.identifiers = new RegExp("^[_A-Za-z][_A-Za-z0-9]*");
         }
-        return CodeMirror.getMode(conf, pythonConf);
+
+        var todoHighlightOverlay = {
+            token: function(stream) {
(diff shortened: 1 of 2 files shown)


takluyver:
cc @jcb91 @jfbercher @juhasch - it looks like a couple of the contrib nbextensions use codemirror overlays, so you might know how best to use them. In particular, do we need to do anything particular to avoid breaking overlays added by extensions? Or to ensure that multiple overlays are applied efficiently?

harshaxnim:
Hey, I think quite a few people use XXX as well. May be you could add a condition to check for that string as well.

blink1073:
Closing this PR as stale as part of housekeeping.  Please feel free to re-open if you'd like to pick it back up @nicolanrizzo.

PR #3647, opened 2018-05-29
Add remote installation guide in README.md

SayakoKondo:
I have installed and launched with reference to README.md, and then have accessed Jupyter notebook remotely. But, since a default URL starts with http://localhost or http://127.0.0.1, I could not access there. And I had a hard time finding the document for remote access.
So, I'd like to add URL of the document for it. It is kind information for beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ Launch with:
 
     $ jupyter notebook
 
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #3655, opened 2018-06-02
Update to jQuery 3.3

takluyver:
The 2.x branch we were using is no longer updated, so we should try to move to a newer version.

I've test-driven this briefly to make sure that it's not completely broken, but if there are compatibility issues with extensions or Javascript output that uses jQuery, we won't find out until we get people using it.


takluyver:
OK, let's merge this and see if any issues come up.

PR #3661, opened 2018-06-05
[WIP] Document details dialog

### LINKED ISSUE #3617: Feature Request - Add "Document details ..." to "File" menu at top bar
During today Software Carpentry workshop, one learner asked how to find the path to the Jupyter notebook they were working. I told the learner to use

```
import os
os.getcwd()
```

but would be very helpful if that information could be obtained without type any code.

# Screenshots

File menu looks like

screenshot from 2018-05-16 11-21-59

Would be helpful to have "Document details" on it.

document-details

The inspiration for it is Google Docs.

screenshot from 2018-05-16 11-23-53

lucasoshiro:
Closes #3617. When clicking on "Document details..." on File menu, shows a dialog with the document name, document path (relative to the Jupyter Notebook root) and the last modification date and time

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -190,11 +190,18 @@ define([
         });
 
         this.element.find('#document_details').click(function () {
+            var months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
             var table = $('<table></table>');
             var body = $('<tbody></tbody>');
+            
+            var lmod = that.notebook.last_modified;
+            var last_modified_str = months[lmod.getMonth()] + ' ' +
+                lmod.getDate() + ', ' + lmod.getFullYear() + ' ' + lmod.getHours() + ':'+ lmod.getMinutes();
+            
             body.append('<tr><td><b>File: </b></td><td>' + that.notebook.notebook_name + '</td></tr>');
             body.append('<tr><td><b>Path: </b></td><td>' + that.notebook.notebook_path + '</td></tr>');
-            body.append('<tr><td><b>Last modified: </b></td><td>' + that.notebook.last_modified + '</td></tr>');
+            body.append('<tr><td><b>Last modified: </b></td><td>' + last_modified_str + '</td></tr>');
+            table.addClass('document_details_table');
             table.append(body);
             dialog.modal({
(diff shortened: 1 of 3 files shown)


lucasoshiro:
I did some changes... Here is a screenshot
image

lucasoshiro:
That's weird... here the tests didn't passed, but they are ok on my fork

takluyver:
May be a random failure. I've restarted it.

notebook/static/notebook/js/menubar.js
```diff
@@ -189,6 +189,27 @@ define([
...
         });
 
+        this.element.find('#document_details').click(function () {
+            var months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+            var table = $('<table></table>');
+            var body = $('<tbody></tbody>');
+            
+            var lmod = that.notebook.last_modified;
+            var last_modified_str = months[lmod.getMonth()] + ' ' +
+                lmod.getDate() + ', ' + lmod.getFullYear() + ' ' + lmod.getHours() + ':'+ lmod.getMinutes();
```
takluyver: We use a package called `moment.js` to format dates - there are different conventions in different places. Have a look at the code for the notebook list here: https://github.com/jupyter/notebook/blob/d9baafaabf23d0c0ba7a05254083f69efc398cb9/notebook/static/tree/js/notebooklist.js#L865-L867

notebook/static/notebook/js/menubar.js
```diff
@@ -189,6 +189,27 @@ define([
...
+            
+            body.append('<tr><td><b>File: </b></td><td>' + that.notebook.notebook_name + '</td></tr>');
```
takluyver: We shouldn't construct HTML like this, because the name and the path might contain characters that are significant in HTML. We should construct the relevant HTML node and then call `.text(that.notebook.notebook_name)` on it to set the text safely.

notebook/static/notebook/less/menubar.less
```diff
@@ -140,3 +140,7 @@ ul#help_menu li a{
 }
 
 //end submenu
+
+.document_details_table {
+    width: 100%;
+}
```
takluyver: I don't think this belongs in `menubar.less`, which is about styling for the menus themselves. I'm not sure what is the right place for it, though.

PR #3664, opened 2018-06-06
properly handle unicode characters in home dirs

vit-tucek:
Jupyter fails to start when the user has Unicode characters in his homedir. This fixes that. See https://github.com/sagemath/sage-windows/issues/10#issuecomment-392137049 and https://github.com/jupyter/jupyter_core/pull/131

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -212,7 +212,7 @@ class NotebookWebApplication(web.Application):
         now = utcnow()
         
         root_dir = contents_manager.root_dir
-        home = os.path.expanduser('~')
+        home = py3compat.str_to_unicode(os.path.expanduser('~'), encoding=sys.getfilesystemencoding()) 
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -213,6 +213,8 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
         
         root_dir = contents_manager.root_dir
         home = os.path.expanduser('~')
+        if sys.version_info.major == 2:
+            home = home.decode(sys.getfilesystemencoding())
```
minrk: I believe we have a `py3compat.str_to_unicode(s, encoding=sys.getfilesystemencoding())`

Related question, though: does `os.path.expanduser(u'~')` do the right thing here, deferring to Python's on path-encoding logic?

minrk:
>  Do you want me to this change?

Yes, that sounds great!

vit-tucek:
Done.

PR #3665, opened 2018-06-06
add kernel_info_timeout traitlet for slow kernel start/restart

mpacer:
This allows you to configure how long the notebook waits for a kernel give its `kernel_info_reply`. 

I looked into adding a test, but I think that this would require a good bit of machinery to test properly. I can do that if we want to, but it looks like we're not testing the functionality of most other traitlets directly.

Note: this is not kernel specific… but I don't think we have any way to do kernel specific configuration for the handler as of today.

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -19,7 +19,7 @@ from tornado.ioloop import IOLoop, PeriodicCallback
 from jupyter_client.session import Session
 from jupyter_client.multikernelmanager import MultiKernelManager
 from traitlets import (Any, Bool, Dict, List, Unicode, TraitError, Integer,
-       Instance, default, validate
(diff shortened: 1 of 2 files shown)


mpacer:
I changed this to WIP because I want to make it also work for the initial startup timeout (as it is in `jupyter_console`).

notebook/services/kernels/kernelmanager.py
```diff
@@ -93,6 +93,22 @@ def _update_root_dir(self, proposal):
         no frontends are connected.
         """
     )
+    
+    kernel_info_timeout = Float(60, config=True,
```
rgbkrk: Excellent!

mpacer:
@Carreau @gnestor @takluyver @minrk Are there any issues with this going forward?

minrk:
Great!

PR #3668, opened 2018-06-07
Add option to override host in displayed url

### LINKED ISSUE #3605: localhost is replaced by the docker container name in 5.5.0
In 5.4.1, the url is printed and works as `localhost`, below

But with 5.5.0, `8bc26195e104` appears. 
With `docker ps`, I can see that the docker Container ID is `8bc26195e104`

```
docker run -p 8889:8888 gcr.io/sixty-capital/base:1.18.6 jupyter-notebook --ip="*" --no-browser --allow-root

Generating a 2048 bit RSA private key
...................................................................................+++
...............+++
writing new private key to '/root/.local/share/jupyter/notebook.pem'
-----
[I 00:24:41.730 NotebookApp] Writing notebook server cookie secret to /root/.local/share/jupyter/runtime/notebook_cookie_secret
[I 00:24:42.030 NotebookApp] Serving notebooks from local directory: /notebooks
[I 00:24:42.030 NotebookApp] 0 active kernels
[I 00:24:42.030 NotebookApp] The Jupyter Notebook is running at:
[I 00:24:42.031 NotebookApp] https://8bc26195e104:8888/?token=7e2e428b6b07c921c0fd16b87b25d89bc46b18c1685a8e84
[I 00:24:42.031 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 00:24:42.032 NotebookApp]

    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:

        https://8bc26195e104:8888/?... <--- here

# in 5.4.1 this is https://localhost:8888/?...

```

For reference, here's the config file:

<details>

```python
# from https://github.com/jupyter/docker-stacks/blob/master/base-notebook/jupyter_notebook_config.py

from jupyter_core.paths import jupyter_data_dir
import subprocess
import os
import errno
import stat

PEM_FILE = os.path.join(jupyter_data_dir(), 'notebook.pem')

c = get_config()
c.NotebookApp.ip = '*'
c.NotebookApp.port = 8888
c.NotebookApp.open_browser = False

# Set a certificate if USE_HTTPS is set to something truthy
if os.environ.get('JUPYTER_USE_HTTPS', None):
    if not os.path.isfile(PEM_FILE):
        # Ensure PEM_FILE directory exists
        dir_name = os.path.dirname(PEM_FILE)
        try:
            os.makedirs(dir_name)
        except OSError as exc:  # Python >2.5
            if exc.errno == errno.EEXIST and os.path.isdir(dir_name):
                pass
            else:
                raise
        # Generate a certificate if one doesn't exist on disk
        subprocess.check_call([
            'openssl', 'req', '-new', '-newkey', 'rsa:2048', '-days', '365', '-nodes', '-x509', '-subj',
            '/C=XX/ST=XX/L=XX/O=generated/CN=generated', '-keyout', PEM_FILE, '-out', PEM_FILE
        ])
        # Restrict access to PEM_FILE
        os.chmod(PEM_FILE, stat.S_IRUSR | stat.S_IWUSR)
    c.NotebookApp.certfile = PEM_FILE

# our code - not copied from the link at top
notebook_dir = os.environ.get('JUPYTER_NOTEBOOK_DIR')

if notebook_dir:
    try:
        os.makedirs(notebook_dir)
    except:
        pass
    c.NotebookApp.notebook_dir = notebook_dir
```

</details>

tomjorquera:
This commit introduces a new alias `connect-host` to override the host
info displayed at launch with a custom string.

It is intended to be used when the app is run in an environment where
the url to display to the users is not detectable reliably (proxified or
containerized setups for example).

Closes #3605

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -666,6 +666,19 @@ class NotebookApp(JupyterApp):
             value = u''
         return value
 
+    custom_display_url = Unicode(u'', config=True,
+        help=_("""Override URL shown to users.
+
+        Replace actual URL, including protocol, address, port and base URL,
+        with the given value when displaying URL to the users. Do not change
+        the actual connection URL. If authentication token is enabled, the
+        token is added to the custom URL automatically.
+
+        This option is intended to be used when the URL to display to the user
+        cannot be determined reliably by the Jupyter notebook server (proxified
+        or containerized setups for example).""")
+    )
+
     port = Integer(8888, config=True,
         help=_("The port the notebook server will listen on.")
     )
@@ -1339,11 +1352,16 @@ class NotebookApp(JupyterApp):
     
     @property
     def display_url(self):
-        if self.ip in ('', '0.0.0.0'):
-            ip = socket.gethostname()
+        if self.custom_display_url:
+            url = self.custom_display_url
+            if not url.endswith('/'):
+                url += '/'
         else:
-            ip = self.ip
-        url = self._url(ip)
+            if self.ip in ('', '0.0.0.0'):
+                ip = socket.gethostname()
+            else:
+                ip = self.ip
+            url = self._url(ip)
         if self.token:
             # Don't log full token if it came from config
             token = self.token if self._token_generated else '...'


takluyver:
We have properties `connection_url` and `display_url` - I think it's surprising if `connect_host` affects *display* but not *connection*. But I don't know whether it's better for it to affect both, or to rename the config option.

@minrk Do you remember why we have two separate URL options? I think `connection_url` gets recorded in the server info file, but I'm not sure why we make two different estimates of the URL. Is `connection_url` meant to be the address you can connect to on the same host that's running the server? If we can't unify them, we should certainly find a way to make this clearer.

minrk:
Maybe this should change `display_url` to a trait with the current logic as a default value generator, rather than a property? Since this is just for display, calling it `display_url` makes more sense to me. `connect` would be right if there actually is logic that uses this to establish connections. If we intend for this to be the case, then calling it `connect_url` does makes sense to me. (pedantry: it's not a 'host', since it includes protocol and url path, it's a full URL. We just ran into an issue where I made this mistake and confused people in jupyterhub/oauthenticator).

takluyver:
I think there probably wants to be a property involved, because something has to add the generated token to the bit you can configure.

We use URLs for:

- Printing in the terminal to tell the user what to open
- Launching a browser when the server and the browser are on the same system
- Writing into the notebook server info file
  - This is used by `jupyter notebook stop` to connect and send an HTTP request
  - It can also be used by tools like [nbmanager](https://github.com/jupyter/nbmanager) to query a running notebook server

Those are all the use cases I can think of at the moment, but there may be others.

tomjorquera:
@takluyver, regarding the different `display_url` and `connection_url` property:

I am not too clear on the inner workings of `jupyter-notebook`,  so I was worried that `connect_url` may be used internally and be required to contains the *actual* locally accessible URL. Maybe it is the case in the two examples you give under "Writing into the notebook server info file" ?

More generally, my concern is about using `jupyter-notebook` as a proxied service (in a docker container, behind a proxy etc.), and needing both a publicly accessible URL for the users, and an internal URL for administering. But maybe it's actually a non-issue? 

@minrk, regarding the URL vs host terminology:

Agreed! I initially kept the name proposed in #3605 for consistency, but I think it would be better to rename it. Just waiting for the discussion to converge on what to do to rename it accordingly.

minrk:
@takluyver good summary. In that case, I think `connect_url` is the right choice, since other programs may use this to connect, so if a human wants to override "use this URL to connect", then I think it's a good choice.

In that case, this PR roughly as-is with `connect_url` as the config option rather than `connect_host` should work.

notebook/notebookapp.py
```diff
@@ -1339,11 +1348,14 @@ def init_webapp(self):
     
     @property
     def display_url(self):
-        if self.ip in ('', '0.0.0.0'):
-            ip = socket.gethostname()
+        if self.connect_host:
+            url = self.connect_host + '/'
```
minrk: Probably don't want to add `+ '/'` here, instead leave `connect_url` untouched.

notebook/notebookapp.py
```diff
@@ -548,6 +548,7 @@ def start(self):
     'notebook-dir': 'NotebookApp.notebook_dir',
     'browser': 'NotebookApp.browser',
     'pylab': 'NotebookApp.pylab',
+    'connect-host': 'NotebookApp.connect_host',
```
minrk: I don't think this is an option that warrants a CLI alias.

takluyver:
You mean replace `connection_url` with a configurable option? I was coming to the opposite conclusion,  that there was a need for a separate display URL and connection URL - the later being valid only on the same host.

minrk:
Yeah, I mean that the URL that should be used to *connect* should be overrideable by the user. That is, the URL written to `notebook list` commands, etc. This may actually be used for connect, etc. This is useful for the cases where the notebook cannot determine its connect URL.

The display URL for humans should use the connect URL, and can be a property adding things like the token.

tomjorquera:
@minrk Sorry if I have misunderstood your proposal, what I got is that you suggest the option to not just override the return of `display_url` but the return of `connection_url` instead.

It seems to me that it would break the functioning of commands such as `notebook stop` (it does in my quick testing but maybe I missed something).

The use case I have in mind is the following:
With a notebook server running in a container, I want:
- as a user, to have a URL I can directly paste in my browser transparently
- as an admin, to be able to go inside the container and use admin commands as usual (`list`, `stop` and so on...)

Its not clear for me if there is a way to make this work with your proposal without breaking the second point (or even if it is a non-issue and that I missed something when testing it locally) .

Any thought ?

minrk:
I guess the question is:

1. should the user be able to specify what URL should be used to connect to the notebook server in general (connection_url), or
2. *only* specify what URL should be printed in the logs (display_url), i.e. what a human should use to open the notebook, but not affect anything that discovers the URL programmatically.

1 would allow things like `jupyter notebook list` to provide useful info for connecting from outside the container, but possibly break things like `jupyter notebook stop` *inside* the container. 2 suggests that we never want the notebook to know its own URL that things should use, e.g. reporting its own connection URL for use outside the container after the initial log.

From your use cases, I think `jupyter notebook list` in the container would work *better* in the first case, because it would give you the info you need to connect from outside the container, rather than having to dig into the logs for that. That said, I don't think that the process-management tools like `notebook list` and `notebook stop` make a great deal of sense in a container, since there's necessarily only one server and it can be terminated normally with `docker stop` commands, etc.

I genuinely don't know the right answer for this. Overriding display_url is certainly the least disruptive since it can't break anything, but overriding connect_url enables solving the same "How should things connect to me" problem in multiple places, not just the logs. I'm genuinely not sure what the right answer is.

Another option is to add a separate 'remote_url' for connections from 'remote' hosts (e.g. the host system in the docker case), and store and display this alongside connection_url, essentially baking in the fact that there can never be only one answer to the URL question when running in containers.

tomjorquera:
Since apparently there is two separate use cases that need different ways to be solved, I wonder if the most straightforward solution wouldn't be to add two separation options.

I am not fond of having lots of small options, but in this case it seems that there is a variety of current and future use cases that could warrant it...

mpacer:
@minrk What is needed for this to be completed?

notebook/notebookapp.py
```diff
@@ -666,6 +667,14 @@ def _valdate_ip(self, proposal):
             value = u''
         return value
 
+    connect_host = Unicode(u'', config=True,
+        help=_("""Custom host to show in displayed URL.
+
+        Replace actual host, including protocol, address, port and base URL,
+        with the given value when displaying URL to the users. Do not change
```
willingc: We probably should clarify this docstring a bit. As it reads now, my understanding is that the connect_host would be displayed to the user, but the notebook would use the actual underlying connection URL.

notebook/notebookapp.py
willingc: Should `connect_host` be renamed to `custom_connect_host` or `custom_display_host`. I'm a bit confused by the naming.

willingc:
Should this be pushed out to 5.7 release?

takluyver:
I'd like to get it in 5.6 if practical - we have had a few bug reports since 5.5.

I suggest that for now, we leave the two separate URL properties in existence, and focus on making `display_url` overridable, since that is what people have been having practical problems with. So let's rename the config option to be clear about that, e.g. something like `custom_display_url` or `display_url_base` (base because the token is appended to make the display URL).

That means that for the time being, we're saying that `connection_url` - which `jupyter notebook list` and `jupyter notebook stop` use - is only intended to be usable from the same host that the server is running on. Depending on the network it may be usable from elsewhere (e.g. your system outside a container), but I don't think we can guarantee that in general.

notebook/notebookapp.py
tomjorquera: Funnily I had a *very* similar docstring amending in store :) I merged the two. Tell me if it seems clearer.

notebook/notebookapp.py
tomjorquera: Renamed

@minrk : regarding the trailing '/', i changed things to append it if it is missing. I think it is the "nicest" approach, but tell me if you have reserves on this.

notebook/notebookapp.py
tomjorquera: Removed

notebook/notebookapp.py
tomjorquera: Updated. tell me if it's still not clear enough (your understanding was correc btw :wink:)

tomjorquera:
Updated the PR to clean things up regarding naming and docstring. So currently it only add the option to override the display URL and not the actual URL.

For what I understand of the code, adding a second option to override the actual URL seems straightforward enough, so I think it can be added to the PR should it be the way to go...

tomjorquera:
(also rebased on top of current master, and updated commit description)

minrk:
@tomjorquera thanks!

PR #3678, opened 2018-06-12
Logging updates

### LINKED ISSUE #3564: Strip logging tokens with "[secret]"
We noticed secret tokens get logged. This is problematic in enterprise systems.

Could we have something similar to what JupyterHub does:
https://github.com/jupyterhub/jupyterhub/blob/0561968fac88dd7ffae94c99dd49cab85ffcd6a8/jupyterhub/log.py#L58

I guess this should be implemented somewhere in:
https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L1251

minrk:
- log non-static-file successes at INFO-level (page views, API requests)
- scrub sensitive fields in url params, headers

Alternatives to increased logging that would result in slightly less log output:

- only include simple successes of API requests
- only include simple successes of non-GET methods (biggest reduction)

closes #3564

### DIFF
--- a/notebook/log.py
+++ b/notebook/log.py
@@ -14,6 +14,9 @@ except ImportError:
     from urllib import urlparse, urlunparse
 
 from tornado.log import access_log
+from tornado.web import StaticFileHandler
+
+
 # url params to be scrubbed if seen
 # any url param that *contains* one of these
 # will be scrubbed from logs
@@ -62,8 +65,12 @@ def log_request(handler):
     """
     status = handler.get_status()
     request = handler.request
-    if status < 300 or status == 304:
-        # Successes (or 304 FOUND) are debug-level
+
+    if (
+        isinstance(handler, StaticFileHandler)
+        and (status < 300 or status == 304)
+    ):
+        # Successes (or 304 FOUND) on static files are debug-level
         log_method = access_log.debug
     elif status < 400:
         log_method = access_log.info
--- a/notebook/log.py
+++ b/notebook/log.py
@@ -6,11 +6,55 @@
 #-----------------------------------------------------------------------------
 
 import json
+
+try:
+    from urllib.parse import urlparse, urlunparse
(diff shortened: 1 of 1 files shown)


notebook/log.py
```diff
@@ -6,43 +6,93 @@
...
+    return uri
+
+
+def _scrub_headers(headers):
+    """scrub auth info from headers"""
+    headers = dict(headers)
+    if 'Authorization' in headers:
+        auth = headers['Authorization']
+        if auth.startswith('token '):
+            headers['Authorization'] = 'token [secret]'
```
takluyver: Should we scrub any auth header? What about cookies?

takluyver:
I'm fairly ambivalent about increased logging, but I do occasionally scroll back up in the terminal to copy the URL with the token that the server prints when it starts. The more logging output there is, the less practical that is.

Do we also have an option to disable printing the token URL on start? How can we make it clear to people who might be capturing logs that it contains sensitive information unless you turn that off?

notebook/log.py
minrk: Scrubbing cookies is a good idea. Will add that.

minrk:
As usual, this is one of the competing use case things. I added this because for service-style deployments like Binder / Docker / etc., it makes more sense for services to log all access, but I recognize that it adds noise to scroll past in the local-user case if you need to recover the token again after having used it for a while. I don't love adding more configuration options, but maybe we should have a 'log_requests' flag that's off by default?

> How can we make it clear to people who might be capturing logs that it contains sensitive information unless you turn that off?

Good question. We only log the token for the cases where the token is auto-generated and thus required to put somewhere to communicate it to users. If a token or password is *specified*, then nothing is logged. We could add to the log itself that this message will not be displayed if a token/password is set via config. What do you think?

blink1073:
Closing this one as stale as part of housekeeping.

PR #3687, opened 2018-06-14
add el & update css in output_area to match run button in input prompt from #3535

mpacer:
Without this change the output and Input areas become unaligned.

This still isn't an ideal fix, but it'll make a release possible without making the styling problematic.

This can be seen as a styling continuation of #3535.

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -370,6 +370,7 @@ define([
     OutputArea.prototype.create_output_area = function () {
         var oa = $("<div/>").addClass("output_area");
         if (this.prompt_area) {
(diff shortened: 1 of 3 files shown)


mpacer:
@takluyver @minrk could I ask for an expedited review on this so that I can proceed with making a release candidate for 5.6?

mpacer:
Alternatively to a expedited review, if we do want to iterate on this PR a bunch, I'd like to suggest reverting #3535 to make a 5.6 release candidate possible, pushing it until 5.7 (or 5.6.1) when we can figure out a better way to handle the styling issues.

gnestor:
Thanks @mpacer! Let me know how I can help with 5.6 efforts 👍

PR #1272, opened 2016-03-29
Set sphinx < 1.4 due to LaTeX build errors

willingc:
Sphinx 1.4 was released yesterday. Build failing on 1.4 due to LaTeX build errors. Drop back to 1.3.

PR #3690, opened 2018-06-15
Unpin Sphinx version for Readthedocs

takluyver:
Undoes #1272 - cc @willingc . I hope whatever issues we were experiencing have been fixed by now.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -8,7 +8,7 @@ dependencies:
 - jupyter_client
 - ipykernel
 - pip:
-  - sphinx<1.4
+  - sphinx
   - nbsphinx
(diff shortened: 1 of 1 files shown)


docs/environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
 - jupyter_client
 - ipykernel
 - pip:
-  - sphinx<1.4
+  - sphinx
```
willingc: OMG. In fact we should move sphinx above the pip section.

takluyver: Thanks, done

willingc:
Thanks @takluyver

PR #3692, opened 2018-06-15
Add "test" npm script

gnestor:
This adds a "test" npm script that will allow us to run JS scripts with a:

```bash
# Run all tests
npm run test

# Run markdown test
npm run test -- markdown
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ config.rst
 /.project
 /.pydevproject
 
+package-lock.json
+
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
   "scripts": {
     "bower": "bower install",
     "build": "python setup.py js css",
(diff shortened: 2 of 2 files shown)


takluyver:
We're working on getting rid of the JS tests by converting them to Selenium, so this doesn't make much sense in the long run. But I assume it doesn't hurt anything if you want it for the short term.

gnestor:
Ok, that's right. Let's just close then...

gnestor:
Oh, the other change in here is adding `package-lock.json` to gitignore. If you are using npm > 5, it will generate a lock file. We can either commit it or ignore. What do you think?

takluyver:
Not sure - I've seen that file appear a few times, and put off thinking about it. Is it the norm to check the lock file in? In my experience with Rust it seems that you do commit the lock file, but I don't know if different languages/tools use it in different ways.

PR #3699, opened 2018-06-18
Fixed possible XSS detected by Fortify webscan

kmadac:
Hi,

request to API endpoint returned non-escaped string:

**Request:** 
```
http://192.168.0.10:8888/api/contents/Untitled%20Folder/<iMg%20SrC%3dvBsCrIpT%3aMsgBox%2878146%29>
```

**Response:**
```
{"message": "No such file or directory: Untitled Folder/<iMg%20SrC%3dvBsCrIpT%3aMsgBox%2878146%29>", "reason": null
```

**Expected Response:**
```
{"message": "No such file or directory: Untitled Folder/&lt;iMg SrC=vBsCrIpT:MsgBox(78146)&gt;", "reason": null}
```

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -28,6 +28,7 @@ except ImportError:
 from jinja2 import TemplateNotFound
 from tornado import web, gen, escape, httputil
 from tornado.log import app_log
+from tornado.escape import xhtml_escape
 
 from notebook._sysinfo import get_sys_info
 
@@ -522,6 +523,7 @@ class APIHandler(IPythonHandler):
                 reply['message'] = 'Unhandled error'
                 reply['reason'] = None
                 reply['traceback'] = ''.join(traceback.format_exception(*exc_info))
+        reply['message'] = xhtml_escape(reply['message'])
         self.log.warning(reply['message'])
         self.finish(json.dumps(reply))
 


takluyver:
I think the message should always be handled as text by the frontend, and not displayed as raw HTML. If that is the case, then escaping it on the server side will lead to the wrong text being displayed.

kmadac:
There is no frontend, because it is an API endpoint and response is a json which will get to browser in unescaped form, and which XSS scanning tool reports as possibly dangerous.

takluyver:
The frontend is what runs in the browser, making the API request and handling the response. I understand why the scan tool reports it as *possibly* dangerous, but I believe that we're protecting against that danger in the frontend, which the scan tool can't introspect.

minrk:
Since it's an API response with content-type application/json, html in json shouldn't be an issue, and html-escaping JSON is not generally required or recommended. If this is getting sent with content-type text/html then it might be an issue, but it's not correct to say that JSON messages should not be allowed to contain HTML.

Is there any evidence that viewing this JSON will result in execution in the browser? Because that would be a bug we should fix.

kmadac:
I did some research and @minrk is correct. As Content-Type is application/json, XSS is not possible on all modern browsers. I'm closing this Pull request as unneeded.

takluyver:
Thanks for investigating it, though :-)

PR #3703, opened 2018-06-21
add `or 127.0.0.1` when listening on all-ips

minrk:
Effectively reverts #3356 due to issues like #3605. Makes it not a copyable URL anymore, but there’s simply no way for a copyable host to work in general in this case. For folks who want a copyable URL, they can now set display_url via #3668.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1358,7 +1358,7 @@ class NotebookApp(JupyterApp):
                 url += '/'
         else:
             if self.ip in ('', '0.0.0.0'):
-                ip = socket.gethostname()
+                ip = "(%s or 127.0.0.1)" % socket.gethostname()
             else:
                 ip = self.ip
             url = self._url(ip)


minrk:
5.4: never connectable, actually the bind address

```
The Jupyter Notebook is running at:
http://0.0.0.0:8888/?token=
```

5.5: connectable *if* hostname is valid (generally true, but not in containers)

```
The Jupyter Notebook is running at:
http://hostname:8888/?token=
```

this PR: not connectable directly via copy/paste due to `(host, ip)` format, but both hosts *are* connectable. I think the lack of copy/paste-ability is *helpful* in the docker case, since it's impossible for it to be reliably correct in containers.

```
The Jupyter Notebook is running at:
http://(hostname or 127.0.0.1):8888/?token=
```

minrk:
@takluyver what do you think of this for addressing the default behavior for 5.6?

takluyver:
For the record, I *can* connect to `0.0.0.0` with Firefox on Linux. Not sure which interface that uses, but:

screenshot from 2018-06-21 14-47-07

But this approach seems reasonable to me.

PR #3712, opened 2018-06-25
fix:can't choose same file twice when in upload file

forbxy:
Clear fileinput value. This is required to reset the form. Otherwise, if you upload a file, delete it and try to upload it again, the changed event won't fire.
the old, replacing fileinput did not work as expected

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -338,11 +338,11 @@ define([
                 reader.onerror = reader_onerror;
             }
         });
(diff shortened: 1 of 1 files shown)


takluyver:
Have you tested the old and the new across different browsers? The old code must have worked for someone at some point, so I want to make sure we don't just go from one solution that works for some people to another that works for some other people.

forbxy:
finally,I test it on safair(11.1), firfox(61),chrome(67).the old code dont't work on chrome, new code work on each browser

Issue #2363, opened 2017-04-04
[feature request] add "Pause" button to Notebook


andreas-h:
I was wondering if there is a way to pause/suspend a running Jupyter kernel from the Notebook interface?  I was looking for a "Pause" button but couldn't find any.

Use case:  Imagine you're working on a laptop, and have a long-running computation at work.  Then it would be nice to suspend the kernel's activity when the battery gets drained.

takluyver:
Thanks for filing. I think this should be straightforward on Unix systems, using SIGSTOP (or SIGTSTP if we want to give the kernel the option to ignore it) and SIGCONT. These could be exposed as kernel actions in the REST API, like interrupt and restart already are. I don't know if Windows has an equivalent, but if someone builds a Unix-only solution and leaves the actions as no-ops on Windows, I don't think we'd hold it up.

vidartf:
Maybe relevant for Windows side: https://msdn.microsoft.com/en-us/library/windows/desktop/ms686342(v=vs.85).aspx

Madhu94:
I ll check this out, if no one is working on it.

takluyver:
I don't think anyone is. The place to start will be in the jupyter_client package, adding API functions to pause and continue a kernel's execution.

sunpope:
Hey, I see there hasn't been any update on this in a while. I've been checking out the jupyter_client package and working out where to add the API pause/continue functions. However, could anyone point me towards the right directory to go about adding a new button to the top bar of a notebook? I think the best way to do it is to have a separate button controlling SIGTSTP and SIGCONT and leave the current Run and Interrupt buttons as is.

takluyver:
I probably wouldn't put this in the toolbar for now: it's likely to be Unix only, and I'm not sure how reliably it will work. In terms of the UI, I'd probably start by making it an action (see `actions.js`) so that people can assign it custom shortcuts. But the APIs have to be built before anyone can add UI for it.

sunpope:
Understandable, I'll go there and check it out. Is there a preference between having one action that both suspends the process if it is running AND continues the process if it has been suspended, or two separate actions to achieve the same thing?

takluyver:
It's not a strong preference, but I'd lean towards two separate actions. We can always add a separate toggle action afterwards.

stas00:
Any updates on this one? @sunpope and @Madhu94 - has either of you made any progress on this?

arovit:
[...]

PR #3739, opened 2018-07-05
[WIP] add "Pause" button to Notebook #2363

arovit:
Here is some work on #2363 ( add "Pause" button to Notebook).  It's not 100% complete yet, and once I get green light on this revision, will update completed revision. I was able to get notebook running with "Pause" and "Resume" in the Menubar under "Kernel" selection. I haven't tested with complex code but simple pausing and resuming a "for" loop works. Thing to note - I am using SIGSTOP and SIGCONT signals for this purpose. Looking forward to the feedback. 

TODO-
1. Add test
2. Add proper icon, currently using "fa-stop" icon.

Snapshot -

screen shot 2018-07-05 at 3 31 40 pm


takluyver:
Nice, I'm impressed how little code it is to make this possible. :-)

A few thoughts:

- I'm pretty sure Windows doesn't have these signals; maybe we should disable/hide the options in that case.
- I'd probably put them lower down the menu, maybe in their own little section after 'Shutdown'. I think you'll want them less often than the restart options, for instance. Putting them in their own section also highlights that they go together.
- When the kernel is paused, I'd like to have a persistent notification to that effect, which is clickable to unpause it. Someone's going to accidentally pause the kernel, and it needs to be clear what's happened and how to get out of it. We already have some way of making clickable notifications, so this shouldn't be hard.
- Testing this will be a pain, I suspect.

takluyver:
Oh, and one more:

When we suspend processes in a terminal by pressing Ctrl-Z, that's actually sending SIGTSTP, which has a similar default effect, but can be handled or ignored by the process. I wonder if we should be using that instead of SIGSTOP. It makes the implementation more complex, because you have to somehow check whether pausing the kernel has had any effect, but it may be useful for kernels that know they shouldn't be paused in a certain operation.

Given that it's a new feature, maybe we can keep SIGSTOP (the un-ignorable one) for now, and see if any problems come up, but it's worth thinking about.

[This article](http://www.linusakesson.net/programming/tty/) is an excellent read on things like this.

arovit:
@takluyver Thanks for reviewing and sharing your thoughts. I was thinking to just show either "Pause" or "Resume" button depending on the state of the notebook like a toggle. How do you feel about that ? On persistent notification, I will read more about the clickable notifications in notebook and try to use them here. For testing, I was thinking to run like a print loop and check affects of pause and resume (using selenium).
Regarding choosing between SIGSTOP and SIGTSTP, I used SIGSTOP because for end users, when they press "pause" they expect their job to pause. Using SIGTSTP,  does not guarantee that and users might be left puzzled for why the pausing did not take place (unless we give them some kind of notification of kernel ignoring the pause). But you are right, we can definitely re-visit this if we see any resistance or complaints.
Again, thanks a lot of reviewing and providing great feedbacks.

arovit:
Just pushed the second iteration of the changes. This set of change mostly deals with following - 
1. Toggling between 'pause' and 'resume' kernel options so only one of them appears at any given time.
2. Notification of 'kernel paused' in the kernel notification bar in the event of kernel pause
3. Skip displaying these 'pause' and 'resume' options on Windows platform
Further work pending: 
1.  As I don't have a windows machine handy right now, I would need some help on testing on windows platform (just want to make sure these option don't show)
2. Add couple of selenium based testcases for pausing and resuming.

arovit:
Pushed set of changes for following - 
1. Added a selenium based testcase for 
-  testing 'pause' and 'resume' kernel functionality using a simple for loop.
-  testing kernel notification bar and toggle options.
2. Bringing the pause/resume option next to 'Interrupt' option in Kernel dropdown
-  Felt its better to have 'pause'/'resume' option next to interrupt because of two things - close alphabetical arrangement and severity of the operation.

Please let me know if there is anything more I need to work on. Looking forward to the feedbacks.

arovit:
@takluyver Please let me know if there is anything I can work on to further improve this.

arovit:
Sorry, it's been a long time. @takluyver Can I please get some traction here ?

MrMauricioLeite:
What is holding this PR back?
Is it something we can help in anyway?

kevin-bates:
Sorry for the frustration -  I agree that a response is warranted.  If this PR is "ready", I would suggest resolving conflicts and removing the "WIP" indicator.  Perhaps that would trigger reviews from appropriate maintainers.

This seems generally useful. I'm curious what happens if the user tries to interrupt, restart, or, more importantly, shutdown a paused kernel?  Seems like these operations would behave as if they're ignored relative to the kernel, but the server doesn't know that.  I would think shutdown, for example, would likely remove the kernel from its list of kernels, and be lost for subsequent interactions (yet, still be "running").  Similarly, restart consists of shutdown and startup; so the shutdown wouldn't really happen, while the startup would, essentially leaking a kernel process.

Likewise, what happens when cells are sent to paused kernels?  

Thanks.

kevin-bates:
@arovit - would you mind updating with master to resolve the current conflicts? 

Also, could you describe the behavior you see when attempting to shutdown and restart a paused kernel - per my previous comment?  I suspect the `shutdown_request` message won't get processed, which then triggers the fallback kill sequence from the kernel manager.

Issue #3682, opened 2018-06-13
Add more Prometheus metrics


yuvipanda:
Now that #3490 has been merged, we should add more prometheus metrics to the notebook server!

Some ideas for metrics to add...

1. Number of kernels running (labeled by type)
2. Number of sessions open (not sure if this is useful?)
3. Number of terminals open
4. Mirror of the [activity tracking](https://github.com/jupyter/notebook/issues/1241) metrics
5. Kernel start / stop latency metrics
6. 0mq metrics

Am sure there's more that I don't know of!

yuvipanda:
Can someone with rights mark this as 'good first issue'? :)

dhirschfeld:
I'm interested in RAM - per kernel and total. Not sure if that's already available though?

rgbkrk:
There's some light amount of collaboration around RAM on a kernel level in https://github.com/jupyter/jupyter/issues/264, though it's on a spec level (especially since the actual kernel make be several child processes deep). @ivanov would probably enjoy having a collaborator on the Python side -- I'm looking forward to the UI portion of using it. The notebook server can use it as well since it has access to the messages as it transports them from ZeroMQ to WebSocket.

yuvipanda:
@dhirschfeld @rgbkrk we could possibly also do it from the default Kernel Manager, since it is just spawning local processes and knows how to collect metrics for them (and their children). This lets other kernel managers report their own metrics as they wish, and works across all kernels without any extra work. It would be complimentary to https://github.com/jupyter/jupyter/issues/264.

GoelJatin:
Hi @yuvipanda ,

If no one is working on this, then I would like to take it up.

I would like to start with the very first one for now.

But have a few doubts.

Do we want the number of kernels running at the time when the API is called / keep on collecting them over the complete period of time since the notebook server was started?

Please let me know accordingly.

CC @Madhu94

GoelJatin:
Guys,

Any update on this?

manuhortet:
Hey @GoelJatin, I'm up to work on this with you.

> I'm interested in RAM - per kernel and total. Not sure if that's already available though?

As I think this will be the most useful metric to add, may I take it? :) 

Will try to do it from the default Kernel Manager, as mentioned by @yuvipanda

GoelJatin:
Hey @manuhortet , sure go ahead.

No concerns from my end. :)

LiryChen:
[...]

PR #3743, opened 2018-07-07
[WIP]: #3682, add metrics

GoelJatin:
Issue: #3682 

Creating this PR to address the first requirement listed in the issue.

Will gradually work on the other requirements as well, and waiting for more details from the issue owner on this one, to address it completely.

CC: @Madhu94 , @yuvipanda

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -22,9 +22,9 @@ except ImportError:
     from httplib import responses
     from Cookie import Morsel
 try:
-    from urllib.parse import urlparse # Py 3
+    from urllib.parse import urlparse  # Py 3
 except ImportError:
-    from urlparse import urlparse # Py 2
+    from urlparse import urlparse  # Py 2
 
 from jinja2 import TemplateNotFound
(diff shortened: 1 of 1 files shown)


minrk:
Hi @GoelJatin, thanks for the PR! It looks like an autoformatter has been run on this code. Can you please revert these changes and rebase without them? It is very hard to review a pull request that has a large amount of these changes, since we can't find which changes are to the behavior and which are to the style. Once those are removed, we can get started reviewing the changes here.

An excellent rule of thumb when contributing to projects in general: If running an auto-formatter on the code produces a large change, then the project probably does not use the autoformatter in question and would likely not welcome the changes the autoformatter would produce. Please avoid 'cleanup' pull requests in general unless they have been explicitly requested.

GoelJatin:
Hi @minrk ,

Sure I understand your point.

Will remove the commits for formatting, and update the PR.

Thanks for the advice!

blink1073:
Closing this PR as stale as part of housekeeping, please feel free to re-open if you'd like to continue working on it @GoelJatin.

PR #3759, opened 2018-07-13
flush iopub before sending restarting/dead status messages

minrk:
to ensure (approximately) that all messages from the stopped kernel are delivered before the dead/restarting message

Since this event is triggered after a process has stopped, the only race is with messages waiting in zmq to be handled, so flushing the zmq stream should eliminate the race in all but the most pathological cases.

cc @jasongrout


takluyver:
Test failures in `js/services` look consistent enough to be something real.

minrk:
Yup, I think I mixed something up. Will fix.

blink1073:
Still wip?

minrk:
Should be fixed now

PR #3765, opened 2018-07-14
WIP Add docs for interacting with kernels

jasongrout:
This fills in the gaps for how interacting with a kernel through the notebook is different than interacting with a kernel directly.

CC @minrk

### DIFF
--- a/docs/source/kernels.rst
+++ b/docs/source/kernels.rst
@@ -1,0 +1,79 @@
+Kernels
+=======
+
+When a kernel is created or connected to via the `RESTful
+API <http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/notebook/master/notebook/services/api/api.yaml>`__, the notebook server sets up a websocket to ZeroMQ bridge for communicating with the kernel. On kernel start, the notebook sends a ``request_kernel_info`` message to the kernel in order to initiate communication and to retrieve the kernel message spec version the kernel implements. The notebook automatically adapts messages from the kernel spec version the kernel implements to the kernel spec implemented by the current version of jupyter_client installed.
+
+Restarting or shutting down a kernel should be done with a RESTful request, not through a kernel message, so that the kernel manager can do the appropriate logic around kernel shutdown, like asking the kernel to shut down first through a kernel message, then forcefully shutting down the kernel if there is no response.
+
+
+Kernel messages
+---------------
+
+A websocket client connecting to a kernel through the notebook server websocket bridge will use messages according to the Jupyter kernel message spec, with the modifications noted below.
+
+Kernel Message Specification Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Message Channels**
+
+The notebook server multiplexes all kernel message channels into a single websocket channel. The new top-level ``channel`` key in each websocket message identifies the channel for the message, such as ``'control'``, ``'shell'``, or ``'iopub'``.
+
+.. note::
+
+    TODO If the channel is missing in a message to the notebook server, it is assumed to be the ``shell`` channel.
+
+    Do we want to explicitly document and rely on this behavior, or just make this an undocumented nicety?)
+
+**Kernel Status**
+
+The notebook server sends several kernel status messages (with ``channel: 'iopub'``) beyond the kernel status messages that are sent by the kernel itself:
+
+1. When a kernel is restarted, an ``execution_state: 'restarting'`` kernel status message is sent.
+2. When a kernel dies, an ``execution_state: 'dead'`` kernel status message is sent.
+
+These status messages will have a different message header ``session`` value than the message header session values in actual kernel messages.
+
+.. note::
+
+    TODO
+
+    should we say that the session value of these is the same as the "session_id" in the kernel connection URL, and corresponds (maybe?) to the session id on the server connecting the filename to the kernel.
+
+.. note::
+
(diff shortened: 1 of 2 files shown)


takluyver:
I stuffed the comms documentation under 'User Documentation' because there was nowhere else for it, and you can't make a section for a single document. But now that we've got two, we should look at separating this more technically focused documentation from the pages that explain how to use the notebook.

jasongrout:
Good idea. I was having a difficult time trying to figure out where to put this, and finally decided next to comms docs was the best place.

docs/source/kernels.rst
```diff
@@ -0,0 +1,79 @@
...
+
+**Message Channels**
+
+The notebook server multiplexes all kernel message channels into a single websocket channel. The new top-level ``channel`` key in each websocket message identifies the channel for the message, such as ``'control'``, ``'shell'``, or ``'iopub'``.
+
+.. note::
+
+    TODO If the channel is missing in a message to the notebook server, it is assumed to be the ``shell`` channel.
+
+    Do we want to explicitly document and rely on this behavior, or just make this an undocumented nicety?)
```
minrk: In the spec, I think we can say that the channel key is required, since we are in undocumented-behavior territory right now.

docs/source/kernels.rst
```diff
@@ -0,0 +1,79 @@
...
+1. When a kernel is restarted, an ``execution_state: 'restarting'`` kernel status message is sent.
+2. When a kernel dies, an ``execution_state: 'dead'`` kernel status message is sent.
+
+These status messages will have a different message header ``session`` value than the message header session values in actual kernel messages.
+
+.. note::
+
+    TODO
+
+    should we say that the session value of these is the same as the "session_id" in the kernel connection URL, and corresponds (maybe?) to the session id on the server connecting the filename to the kernel.
```
minrk: > should we say that the session value of these is the same as the "session_id" in the kernel connection URL

We could do this, but I'm not sure we should commit to it. I think it's mostly an accident that it happens to be this way. We may come up with a better way to signal that a message came 'from the notebook server' than using the client's own session_id in the future. We could mention it as a current implementation detail, though.

>  corresponds (maybe?) to the session id on the server connecting the filename to the kernel.

It definitely doesn't. session_id is the per-client-connection id used in message headers, not related to the kernel:document mapping.

docs/source/kernels.rst
```diff
@@ -0,0 +1,79 @@
...
+
+If a message has one or more associated binary buffers, the message and the buffers are serialized to a binary websocket frame. This frame starts with the following header:
+
+* 4 bytes: number of frame parts (``parts``) as a big-endian unsigned 32-bit integer
+* ``4 * parts`` bytes: offset for each frame part as a big-endian unsigned 32-bit integer. Offsets are from the start of the frame, including the header.
+
+The first frame part is the message serialized to JSON and encoded as UTF-8. The rest of the frame parts are the binary buffers associated with the message.
+
+.. note::
+    Should we explicitly say how we are serializing dates?
```
minrk: Since the message is already JSON, it's part of the message spec that dates are ISO8601 strings. At the binary serialization level, everything is JSONable, so just number,str,list,dict. Date handling would be occurring at the *library* level, e.g. ipywidgets/ipykernel, but not in the message spec.

I think we do already have a note that dates are ISO8601 in the kernel messaging doc

docs/source/kernels.rst
```diff
@@ -0,0 +1,79 @@
...
+If all websocket clients have disconnected from a kernel, the notebook server will temporarily buffer messages from the kernel to be delivered to the first websocket client that connects to the kernel.
+
+.. note::
+    In the classic notebook client and JupyterLab, requesting a kernel restart immediately closes all websocket connections to the kernel, so kernel buffering starts. When a new websocket connection is created connecting to the kernel, the notebook server transmits all of the messages buffered from the kernel. For the IPython kernel, this means the new websocket connection will start with receiving status busy, shutdown_reply, and status idle messages on the iopub channel from before the restart.
+
+.. note::
+
+    TODO
+    
+    Document the session URL parameter used in kernel connections. Is that created every time we request a kernel with a post request? Is it tied to just creating new sessions with the session rest api?
```
minrk: Yes, we should document it. sessions_id is associated with each specific client. It's entirely unrelated to the session rest API. The session_id here is the message protocol `header.session` field, which is used on replies and iopub to identify the original sender of the message. The main *important* use of this session_id is setting zmq.IDENTITY on reconnects to ensure that replies intended for this client still arrive there after reconnects. It's also used to build messages created in the notebook server for this client, but I'm not sure we should commit to or rely on that yet.

echarles:
This is a nearly 4 years PR. Are the proposed changes still applicable?

jasongrout:
> This is a nearly 4 years PR. Are the proposed changes still applicable?

Yes, very much so. https://github.com/jupyter/jupyter_client/pull/388 was merged long ago, and this PR has documentation that was taken out of the docs there and put here as a more fitting place (for example, the kernel restarting messages on the websocket).

jasongrout:
> Yes, very much so. [jupyter/jupyter_client#388](https://github.com/jupyter/jupyter_client/pull/388) was merged long ago, and this PR has documentation that was taken out of the docs there and put here as a more fitting place (for example, the kernel restarting messages on the websocket).

However, looking at the documentation in the main branch, for notebook 7, I think this content belongs in the jupyter server docs now, where the websocket connection is actually handled.

jasongrout:
I opened up a PR in Jupyter Server: https://github.com/jupyter-server/jupyter_server/pull/869

echarles:
Thank you @jasongrout

Issue #3754, opened 2018-07-12
"ValueError: 'fe80::1%lo0' does not appear to be an IPv4 or IPv6 address" error when running notebook


gnestor:
Follow up from https://github.com/jupyter/notebook/pull/3751#issuecomment-404515819:

I'm getting a strange error when trying to run the notebook against master:

```
$ jupyter notebook
[W 13:07:55.769 NotebookApp] server_extensions is deprecated, use nbserver_extensions
[I 13:07:55.804 NotebookApp] It looks like you're running the notebook from source.
        If you're working on the Javascript of the notebook, try running
    
        npm run build:watch
    
        in another terminal window to have the system incrementally
        watch and build the notebook's JavaScript for you, as you make changes.
Traceback (most recent call last):
  File "/Users/grant/anaconda/lib/python3.6/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'allow_remote_access'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/grant/Sites/jupyter/notebook/notebook/notebookapp.py", line 856, in _default_allow_remote
    addr = ipaddress.ip_address(self.ip)
  File "/Users/grant/anaconda/lib/python3.6/ipaddress.py", line 54, in ip_address
    address)
ValueError: 'localhost' does not appear to be an IPv4 or IPv6 address

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/grant/anaconda/bin/jupyter-notebook", line 11, in <module>
    load_entry_point('notebook', 'console_scripts', 'jupyter-notebook')()
  File "/Users/grant/anaconda/lib/python3.6/site-packages/jupyter_core/application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/Users/grant/anaconda/lib/python3.6/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-7>", line 2, in initialize
  File "/Users/grant/anaconda/lib/python3.6/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/Users/grant/Sites/jupyter/notebook/notebook/notebookapp.py", line 1598, in initialize
    self.init_webapp()
[...]

PR #3766, opened 2018-07-14
Disable the host check for 5.6 because of gh-3754

takluyver:
I intend to re-enable this again straight after the release, but for now we want to get 5.6 out without breaking things for lots of Mac users.


gnestor:
Confirmed that this resolves https://github.com/jupyter/notebook/issues/3754 👍

takluyver:
To be clear, this is a workaround so I could get a release out quickly - it disables the check for everyone, so I definitely want to find a better solution before the next release.

PR #3767, opened 2018-07-14
Re-enable Host header check after 5.6

takluyver:
We need to figure out the Mac issues in #3754, but I want to ensure this doesn't linger indefinitely.


takluyver:
OKed by @gnestor

raabf:
This is causing problems again, see #3946

PR #3768, opened 2018-07-14
adding selenium to test requirements

munkm:
I noticed when I tried to run the tests for the notebook that selenium wasn't in the requirements for the test environment, so I got a selenium error.  I added selenium to the test list in the extras_require dict.


mpacer:
Looks good!

Issue #3772, opened 2018-07-16
Removing the play button introduced in 5.6.0?


ivanov:
I understand the rationale of trying this feature out, but I worry that it becomes a feature that if we keep this change now we will have a harder time deprecating it in the future. We don't have enough information about it to keep it as the new default, and in a day of usage, I find it to be visual noise and frustrating. 

One particular problem with the current implementation is target area for executing a cell in place fills the entire vertical column, but the icon is only at the top.

Alternatives paths forward could be to provide this as an option, or to patch the release taking this feature out and put it back in in `master` to let it bake for longer. 

I feel that a UI change like this is significant and I think that the cost to current users who do not need it is too great for the benefit it provides to those who want such a feature.

takluyver:
I imagine you can use some custom CSS to hide them. I don't buy that there's a significant cost to people who're not using it.

jasongrout:
I've used this a short time, and I find the hover behavior *very* distracting - as my mouse moves over the notebook, the apparent movement in the gutter draws my eye to the play button that keeps moving around. 

Also, the fact that the play button follows the mouse hover instead of the current cell is very confusing. If I leave my mouse over a cell, then move around with the keyboard, the play button stays on the cell hovered over, even when the mouse pointer disappears - then it seems like there is this arbitrary play button next to one of the cells, which isn't the current cell, but does change as I scroll the notebook in the page.

That's not to say that we can't fix these sorts of issues, of course. However, I'd definitely be +1 on removing this from the current released notebook and letting this behavior bake more in master, with wider UX testing before committing to it.

takluyver:
Sorry to have hurt you. I'm frustrated by this issue because:

* The feature was worked on in two pull requests which no-one besides me reviewed (#3437, #3535).
* I merged it nearly two months ago, and I'm not aware of any complaints while it was in master.
* M improved on it in a third pull request, still without any real review (#3687).
* We made a release candidate a few weeks back. No feedback.
* I find the time to get a real release out, and within hours I've got a complaint saying it should be reverted.

It feels like there's a huge imbalance between how many people care about the notebook and how many people are willing to be involved in the development process.

Furthermore, this feels like the kind of complaint provoked by any UI change: it's not what we're used to, we'll find a reason that it's objectively bad, and then demand that it be reverted. The fact that it comes so soon after a release reinforces this impression.

[...]

PR #3776, opened 2018-07-17
Update styles for inline run button

gnestor:
Closes https://github.com/jupyter/notebook/issues/3772
This does not display the inline run button on desktop (non-mobile) because it has proven to be distracting and not really useful for desktop.

It also makes the input prompt and run button inline so that they don't take up 2 separate lines.

Before: 

image

After:

image

### DIFF
--- a/notebook/static/notebook/js/codecell.js
+++ b/notebook/static/notebook/js/codecell.js
@@ -161,6 +161,8 @@ define([
         var input = $('<div></div>').addClass('input');
         this.input = input;
 
+        var prompt_container = $('<div/>').addClass('prompt_container');
+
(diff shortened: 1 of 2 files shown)


minrk:
In testing this PR, I noticed a weird alignment issue:


screen shot 2018-07-17 at 22 17 31

As cells are executed, the prompt changes width.

gnestor:
I'm seeing this in master too:

image

gnestor:
I just pushed a fix that should solve the problem as long as the execution count is under 10,000 😺

image

takluyver:
I didn't see the misalignment on this, but while I was checking, I did discover a misalignment between code and markdown cells: #3781.

PR #3778, opened 2018-07-18
[FIX] notebookapp, auth: `get_secure_cookie` kwargs

beledouxdenis:
Per Tornado's documentation:

>By default, Tornado’s secure cookies expire after 30 days.
>To change this, use the expires_days keyword argument to
>set_secure_cookie and the max_age_days argument to get_secure_cookie.
>These two values are passed separately so that you may
>e.g. have a cookie that is valid for 30 days for most purposes,
>but for certain sensitive actions
>(such as changing billing information)
>you use a smaller max_age_days when reading the cookie.

With the current implementation in `auth/login.py`,
this is possible to pass the `expires_days` option
but not possible to enforce it as this is not possible
to pass `max_age_days` to `get_secure_cookie`

This makes impossible to set the cookie expiration without
using a custom `LoginHandler`.

This revision is about adding the possibility to pass options
to Tornado's `get_secure_cookie` method,
so it can be possible to set the cookies expiration,
among others.

In my opinion,  `get_cookie_options` is a weird naming given the options to pass to `set_secure_cookie` are called `cookie_options`. That said, I am not sure what is your policy regarding the retro-compatibility of  the settings, if `cookie_options` can be renamed `set_cookie_options` or not. Anyway we first have to discuss the feasibility of this change.

### DIFF
--- a/notebook/auth/login.py
+++ b/notebook/auth/login.py
@@ -168,7 +168,8 @@ class LoginHandler(IPythonHandler):
             return handler._user_id
         user_id = cls.get_user_token(handler)
         if user_id is None:
-            user_id = handler.get_secure_cookie(handler.cookie_name)
+            get_secure_cookie_kwargs  = handler.settings.get('get_secure_cookie_kwargs', {})
+            user_id = handler.get_secure_cookie(handler.cookie_name, **get_secure_cookie_kwargs )
         else:
             cls.set_login_cookie(handler, user_id)
             # Record that the current request has been authenticated with a token.
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -945,6 +945,10 @@ class NotebookApp(JupyterApp):
         help=_("Extra keyword arguments to pass to `set_secure_cookie`."
              " See tornado's set_secure_cookie docs for details.")
     )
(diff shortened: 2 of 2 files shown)


minrk:
`get_cookie_options` is a little funky, since I would expect it to be a callable itself on first glance. For passthrough to other APIs, we often give configurables names like `get_secure_cookie_kwargs` to communicate "these are kwargs that will be passed unmodified to get_secure_cookie." The other option is to expose just a single Float configurable for `cookie_max_age_days`, since there aren't as many useful options to expose for get_cookie as there are for set_cookie.

beledouxdenis:
Thank you for your review :)

I updated the PR.

I chose the `get_secure_cookie_kwargs` option: To be future-proof in case Tornardo adds new arguments or in case someone is interested by the already existing `min_version` argument.
http://www.tornadoweb.org/en/stable/_modules/tornado/web.html#RequestHandler.get_secure_cookie

That said, for my own purposes, I am indeed only interested to the `max_age_days` argument. Therefore, if you rather like the `cookie_max_age_days` naming, just let me know and I will update the PR.

minrk:
Thanks!

beledouxdenis:
Thank you for having considered the PR : )

PR #3789, opened 2018-07-20
modify contributing instructions to use pytest for testing

munkm:
At the scipy sprints it was suggested that I use pytest rather than nose to run the jupyter notebook tests. I had been following the contributing documentation that directs using nose. This PR
- modifies the contributing instructions to use pytest when running the notebook tests
- adds pytest and pytest-cov to the the test key in extras_require so that pytest and the coverage package are included in the test dependencies.


minrk:
Thanks!

PR #3793, opened 2018-07-24
Implement password hashing with bcrypt

### LINKED ISSUE #2426: default password hash of sha-1 is insecure
see https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html

would be better to use sha256, or even better something like scrypt or bcrypt (which are actually meant for password hashing)

remram44:
Fixes #2426

The `hashlib` hashing method could probably be removed from `passwd()` since nothing ever specifies the `algorithm` argument. Should probably be kept in `passwd_check()` until people update their configs.

### DIFF
--- a/notebook/auth/security.py
+++ b/notebook/auth/security.py
@@ -21,7 +21,7 @@ from jupyter_core.paths import jupyter_config_dir
 salt_len = 12
 
 
-def passwd(passphrase=None, algorithm='sha1'):
+def passwd(passphrase=None, algorithm='bcrypt'):
     """Generate hashed password and salt for use in notebook configuration.
 
     In the notebook configuration, set `c.NotebookApp.password` to
@@ -59,11 +59,17 @@ def passwd(passphrase=None, algorithm='sha1'):
         else:
             raise ValueError('No matching passwords found. Giving up.')
 
-    h = hashlib.new(algorithm)
-    salt = ('%0' + str(salt_len) + 'x') % random.getrandbits(4 * salt_len)
-    h.update(cast_bytes(passphrase, 'utf-8') + str_to_bytes(salt, 'ascii'))
+    if algorithm == 'bcrypt':
+        import bcrypt
+        h = bcrypt.hashpw(cast_bytes(passphrase, 'utf-8'), bcrypt.gensalt())
 
-    return ':'.join((algorithm, salt, h.hexdigest()))
+        return ':'.join((algorithm, cast_unicode(h, 'ascii')))
+    else:
+        h = hashlib.new(algorithm)
+        salt = ('%0' + str(salt_len) + 'x') % random.getrandbits(4 * salt_len)
+        h.update(cast_bytes(passphrase, 'utf-8') + str_to_bytes(salt, 'ascii'))
+
+        return ':'.join((algorithm, salt, h.hexdigest()))
 
 
 def passwd_check(hashed_passphrase, passphrase):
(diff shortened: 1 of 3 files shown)


nealmcb:
Thank you - I agree that the current non-iterated use of SHA-1 is awful.
But as I noted at https://github.com/jupyter/notebook/issues/2426#issuecomment-529151833, I think `hashlib.pbkdf2_hmac` which is already in standard Python might be more appropriate.

I also think it is now clear that Argon2 is better than bcrypt, and I wonder if using it (and adding a dependency on it) instead of bcrypt might be more appropriate now.

Or maybe Argon2 will come to the standard Python3 library soon?
Anyway, thanks!

remram44:
I'm happy to make further changes and look into stdlib facilities if/when a project member expresses interest in this.

blink1073:
Hi @remram44, we're catching up on PR review.  I'd support the use of `hashlib.pbkdf2_hmac` as a compromise that improves the hashing without taking on an external dependency.   Are you able to pick this work back up?

nealmcb:
I wonder if scrypt is appropriate by now.  scrypt is significantly better than pbkdf2, since it is memory-hard, and is supported in Python 3.6 ([scrypt](https://docs.python.org/3/library/hashlib.html#hashlib.scrypt)), and requires OpenSSL 1.1+

I also note this interesting reflection on argon2: https://twitter.com/Sc00bzT/status/1149950559258140673

blink1073:
I'll defer to your expertise here @nealmcb.  Given that argon2 is available on [conda-forge](https://anaconda.org/conda-forge/argon2_cffi) I'd support taking it on as a dependency.

nealmcb:
With either scrypt or argon2, I guess the next question is how much memory overhead we're comfortable with. The default `memory_cost` is 105 MB (102400 KiB, from the handy `python -m argon2` command). And the parameter tuning tips at the latest IETF internet-draft on the topic 

* [draft-irtf-cfrg-argon2-10 - The memory-hard Argon2 password hash and proof-of-work function](https://tools.ietf.org/html/draft-irtf-cfrg-argon2-10#section-4)

currently emphasize  maximizing memory (*3.  Figure out the maximum amount m of memory that each call can afford.*)

But the initial size of Jupyter seems more like 50 MB, based on my laptop running Ubuntu 20.04, if I'm interpreting the output of `smem` correctly. After startup that is likely to grow, but do we want logging in to always take twice the initial size?

What is the maximum (momentary) extra memory hit you're comfortable with? I guess login most often happens at startup, but if the system is really short on memory, you'd hate to cause problems for users trying to log in to a notebook to gracefully shut it down.

So perhaps we go with a smaller value like `memory_cost=10240`, and a correspondingly higher number of iterations, like `time_cost=18` (which takes the same time per password verification as the default parameters on my machine: ~100 ms).

We could also consider different values for the maximum number of threads to use, or pick a different maximum real time.

remram44:
This can easily be updated to another method, if a consensus is reached.

blink1073:
@remram44 if you make these changes I'll go ahead and merge: https://github.com/jupyter/notebook/issues/2426#issuecomment-650399655

remram44:
I rebased, and switched to argon2-cffi as requested.

A point that was mentioned by @nealmcb was making the Argon2 parameters configurable via `jupyter_notebook_config.py`, however I am not sure that loading that config file in `passwd()` is appropriate. Also I am a little confused about the different approaches in dealing with the config file ([this](https://github.com/jupyter/notebook/blob/0df10dee3ee7e8c0f56f879b69d37070fccca1c7/notebook/auth/security.py#L125-L134) vs [this](https://github.com/jupyter/notebook/blob/0df10dee3ee7e8c0f56f879b69d37070fccca1c7/notebook/auth/__main__.py#L21-L26)).

blink1073:
It looks like `auth/__main__.py` should be calling out to `auth/secruity.py` (needs a refactor).
I think we can take up optimizations in further PRs, let's merge and iterate.

CaselIT:
Any reason this was not also added to jupyter_server and ipython that have the same implementation of these functions?

remram44:
I don't think jupyter_server existed when I opened this PR. It took 2 years to merge.

nealmcb:
Thanks to everyone for getting this wrapped up last year.
It might help to change the title line of this issue, to clarify that we ended up using argon2_cffi, not bcrypt. I guess perhaps only the owner can do that?

PR #3799, opened 2018-07-26
Fill in Checkpoints section

aaronchall:
(I think it needs more complete examples like ContentsManager has regarding what is returned by these methods.)
Created as per https://gitter.im/jupyter/notebook?at=5b50fad79ddf5f4aadf6fa42


minrk:
Thanks!

PR #3809, opened 2018-07-31
ip_address only accepts unicode on Python 2

minrk:
ipaddress.ip_address('127.0.0.1') fails with ValueError on Python 2

need to decode it, otherwise 127.0.0.1 won't be treated as local

workaround in the meantime should be to use 'localhost' instead of '127.0.0.1'

closes https://github.com/jupyter/nbdime/issues/408

cc @vidartf

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -35,7 +35,7 @@ from notebook._sysinfo import get_sys_info
 
 from traitlets.config import Application
 from ipython_genutils.path import filefind
-from ipython_genutils.py3compat import string_types
+from ipython_genutils.py3compat import string_types, PY3
 
 import notebook
 from notebook._tz import utcnow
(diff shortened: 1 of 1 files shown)


notebook/base/handlers.py
```diff
@@ -427,11 +427,15 @@ def check_host(self):
...
+        if not PY3:
+            # ip_address only accepts unicode on Python 2
+            host = host.decode('utf8', 'replace')
+
         try:
             addr = ipaddress.ip_address(host)
         except ValueError:
             # Not an IP address: check against hostnames
-            allow = host in self.settings.get('local_hostnames', [])
+            allow = host in self.settings.get('local_hostnames', ['localhost'])
```
minrk: additionally, added 'localhost' as the default here so that it is defined by default when used outside NotebookApp, e.g. nbdime.

Issue #3769, opened 2018-07-14
Unknown exporter "ipynb": Did you mean "notebook"?


rsignell-usgs:
When I do  `File=>Download as=>Notebook` from my JupyterHub, the notebook downloads fine, but I get a "500 : Internal Server Error" displayed on the screen, and in the log, I see:
```python-traceback
File "/opt/conda/lib/python3.6/site-packages/nbconvert/exporters/base.py", line 110, in get_exporter
        % (name, ', '.join(get_export_names())))
    ValueError: Unknown exporter "ipynb", did you mean one of: asciidoc, custom, html, latex, markdown, notebook, pdf, python, rst, script, slides?
```
I notice that someone else had the [same problem on Stackoverflow](https://stackoverflow.com/q/51014351/2005869). 

Here's my conda package list: 
<details>

```
# packages in environment at /opt/conda:
#
# Name                    Version                   Build  Channel
absl-py                   0.2.2                      py_0    conda-forge
affine                    2.2.1                      py_0    conda-forge
alembic                   0.9.9                      py_0    conda-forge
appdirs                   1.4.3                      py_1    conda-forge
asciitree                 0.3.3                    py36_1    conda-forge
asn1crypto                0.24.0                   py36_0    conda-forge
astor                     0.6.2                      py_0    conda-forge
async_generator           1.9                           0    conda-forge
attrs                     18.1.0                     py_1    conda-forge
automat                   0.7.0                    py36_0    conda-forge
backcall                  0.1.0                      py_0    conda-forge
beautifulsoup4            4.6.0                    py36_0    conda-forge
blas                      1.1                    openblas    conda-forge
bleach                    1.5.0                    py36_0    conda-forge
blinker                   1.4                        py_1    conda-forge
blosc                     1.14.0                        1    conda-forge
bokeh                     0.12.16                  py36_0    conda-forge
boost                     1.66.0                   py36_1    conda-forge
boost-cpp                 1.66.0                        1    conda-forge
boto3                     1.7.48                     py_0    conda-forge
botocore                  1.10.48                    py_0    conda-forge
bottleneck                1.2.1            py36h7eb728f_1    conda-forge
bzip2                     1.0.6                         1    conda-forge
ca-certificates           2018.4.16                     0    conda-forge
cachetools                2.0.1                      py_0    conda-forge
cairo                     1.14.10                       0    conda-forge
cartopy                   0.16.0                   py36_0    conda-forge
[...]

PR #873, opened 2015-12-15
Add sidebar box in doc's index for 4.1.x

willingc:
Adds a release link to changelog.rst and a sidebar box to index.rst. Displays as below:
screen shot 2015-12-14 at 8 16 37 pm

PR #3832, opened 2018-08-08
Download and export

actionshrimp:
This is a fix for #3769.

Previously the 'Download as > Notebook (ipynb)' menu item was hooked up to two event handlers, one which downloaded the current file 'as is' off disk (and seemed to work ok), and one which tried to pass it through `nbconvert` as an `ipynb` file (derived from the menu item's ID `download_ipynb`), and displayed a 500.

I've instead added a new menu item 'Download' which downloads the file 'as is', and fixed the ID as `export_notebook` to align with what `nbconvert` expects (and renamed the `download` menu markup items to `export` to align with the `exporters` that are enumerated at the bottom).

Even though for `ipynb` files the behaviour should be similar, if you've got a custom ContentsManager or something installed you might be editing a different file on disk to an actual ipynb file, so you may want to be able to choose the behaviour you want.

### DIFF
--- a/notebook/static/base/js/utils.js
+++ b/notebook/static/base/js/utils.js
@@ -1183,6 +1183,17 @@ define([
         document.head.appendChild(link);
     };
 
+    var file_extension = function(path){
+        /**
+         *  return the last pat after the dot in a filepath
+         *  or the filepath itself if no dots present.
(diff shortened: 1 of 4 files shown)


blink1073:
This behavior was fixed in #873.  Thanks!

PR #3846, opened 2018-08-10
Load notebook after extensions have loaded.

dalejung:
Is it possible that we can enforce the notebook loading after extensions have run?

I run this branch locally because I have a nbextension that needs to modify the notebook js classes before they're instantiated.

I might be the only one i need of this but I figure it wouldn't hurt to ask :/


blink1073:
I don't think we want to make this substantive of a change, since an extension could take arbitrarily long to load, blocking the document from loading.  I think instead you could wait on an event to fire.

PR #3878, opened 2018-08-24
Fix for broken/permission denied symlinks

scanon:
This is an alternate fix for #3151.  This will still show the broken link.
If the user clicks on the link, they will see the standard permission denied error.
A more complete fix would some how disable the click or add some icon for it.
But this will address the problem where the user doesn't see any listing due
to a single broken link in a directory.

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -298,7 +298,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
 
         if not os.path.isdir(os_path):
             raise web.HTTPError(404, four_o_four)
-        elif is_hidden(os_path, self.root_dir) and not self.allow_hidden:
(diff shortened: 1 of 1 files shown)


minrk:
I think you mentioned adding a test for this. Can you add a test that includes both a broken symlink and a good symlink with a hidden name? e.g. if `~/.ssh` is a symlink, it should trigger this hidden logic, and a broken symlink should avoid causing problems for its parent, at least.

flying-sheep:
I assume this will fix cases like mine: https://github.com/jupyterlab/jupyterlab/issues/5866#issue-399275656

scanon:
I'm going to close this since I'm not convinced this is addressing the real issue for us.  I have another fix that is for recursive symlinks (#4669) that I think was the real culprit for us.

PR #3879, opened 2018-08-25
Get the list of exporters from entrypoints

starcruiseromega:
`exporter_map` is deprecated, so let's use the list of exporters fetched
from the installed entrypoints.

There's a supposed attribute `export_from_notebook` that should be set
to a friendly string name if the exporter should be exposed in the
front-end. However, the exporters defined in `nbconvert` don't have it
set, so I haven't used it to determine inclusion in the list. Instead,
I've used the entrypoint name as the friendly name, which looks like it
was the intention from the way they are named.

I ran the unit tests and tried starting up the notebook server and
accessing the API endpoint to verify the JSON looked correct.

### DIFF
--- a/notebook/services/nbconvert/handlers.py
+++ b/notebook/services/nbconvert/handlers.py
@@ -4,18 +4,32 @@ from tornado import web
 
 from ...base.handlers import APIHandler
 
+
 class NbconvertRootHandler(APIHandler):
 
     @web.authenticated
     def get(self):
         try:
-            from nbconvert.exporters.export import exporter_map
+            from nbconvert.exporters import base
         except ImportError as e:
             raise web.HTTPError(500, "Could not import nbconvert: %s" % e)
         res = {}
-        for format, exporter in exporter_map.items():
-            res[format] = info = {}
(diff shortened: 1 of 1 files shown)


starcruiseromega:
The JSON from my test run at http://localhost:8888/api/nbconvert looks like this:

```json
{
	"asciidoc": {
		"output_mimetype": "text/asciidoc"
	},
	"custom": {
		"output_mimetype": ""
	},
	"html": {
		"output_mimetype": "text/html"
	},
	"latex": {
		"output_mimetype": "text/latex"
	},
	"markdown": {
		"output_mimetype": "text/markdown"
	},
	"notebook": {
		"output_mimetype": "application/json"
	},
	"pdf": {
		"output_mimetype": "text/latex"
	},
	"python": {
		"output_mimetype": "text/x-python"
	},
	"rst": {
		"output_mimetype": "text/restructuredtext"
	},
	"script": {
		"output_mimetype": ""
	},
	"slides": {
		"output_mimetype": "text/html"
	}
}
```

starcruiseromega:
@ivanov @mpacer @dharmaquark

Issue #3781, opened 2018-07-19
Left edge of code and markdown cells misaligned when cell toolbar visible


takluyver:
Turning on any cell toolbar highlights an alignment issue at the left-hand side of code cells and markdown cells:

Firefox:

![screenshot from 2018-07-19 09-43-10](https://user-images.githubusercontent.com/327925/42928728-66520d64-8b38-11e8-8175-e794d1acc944.png)

Chromium:

![screenshot from 2018-07-19 09-45-22](https://user-images.githubusercontent.com/327925/42928771-8357de34-8b38-11e8-84f6-4dabe5300676.png)

PR #3913, opened 2018-09-13
fix uneven input/output prompt alignment

minrk:
Something got a little wonky with prompt and output alignment (see #3781 and #3784). This is master:

screen shot 2018-09-13 at 16 34 21

and this is with this PR, where everything is back where it belongs:

screen shot 2018-09-13 at 16 34 49


minrk:
This was actually fixed by #3772. This PR now just makes it harder to get into this situation by removing the separate sources for input and output prompt width, since they must always be the same.

PR #3925, opened 2018-09-15
Try rebuild translation js

Carreau:
Having translations in the core if we don't package them does not make much sens IMHO. 

THis is an attempt at fixing that (need to figure out if we should unconditionally depend on Babel to build the SDist on user machine).


Carreau:
Thanks !

PR #3942, opened 2018-09-19
[WIP] use Accept-Language header in jinja templates

minrk:
solves translation issue where jinja used server-side LANG environment and javascript used browser configuration via the Accept-Language header, which could result in multiple languages on the page. Now, all browser messages should be consistent with the browser settings.

Introduces caching of gettext translation objects to ensure that we don't keep loading the translation config on every request.

cc @JCEmmons

TODO:

- selenium test with Accept-Language header to verify that it's loading the right thing (verified manually, but we should have a real test)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ __pycache__
 .coverage
 .pytest_cache
 src
+.pytest_cache
 
 *.swp
 *.map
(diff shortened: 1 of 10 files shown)


takluyver:
This has a merge conflict now, if you have time to rebase it.

jtpio:
Closing as stale.

Please feel free to continue in https://github.com/jupyter-server/jupyter_server if still needed, thanks!

PR #4030, opened 2018-09-24
Prompt for folder name when creating a new folder

### LINKED ISSUE #3928: Prompt for folder name when creating a new folder
On the server homepage, clicking **New -> Folder** creates a new folder called 'Untitled Folder' in the current path.

jupyter

It would be useful to have a prompt ask for folder name before creating. I would like to work on it if possible.

gustavoefeiche:
**Description**
resolve #3928 

On the server homepage, prompt for the folder name when creating a new folder instead of creating one with the name 'Untitled Folder'.

Currently the name is set using the already existing method `new_untitled`, which previously would create a new untitled {folder | file | notebook}. It does not make much sense, since we are setting a new name to the folder and, well, it is not untitled anymore. What do you suggest? Do I create a new function? Or rename the existing one?

**EDIT 2018-10-16**
Renamed new_untitled to new_entity and rebuilt naming logic

**Modified Files**
- `notebook/services/contents/handlers.py`
- `notebook/services/contents/manager.py`
- `notebook/static/services/contents.js`
- `notebook/static/tree/js/notebooklist.js`
- `notebook/static/edit/js/menubar.js`
- `notebook/static/notebook/js/kernelselector.js`
- `notebook/static/tree/js/newnotebook.js`

**To Do**
- [x] **Create new function instead of using `new_untitled`. (?)**
- [x] Show Modal DIalog when prompting for folder name instead of native browser prompt.
- [x]  Tests

### DIFF
--- a/notebook/services/contents/handlers.py
+++ b/notebook/services/contents/handlers.py
@@ -148,10 +148,10 @@ class ContentsHandler(APIHandler):
         self._finish_model(model)
     
     @gen.coroutine
-    def _new_untitled(self, path, type='', ext='', title=''):
-        """Create a new, empty untitled entity"""
+    def _new_entity(self, path, type='', ext='', title=''):
+        """Create a new, empty entity"""
         self.log.info(u"Creating new %s in %s", type or 'file', path)
-        model = yield gen.maybe_future(self.contents_manager.new_untitled(path=path, type=type, ext=ext, title=title))
+        model = yield gen.maybe_future(self.contents_manager.new_entity(path=path, type=type, ext=ext, title=title))
         self.set_status(201)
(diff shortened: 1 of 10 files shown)


minrk:
Thanks for the PR! So sorry for not reviewing this for so long. This shouldn't have needed a change to the Python API at all I believe (and this contains backward-incompatible changes such as removing the `new_untitled` method). We already have a "new entity" method, it's called `new` and we have an API for it (PUT to a path). I think this can be replaced with purely a javascript change to use PUT after prompting instead of any of the `new_entity` changes.

blink1073:
Closing this PR as out of date.  Please feel free to re-open @gustavoefeiche if you'd like to pick it back up!

PR #4034, opened 2018-09-26
When adapting also log the version jupyter_client is expecting.

Carreau:
It's still (IMHO) unclear how the adaptation goes. Like does "adapting
to protocol X.Y.Z." mean the end-result is X.Y.Z or what you
got is X.Y.Z and it's X.Y.W after adaptation.


blink1073:
What if we said instead "adapting from protocol version XXX to YYY"?  I agree that the message is confusing, I had no idea it meant which version we were adapting from.

Carreau:
> "adapting from protocol version XXX to YYY"

Yeah, even that is ambiguous 

`adapting from protocol version XXX (kernel) to YYY (client)` ?

blink1073:
Sounds good!

Carreau:
I've tried something from the github UI, let's see if it works.

jasongrout:
Thanks! That message looks a lot more understandable to me.

PR #4036, opened 2018-09-26
Added metrics for currently running terminals and labeled by type kernels

Hyaxia:
Moved the 'metrics' file to a new package named 'promethues'.
Then split the file into two files, one containing the metrics themselves, the other containing any function that have something to do with prometheus.

Finally, Added new metrics into the prometheus.metrics file that represent the number of running terminals and added the functionality for that metric to be recorded in the terminal.api_handlers file.

relevant issue #3682

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ before_install:
         tar -xzf geckodriver-v0.19.1-linux64.tar.gz -C geckodriver
         export PATH=$PATH:$PWD/geckodriver
       fi
+    - pip install "attrs>=17.4.0"
 
 install:
     - pip install --pre .[test]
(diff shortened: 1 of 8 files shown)


Hyaxia:
Its wierd that it failed...
All i have changed way the python code and I am able to run the jupyter notebook perfectly fine...
All of the errors say that it has failed running `python -m notebook.jstest notebook`, no idea why that is...
Can anyone explain?

Hyaxia:
Feels like the problem is not with my code but rather the way i uploaded the changes...
Can anybody explain me how to do it in the right way? there is no reason for this error to occur because of my code... (or atleast i cant find one)

Carreau:
I've pushed the fix of #4039 to your branch to hopefully fix the test suite.

blink1073:
LGTM, thanks!

Issue #4038, opened 2018-09-26
PR tests fail because of jsonschema


Hyaxia:
Hi everyone.
I have submitted a PR and waited for the tests to complete - #4036
The tests on travis failed, resulting in a long output that repeats itself saying:
```
The command "symlinks=$(find . -type l| grep -v './node_modules/' | grep -v './git-hooks')
if [[ $(echo $symlinks) ]]; then
    echo "Repository contains symlinks which won't work on windows:"
    echo $symlinks
    echo ""
    false
else
    true
fi
" exited with 0.
5.04s$ if [[ $GROUP == js* ]]; then travis_retry python -m notebook.jstest ${GROUP:3}; fi
Test group: base
Notebook failed to start: 
['/home/travis/virtualenv/python3.6.3/bin/python', '-m', 'notebook', '--no-browser', '--notebook-dir', '/tmp/tmpfsrwphhz', '--NotebookApp.token=', '--NotebookApp.base_url=/a@b/', '--KernelManager.transport=ipc']
Traceback (most recent call last):
  File "/opt/python/3.6.3/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/python/3.6.3/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/notebook/__main__.py", line 4, in <module>
    from notebook import notebookapp as app
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/notebook/notebookapp.py", line 82, in <module>
    from .services.contents.manager import ContentsManager
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/notebook/services/contents/manager.py", line 17, in <module>
    from nbformat import sign, validate as validate_nb, ValidationError
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/nbformat/__init__.py", line 33, in <module>
    from .validator import validate, ValidationError
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/nbformat/validator.py", line 12, in <module>
    from jsonschema import ValidationError
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/jsonschema/__init__.py", line 21, in <module>
    from jsonschema._types import TypeChecker
[...]

PR #4039, opened 2018-09-26
try to fix tests

Carreau:
Json schema pre seem to require attrs>17.3.0. 

So pin it in tests. 

Should solve issues of people like @Hyaxia in #4038 

Let's see if the test pass only with this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ before_install:
         tar -xzf geckodriver-v0.19.1-linux64.tar.gz -C geckodriver
         export PATH=$PATH:$PWD/geckodriver
       fi
+    - pip install "attrs>=17.4.0"
 
 install:
     - pip install --pre .[test]


Hyaxia:
Looks like travis gives the same errors as the one I'm having at #4038

Carreau:
> Looks like travis gives the same errors as the one I'm having at #4038

Yeah I pinned to >17.3 instead of >=17.4. I've pushed changes.

Hyaxia:
Hey, it looks like the tests are doing problems again.
#4036 I commited few other changes and the tests now fail in appveyor.
Fail at the package installation part, at pip install .[test].

It looks like the problem is something with the setuptools or the pip itself.

Hyaxia:
@jph00 see my temporary fix for the bug.
It seems like there is a problem with the current of `nbconvert` or its dependencies that
anaconda installs, so we just use pip instead.

jph00:
Thanks for the tip!

blink1073:
Thanks @Carreau, forging ahead!

PR #4043, opened 2018-09-27
closes #4029: implement optional markdown header and footer files

### LINKED ISSUE #4029: RFC: Implement header and footer in dashboard file viewer
## Goal
Enable notebook authors to provide per-folder information by enabling header and/or footer documents in the file viewer.

## Proposal
* The header will be placed just below `#tabs` div and above `#notebook_toolbar` div (i.e., above "Select items to perform actions"). The rationale for this placement is that the folder scope applies only to files and not to Running or Clusters tabs. 
* The footer will be placed below the `#notebook_list` div. 
* The header and footer are both optional.
* The header will be obtained from the first available file in this order: `.header.html`, `.header.md`, `header.html`, `header.md`. 
* The footer will be obtained from the first available file in this order: `.footer.html`, `.footer.md`, `footer.html`, `footer.md`. 
* If multiple header or footer files are found, only the first in the above order will be used.

## Comments/Questions
* The header and footer should be static documents.
* README-based filenames (e.g., README.md) should be reserved for longer explanatory documents.
* index-based filenames (e.g., index.ipynb) should be reserved for main content.
* How important is it that the header and footer be editable within the folder? Other options are to open new markdown editor page or to require editing externally.
* Should the proposed header and footer be moved to the top and bottom rows of the `#notebook_list` instead (within `list_item` divs)? Doing so provides better scope of the header and footer under the indicated path, but clutters the file display.

### Outside notebook_list
selection_171
### Inside notebook_list
selection_173

Related: Brief exchange with @minrk at https://gitter.im/jupyter/notebook?at=5ba89537e5c2cc56add5adca

Suggested labels: `component:File Browser`, `status:Needs Discussion`, `type:Enhancement`, `tag:Design and UX`

reece:
I took a stab at implementing header and footer support from #4029. Comments appreciated. I'll add documentation to this PR after I get feedback that this change is welcomed.

```
$ jupyter notebook --debug docs/source/examples/Notebook/
```

selection_176

### DIFF
--- a/docs/source/examples/Notebook/header.md
+++ b/docs/source/examples/Notebook/header.md
@@ -1,0 +1,4 @@
+# Notebook Examples
+
+The pages in this section are all converted notebook files. You can also
+[view these notebooks on nbviewer](http://nbviewer.jupyter.org/github/jupyter/notebook/blob/master/)
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -10,8 +10,9 @@ define([
     'base/js/events',
     'base/js/keyboard',
     'moment',
-    'bidi/bidi'
-], function($, IPython, utils, i18n, dialog, events, keyboard, moment, bidi) {
+    'bidi/bidi',
+    'components/marked/lib/marked'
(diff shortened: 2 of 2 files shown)


Hyaxia:
Hey, check #4039 for the fix for the tests that are failing.
Enjoy!

PR #4075, opened 2018-10-07
Kernel ram metric

Hyaxia:
I have added a metric for the RAM per kernel.
Each kernel appears alone identified by his id and type.
I have chose to implement the monitoring by a background task with the tornado framework.
As soon as you start the notebook it will all the running kernels.
Every 0.2 seconds it will set the RAM usage of the specific kernel to what is uses at that moment.
Additionally, the way i check the ram of each process is with the `psutil` package.
I have found that this was the easiest way, if you guys have something else in mind please 

relevant issue #3682

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -11,10 +11,12 @@ from collections import defaultdict
 from datetime import datetime, timedelta
 from functools import partial
 import os
+import psutil
 
 from tornado import gen, web
 from tornado.concurrent import Future
 from tornado.ioloop import IOLoop, PeriodicCallback
+from tornado.stack_context import run_with_stack_context, NullContext
 
 from jupyter_client.session import Session
 from jupyter_client.multikernelmanager import MultiKernelManager
@@ -27,6 +29,7 @@ from notebook._tz import utcnow, isoformat
 from ipython_genutils.py3compat import getcwd
 
 from notebook.prometheus.metrics import KERNEL_CURRENTLY_RUNNING_TOTAL
+from notebook.prometheus.metrics import KERNEL_RAM_USAGE_MB
 
 
 class MappingKernelManager(MultiKernelManager):
@@ -118,6 +121,7 @@ class MappingKernelManager(MultiKernelManager):
 
     def __init__(self, **kwargs):
         super(MappingKernelManager, self).__init__(**kwargs)
+        self.kernel_ram_monitoring()
         self.last_kernel_activity = utcnow()
 
     #-------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


notebook/services/kernels/kernelmanager.py
```diff
@@ -463,3 +467,38 @@ def cull_kernel_if_idle(self, kernel_id):
                 self.log.warning("Culling '%s' kernel '%s' (%s) with %d connections due to %s seconds of inactivity.",
                                  kernel.execution_state, kernel.kernel_name, kernel_id, connections, idle_duration)
                 self.shutdown_kernel(kernel_id)
+
+    @gen.coroutine
+    def kernel_ram_monitoring(self):
+        # Arbitrary time for waiting till next metric collection
+        yield gen.sleep(0.2)
```
kevin-bates: Five passes per second seems a little excessive to me.  I suspect that on highly loaded servers (with lots of kernels running), the set of kernels may not get completed before the next iteration will occur.  If this interval could be configured, that would be nice.  Seems like a default of 1 second would be frequent enough for the normal case (of a couple kernels).

notebook/services/kernels/kernelmanager.py
```diff
@@ -463,3 +467,38 @@ def cull_kernel_if_idle(self, kernel_id):
...
+
+        # Update the relevant kernel's ram usage metric
+        def update_ram_metric():
+            pid = kernel.kernel.pid
+            kernel_process = psutil.Process(pid)
```
kevin-bates: This is assuming that all kernels are local to the server - which is not the case for Enterprise Gateway - which derives its kernel management from these classes.  In this case, EG will break since the launching process is no longer running.

I would recommend moving the code that gets the memory usage to a method on the individual kernel manager class - this way child classes can do their own means of capturing memory usage.  Unfortunately, Notebook doesn't define a kernel manager class.  If one was added, it would derive from jupyer_client's [`IOLoopKernelManager`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/ioloop/manager.py#L26) - then we'd need to fix up the hierarchy usage in Jupyter Kernel and Enterprise Gateway.  I'd be happy to help with this if others agree.  The other option is to abstract the `kernel` class that represents the process.  We do this in Enterprise Gateway with our `process proxy` classes, but that's a bit more work.  We'd then have a 'hook' to place capture code for our remote kernels.

Any way we can make the capture of per-kernel info more abstract would be great.

In the meantime, we should probably make this feature optional and, preferably, off by default so as to not break applications.

minrk:
This should probably be implemented at the lower, KernelManager level in jupyter_client, so that when an implementation is swapped out, it can override the appropriate method.

Hyaxia:
Thabk you both, @minrk & @kevin-bates.
I will check the jupyter client.

PR #4085, opened 2018-10-10
Fixed some French translations and typo

vgalisson:
Suppressed some capital letters because doesn't look right in French
Changed "De Confiance" to "Fiable" because the formulation is odd, "De Confiance" sounds a bit like "Is Trusted" , when "Fiable" can directly be translated to "Trusted" (and therefore the change from "Sans confiance" to "Non Fiable")
Changed from "de" to "du" when it was uncorrectly used


takluyver:
@Carreau want to have a look?

Carreau:
Looks good to me.

Carreau:
Thanks !

PR #4101, opened 2018-10-14
Translated kernel menu javascript test to selenium

dubielt1:
This is a contribution to issue #3335 converting js tests to selenium. This contains the conversion of the file notebook/tests/notebook/kernel_menu.js


dubielt1:
@takluyver I looked through the travis-ci logs and my test was the only one with this error: `Error opening stream: HTTP 404: Not Found (Kernel does not exist: 73b65ed7-2d99-4549-af2f-35fb69a9b70e)`

It seems to say a kernel connection wasn't established, so when the test code clicked the shutdown kernel button, the confirmation dialog never appeared, causing the `TimeoutException`

Maybe we can try re-running the tests? Let me know if I've overlooked anything

takluyver:
I've hit restart. For future reference you can usually restart tests by closing and reopening the pull request on Github (no need to open a new one - after you close, there's a button to reopen it).

It looks like the failure was in the test you converted, though, so it's worth looking into even if it's an intermittent failure.

takluyver:
Still failing. From the log, it looks like the kernel is started, then shutdown, and then something tries to connect to it after it is shut down.

notebook/tests/selenium/test_kernel_menu.py
```diff
@@ -0,0 +1,79 @@
...
+        browser.find_element_by_css_selector(shutdown_selector).click()
+
+        wait_for_selector(browser, confirm_selector)
+        browser.find_element_by_css_selector(confirm_selector).click()
+
+        WebDriverWait(browser, 3).until(
+            EC.element_to_be_clickable((By.CSS_SELECTOR, kernel_menu_selector)))
+
+        while check_modal_still_open(notebook):
+            pass
```
takluyver: I'm not super keen on the tight loop - can it have a `sleep(0.01)` or something so it's not just polling as fast as it can.

dubielt1: For sure

dubielt1:
> I've hit restart. For future reference you can usually restart tests by closing and reopening the pull request on Github (no need to open a new one - after you close, there's a button to reopen it).
> 
> It looks like the failure was in the test you converted, though, so it's worth looking into even if it's an intermittent failure.

Thanks for the tip.

I'll work on reproducing the failure locally this evening, but in my initial runs it passed.

takluyver:
@dubielt1 Do you think you'll have time to look at this again?

dubielt1:
Yes, I'll give it another look this week

takluyver:
Happy new year! Do you think you'll have time to look at this pull request again?

takluyver:
I've rebased and updated this to try to get it finished.

Issue #3947, opened 2018-09-20
Convenient notebook URL for copy-pasting is not convenient when --ip flag specified


ned2:
When passing the --ip flag (in my case, providing --ip 0.0.0.0 to make the server available on an intranet) to Jupyter notebook, the resulting URL string that is printed to the terminal for copy-pasting into the browser has the following form:

    https://(PUBLIC_IP or 127.0.0.1):5000/?token=TOKEN

The host part of the URL has the disjunction `(PUBLIC_IP or 127.0.0.1)` which makes the URL invalid and not suitable for copy-pasting, rendering this feature somewhat broken in this context.

What about changing the behaviour to produce the following two alternative URLs when the --ip flag is specified:

    https://PUBLIC_IP:5000/?token=TOKEN
    https://127.0.0.1:5000/?token=TOKEN

minrk:
Displaying both URLs might be the best choice, rather than where we are now. We've made several compromises on this feature, to nobody's satisfaction. See #3356, #3605, #3668, and ultimately #3703. This is a tough compromise, and the answer is simply that it is impossible for one URL to be correct, so I made the deliberate choice to force people to think about it, since the right link cannot possibly be known by the notebook application. But picking from two full URLs could definitely be better than forcing people to edit. In some cases (starting server locally), 127.0.0.1 is right, in others (remote server, docker) it doesn't work. In some cases (everything but docker), hostname/public ip is right, in others (everything but docker), it doesn't work.

nfultz:
@minrk Thank you for linking to the other tickets, so that I could start to understand why this is now happening. 

The other tickets do mention a new feature to work around this, `display_url` and `connect_url`, but I don't understand how to use them correctly:

```
(jupyterlab) [nfultz@spaceaware-compute-01 ~]$ jupyter notebook --version
5.7.4
(jupyterlab) [nfultz@spaceaware-compute-01 ~]$ jupyter lab --version
0.35.3
(jupyterlab) [nfultz@spaceaware-compute-01 ~]$ jupyter lab --ip=0.0.0.0 --display_url=$(hostname)
[W 10:34:53.890 LabApp] Unrecognized alias: '--connect_url=spaceaware-compute-01.ads.isi.edu', it will probably have no effect.
[I 10:34:54.267 LabApp] JupyterLab extension loaded from /nas/home/nfultz/miniconda3/envs/jupyterlab/lib/python3.7/site-packages/jupyterlab
[...]

PR #4103, opened 2018-10-15
Fix copy/paste-ability of default URL presented on startup

einon:
Currently the default URL message given on the console on startup is:

Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://(myip.com or 127.0.0.1):8888/?token=8fdc8 ...

This will always need editing to use. Replace brackets/ 'or' part with host IP (e.g. 'myip.com')
to make it copy/pastable again.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1439,7 +1439,9 @@ class NotebookApp(JupyterApp):
         if self.token:
             # Don't log full token if it came from config
             token = self.token if self._token_generated else '...'
-            url = url_concat(url, {'token': token})
+            url = (url_concat(url, {'token': token})
+                  + '\n or '
+                  + url_concat(self._url('127.0.0.1'), {'token': token}))
         return url
 
     @property
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1432,7 +1432,7 @@ class NotebookApp(JupyterApp):
                 url += '/'
         else:
             if self.ip in ('', '0.0.0.0'):
-                ip = "(%s or 127.0.0.1)" % socket.gethostname()
+                ip = "%s" % socket.gethostname()
             else:
                 ip = self.ip
             url = self._url(ip)


minrk:
This is the same as #3356, which was reverted because it caused problems. See #3605, #3668, #3703 for why we don't do this, and various discussions about what to do and when.

The latest proposal in #3947 is to display *both* the hostname and localhost URL in copy-pasteable form, which might be the best so far. Would you like to take a stab at doing that?

einon:
Proposal stabbed at, comments welcome.

Carreau:
Thanks, 

Your editor seem to be configured to trim whitespaces, this adds a lot of noise in the diff, and also often is the source of issues when trying to backport Pull-requests / use Git blame, or git-bisect. 

We in general are in favor of removing whitespace, but only in proximity of modified code. 

How comfortable are you in rebasing without the whitespace changes ? Otherwise we can do that for you.

einon:
Hi, thanks for looking at this PR.

The whitespace changes were in a separate commit, so I've just reverted that.

I've also rebased my local branch to the head of master and force-pushed, which I assume was causing the previous merge error. (Does github always take a few hours to update it's UI after a force-push?)

Thanks,

Mark

einon:
Hi, thanks for looking at this PR.

The whitespace changes were in a separate commit, so I've just reverted that.

I've also rebased my local branch to the head of master and force-pushed, which I assume was causing the previous merge error. (Does github always take a few hours to update it's UI after a force-push?)

Thanks,

Mark

einon:
The CI pipeline looks to be stuck on this PR, given that the same code changes passed a few days ago. Can it be restarted?

einon:
Hi @Carreau, just checking in - is there anything I can do to help merge this PR? I don't believe the current CI checks are sane.

takluyver:
@einon generally closing and reopening a pull request will re-run the CI. I'll try that now.

notebook/notebookapp.py
```diff
@@ -1432,14 +1432,16 @@ def display_url(self):
...
             else:
                 ip = self.ip
             url = self._url(ip)
         if self.token:
             # Don't log full token if it came from config
             token = self.token if self._token_generated else '...'
-            url = url_concat(url, {'token': token})
+            url = (url_concat(url, {'token': token})
+                  + '\n or '
+                  + url_concat(self._url('127.0.0.1'), {'token': token}))
```
takluyver: We've previously been trying to show only the real IP if there's one set, and make the 'or 127.0.0.1' a special case when you're listening on all IPs. This would show the 127.0.0.1 option all the time, even when you're not listening on that IP, and even when it's the same as the first address. But maybe that's better than trying to be clever about it?

einon: Hi @takluyver, thanks for the update.
In my experience it would be an unusual system not to be listening on 127.0.0.1, the loopback address. Is that an existing use case for jupyter notebooks?

takluyver: It's definitely possible to listen only on an external-facing interface, though I don't know how many people do it. But even when you do that, it might be e.g. the external network interface of a container, which is mapped to the same port on localhost outside the container. There's no way in general to know what address the server is actually accessible at.

So I'm inclined to merge this, and we can always reconsider it again later.

AlJohri:
Thanks for this! I was just about to make an issue regarding the copy/pastability.

Markus92:
Would it be possible to apply this patch to the 5.7.x branch? It'd allow me to put it into production right away (and make our users very happy!).

Issue #4116, opened 2018-10-18
feature request: md link to the same notebook in the same tab


ggrrll:
Hi, 

an idea  I just had while using my notebooks -- not sure if it's already under development:

whenever I have a internal reference to another section of the same notebook, in a md cell, would be nice if after clicking it, the following action would take you to the right point _in the same tab_ , instead of opening the same notebook in a new tab.

Thanks

ps: this is maybe related to #4115

minrk:
This is the behavior I see when I click markdown links to local anchors. What do your links look like that open new tabs?

The following links result in local navigation of the current tab:

```markdown
[click to scroll](#scroll-target)
```

and in a later cell:

```markdown
<a id="scroll-target">you found me!</a>
```

ggrrll:
ok, 

 I see -- what  I wrote in the title is what happes when one puts a url which contains before something else, than just `#scroll-target` , like `localhost:8888/my_notebook.ipynb#find-best-alpha`

so sorry, my fault -- thanks for clarifying  @minrk

PR #4120, opened 2018-10-19
Notebook opens in same tab

alekhya1891:
Issue #4116 

Open for collaborations


minrk:
Thanks for the PR! These links use `IPython._target on purpose, because there are two modes for this: `_blank` opens in a new tab, while `_self` opens in the current tab. This can be customized in `custom.js`. This PR removes the option, which I don't think we want to do. Additionally, #4116 doesn't appear to be an issue in the current version of the notebook.

PR #4122, opened 2018-10-19
Corrected alt-text of "replace in all cells"

thomasaarholt:
When using the (great) "Find and Replace" function in the notebook, there is a button one can click to replace text in **all** cells, not only the selected one.

It's alt-text ("hover-text") says that it does the opposite of what it does, as seen in the image. In the image the button is not been selected and only finds one of the two "hello" instances.

image


minrk:
Thanks!

Issue #3024, opened 2017-11-07
Markdown table alignment


GreatBahram:
Hi
I want to change the default alignment in jupyter notebooks. I realized the default markdown change alignment like `:----` and `:---:` does not work inside jupyter notebook and the table I try to show inside notebook looks ugly because everything is by default align in right side :-(. 
This is my table 

Resource record Type | Resource record name | Description  
--- | :---: | ---
*1* | `A` | *A specifies IP address (IPv4) for given host*
*28* | `AAAA` | *AAAA (also quad-A record) specifies IPv6 address for given host.*
*15* | `MX` | *The MX resource record specifies a mail exchange server for a DNS domain name.*
*12* | `PTR` | *The PTR record is used to look up domain names based on an IP address.*
*16* | `TXT` | *The text record can hold arbitrary non-formatted text string.*
*33* | `SRV` | *The SRV resource record identifies the host(s) that will support a particular service.*
*2* | `NS` | *The NS record specifies an authoritative name server for given host.*
*6* | `SOA` | *The record specifies core information about a DNS zone, including the primary name server, 0the email of the domain administrator, the domain serial number, and several timers relating to refreshing the zone. *

![What jupyter shows to me!](https://user-images.githubusercontent.com/14103831/32486994-8d786cf0-c3bd-11e7-8ec2-aa4ada31814b.png)

I want to some of it columns be center align and other left or right, How do I have to change this behavior in the right way ?

HaiKju:
Same problem with me. Is there anybody who can help?

takluyver:
Inspecting the HTML in the browser, the alignment information does not appear to be coming through the markdown parser. But I can see code there to process it in marked.js . So it looks like it's either a bug in the version of marked we bundle, or a problem with the way we're passing data into marked.

HaiKju:
If you "Download as HTML, LaTeX or PDF via LaTeX" the output is correct. So it seems to be a bug in the parser.

takluyver:
Ah, no, I know. The HTML in markdown cells is sanitized, and that includes stripping off style attributes. Not sure what to do about that at the moment.

When you 'download as', the markdown is processed by different parsers (mistune for HTML, pandoc for Latex).

HaiKju:
[...]

PR #4130, opened 2018-10-22
Default to right table alignment but allow table local overrides

### LINKED ISSUE #3919: Markdown table alignment stopped working
| col1 | col2 | col3 |
|:-|-:| :-: |
| left aligned | right aligned | centered |
| A | A | A |

In Jupyter, now it's always right aligned!

It's one thing enforce right-alignment for consistency reasons as described in jupyter/notebook/pull/2534. It's another to stop supporting features in Markdown as described by the link in the Help bar, currently pointing to:

https://help.github.com/articles/getting-started-with-writing-and-formatting-on-github/
https://help.github.com/articles/organizing-information-with-tables/#formatting-content-within-your-table

Furthermore, the user gets no messaging to let them know that they're using unsupported features, i.e. the :---, :---:, and ---: tokens to describe justification.

SpencerPark:
Allow markdown tables to specify their alignment with the `:--`, `:--:`, `--:` syntax while still allowing `---` to default to `right` alignment.

This reverts a previous change (part of #2534) that stripped the alignment information from the rendered HTML coming out of the marked renderer. To be clear this does not change the `right` alignment default but does bring back support for overriding this alignment per column by using the gfm table alignment syntax.

Since the default is defined as a css rule, the `align` attribute would no longer override that and as such this PR uses an inline `style` with the appropriate `text-align` instead. Caja by default whitelists `text-align` so the style is not stripped during sanitation. 

The alignment is properly captured in display data:

```python
from IPython.core.display import Markdown

display(Markdown('''
| None | Right | Left | Center |
|------|------:|:-----|:------:|
| 0    | 1     | 2    | 3      |
'''))
```

as well as in a markdown cell:

```markdown
| None | Right | Left | Center |
|------|------:|:-----|:------:|
| 0    | 1     | 2    | 3      |
```

both render as

| None | Right | Left | Center |
|-----:|------:|:-----|:------:|
| 0    | 1     | 2    | 3      |

Fixes #3919 (and the prior #3024).

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -732,7 +732,8 @@ define([
         var renderer = new marked.Renderer();
         renderer.tablecell = function (content, flags) {
           var type = flags.header ? 'th' : 'td';
-          var start_tag = '<' + type + '>';
+          var style = flags.align == null ? '': ' style="text-align: ' + flags.align + '"';
+          var start_tag = '<' + type + style + '>';
           var end_tag = '</' + type + '>\n';
           return start_tag + content + end_tag;
         };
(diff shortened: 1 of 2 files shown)


SpencerPark:
Is there anything I can do to help this along?

minrk:
Thanks!

PR #4141, opened 2018-10-26
Migrate image display test to selenium

askerry:
Migrates a single js test (for image display functionality) to selenium as suggested in issue #3335.

Test Plan:
py.test -v notebook/tests/selenium/test_display_image.py


notebook/tests/selenium/test_display_image.py
```diff
@@ -0,0 +1,71 @@
...
+        'b64data = %s' % b64data,
+        'data = base64.decodebytes(b64data)',
+        'display(Image(data, retina=%s))' % retina
+    ]
+    notebook.append("\n".join(commands))
+    notebook.execute_cell(cell_index)
+
+    # Find the image element that was just displayed
+    import time
+    time.sleep(1)
```
takluyver: Do we need the sleep here?

notebook/tests/selenium/test_display_image.py
```diff
@@ -0,0 +1,71 @@
...
+
+    width = img_element.size["width"]
+    height = img_element.size["height"]
+    width_attr = img_element.get_attribute("width")
+    height_attr = img_element.get_attribute("height")
+    expected_size = 1 if retina else 2
+    assert width == expected_size
+    assert height == expected_size
+    assert width_attr == str(expected_size)
+    assert height_attr == str(expected_size)
```
takluyver: I would condense these lines with the definitions just above, e.g. `assert img_element.size["width"] == ...` - one less thing to look at if you're investigating a test failure.

notebook/tests/selenium/test_display_image.py
askerry: Derp, no. That was a temp hack I'd been using before I implemented `wait_for_tag` and I forgot to delete it.

takluyver:
Thanks!

PR #4142, opened 2018-10-26
Migrate test for merge cells api to selenium

askerry:
Migrates the test for merging cells to use selenium as suggested in issue #3335.

Test Plan:
py.test -v notebook/tests/selenium/test_merge_cells.py


takluyver:
Thanks! It would be good to also add a test for `merge_selected_cells`, but that can be done later.

takluyver:
Opened issue #4177 for `merge_selected_cells`.

PR #4147, opened 2018-10-26
Updated the description of Heading cells, which are no longer used

jakelevirne:
Heading cells don't exist anymore, so I've updated the "What is the Jupyter Notebook" example to describe them in the same way that jupyter describes them when you try to set a cell to type "Heading"


takluyver:
Thanks! Let's go one step further, though - we can remove heading cells from that list and say "There are three basic cell types". We don't need the history in this introductory document.

PR #4148, opened 2018-10-26
Cleared all outputs in Running Code example

jakelevirne:
The [Documentation Contributor Guide](https://jupyter.readthedocs.io/en/latest/contrib_docs/getting-started.html) says to please clear the output cells on notebooks.  Seems important for this example notebook in particular, since it guides the user through running cells.


takluyver:
Thanks! I wonder if this notebook should be an exception, though. Clearing the output means it's run when building the documentation, and that notebook has a couple of `time.sleep()` calls in it, so it would make the docs build pause for an extra 14 seconds.

Issue #4023, opened 2018-09-22
Markdown blocks: Typing parentheses/quotes around selected text should surround, not replace


jimbo1qaz:
When in a Python cell, typing parentheses/quotes around selected text surrounds with parentheses. I think the same behavior should also occur for Markdown/etc. cells.

PR #4173, opened 2018-11-06
Add bracket autocompletion to codemirror config of text cell

timporath:
Typing parentheses/quotes around selected text in markdown or raw text cell surrounds it with parentheses. Brackets will also be auto-completed and matching will be highlighted. 

Fix for Issue #4023


blink1073:
We've seen other users wanting the opposite behavior, and it can be configured using `custom.js`.  Closing, but thank you for the suggestion, @timporath!

PR #4184, opened 2018-11-08
allow setting event loop policy

### LINKED ISSUE #4183: Hang when setting an asyncio event loop policy in server extension
I believe that if a server extension installs a new asyncio event loop policy that Jupyter will hang.  Two common cases where this comes up:

1.  Setting Tornado's `AnyThreadEventLoopPolicy` allowing other threads to be used for IOLoops

    ```python
    import asyncio
    import tornado.platform.asyncio
    asyncio.set_event_loop_policy(tornado.platform.asyncio.AnyThreadEventLoopPolicy())
    ```  

    Dask uses this

2.  Using uvloop

    ```python
    import uvloop
    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
    ```

My guess is that these interfere with the use of `pyzmq`'s event loop policy within the Jupyter server, but I'm not sure.

This is causing problems with some work for the dask-labextension.  See https://github.com/dask/dask-labextension/pull/31

minrk:
can be an import string to a class or an instantiated policy

additionally, avoid calling deprecated ioloop.install with pyzmq 17. This can complicate tornado eventloop instantiation.

closes #4183 as best I can think of, since I don't think we can in general load server extensions early enough that they could do this themselves.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -107,8 +107,10 @@ from jupyter_client.session import Session
 from nbformat.sign import NotebookNotary
 from traitlets import (
     Any, Dict, Unicode, Integer, List, Bool, Bytes, Instance,
-    TraitError, Type, Float, observe, default, validate
+    TraitError, Type, Float, observe, default, validate,
+    import_item,
 )
+
 from ipython_genutils import py3compat
 from jupyter_core.paths import jupyter_runtime_dir, jupyter_path
 from notebook._sysinfo import get_sys_info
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -43,8 +44,12 @@
...
 # Install the pyzmq ioloop. This has to be done before anything else from
 # tornado is imported.
-from zmq.eventloop import ioloop
-ioloop.install()
+import zmq
+if V(zmq.__version__) < V('17'):
+    from zmq.eventloop import ioloop
+    ioloop.install()
+else:
+    ioloop = None
```
takluyver: I'd be inclined to put the `from tornado` import here directly, so you can see both cases in one snippet.

notebook/notebookapp.py
```diff
@@ -1116,9 +1125,37 @@ def _update_mathjax_config(self, change):
...
+        will be imported and instantiated.
+        """
+        ),
     )
+    @observe('event_loop_policy')
+    def _event_loop_policy_changed(self, change):
+        policy = change.new
+        if isinstance(policy, str):
+            # if it's a str, it should be an import string
+            policy = import_item(policy)()
```
takluyver: Doesn't the `Type` traitlet allow importable strings already? Are there cases where it's important to pass an instance so you can instantiate it with non-default arguments?

takluyver:
More importantly, is this actually useful? The problem motivating it in #4183 isn't really solved by making it configurable - as you point out, extensions are loaded too late to influence creating the event loop. @mrocklin has a workaround for dask (https://github.com/dask/distributed/pull/2343), and I'm not aware of anyone else who has asked for this.

How the event loop is created seems like squarely an application concern, so I'm -1 on making it configurable unless there's a clear use case that's not better solved another way. The more bits that can be configured, the harder it is to change the underlying code later.

minrk:
While putting this together, I did realize that `jupyter_notebook_config.py` can also just do:

```python
asyncio.set_event_loop_policy(Policy())
```

Which isn't really any more complex than

```python
c.NotebookApp.event_loop_policy = Policy()
```

The main difference would be that having it in config makes it explicit that users can set the event loop policy, while the other just happens to work.

I think we should allow users to request running with uvloop, etc.

blink1073:
Closing this one as stale, cheers!

PR #4182, opened 2018-11-08
Add test for JS function "merge_selected_cells" in test_merge_cells.py

AndresSan6:
This PR add test requested in #4177

PR #4185, opened 2018-11-08
Add wait_for_tag to avoid race condition in display isolation test

askerry:
Attempts to fix flakiness in `test_display_isolation`. We now ensure the iframe has been added to the dom before calling the selector. To make this work, we clean up the iframe cells (and all other cells) at the end of each test. I'm not 100% positive this fixes the issue, since I haven't been able to reproduce the failure. But the hope is that this fixes the intermittent test failing seen in https://github.com/jupyter/notebook/pull/4182.

### DIFF
--- a/notebook/tests/selenium/test_display_isolation.py
+++ b/notebook/tests/selenium/test_display_isolation.py
@@ -3,6 +3,7 @@
 An object whose metadata contains an "isolated" tag must be isolated
 from the rest of the document.
 """
+from .utils import wait_for_tag
 
 
 def test_display_isolation(notebook):
@@ -26,7 +27,7 @@ def isolated_html(notebook):
     """
     red = 'rgb(255, 0, 0)'
     blue = 'rgb(0, 0, 255)'
-    test_str = "<div id='test'>Should be red from non-isolation</div>"
+    test_str = "<div id='test'>Should turn red from non-isolation</div>"
     notebook.add_and_execute_cell(content="display(HTML(%r))" % test_str)
     non_isolated = (
         "<style>div{color:%s;}</style>" % red +
@@ -40,6 +41,8 @@ def isolated_html(notebook):
     display_i = "display(HTML(%r), metadata={'isolated':True})" % (
         isolated)
     notebook.add_and_execute_cell(content=display_i)
+
+    wait_for_tag(notebook.browser, "iframe")
 
     # The non-isolated div will be in the body
     non_isolated_div = notebook.body.find_element_by_id("non-isolated")
@@ -55,6 +58,9 @@ def isolated_html(notebook):
     isolated_div = notebook.browser.find_element_by_id("isolated")
     assert isolated_div.value_of_css_property("color") == blue
     notebook.browser.switch_to.default_content()
+    # Clean up the html test cells
+    for i in range(1, len(notebook.cells)):
+        notebook.delete_cell(1)
 
 
 def isolated_svg(notebook):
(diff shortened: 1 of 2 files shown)


askerry:
@takluyver I'm hoping this fixes the failure in https://github.com/jupyter/notebook/pull/4182

notebook/tests/selenium/test_display_isolation.py
```diff
@@ -73,16 +79,20 @@ def isolated_svg(notebook):
         content="display_svg(SVG(s1), metadata=dict(isolated=True))")
     notebook.add_and_execute_cell(
         content="display_svg(SVG(s2), metadata=dict(isolated=True))")
+    wait_for_tag(notebook.browser, "iframe")
     iframes = notebook.body.find_elements_by_tag_name("iframe")
```
takluyver: Here we're waiting for an `iframe` tag to exist, and then getting a list of iframes, which must contain 2 iframes. I wonder if the list is sometimes captured when only one is on the page. Can we figure out a way to wait for two, or to wait for the one we know will appear second?

askerry: hmm true. I can probably give the iframes ids and wait for both of the ids to be safe.

askerry: setting ids turned out to be non-trivial but I think I can probably write a custom `wait_for_n` selenium expectation.

notebook/tests/selenium/utils.py
```diff
@@ -33,6 +54,36 @@ def _wait_for(driver, locator_type, locator, timeout=10, visible=False, single=F
...
+        try:
+            elements = driver.find_elements(locator_type, locator)
+        except WebDriverException as e:
+            raise e
+        if len(elements) < wait_for_n:
+            return False
+        if visible:
+            for element in elements:
+                if not element.is_displayed():
+                    return False
```
takluyver: I think this might need to be refined slightly. E.g. if I say `wait_for_tag('h2', wait_for_n=2, visible=True)` when there are 2 visible `h2` tags on the page and one invisible, it will never finish waiting.

Alternatively, we could leave out visibility from this code for now, and wait until there's a need for it.

notebook/tests/selenium/utils.py
```diff
@@ -12,13 +13,33 @@
 pjoin = os.path.join
 
 
-def wait_for_selector(driver, selector, timeout=10, visible=False, single=False):
+def wait_for_selector(driver, selector, timeout=10, visible=False, single=False, wait_for_n=1):
+    if wait_for_n > 1:
+            return _wait_for_multiple(
```
takluyver: Can we use the normal four-space indentation for these?

notebook/tests/selenium/utils.py
```diff
@@ -33,6 +54,36 @@ def _wait_for(driver, locator_type, locator, timeout=10, visible=False, single=F
...
+        wait_for_n: wait until this number of matching elements are present/visible
+        visible: if True, require that elements are not only present, but visible
+    """
+    wait = WebDriverWait(driver, timeout)
+
+    def multiple_found(driver):
+        try:
+            elements = driver.find_elements(locator_type, locator)
+        except WebDriverException as e:
+            raise e
```
takluyver: I see a similar pattern in some of the selenium code, but as I understand it, the try/except is redundant here - reraising the exception is the same as not catching it.

askerry: yeah I was confused by it too, but included because the selenium conditionals did. happy to remove!

notebook/tests/selenium/utils.py
askerry: ah good point! I'll adjust to just look for the specified number visible.

takluyver:
Thanks :-)

PR #4194, opened 2018-11-12
Fix typo in nbjs.po

hansnow:
`出` is a typo of `处` in Chinese with same pronunciation. It's a common mistake with some Chinese input methods.


takluyver:
Thanks, we'll trust you on this!

PR #4196, opened 2018-11-13
generate Python API in docs

Carreau:


takluyver:
Who are we doing this for? Our experience of generated API docs in IPython hasn't been great; we end up with all sorts of pieces documented that we wouldn't really want to consider public API. I was quite pleased when we did the big split that it let us cast off generated API docs for most of the new projects.

If the target audience is people writing server extensions, then also a big part of the API they care about will be the structure of instances in the running application, which the generated docs aren't going to capture.

Carreau:
> Who are we doing this for?

So far, me. I wanted to have them locally, and as I had it, I sent the PR. 
Mostly I'm writing a FileContentManager, and jumping through the code to find all the methods start to be hard. 

I'm ok not merging that – as it is true that the API docs are not supper pretty, and I miss some options like showing all methods including the ones from `super()`.

minrk:
I think there's two things to use auto-doc for:

1. actual API generation (API consumers can look here for the public API)
2. sourcecode reference (potentially useful for contributors working on this repo itself)

I think it's fine to do bulk generation for 2, but 1 should be done with deliberate whitelisting of which APIs are public and supported. For the writing a contents manager case, it sounds like 1 is what we want. @Carreau would you like to go through and whitelist the public APIs for contents managers as a starting point?

PR #4210, opened 2018-11-13
Enable kernel message filtering

SylvainCorlay:
This is the counterpart to https://github.com/jupyter/jupyter_server/pull/22 enabling the kernel message filtering.

This is useful for running a notebook server for kernel extensions that should not allow arbitrary code execution, such as [voila](https://github.com/QuantStack/voila/).


.pytest_cache/v/cache/nodeids
```diff
@@ -0,0 +1,255 @@
+[
+  "notebook/auth/tests/test_security.py::test_passwd_structure",
```
blink1073: Checked in by accident?

SylvainCorlay: Fixed!

PR #4216, opened 2018-11-15
Add shutdown hook

darkocerdic:
This patch adds the ability to specify a script that will be called if a notebook server is killed due to inactivity. 

The script path is set in configuration variable: `c.NotebookApp.on_inactivity_shutdown_execute`

For the script to be executed these variables need to be set:
* `c.MappingKernelManager.cull_idle_timeout`
* `c.NotebookApp.shutdown_no_activity_timeout`
* `c.NotebookApp.on_inactivity_shutdown_execute`

The rationale for this functionality is that sometimes the jupyter notebooks are executed on expensive cloud VM's that are supposed to be shut down after the work in notebook has been completed. People tend to forget to stop them after completing the work, which incurs unneeded expenses. This patch enables setting the server shutdown script to be executed after some period of inactivity.

Example config file:
```python
# Cull inactive kernels after 30 minutes
c.MappingKernelManager.cull_idle_timeout = 1800

# Shutdown jupyter server after 10 minutes of no active kernels
c.NotebookApp.shutdown_no_activity_timeout = 600

# Execute system shutdown when shutting down jupyter server
c.NotebookApp.on_inactivity_shutdown_execute = "/usr/bin/stop_vm"
```

`/usr/bin/stop_vm` script:
```bash
#! /bin/sh
sudo poweroff
```
(system should be set to accept running `sudo poweroff` without password in this case)

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -25,6 +25,7 @@ import re
 import select
 import signal
 import socket
+import subprocess
 import sys
 import threading
 import time
@@ -1283,6 +1284,14 @@ class NotebookApp(JupyterApp):
               "0 (the default) disables this automatic shutdown.")
     )
 
+    on_inactivity_shutdown_execute = Unicode(u'', config=True,
+        help=("Path of the script to execute before shutting down due to inactivity."
+              "This is intended to be used together with culling idle kernels"
+              "(MappingKernelManager.cull_idle_timeout) and automatic shutting"
+              "down the server after inactivity (NotebookApp.shutdown_no_activity_timeout)"
+              "configuration options")
+    )
+
     terminals_enabled = Bool(True, config=True,
          help=_("""Set to False to disable terminals.
 
@@ -1617,7 +1626,16 @@ class NotebookApp(JupyterApp):
         if seconds_since_active > self.shutdown_no_activity_timeout:
             self.log.info("No kernels or terminals for %d seconds; shutting down.",
                           seconds_since_active)
+            if self.on_inactivity_shutdown_execute:
+                self.execute_pre_shutdown_script(self.on_inactivity_shutdown_execute)
             self.stop()
+
+    def execute_pre_shutdown_script(self, path):
+        self.log.info("Executing pre-shutdown script: %s", path)
+        try:
+            subprocess.call(path, shell=True)
+        except:
+            self.log.info("Error running script: %s", path)
 
     def init_shutdown_no_activity(self):
         if self.shutdown_no_activity_timeout > 0:


takluyver:
Thanks. The use case makes sense, but I think it be tidier to do it in the parent process that launches the notebook - wait for it to finish, then shut the server down. That will catch any possible way that the server might terminate.

If you want to do it from inside Python, you can use the [atexit module](https://docs.python.org/3/library/atexit.html) to call a function on exit. You can do that in your `.py` config file, or in a server extension. This will catch most ways that the server can terminate, but not e.g. if the server process segfaults or you `kill -9` it.

darkocerdic:
@takluyver Your proposal certainly works but if I use the parent process then it will trigger a VM shutdown even if kill the notebook intentionally without wanting to power off the VM. This commit makes it explicit that it should only trigger the VM shutdown script in case of inactivity, not a manual killing of a notebook. In my work I usually start and stop a notebook server several times and having to worry about a process invoking a VM stop each time I manually stop the notebook is cumbersome.

I'm aware that a notebook server might segfault and then the script would not be triggered but that is an edge case that many people would not be worried about.

darkocerdic:
I understand that you are reluctant to add just any change some random user wants but I do think this one makes sense. The benefit of this approach is that it doesn't require adding additional infrastructure to the user setup. It only requires one-time setup to config file and from there on any notebook server behaves as expected. That is in contrast to having to come up with scripts and aliases to start the server and stop the VM after exit (even if you add a special `timeouted` exit code for the notebook server). It lowers the barrier to enabling and using this functionality.

> should it run a subprocess with or without a shell, or call a Python function? 

I propose to run an external process within a shell since that is the most likely way it would be used.

> What happens if the shutdown hook itself never finishes? If we're going to have a time limit, how long is too long?

Then I propose to start a detached process with:
```python
subprocess.Popen(path, shell=True)
```
instead of:
```python
subprocess.call(path, shell=True)
```
That way the notebook server shutdown will continue immediately and independently of external shutdown process.

> What kinds of shutdown should it fire on?

Only on shutdown because of inactivity. That is already in this PR.

takluyver:
It was clear what answers you want to those questions. But someone else might want a different set of answers to all of them. That's what I meant about getting 30 different options: someone else might want to run a process on inactivity and on manual shutdown, and wait for it to finish, and... each use case is straightforward to write the code for, but it adds up to a lot of config options.

So the attraction of doing it externally is flexibility. So long as you have the information to distinguish the cases that you're interested in, you can write code that does whatever you like when the notebook server shuts down. And lots of people already know how to write bash scripts or Python scripts to interact with this, whereas a new config option (or set of options) is something everyone will have to find out about.

darkocerdic:
Ok, I'm really sad to see that you are so against this change but I guess there is not much I can do about it. 😞 

So how do you propose I get this functionality at this moment?
Is it at all possible?

blink1073:
I agree this behavior should not be baked into core.  Thanks for the proposal and the discussion, @darkocerdic!

PR #4217, opened 2018-11-15
Documenation error fix

andrea-null-swanson:
Fixes documentation error that insinuates that the shortcuts require uppercase letters, when they actually require lowercase letters


jasongrout:
Thanks for contributing where you found something confusing.

The keyboard shortcuts show what is printed on the keys, which typically is an upper-case letter, i.e., the keyboard shortcut illustrates out what keys to press, not what text to type. Thus (for a US keyboard layout) a keyboard shortcut is `Shift 4` instead of `$`. If the printed keyboard shortcut was confusing, let's add an explanation.

(It seems that you also accidentally committed a lot more than maybe you meant to?)

andrea-null-swanson:
Yes I did commit more than I meant to-whoops.
I saw this issue listed as a documentation issue and went ahead and changed it. Do you think that the shortcut keys should have an explanation or are you set on leaving everything as an uppercase letter?
I am working on some bugs (and this documentation "issue") to get an A in one of my classes and have never worked in open source. I would love to contribute and if there is anything I could do to maybe provide clarification for the shortcuts that I could send a pull request in for or if you maybe could point me in a direction of a bug that would be a good start for me. I would greatly appreciate any help. Thanks.

blink1073:
Closing this one as stale.

PR #4219, opened 2018-11-16
Update session_exists() to account for invalid sessions due to culling

### LINKED ISSUE #4209: Opening notebooks occasionally yields kernel error, KeyError
Observed behavior: Users periodically open notebooks to a red "Kernel Error" status marker at top right, along with the bomb icon by the Python 3 kernel indicator. In the devtools console is this stack:

```
[E 14:13:17.077 NotebookApp] Uncaught exception POST /notebook/api/sessions (10.255.0.4)
    HTTPServerRequest(protocol='http', host='<our host here>', method='POST', uri='/notebook/api/sessions', version='HTTP/1.1', remote_ip='<our remote ip>')
      File "/anaconda/envs/py3/lib/python3.6/site-packages/tornado/web.py", line 1543, in _execute
        result = yield result
      File "/anaconda/envs/py3/lib/python3.6/site-packages/tornado/gen.py", line 1099, in run
        value = future.result()
      File "/anaconda/envs/py3/lib/python3.6/site-packages/tornado/gen.py", line 1113, in run
        yielded = self.gen.send(value)
      File "/anaconda/envs/py3/lib/python3.6/site-packages/notebook/services/sessions/handlers.py", line 67, in post
        model = yield gen.maybe_future(sm.get_session(path=path))
      File "/anaconda/envs/py3/lib/python3.6/site-packages/notebook/services/sessions/sessionmanager.py", line 170, in get_session
        return self.row_to_model(row)
      File "/anaconda/envs/py3/lib/python3.6/site-packages/notebook/services/sessions/sessionmanager.py", line 209, in row_to_model
        raise KeyError
    KeyError
[W 14:13:17.078 NotebookApp] Unhandled error
```

On page refresh everything works fine.

That trace points to this: https://github.com/jupyter/notebook/blob/4135e3ffe7df627cd9f264f600b2990145af17f6/notebook/services/sessions/sessionmanager.py#L204-L206

We cull idle notebooks after 24 hours. I don't know if that's related. Either way, is this the desired behavior, that they would just get a kernel error when they land on the page? Is there some way to avoid this or trigger auto-refresh of the session, etc.?

kevin-bates:
When kernels are culled, the kernel is terminated in the background,
unbeknownst to the session management.  As a result, invalid sessions
can be produced that appear to exist, yet cannot produce a model from
the persisted row due to the associated kernel no longer being active.
Prior to this change, these sessions, when encountered via a subsequent
call to `get_session()`, would be deleted and a KeyError would be raised.

This change updates the existence check to tolerate those kinds of sessions.
It removes such sessions (as would happen previously), but rather than
raise a KeyError when attempting to convert the row to a dictionary,
it logs a warning and returns None, which then allows `session_exists()`
to return False since the session was removed (as was ultimately the
case previously).

Calls to `get_session()` remain just as before and have the potential
to raise `KeyError` in such cases.  The difference now being that the
`KeyError` is accompanied by a message indicating the cause.

Fixes #4209

### DIFF
--- a/notebook/services/sessions/sessionmanager.py
+++ b/notebook/services/sessions/sessionmanager.py
@@ -59,10 +59,17 @@ class SessionManager(LoggingConfigurable):
     def session_exists(self, path):
         """Check to see if the session of a given name exists"""
         self.cursor.execute("SELECT * FROM session WHERE path=?", (path,))
-        reply = self.cursor.fetchone()
-        if reply is None:
+        row = self.cursor.fetchone()
+        if row is None:
             return False
         else:
+            # Note, although we found a row for the session, the associated kernel may have
+            # been culled or died unexpectedly.  If that's the case, we should delete the
+            # row, thereby terminating the session.  This can be done via a call to
+            # row_to_model that tolerates that condition.  If row_to_model returns None,
+            # we'll return false, since, at that point, the session doesn't exist anyway.
+            if self.row_to_model(row, tolerate_culled=True) is None:
+                return False
             return True
 
     def new_session_id(self):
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #4221, opened 2018-11-17
Tweak upload button to give visual feedback

### LINKED ISSUE #4121: No visual feedback for slow file uploads
Is it possible to configure the [file size that marks the boundary](https://github.com/jupyter/notebook/blob/4f9a6cbdd81ed26a7979d993766d7223e1c72e5c/notebook/static/tree/js/notebooklist.js#L135) between "large" and "not large" files in the upload widget of the notebook?

Currently it is set to 25MB, which even on a fast-ish connection to a remote JupyterHub means I get no visual feedback when uploading 10MB files which is confusing because it takes an appreciable amount of time for the file to upload during which the GUI gives the impression that I didn't click the upload button or need to click again or ...?

What is the downside to always showing the progress bar even for small files? That could be an alternative to lowering/changing the big file size limit.

betatim:
For files below 25MB there was no visual feedback to the user when
uploading a file. This leads to confusion when uploading files that are
big but not huge over a slow network connection.

Closes #4121 not by changing the "large file" limit or introducing a progressbar. Instead it changes the label of the button to "Uploading ..." once the user clicks on it.

kapture 2018-11-17 at 19 33 01

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -425,7 +425,7 @@ define([
             }
         );
     };
-    
+
     NotebookList.prototype.update_location = function (path) {
         this.notebook_path = path;
         $('body').attr('data-notebook-path', path);
@@ -525,7 +525,7 @@ define([
         var item = $("<div/>")
             .addClass("col-md-12")
             .appendTo(row);
-        
+
         var checkbox;
         if (selectable !== undefined) {
             checkbox = $('<input/>')
@@ -624,7 +624,7 @@ define([
       var ipynb_extensions = ['ipynb'];
       return includes_extension(model.path, ipynb_extensions);
     };
-    
+
     NotebookList.prototype._is_editable = function(model) {
       // Allow any file to be "edited"
       // Non-text files will display the following error:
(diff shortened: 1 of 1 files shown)


notebook/static/tree/js/notebooklist.js
```diff
@@ -425,7 +425,7 @@ define([
             }
         );
     };
-    
+
```
betatim: Should I reinstate this trailing white-space? I got automatically removed by my editor and I think that is what you want but happy to revert it.

notebook/static/tree/js/notebooklist.js
```diff
@@ -1434,6 +1434,7 @@ define([
         var upload_button = $('<button/>').text(i18n.msg._("Upload"))
             .addClass('btn btn-primary btn-xs upload_button')
             .click(function (e) {
+                item.find('.upload_button').text("Uploading...");
```
betatim: This is the actual change. My JS is weak so if there is some more idiomatic way of doing this please show me :)

lwasser:
THANK YOU @betatim !! all - uploading files is a real challenge for my students. it's hard to know if the upload if failing or just not working or what is happening and how long to wait. Visual feedback that something is happening would be STELLAR!!! :)

minrk:
Nice! Next step: progress bar?

lwasser:
A progress bar would be awesome @m3nu ... i wish i knew javascript

betatim:
I looked at the progress bar code that is used for large files but then decided that it would be a bigger operation/I'd end up mostly copy&pasting code because I don't understand it well enough. What I was thinking instead is to make the "large file" limit configurable so that for JupyterHub deployments you can lower it so that for uploads that take "a short amount of time" you get "Uploading ..." and for longer ones the progress bar.

Any idea/feeling on which option (progress bar by default or tunable limit) would actually be more work?

PR #4236, opened 2018-11-26
Fix ui-icons from blocking tooltip text

### LINKED ISSUE #4230: Large buttons block view of docstring
Hi, 

Few months ago I upgraded my version of the Jupyter Notebook to 5.7.2. I noticed that when I few docstrings, these large buttons appear (please see screen-shot below), often blocking important keywords. I find myself that I remember many of the keywords first few keywords, but I like viewing the docstring after that, and this makes it a bit harder. 

I've have the same behaviour on Firefox, and Safari (macOS High Sierra). I am not a Chrome user so I can't comment on that. 

Thank you.

buttons-blocking-docstring

steaward:
closes #4230.

screenshot from 2018-11-26 17-38-18

Fix looks like the above.

### DIFF
--- a/notebook/static/notebook/less/tooltip.less
+++ b/notebook/static/notebook/less/tooltip.less
@@ -85,7 +85,9 @@
 .tooltiptext
 {
     /*avoid the button to overlap on some docstring*/
-    padding-right:30px
+    padding-right:30px;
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

Issue #4192, opened 2018-11-11
Is there a keyboard shortcut of next option in completion list of Jupyter Notebook?


roachsinai:
> When I type a '.' after an object name and press the TAB key I am able to see the suggested completions. To confirm and insert a suggested completion I have to use the arrow keys to select the completion I want and then press the ENTER key to confirm my selection. Is there a way to change this so that the TAB key confirms my selection?

Same situation as above which discribed [How do I customize Jupyter Notebook autocomplete accept suggestion keyboard shortcut?](https://stackoverflow.com/questions/43711112/how-do-i-customize-jupyter-notebook-autocomplete-accept-suggestion-keyboard-shor) and [What's the keyboard shortcut of next option in completion list of Jupyter Notebook?](https://stackoverflow.com/questions/48494085/whats-the-keyboard-shortcut-of-next-option-in-completion-list-of-jupyter-notebo)

timporath:
If this is supposed to be the desired behavior, as it is in other environments, I'd like to try and implement this.

roachsinai:
I think <kbd>Tab</kbd> trigger the completions, and second <kbd>Tab</kbd> move selection to next tips is need.

timporath:
You can sorta implement this functionality by changing how Tab behaves in the completer.js file, but there is still a problem with it being reset after it completes, probably by some global handler.

VitanovG:
I know its an old issue but as far as I could find still no such feature got implemented. I at least would love to use your solution on my local machine but I would need some help to point me in the right direction on how? I see only the completer.js file is changed in your commits. I tried replacing all completer.js files on my machine with yours but it did not work. I also tried swapping the enter and tab behavior from the original files, but still no change in behavior when testing. Can someone tell me what am I missing?

PR #4246, opened 2018-11-30
Pressing Tab to cycle through autocomplete selection

timporath:
<kbd>TAB</kbd> can now be used to cycle through the selections of the autocompletion. Its original functionality will still be invoked if a shared start can be found. 
<kbd>TAB</kbd> will also not call carry_on_completion with the true parameter, as in the first invocation of the completer, if the autocompletion is already open.

Mentioned in Issue #4192

### DIFF
--- a/notebook/static/notebook/js/completer.js
+++ b/notebook/static/notebook/js/completer.js
@@ -96,9 +96,8 @@ define([
          * special behavior like autopicking if only one completion available.
          */
         
-        if (this.editor.somethingSelected() || this.editor.getSelections().length > 1) return;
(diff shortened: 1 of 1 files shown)


fhgd:
I would love to use this feature. Please merge it, many thanks in advance :-)

blink1073:
I'm hesitant to change the tab behavior, as it is a major UX change.  Thank you for the suggestion though, @timporath!

fhgd:
What are the reasons against that (major) UX change?

BTW: In IPython is this wanted behaviour standard (tab cycles throuh the autocompletion).

blink1073:
We are minimizing changes to the classic notebook in favor of development of JupyterLab.

PR #4250, opened 2018-12-05
Respect parent template headercontainer in child pages

vilhelmen:
Child pages currently override `headercontainer` blocks from `page`. If, for example, you were to use jupyterhub's `page` template override to label a container image, it won't be shown as all child pages override `headercontainer` without respecting the parent's block. Adding `{{super()}}` allows you to set `headercontainer` in `page` and have it persist in child pages.

Extremely poorly aligned, unformatted example after adding `{{super()}}`:
image

----

Alternatively, you could use imagemagick to build a text image from a build arg in docker and mosaic it with a jupyterhub logo and then replace jupyterhub's `{{logo_url}}` with that image in base64. It works perfectly, but it adds 13KB to each page load in my case and I'd prefer something more minimal like plaintext. But it is a hilarious solution and I love it, so here it is anyway:
```
ARG BASE_CONTAINER=jupyter/minimal-notebook
FROM $BASE_CONTAINER

USER root

RUN apt-get update && apt-get install -yq --no-install-recommends \
   imagemagick pngcrush \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/*
ARG IMAGE_LABEL="UNLABELD CONTAINER IMAGE 🚱"
ADD logo.png /tmp/logo.png
RUN convert -font Courier -pointsize 42 -background none "label:$IMAGE_LABEL" /tmp/label.png \
	&& montage /tmp/logo.png -label '' /tmp/label.png -background none -geometry +1+1 /tmp/built_logo.png \
	&& pngcrush /tmp/built_logo.png /tmp/crushed_logo.png \
	&& sed -e "s|{{logo_url}}|data:image/png;base64,$(base64 /tmp/crushed_logo.png | tr -d '\n')|" -i $(find "$CONDA_DIR/lib/" -path '*/jupyterhub/singleuser.py' | head -n 1)
```
image

### DIFF
--- a/notebook/templates/edit.html
+++ b/notebook/templates/edit.html
@@ -19,13 +19,13 @@ data-file-path="{{file_path}}"
 {% endblock %}
 
 {% block headercontainer %}
-{{super()}}
 
 <span id="save_widget" class="pull-left save_widget">
     <span class="filename"></span>
     <span class="last_modified"></span>
 </span>
 
+{{super()}}
 {% endblock %}
 
 {% block header %}
--- a/notebook/templates/edit.html
+++ b/notebook/templates/edit.html
@@ -19,6 +19,7 @@ data-file-path="{{file_path}}"
 {% endblock %}
 
 {% block headercontainer %}
+{{super()}}
 
 <span id="save_widget" class="pull-left save_widget">
     <span class="filename"></span>
--- a/notebook/templates/notebook.html
+++ b/notebook/templates/notebook.html
@@ -36,7 +36,6 @@ data-notebook-path="{{notebook_path | urlencode}}"
 
 
 {% block headercontainer %}
-{{super()}}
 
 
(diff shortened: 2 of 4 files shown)


vilhelmen:
For better styling, we may want to put `{{super()}}` calls after the flex-span / save widget so users don't confuse the content with the save widget. Opinions?

vilhelmen:
Left and right seem like a good idea, but I'm not sure how much of the templates will have to be messed with to accommodate that. I'm told my site design is very "developer," so I'm probably not the best for this either.

In the meantime, I'm rather pleased with my logo hacks. I've offloaded it from the page contents so it an be cached properly now. If anyone wants/needs it, I've got it over [here](https://github.com/vilhelmen/jh-k8-etc/tree/master/singleuser-shim).

Issue #4244, opened 2018-11-29
Markdown - trying to escape two dollar symbols


billtubbs:
I am trying to add dollar symbols '$' to a markdown cell in a jupyter notebook using the escape character '/'.  Adding one is fine but when you add the second something strange happens to the text in between.  Images below show two cells in edit mode and published mode.

<img width="1092" alt="screen shot 2018-11-29 at 09 15 02" src="https://user-images.githubusercontent.com/7958850/49235660-d43afb00-f3b7-11e8-87e6-a1367983dc47.png">

<img width="1084" alt="screen shot 2018-11-29 at 09 14 50" src="https://user-images.githubusercontent.com/7958850/49235666-d604be80-f3b7-11e8-9d66-27694a507b82.png">

Am I doing something wrong or is this a bug?

syockit:
Using double backslashes solved the problem for me.

tk3369:
Adding double backslashes works for the notebook itself but the extra slash gets in the way when converting to Latex/PDF.  Is there any other workaround?

kwon8543:
In iPython Jupyter notebooks I find that whether you need one backslash vs two backslashes is also dependent on what you've done  in the enclosing markdown. e.g. Headings created with the ## tag at the beginning of the line, need two backslashes to escape, say, a dollar sign, whereas headings created using the `<h1>` html tag only need one backslash to escape, the dollar sign, for example.

So maybe tk3369, you can try avoiding the ## markdown syntax and just use HTML syntax for your comments?

Both of these markdown examples below work fine in Jupyter notebooks:

`<h1>`19. Model performance on products that cost between `\$`50 and `\$`100 `</h1>`

`##` 20. Model performance on products that cost between `\\$`100 and `\\$`200

Interestingly, just while writing the second example, the editor autodetects the  double backslash as a URL in the tooltip (even though it's not a URL, it's escaping a dollar sign)

PR #4257, opened 2018-12-09
Add Documentation regarding backslash escaping in Markdown File

timporath:
Adds a small section on how to generate literal characters of the characters used in the Markdown Syntax.
Also includes how to escape the $ symbol as mentioned in Issue #4244.


minrk:
Thanks!

PR #4266, opened 2018-12-13
Documentation fix lower case

andrea-null-swanson:


notebook/static/notebook/js/quickhelp.js
```diff
@@ -139,9 +139,9 @@ define([
     var special_case = { pageup: i18n.msg._("PageUp"), pagedown: i18n.msg._("Page Down") };
     
     function humanize_key(key){
-        if (key.length === 1){
-            return key.toUpperCase();
```
minrk: Thanks for the PR! These are uppercase on purpose, I think. What is the result before and after this change?

lresende:
Closing based on @minrk comment and lack of updates

PR #4271, opened 2018-12-14
bootstrap 3.4

minrk:
the latest 3.x bugfix release, since we are preparing 5.7.3

and switch from components/ to canonical bootstrap repo, which puts bootstrap.min.js into a `dist` subdir because components hasn't received the 3.4 update.


takluyver:
I'm assuming that the CI would fail if there was a problem with the installation code.

takluyver:

Issue #4270, opened 2018-12-14
When downloaded in .ipynb format notebook is downloaded twice


prajwalMR:
Jupyter notebook is downloaded twice when downloaded in .ipynb format
Jupyter Version : 5.6.0
OS : MacOS Mojave
<img width="1680" alt="screenshot 2018-12-14 at 11 46 52 am" src="https://user-images.githubusercontent.com/15127803/49987040-fb88ef00-ff97-11e8-880e-ca7210cc5e57.png">
<img width="1680" alt="screenshot 2018-12-14 at 11 45 09 am" src="https://user-images.githubusercontent.com/15127803/49987043-fd52b280-ff97-11e8-9f8e-7b8f86f2f48f.png">

PR #4275, opened 2018-12-15
remove multiple download links for the #download_ipynb click

steaward:
This PR is with respect to issue #4270.

A user noticed that when you click the option to download the notebook, it pops open two different windows. One for downloading the .ipynb file, and the other for a .json file. Both of which seem to be the same thing? 

I noticed there was an extra event call for this element compared to all the other drop down items. Removing this call also removed the popup for the 'json' file, but users are still prompted to download the .ipynb file. I believe this is what should occur? Furthermore, it appears that [this](https://github.com/jupyter/notebook/blob/46a887f6e0ecf50f3af5ecadaa05a6ac34675fbb/notebook/static/notebook/js/menubar.js#L133) function handles downloads in all other cases, including the `#download_ipynb` element.

### DIFF
--- a/notebook/static/notebook/js/menubar.js
+++ b/notebook/static/notebook/js/menubar.js
@@ -174,20 +174,6 @@ define([
             that.notebook.save_notebook_as();
             return false;
         });
-        this.element.find('#download_ipynb').click(function () {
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #4287, opened 2018-12-17
updating file tree in "extending notebook" section

choldgraf:
I was going through the "extending the notebook" section and noticed that there's a file tree that's being rendered incorrectly which is a bit confusing. This cleans that up.

Here's the page w/ the bug now: https://jupyter-notebook.readthedocs.io/en/latest/examples/Notebook/Distributing%20Jupyter%20Extensions%20as%20Python%20Packages.html#Automatically-enabling-a-server-extension-and-nbextension

While I'm at it, is it possible to automatically activate a bundler w/ install in the same way following these instructions? They seem to be scoped for nbserver / notebook extensions only

### DIFF
--- a/docs/source/examples/Notebook/Distributing Jupyter Extensions as Python Packages.ipynb
+++ b/docs/source/examples/Notebook/Distributing Jupyter Extensions as Python Packages.ipynb
@@ -284,12 +284,15 @@
    "metadata": {},
    "source": [
     "In addition to the `my_fancy_module` file tree, assume:\n",
-    "- `jupyter-config/`\n",
-    "  - `jupyter_notebook_config.d/`\n",
-    "    - `my_fancy_module.json`\n",
-    "  - `nbconfig/`\n",
-    "    - `notebook.d/`\n",
-    "      - `my_fancy_module.json`"
+    "\n",
+    "```\n",
+    "jupyter-config/\n",
+    "├── jupyter_notebook_config.d/\n",
+    "│   └── my_fancy_module.json\n",
+    "└── nbconfig/\n",
+    "    └── notebook.d/\n",
+    "        └── my_fancy_module.json\n",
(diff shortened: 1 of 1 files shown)


parente:
Haven't tested to see if bundler extensions can be activated the same way. I suspect it'll "just work" since they derive from the same extension config base classes.

/me makes a note to try sometime

choldgraf:
@parente if you could give me a free-form set of steps to try, I can give it a shot and see if it works, and if so, can make a PR to the docs for this...

parente:
Took a quick peek to refresh my memory. Doesn't look like it works today. I ran the classic commands near the bottom of https://jupyter-notebook.readthedocs.io/en/stable/extending/bundler_extensions.html to enable the zip bundler included in the notebook package.

```
jupyter bundlerextension enable --py notebook.bundler.zip_bundler --sys-prefix
```

This wrote a JSON file in `$PREFIX/etc/jupyter/nbconfig/notebook.json` with content:

```
{
  "bundlerextensions": {
    "notebook_zip_download": {
      "label": "IPython Notebook bundle (.zip)",
      "module_name": "notebook.bundler.zip_bundler",
      "group": "download"
    }
  }
}
```

Moving that JSON file to `$PREFIX/etc/jupyter/jupyter_notebook_config.d/zip_bundler.json` didn't appear to work. `jupyter bundlerextension list` shows nothing and I don't see the menu item when I fire up `jupyter notebook`.

Maybe the JSON config structure needs to change? Maybe we need a traitlet like for bundlers  https://github.com/jupyter/notebook/blob/cf1b7b36284dfff8ae55004d9e64789f79e842b5/notebook/notebookapp.py#L1254?

choldgraf:
hmmm - I'm gonna pop this off into an issue :-)

PR #4294, opened 2018-12-19
Add --whole-filesystem command line flag

takluyver:
For security, we don't currently serve files above the notebook directory, normally the working directory where you launch the server. Someone who gets full access can still access the full filesystem by starting a kernel or a terminal, but it limits the possible damage if the attacker can only make plain HTTP requests.

However, this restriction is often frustrating, and people would like to be able to navigate up from the start point, trusting other layers of security. This can be achieved with the correct configuration, but it's not very convenient.

This PR adds a `--whole-filesystem` command line flag which uses the filesystem root `/` as the notebook directory, but opens the browser to the CWD where you launched the notebook. It's meant as a proposal for discussion, not a finished work.

I haven't thought yet about what this does on Windows, or with custom contents managers which don't use a filesystem.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -108,7 +108,9 @@ from jupyter_core.paths import jupyter_runtime_dir, jupyter_path
 from notebook._sysinfo import get_sys_info
 
 from ._tz import utcnow, utcfromtimestamp
-from .utils import url_path_join, check_pid, url_escape, urljoin, pathname2url
+from .utils import (url_path_join, check_pid, url_escape, urljoin, pathname2url,
+                    path2url)
+
 
 #-----------------------------------------------------------------------------
 # Module globals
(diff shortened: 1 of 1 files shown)


takluyver:
Ping @minrk @Carreau for possible interest in this

Carreau:
Sounds reasonable to me. And I'm happy to have it as a linux only, at least as a start.
That might confuse the lab extension, that start to set the starting url to /lab though. 
Maybe wee want a "default location" (equivalent of $HOME) which would be sent to the UI.

dhirschfeld:
I set the `notebook_dir` to the root of the filesystem but direct the `default_url` to users' home directory. I do this because I want my users to have access to the source files in `site-packages` for debugging purposes.
```python
c.DockerSpawner.notebook_dir = 'C:/'
c.DockerSpawner.default_url = '/user/{username}/lab/tree/jupyter'
c.DockerSpawner.volumes = {'J:/{username}': 'C:/jupyter'}
```

I guess the `--whole-filesystem` flag would just be a convenience for the above configuration?

minrk:
@dhirschfeld that's right. This adds a flag that is shorthand for setting two other already existing options in a common configuration. It's not new functionality, just new convenience.

Doing it exactly like this could have some slightly tricky interactions, e.g. with base_url, but it seems a sensible option to me.

takluyver:
Yup, it's convenience, plus a signal that this is a reasonable thing to do.

People seem to be happy with the idea, so I'll try to work out how best to implement it.

blink1073:
@takluyver, should we close this one or open an issue on `jupyter_server`?

jtpio:
Closing as handled in Jupyter Server via `root_dir` + `preferred_dir`.

Thanks all!

PR #4328, opened 2019-01-08
List hidden files if allowed

### LINKED ISSUE #3812: [Feature Request] List hidden files if ContentsManager.allow_hidden=True
***xref:*** https://github.com/jupyterlab/jupyterlab/issues/2049

maxmouchet:
This is an attempt to fix #3812 and jupyterlab/jupyterlab#2049.

If the listing of hidden files and directory is allowed, then we return them in the `api/contents/:` call so that they are displayed in the file manager of the notebook and lab interfaces.

I am not sure about the code style. Feel free to suggest a better way, if any.


notebook/services/contents/filemanager.py
```diff
@@ -334,7 +334,7 @@ def _dir_model(self, path, content=True):
                     self.log.debug("%s not a regular file", os_path)
                     continue
 
-                if self.should_list(name) and not is_file_hidden(os_path, stat_res=st):
+                if self.should_list(name) and ((not is_file_hidden(os_path, stat_res=st)) or self.allow_hidden):
                     contents.append(self.get(
                         path='%s/%s' % (path, name),
                         content=False)
```
dhirschfeld: The logic might be clearer split over a couple of lines:
```python
if self.should_list(name):
    if self.allow_hidden or not is_file_hidden(os_path, stat_res=st):
        contents.append(
            self.get(path='%s/%s' % (path, name), content=False)
        )
```

dhirschfeld:
ping @yuvipanda!

Since you asked on [gitter](https://gitter.im/jupyterlab/jupyterlab?at=5c6713498328315decfea8e1) I think all that is required is for someone to review and merge this PR.

minrk:
Thanks!

PR #4340, opened 2019-01-11
Fix filename to open in notebookapp.py after #4260 and #4265

### LINKED ISSUE #4303: Cannot start jupyter notebook
When I try to start (any) jupyter notebook directly, I get the following error:

      File "/usr/bin/jupyter-notebook", line 11, in <module>
        sys.exit(main())
      File "/usr/lib/python2.7/site-packages/jupyter_core/application.py",     line 266, in launch_instance
        return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
      File "/usr/lib/python2.7/site-    packages/traitlets/config/application.py", line 658, in launch_instance
        app.start()
      File "/usr/lib/python2.7/site-packages/notebook/notebookapp.py",     line 1784, in start
        self.launch_browser()
      File "/usr/lib/python2.7/site-packages/notebook/notebookapp.py",     line 1742, in launch_browser
    TypeError: coercing to Unicode: need string or buffer, int found

However, if I only start jupyter on some directory, I can open notebooks through browser just fine.
I installed jupyter through pip (not using anaconda).

jupyter version: 4.4.0
python version: 2.7.15
OS: Fedora 29

slel:
This also needs to be backported to the 5.7.x branch and should fix #4303.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1747,7 +1747,7 @@ class NotebookApp(JupyterApp):
 
             # Write a temporary file to open in the browser
             fd, open_file = tempfile.mkstemp(suffix='.html')
-            with open(fd, 'w', encoding='utf-8') as fh:
+            with open(open_file, 'w', encoding='utf-8') as fh:
                 self._write_browser_open_file(uri, fh)
         else:
             open_file = self.browser_open_file


takluyver:
That wasn't a mistake, `open()` on Python 3 takes a file descriptor, and closes it when it's done.

We should use [io.open()](https://docs.python.org/2/library/io.html#io.open) for Python 2, which is effectively the Python 3 open() function, and accepts a file descriptor. Can you submit a PR to do this against the 5.7.x branch (master has already dropped Python 2 support, so we don't need any change there).

slel:
Thanks for clarifying. I'll open a PR against the 5.7.x branch
changing `with open` to `with io.open` at line 1742 of the file
`notebook/notebookapp.py`.

What about the occurrence at line 1674, does that one stay as is?

https://github.com/jupyter/notebook/blob/5.7.x/notebook/notebookapp.py#L1742
https://github.com/jupyter/notebook/blob/5.7.x/notebook/notebookapp.py#L1674

PR #4348, opened 2019-01-16
Use io.open in notebookapp.py to fix #4303

### LINKED ISSUE #4303: Cannot start jupyter notebook

slel:
Following the discussion at #4340, in Python 2 we need `io.open`
rather than `open` if using a file descriptor rather than a file name.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1671,7 +1671,7 @@ class NotebookApp(JupyterApp):
     def write_server_info_file(self):
         """Write the result of server_info() to the JSON file info_file."""
         try:
-            with open(self.info_file, 'w') as f:
+            with io.open(self.info_file, 'w') as f:
                 json.dump(self.server_info(), f, indent=2, sort_keys=True)
         except OSError as e:
             self.log.error(_("Failed to write server-info to %s: %s"),
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1739,7 +1739,7 @@ class NotebookApp(JupyterApp):
 
             # Write a temporary file to open in the browser
             fd, open_file = tempfile.mkstemp(suffix='.html')
-            with open(fd, 'w', encoding='utf-8') as fh:
+            with io.open(fd, 'w', encoding='utf-8') as fh:
                 self._write_browser_open_file(uri, fh)
         else:
             open_file = self.browser_open_file
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1671,7 +1671,7 @@ class NotebookApp(JupyterApp):
     def write_server_info_file(self):
         """Write the result of server_info() to the JSON file info_file."""
         try:
-            with io.open(self.info_file, 'w') as f:
+            with io.open(self.info_file, 'w', encoding='utf-8') as f:
                 json.dump(self.server_info(), f, indent=2, sort_keys=True)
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -1671,7 +1671,7 @@ def server_info(self):
     def write_server_info_file(self):
         """Write the result of server_info() to the JSON file info_file."""
         try:
-            with open(self.info_file, 'w') as f:
+            with io.open(self.info_file, 'w') as f:
```
takluyver: Can you add `encoding='utf-8'` here? It's not the same problem, but it would be good to do it while we're looking at this code.

takluyver:
The test failures are because of tornado 6.0a1. I've committed to 5.7.x to limit it to tornado 5.x; I'll close and reopen this pull request to test it again.

notebook/notebookapp.py
```diff
@@ -1671,7 +1671,7 @@ def server_info(self):
     def write_server_info_file(self):
         """Write the result of server_info() to the JSON file info_file."""
         try:
-            with open(self.info_file, 'w') as f:
+            with io.open(self.info_file, 'w', encoding='utf-8') as f:
```
takluyver: Sorry, I think I told you to change the wrong one. This one needs to stay as regular `open()`, because the JSON module on Python 2 produces bytes, not unicode.

I thought we were looking at this line, but it already uses `io.open()`:

https://github.com/jupyter/notebook/blob/0579bc09699e795a8900ddd02105848f5fcb7fcc/notebook/notebookapp.py#L1699

slel:
Grepping for `with open` and `with io.open` in the 5.7.x branch
gives about 30 of each... it might make sense to go through them
systematically and check whether more `open` should become `io.open`
and whether more `encoding='utf-8'` should be added.

```
$ grep -r "with open" .
$ grep -r "with io.open" .

$ grep -r "with open" . | wc -l
$ grep -r "with io.open" . | wc -l
```

notebook/notebookapp.py
slel: If commit f8d7d40 has all that is needed, can you use only that commit
and leave out the further two commits in this branch when you merge?

takluyver: Yup. I'll open a new PR so it re-runs the CI tests with just that.

takluyver: See #4349

takluyver:
Merged #4349

PR #4355, opened 2019-01-20
Fix case sensitive notebook file extension 

EvgeniDubov:
In case a notebook file extension is not all lower case (e.g. NOTE.IPYNB) then notebook will identify it as plain text.
My fix is making the notebook file extension check to be case insensitive

### DIFF
--- a/notebook/files/handlers.py
+++ b/notebook/files/handlers.py
@@ -57,7 +57,7 @@ class FilesHandler(IPythonHandler):
             self.set_attachment_header(name)
         
         # get mimetype from filename
-        if name.endswith('.ipynb'):
+        if name.lower().endswith('.ipynb'):
             self.set_header('Content-Type', 'application/x-ipynb+json')
         else:
             cur_mime = mimetypes.guess_type(name)[0]


takluyver:
We might need to do something to make this work correctly in Turkish locales, where there are [two different kinds of i](https://en.wikipedia.org/wiki/Dotted_and_dotless_I), becauses I think `IPYNB` there will be lowercased to `ıpynb` (missing the dot). I'm not sure how to easily work around that.

Otherwise, this seems reasonable to me, I'll give it a day or two unless anyone else has a different opinion.

To be clear, I'd still regard lower-case `.ipynb` as the canonical file extension, and I'd recommend that any tools writing a notebook file do so with that extension.

EvgeniDubov:
Thanks for the feedback @takluyver 

Interesting point about the Turkish latter, thank you.

Regarding the upper case extension, I totally agree with your recommendation.
The way I found the issue was that someone sent me a notebook file with lower case extension by email but outlook changed the attached file extension to upper case.
You can see a discussion on that issue here:
[https://superuser.com/questions/1207156/disable-attachment-file-name-extension-case-conversion-in-outlook](https://superuser.com/questions/1207156/disable-attachment-file-name-extension-case-conversion-in-outlook)

PR #4374, opened 2019-01-30
Improve comms.rst example

mlucool:
This improves the example by making it more copy/pastable out of the box and demoing how you can send a message at various places. It also fixes a poor usage of shadowed variables


takluyver:
Thanks!

PR #4376, opened 2019-01-30
surface config.d nbserver_extensions to the NotebookApp config object

mpacer:
This PR comes from a discussion that arose around surfacing information about whether a server_extension is enabled from within another server_extension when the server_extension has been automatically enabled through the use of a `jupyter_notebook_config.d/` directory. 

Basically if you autoenable a server_extension, it's never discovered by the core JupyterApp, and so while the server_extension is enabled, this information doesn't appear in the resulting config object surfaced within `config.NotebookApp.nbserver_extensions`.

This arose when trying to surface whether [bookstore](https://github.com/nteract/bookstore)(a s3 publishing service) is enabled in nteract (see discussion in https://github.com/nteract/nteract/pull/4144). 

- separates nbserver_extension config loading into new 
init_server_extension_config method 
- adds init_server_extension_config to the initialize funciton before 
the init_webapp call
- adds nbserver_extension configuration found in config.d files (by the 
BaseJSONConfigLoader) to both the underlying NotebookApp config object 
and the self.nbserver_extensions value
- makes self.nbserver_extensions the canonical location for identifying 
all nbserver_extensions rather than temporary extensions variable

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1539,15 +1539,13 @@ class NotebookApp(JupyterApp):
         # TODO: this should still check, but now we use bower, not git submodule
         pass
 
-    def init_server_extensions(self):
-        """Load any extensions specified by config.
+    def init_server_extension_config(self):
+        """Consolidate server extensions specified by all configs.
 
-        Import the module, then call the load_jupyter_server_extension function,
-        if one exists.
+        The resulting list is stored on self.nbserver_extensions and updates config object.
         
         The extension API is experimental, and may change in future releases.
         """
-        
         # TODO: Remove me in notebook 5.0
         for modulename in self.server_extensions:
             # Don't override disable state of the extension if it already exist
@@ -1566,15 +1564,23 @@ class NotebookApp(JupyterApp):
         manager = ConfigManager(read_config_path=config_path)
         section = manager.get(self.config_file_name)
         extensions = section.get('NotebookApp', {}).get('nbserver_extensions', {})
-
-        for modulename, enabled in self.nbserver_extensions.items():
-            if modulename not in extensions:
-                # not present in `extensions` means it comes from Python config,
-                # so we need to add it.
-                # Otherwise, trust ConfigManager to have loaded it.
-                extensions[modulename] = enabled
-
+        
         for modulename, enabled in sorted(extensions.items()):
+            if modulename not in self.nbserver_extensions:
+                self.config.NotebookApp.nbserver_extensions.update({modulename: enabled})
+                self.nbserver_extensions.update({modulename: enabled})
+
+    def init_server_extensions(self):
+        """Load any extensions specified by config.
+
+        Import the module, then call the load_jupyter_server_extension function,
+        if one exists.
+        
+        The extension API is experimental, and may change in future releases.
+        """
+        
+
+        for modulename, enabled in sorted(self.nbserver_extensions.items()):
             if enabled:
                 try:
                     mod = importlib.import_module(modulename)
@@ -1632,6 +1638,7 @@ class NotebookApp(JupyterApp):
         if self._dispatching:
             return
         self.init_configurables()
+        self.init_server_extension_config()
         self.init_components()
         self.init_webapp()
         self.init_terminals()
--- a/notebook/tests/test_serverextensions.py
+++ b/notebook/tests/test_serverextensions.py
@@ -147,6 +147,7 @@ class TestInstallServerExtension(MockEnvTestCase):
         toggle_serverextension_python('mockext_both', enabled=False, user=True)
 
         app = NotebookApp(nbserver_extensions={'mockext_py': True})
+        app.init_server_extension_config()
         app.init_server_extensions()
 
         assert mock_user.loaded


notebook/notebookapp.py
```diff
@@ -1539,15 +1539,13 @@ def init_components(self):
         # TODO: this should still check, but now we use bower, not git submodule
         pass
 
-    def init_server_extensions(self):
-        """Load any extensions specified by config.
+    def init_server_extension_config(self):
+        """Consolidate server extensions specified by all configs.
```
rgbkrk: This wording is much clearer to me. 👍

notebook/notebookapp.py
```diff
@@ -1566,15 +1564,23 @@ def init_server_extensions(self):
         manager = ConfigManager(read_config_path=config_path)
         section = manager.get(self.config_file_name)
         extensions = section.get('NotebookApp', {}).get('nbserver_extensions', {})
+        
+        for modulename, enabled in sorted(extensions.items()):
+            if modulename not in self.nbserver_extensions:
+                self.config.NotebookApp.nbserver_extensions.update({modulename: enabled})
+                self.nbserver_extensions.update({modulename: enabled})
```
rgbkrk: 👍

notebook/notebookapp.py
```diff
@@ -1566,15 +1564,23 @@ def init_server_extensions(self):
         manager = ConfigManager(read_config_path=config_path)
         section = manager.get(self.config_file_name)
         extensions = section.get('NotebookApp', {}).get('nbserver_extensions', {})
+        
+        for modulename, enabled in sorted(extensions.items()):
+            if modulename not in self.nbserver_extensions:
+                self.config.NotebookApp.nbserver_extensions.update({modulename: enabled})
```
minrk: This makes it very likely that `self.config.NotebookApp.nbserver_extensions` will be a LazyConfigValue, which is not meant to be interrogated directly, so you will still have trouble if you ever try to talk to the config object directly.

notebook/notebookapp.py
```diff
@@ -1566,15 +1564,23 @@ def init_server_extensions(self):
         manager = ConfigManager(read_config_path=config_path)
         section = manager.get(self.config_file_name)
         extensions = section.get('NotebookApp', {}).get('nbserver_extensions', {})
+        
+        for modulename, enabled in sorted(extensions.items()):
+            if modulename not in self.nbserver_extensions:
```
minrk: I think this inverts priority when config is defined at both conf.d and top-level. What we had before, was:

`if in nbserver_extensions and not in extensions, load it` (conf.d extensions has higher priority)

What we have here is:

`if in extensions and not in self.nbserver_extensions` (conf.d extensions has *lower* priority)

This resolves differently when a given config value is defined at both levels.

If you'd like to add a test that verifies this (would be super helpful!), try loading config where `jupyter_notebook_conifg.py` enables an extension and `jupyter_notebook_config.d/test.json` disables it. The result should be a disabled extension.

notebook/notebookapp.py
minrk: 👍 This is the most important change, I think. `self.nbserver_extensions` is the thing to check, and that wasn't properly updated. It should never be assumed that `self.config.NotebookApp.nbserver_extensions` contains the final value of the extensions to be loaded.

notebook/notebookapp.py
mpacer: when you say priority do you mean literal priority in terms of the Dict itself or in terms of overriding?

I had thought that user settings should override the defaults. 

If self.nbserver_extensions already has this modulename then we defer to the user's configuration loaded from the app. Otherwise we adhere to what we find in the `jupyter_notebook_config.d/` directory.

mpacer: > If you'd like to add a test that verifies this (would be super helpful!), try loading config where `jupyter_notebook_conifg.py` enables an extension and `jupyter_notebook_config.d/test.json` disables it.

That seems backwards to me, that means that as long as that library is installed I can never override its values.

Specifically in the case I'm imagining

`jupyter_notebook_conifg.py` disables an extension and `jupyter_notebook_config.d/test.json` enables the extension… If we always enable the extension there is no way for a user to override that library's default to enabling when it is installed.

rgbkrk: Where I think this gets murky is if a user's config ends up making their notebook servers not launchable by orchestrators like JupyterHub. As an operator I like to know that a deployment will reliably come up and if there are issues they get surfaced so that we can resolve them. In the current state of affairs like the inconsistency that's being addressed in this PR, the lack of config either doesn't surface (a jupyter server extension doesn't hard stop the server if improperly configured) or ends up buried away in the notebook server logs.

mpacer: > Where I think this gets murky is if a user's config ends up making their notebook servers not launchable by orchestrators like JupyterHub.

That feels like a different case in my mind than what is happening in the `jupyter_notebook_config.d` which is often not defined by deployers but by library maintainers as part of their data_files to their `setuptools.setup` call.

I would agree that there should be a case for some way to override user configuration available to deployers. But that shouldn’t be the same convention that libraries are using to ensure their extensions are loaded. 

Consider that if the person were to upgrade a library that autoenabled an extension by default (an extension that had been explicitly disabled by the deployed).  The user would then overwrite the configuration placed as part of deployment in the `jupyer_notebook_config.d`, and that would mean a user could inadvertently change deployment config without having any clue that it was happening.

mpacer: However I didn’t intend for this to be a change in behaviour so I’ll switch it back to the previous logic even if I disagree with that logic’s appropriateness to the purpose of this mechanism.

rgbkrk:
What do you think now @minrk?

notebook/notebookapp.py
minrk: Thanks for thinking through it. I think it's tricky to say whether conf.d vs top-level is 'user config' vs not. The priority was based on other conf.d-type systems (profile.d, etc.) where the top-level config is loaded first, followed by `.d/*`, thus giving .d higher priority. That said, in practice today, conf.d is typically written by installers, while config.{py|json} is written by user actions. It does indeed make sense for that to have highest priority in that case.

I'll go ahead and merge this one.

PR #4412, opened 2019-02-19
Enable restart_kernel for async usage

kevin-bates:
Converted `MappingKernelManager.restart_kernel` to a coroutine so that projects that take advantage of async kernel startup can also realize appropriate behavior relative to restarts.  To take full advantage of async kernel startup (with restarts), this PR along with its "[sibling PR](https://github.com/jupyter/jupyter_client/pull/425)" will be required.

Please note, however, that each of these PRs can be independently installed w/o affecting today's notebook applications (as far as I can determine via testing the possible combinations).  That said, BOTH of these PRs will be required for usage by Enterprise Gateway - which incurs very long kernel startup times - or other projects that require concurrent kernel starts.

It would be ideal to have this PR back-ported to python 2.7-relative release since EG has an immediate need.

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -302,10 +302,11 @@ class MappingKernelManager(MultiKernelManager):
 
         return super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now)
 
+    @gen.coroutine
     def restart_kernel(self, kernel_id):
         """Restart a kernel by kernel_id"""
         self._check_kernel_id(kernel_id)
-        super(MappingKernelManager, self).restart_kernel(kernel_id)
+        yield gen.maybe_future(super(MappingKernelManager, self).restart_kernel(kernel_id))
         kernel = self.get_kernel(kernel_id)
         # return a Future that will resolve when the kernel has successfully restarted
         channel = kernel.connect_shell()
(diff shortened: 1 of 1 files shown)


kevin-bates:
@minrk - thank you for merging.  

What is the best way to request a back-port to 5.7.x since it would be ideal if this could be used on Python 2.x (along with its sibling PR https://github.com/jupyter/jupyter_client/pull/425)?

I'd be happy to submit a PR on the 5.7.x branch if that's all that is necessary?

cc: @takluyver

minrk:
I started backporting it to 5.7.x, but since it's a backward-incompatible API change, that wouldn't be appropriate, so I've marked it for 6.0.

kevin-bates:
This is really unfortunate.  This essentially means that our gateway 1.x line can never support async kernel startup since those customers still require Python 2 which can't coexist with NB 6. 

Do you know of applications that call `restart_kernel directly` besides Enterprise Gateway (which does so via inheritance)?  I understand you can't know the complete answer, just curious if its non-zero.

minrk:
Ah! I misread this PR. `restart_kernel` is already returning a Future, so we should be okay backporting this one. The jupyter-client PRs are still more complicated, but I think this one will be okay. I have to go, but I'll look into this one a bit more in the morning.

kevin-bates:
Thanks great news Min - thank you.  Do you have a timeframe for the 6.0 releases for NB and JC?  We'd like to coordinate our EG 2.0 GA with those if they're in the ballpark of end of March.

PR #4425, opened 2019-02-25
support async kernel shutdown

kevin-bates:
This is an "add-on" commit to PR #4412 in which kernel startup was addressed.  As with that PR, this should be considered a "sibling PR" to the same [jupyter_client PR](https://github.com/jupyter/jupyter_client/pull/425) - which I just updated with a shutdown-related commit.

I should have looked into shutdowns sooner.  I thought it would be more involved than it was. Sorry about that.

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -283,6 +283,7 @@ class MappingKernelManager(MultiKernelManager):
             self.log.info("Discarding %s buffered messages for %s",
                 len(msg_buffer), buffer_info['session_key'])
 
+    @gen.coroutine
     def shutdown_kernel(self, kernel_id, now=False):
         """Shutdown a kernel by kernel_id"""
         self._check_kernel_id(kernel_id)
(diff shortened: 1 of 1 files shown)


kevin-bates:
@minrk - since it sounds like we're going to need a release to cap tornado < 6.x, I'm hoping this PR could be included into a 5.7.x tag.  We really need this, along with #4412 (and the jupyter_client PR) for our 1.x line in Enterprise Gateway.

lresende:
I believe what @kevin-bates means is: could this be also merged into 5.7.x branch and be part of the possible release that is going to handle the tornado version cap.

kevin-bates:
I may be seeing some issues relative to the shutdown changes during restarts.  Need to take a closer look.  Sorry.

notebook/services/kernels/kernelmanager.py
```diff
@@ -300,7 +301,8 @@ def shutdown_kernel(self, kernel_id, now=False):
             type=self._kernels[kernel_id].kernel_name
         ).dec()
 
-        return super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now)
+        ret = yield gen.maybe_future(super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now))
+        raise gen.Return(ret)
```
minrk: This already supported async shutdown_kernel without any change, so I don't think we should merge this. Since `shutdown_kernel` is not itself async, returning the `super().shutdown_kernel()` works if shutdown_kernel is async, since it will return a Future or coroutine object, which gets passed through to the caller of this method.

kevin-bates:
That's fine.  If need be we can revisit since I had to back out the jupyter_client portion of this since there were restart side-effects that require further investigation.  Closing.

PR #4431, opened 2019-02-26
Update gateway support with recent changes

kevin-bates:
Convey notebook working directory to the gateway ([nb2kg pr-21](https://github.com/jupyter/nb2kg/pull/21))
- This allows users of containerized kernels to mount home directories on the containers/pods so as to mimic behaviors as if the kernel was local.

Support retrieval of kernelspec resources from the gateway ([nb2kg pr-23](https://github.com/jupyter/nb2kg/pull/23))
- This simply enables the three kinds of kernelspec resources (kernel.js, kernel.css, logo-*) to be returned when running against a gateway. 
- Since gateway returns kernelspecs in _model form_, we now test if the kernel spec info meets the model criteria (rather than add additional (and duplicate) handlers).
- Requires Jupyter Kernel Gateway version 2.2.0 otherwise no resources are returned (as before the change).

### DIFF
--- a/notebook/gateway/managers.py
+++ b/notebook/gateway/managers.py
@@ -91,6 +91,16 @@ class GatewayClient(SingletonConfigurable):
     def _kernelspecs_endpoint_default(self):
         return os.environ.get(self.kernelspecs_endpoint_env, self.kernelspecs_endpoint_default_value)
 
+    kernelspecs_resource_endpoint_default_value = '/kernelspecs'
+    kernelspecs_resource_endpoint_env = 'JUPYTER_GATEWAY_KERNELSPECS_RESOURCE_ENDPOINT'
+    kernelspecs_resource_endpoint = Unicode(default_value=kernelspecs_resource_endpoint_default_value, config=True,
+        help="""The gateway endpoint for accessing kernelspecs resources
+            (JUPYTER_GATEWAY_KERNELSPECS_RESOURCE_ENDPOINT env var)""")
+
+    @default('kernelspecs_resource_endpoint')
+    def _kernelspecs_resource_endpoint_default(self):
+        return os.environ.get(self.kernelspecs_resource_endpoint_env, self.kernelspecs_resource_endpoint_default_value)
+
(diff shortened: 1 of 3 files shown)


notebook/gateway/managers.py
```diff
@@ -540,9 +551,32 @@ def get_kernel_spec(self, kernel_name, **kwargs):
...
+        try:
+            response = yield gateway_request(kernel_spec_resource_url, method='GET')
+        except HTTPError as error:
+            if error.code == 404:
+                kernel_spec_resource = None
+            else:
+                raise
+        else:
+            kernel_spec_resource = response.body
+        raise gen.Return(kernel_spec_resource)
```
minrk: FWIW, since we require Python 3, we can use `return kernel_spec_resource`

notebook/gateway/managers.py
```diff
@@ -91,6 +91,16 @@ def _kernels_endpoint_default(self):
     def _kernelspecs_endpoint_default(self):
         return os.environ.get(self.kernelspecs_endpoint_env, self.kernelspecs_endpoint_default_value)
 
+    kernelspecs_resource_endpoint_default_value = '/kernelspecs'
```
minrk: Does every endpoint need to be separately configurable? It seems to me like a single base path should be everything that's needed.

notebook/gateway/managers.py
kevin-bates: Ok - thanks.  Since this is one of many gen.Returns, I think it makes sense to change these whenever we "make the sweep" to move to async/await - if that's okay with you.

notebook/gateway/managers.py
kevin-bates: I believe these are configurable (in NB2KG is done via just envs) for cases when the gateway server (JKG or JEG) is behind a reverse proxy.  This should be the last of them. (yeah, famous last words, I know. :smile:)

minrk: ok, let's switch this over to a single configurable prefix before adding any more after this one.

PR #4479, opened 2019-03-14
Add support for async kernel management

kevin-bates:
This PR adds support for asynchronous kernel management in an optional manner.  This approach will allow the project as a whole to move forward while applications that wish to leverage the new functionality can do so at their convenience until the appropriate time.  At some point, we should converge the implementations to async only.

Async kernel management is dependent on changes in [jupyter_client PR 428](https://github.com/jupyter/jupyter_client/pull/428).  In that PR, both `MultiKernelManager` and `KernelManager` have been extended with direct subclasses `AsyncMultiKernelManager` and `AsyncKernelManager`.  Same goes for the pair of IOLoop classes.  Please view that PR for details.

With these changes, the MappingKernelManager has been refactored as follows:
1. A separate, and disjoint, class has been introduced named `AsyncMappingKernelManager`.  This class is essentially the equivalent of `MappingKernelManager`, but for complete async behavior.  It is disjoint because it must derive from `AsyncMultiKernelManager`, so cannot derive from notebook's `MappingKernelManager` (which derives from `MultiKernelManager`).
2. To minimize duplication, the buffering, activity monitoring and culling pieces have been factored into separate internal classes such that the two respective mapping kernel manager classes have a _has a_ relationship to each.  Note that in order to facilitate any existing subclasses of `MappingKernelManager`, the existing methods relative to each of these classes remain in place in case they were overridden in subclasses.  It would be good to determine if there are overrides (somehow).  If none exist, those place-holder methods should be removed.
3. To further minimize duplication, a second superclass has been introduced - `MappingKernelManagerBase`. This class is needed for a couple reasons.
	1. It holds config properties and methods that do not require access to the true super classes (since those are disjoint).  
	2. It serves has the required class for class-based traits relative to either of the MappingKernelManager classes.
	Once we converge back to a single mapping kernel manager, we can remove this class - moving its contents back onto the one mapping kernel manager class.

To switch to using `AsyncMappingKernelManager`, the following command-line option is necessary:

```
--NotebookApp.kernel_manager_class=notebook.services.kernels.kernelmanager.AsyncMappingKernelManager
```

leveraging the existing ability to customize kernel manager classes.

Also note, that code exists to NOT require upgrade of `juptyer_client` until the project is ready to update dependencies.  As a result, no upgrade of jupyter_client is required unless the kernel_manager_class override is in place.  In addition, should someone enable that option without having upgraded jupyter_client, the following RuntimeError will be produced (along with its traceback) when Notebook is started:

_RuntimeError: Using `AsyncMappingKernelManager` without an appropriate jupyter_client installed!  Upgrade jupyter_client and try again._

Similarly for the notebook kernel api tests.  If the appropriate jupyter_client is not available, the async-related tests will be skipped - logging an appropriate message.  Those tests derive directly from the non-async tests by merely setting the `NotebookApp.kernel_manager_class`.

Note: If `AsyncMappingKernelManager` is in use, then any command-line or configuration options referencing `MappingKernelManager` need to be replaced with `AsyncMappingKernelManager`.  If you need to switch back and forth, it may be easier to just reference `MappingKernelManagerBase` instead.

Here's a class diagram, extending the one shown in the jupyter_client PR.
notebook_kernel_manager_hierarchy

Although this has been tested with and without the appropriate jupyter_client present, I'd like to keep this PR as WIP until a POC with Enterprise Gateway has been completed and the jupyter_client PR has been resolved.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -107,7 +107,7 @@ from jupyter_core.paths import jupyter_runtime_dir, jupyter_path
 from notebook._sysinfo import get_sys_info
 
 from ._tz import utcnow, utcfromtimestamp
-from .utils import url_path_join, check_pid, url_escape, urljoin, pathname2url
+from .utils import url_path_join, check_pid, url_escape, urljoin, pathname2url, run_sync
 
 # Check if we can use async kernel management
 try:
@@ -1801,11 +1801,7 @@ class NotebookApp(JupyterApp):
         n_kernels = len(self.kernel_manager.list_kernel_ids())
         kernel_msg = trans.ngettext('Shutting down %d kernel', 'Shutting down %d kernels', n_kernels)
         self.log.info(kernel_msg % n_kernels)
-        # If we're using async kernel management, we need to invoke the async method via the event loop.
-        if isinstance(self.kernel_manager, AsyncMappingKernelManager):
-            asyncio.get_event_loop().run_until_complete(self.kernel_manager.shutdown_all())
-        else:
-            self.kernel_manager.shutdown_all()
+        run_sync(self.kernel_manager.shutdown_all())
 
     def notebook_info(self, kernel_count=True):
         "Return the current working directory and the server url information"
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1390,14 +1390,16 @@ class NotebookApp(JupyterApp):
             connection_dir=self.runtime_dir,
             kernel_spec_manager=self.kernel_spec_manager,
         )
-        #  Ensure the appropriate jupyter_client is in place.
+        #  Ensure the appropriate version of Python and jupyter_client is available.
         if isinstance(self.kernel_manager, AsyncMappingKernelManager):
+            if sys.version_info < (3, 6):
+                raise ValueError("You are using `AsyncMappingKernelManager` in Python 3.5 (or lower) "
+                                 "which is not supported. Please upgrade Python to 3.6+ or change kernel managers.")
             if not async_kernel_mgmt_available:
                 raise ValueError("You are using `AsyncMappingKernelManager` without an appropriate "
-                                 "jupyter_client installed!  Upgrade jupyter_client or change kernel managers.")
-            else:
-                self.log.info("Asynchronous kernel management has been configured to use '{}'.".
-                              format(self.kernel_manager.__class__.__name__))
+                                 "jupyter_client installed!  Please upgrade jupyter_client or change kernel managers.")
+            self.log.info("Asynchronous kernel management has been configured to use '{}'.".
+                          format(self.kernel_manager.__class__.__name__))
 
         self.contents_manager = self.contents_manager_class(
             parent=self,
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1392,10 +1392,10 @@ class NotebookApp(JupyterApp):
         )
         #  Ensure the appropriate version of Python and jupyter_client is available.
         if isinstance(self.kernel_manager, AsyncMappingKernelManager):
-            if sys.version_info < (3, 6):
+            if sys.version_info < (3, 6):  # Can be removed once 3.5 is dropped.
                 raise ValueError("You are using `AsyncMappingKernelManager` in Python 3.5 (or lower) "
                                  "which is not supported. Please upgrade Python to 3.6+ or change kernel managers.")
-            if not async_kernel_mgmt_available:
+            if not async_kernel_mgmt_available:  # Can be removed once jupyter_client >= 6.1 is required.
                 raise ValueError("You are using `AsyncMappingKernelManager` without an appropriate "
                                  "jupyter_client installed!  Please upgrade jupyter_client or change kernel managers.")
             self.log.info("Asynchronous kernel management has been configured to use '{}'.".
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -75,7 +75,7 @@ from notebook import (
 
 from .base.handlers import Template404, RedirectWithParams
 from .log import log_request
-from .services.kernels.kernelmanager import MappingKernelManager
+from .services.kernels.kernelmanager import MappingKernelManager, AsyncMappingKernelManager, MappingKernelManagerBase
 from .services.config import ConfigManager
 from .services.contents.manager import ContentsManager
 from .services.contents.filemanager import FileContentsManager
@@ -1177,6 +1177,7 @@ class NotebookApp(JupyterApp):
 
     kernel_manager_class = Type(
         default_value=MappingKernelManager,
+        klass=MappingKernelManagerBase,
         config=True,
         help=_('The kernel manager class to use.')
     )
@@ -1380,6 +1381,13 @@ class NotebookApp(JupyterApp):
             connection_dir=self.runtime_dir,
             kernel_spec_manager=self.kernel_spec_manager,
         )
+        #  Ensure the appropriate jupyter_client is in place.
+        #  TODO: remove once dependencies are updated.
+        if isinstance(self.kernel_manager, AsyncMappingKernelManager):
+            if not hasattr(self.kernel_manager, 'list_kernel_ids'):
+                raise RuntimeError("Using `AsyncMappingKernelManager` without an appropriate "
+                                   "jupyter_client installed!  Upgrade jupyter_client and try again.")
+
         self.contents_manager = self.contents_manager_class(
             parent=self,
             log=self.log,
@@ -1897,8 +1905,8 @@ class NotebookApp(JupyterApp):
             assembled_url = urljoin('file:', pathname2url(open_file))
         else:
             assembled_url = url_path_join(self.connection_url, uri)
-        
-        b = lambda: browser.open(assembled_url, new=self.webbrowser_open_new)                            
+
+        b = lambda: browser.open(assembled_url, new=self.webbrowser_open_new)
         threading.Thread(target=b).start()
 
     def start(self):
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -7,6 +7,7 @@
 from __future__ import absolute_import, print_function
 
 import notebook
+import asyncio
 import binascii
 import datetime
 import errno
@@ -583,7 +584,7 @@ class NotebookApp(JupyterApp):
     flags = flags
     
     classes = [
-        KernelManager, Session, MappingKernelManager, KernelSpecManager,
+        KernelManager, Session, MappingKernelManager, AsyncMappingKernelManager, KernelSpecManager,
         ContentsManager, FileContentsManager, NotebookNotary,
         GatewayKernelManager, GatewayKernelSpecManager, GatewaySessionManager, GatewayClient,
     ]
@@ -1392,10 +1393,10 @@ class NotebookApp(JupyterApp):
         #  Ensure the appropriate jupyter_client is in place.
         if isinstance(self.kernel_manager, AsyncMappingKernelManager):
             if not async_kernel_mgmt_available:
-                raise ValueError("You're using `AsyncMappingKernelManager` without an appropriate "
+                raise ValueError("You are using `AsyncMappingKernelManager` without an appropriate "
                                  "jupyter_client installed!  Upgrade jupyter_client or change kernel managers.")
             else:
-                self.log.info("Asynchronous kernel management has been configured via '{}'.".
+                self.log.info("Asynchronous kernel management has been configured to use '{}'.".
                               format(self.kernel_manager.__class__.__name__))
 
         self.contents_manager = self.contents_manager_class(
@@ -1800,7 +1801,11 @@ class NotebookApp(JupyterApp):
         n_kernels = len(self.kernel_manager.list_kernel_ids())
         kernel_msg = trans.ngettext('Shutting down %d kernel', 'Shutting down %d kernels', n_kernels)
         self.log.info(kernel_msg % n_kernels)
-        self.kernel_manager.shutdown_all()
+        # If we're using async kernel management, we need to invoke the async method via the event loop.
+        if isinstance(self.kernel_manager, AsyncMappingKernelManager):
+            asyncio.get_event_loop().run_until_complete(self.kernel_manager.shutdown_all())
+        else:
+            self.kernel_manager.shutdown_all()
 
     def notebook_info(self, kernel_count=True):
         "Return the current working directory and the server url information"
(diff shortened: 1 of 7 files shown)


kevin-bates:
The POC with Enterprise Gateway has completed so I've removed the [WIP] status.  There are some test failures in python 3.7 that I'm struggling to figure out since I can't reproduce them in my 3.7 env.  In addition, they appear to be socket related - so I'm struggling to see the link.  Any help/guidance would be appreciated.

meeseeksmachine:

https://discourse.jupyter.org/t/scalable-enterprise-gateway/2014/2

davidbrochart:
That's still a lot of code duplication between `MappingKernelManager` and `AsyncMappingKernelManager`:
- `start_kernel`: identical in both classes, except that one is using newer async/await syntax.
- `restart_kernel`: same as above.
- `kernel_mode`: identical.
- `list_kernels`: identical.
- `_check_kernel_id`: identical.
- `shutdown_kernel`: this one is blocking in `MappingKernelManager` and async in `AsyncMappingKernelManager`, but I think there is no reason not to make it async in `MappingKernelManager` too.

Instead of having two classes, I was thinking of keeping one `MappingKernelManager` class, that would be a subclass of a `MaybeAsyncMultiKernelManager`, which would be set to `AsyncMultiKernelManager` if jupyter_client has it, or `MultiKernelManager` if not.
What do you think? Am I missing something?

kevin-bates:
I wish I could do that, and perhaps there's some fancy python way of doing this, but this has to do with the jupyter_client classes that they must derive from. I mention this in item 1 of the PR description:
> A separate, and disjoint, class has been introduced named AsyncMappingKernelManager. This class is essentially the equivalent of MappingKernelManager, but for complete async behavior. It is disjoint because it must derive from AsyncMultiKernelManager, so cannot derive from notebook's MappingKernelManager (which derives from MultiKernelManager).

davidbrochart:
Maybe I'm not understanding well. I was thinking about something simple like:
```python
try:
    from jupyter_client import AsyncMultiKernelManager
    MaybeAsyncMultiKernelManager = AsyncMultiKernelManager
except ImportError:
    MaybeAsyncMultiKernelManager = MultiKernelManager

class MappingKernelManager(MaybeAsyncMultiKernelManager):
    ...
```
Wouldn't it work?

kevin-bates:
Sorry, I guess I missed that portion of your original response.  This is interesting and one of the things about python I just can't get used to.  I think it would work - or could be made to work.  

The issue is that it doesn't provide the ability for the user to opt-in to using async - which we need.  Another issue is that any custom managers that derive from the sync `MappingKernelManager` would break when suddenly (and unknowingly) derived from the async form of `MappingKernelManager`. Both of these issues would occur the moment the updated `jupyter_client` is installed.

However, it does seem that `kernel_model()`, `list_kernels()` and `_check_kernel_id()` could move to the mixin (`MappingKernelManagerBase`).

Btw, I think I might have the mixin class declared in the wrong position.  I believe it needs to proceed each of the other juptyer_client superclasses.  Is that correct?

davidbrochart:
Good point, I forgot about async being an opt-in.

Zsailer:
> Note: If AsyncMappingKernelManager is in use, then any command-line or configuration options referencing MappingKernelManager need to be replaced with AsyncMappingKernelManager. If you need to switch back and forth, it may be easier to just reference MappingKernelManagerBase instead.

Is this by design or consequence of the inheritance problem? 

In theory, we could make a "shim" class for `MappingKernelManager` when users opt in to `AsyncMappingKernelManager`. Do we expect that this is going to cause a lot of headache for users? 

More importantly, will traits with the `MappingKernelManager` prefix be ignored silently when using `AsyncMappingKernelManager`?

notebook/notebookapp.py
```diff
@@ -576,7 +584,7 @@ class NotebookApp(JupyterApp):
     flags = flags
     
     classes = [
-        KernelManager, Session, MappingKernelManager, KernelSpecManager,
+        KernelManager, Session, MappingKernelManager, AsyncMappingKernelManager, KernelSpecManager,
```
Zsailer: Are all traits between `MappingKernelManager` and `AsynMappingKernelManager` the same? It seems confusing to always expose both sets of traits to the CLI if you can only use one or the other.

notebook/services/kernels/kernelmanager.py
```diff
@@ -19,33 +19,42 @@
...
+       Float, Instance, default, validate)
+from traitlets.config.configurable import LoggingConfigurable
 
 from notebook.utils import maybe_future, to_os_path, exists
 from notebook._tz import utcnow, isoformat
 from ipython_genutils.py3compat import getcwd
 
 from notebook.prometheus.metrics import KERNEL_CURRENTLY_RUNNING_TOTAL
 
+# Since use of AsyncMultiKernelManager is optional at the moment, don't require appropriate juptyer_client.
```
Zsailer:
```suggestion
# Since use of AsyncMultiKernelManager is optional at the moment, don't require appropriate jupyter_client.
```

Zsailer:
> I wish I could do that, and perhaps there's some fancy python way of doing this, but this has to do with the jupyter_client classes that they must derive from. I mention this in item 1 of the PR description:

I absolutely hate to admit this, but I think this can be solved using metaclasses. 

Normally, I would strongly avoid metaclasses; but because of the confusion this might create in config (thanks, traitlets), think metaclasses might shrink this code down a ton.

Zsailer:
@kevin-bates, let me take a crack at it today and I'll PR against your branch.

notebook/notebookapp.py
kevin-bates: With the exception of the default value for [`kernel_manager_class`](https://github.com/kevin-bates/notebook/blob/async-subclasses/notebook/services/kernels/kernelmanager.py#L266) - yes - they are the same.  All traits are physically defined on the `MappingKernelManagerBase` mixin.

My concern was not to impact existing configs - which will be using `MappingKernelManager` as the "trait class".

I'm sure there's some trailet voodoo that we can take advantage of but I'm not up on how to go about doing this.  How can a trait associated with `MappingKernelManager` (albeit defined on the mixin) be available to `AsyncMappingKernelManager`?

notebook/services/kernels/kernelmanager.py
kevin-bates: :+1:

kevin-bates:
Thanks Zach.  I find metaclass stuff more confusing than they're worth, but please feel free.  I will always try to err on the side of maintainability.

I guess I'm not quite following what confusion there is.  No one is using `AsyncMappingKernelManager` today.  To use `AsyncMappingKernelManager` they must "opt-in" by setting `c.NotebookApp.kernel_manager_class = AsyncMappingKernelManager`.  Doesn't it make sense that they would then need to adjust the relative traits to that class?  If they switch back, they might want different values for each class type and both can be defined, but only the "in-use" class's traits will take effect - is that right?

notebook/notebookapp.py
Zsailer: I'm diving into the code now to explore this further. We need `MappingKernelManager` to be inserted somewhere in `AsyncMappingKernelManager`'s MRO, but we need to do this cautiously. I know you've explored this already, so I likely will reach your same conclusion haha. But maybe it's worth having someone else look at the code more deeply.

kevin-bates: ok - thanks.  That doesn't feel like the right thing to do from a class hierarchy perspective, but let's see where it takes this.

Zsailer:
> To use AsyncMappingKernelManager they must "opt-in" by setting c.NotebookApp.kernel_manager_class = AsyncMappingKernelManager. Doesn't it make sense that they would then need to adjust the relative traits to that class?

Hm, you're right. The main confusion is that both mapping kernel managers are listed in `NotebookApp` classes list, and therefore listed in `--help-all` list at the same time. That's a pretty small price, I agree, but it does create some cognitive load on users exploring the traits in our help docs.

Right now, if you write a custom kernel_manager_class, you would inherit and override `MappingKernelManager` methods, so the traits would be handled properly and the help documentation would still be consistent.

I think we should follow this pattern here. AsyncMappingKernelManager should essentially be a subclass of MappingKernelManager.

Because you have the issue of a different class you're inheriting from jupyter_client, namely `AsyncMultiKernelManager`, we will have to build this new class using metaclasses, where we swap out the `MultiKernelManager` for `AsyncMultiKernelManager`.

davidbrochart:
@Zsailer, @kevin-bates, just to let you know, I'm experimenting over in https://github.com/jupyter/jupyter_server/pull/191 with similar ideas.

kevin-bates:
@Zsailer - I believe @davidbrochart has solved this in an elegant manner in https://github.com/jupyter/jupyter_server/pull/191.  I think it would be good that the two repos essentially use the same code in this area so let's just replace this PR with David's.

Zsailer:
Sounds good. Thanks, Kevin and David!

kevin-bates:
Hmm - since the tests are a bit different between server (pytest) and notebook (nose), it might be best to just adjust the class hierarchy of he multi-kernel managers in this PR and keep everything else as is.

@davidbrochart - do you mind if I do that and co-author you in on this PR?

davidbrochart:
Yes, please go ahead. No need to co-author me in.

kevin-bates:
I think we need to revisit David's proposed class hierarchy approach: https://github.com/jupyter/jupyter_server/pull/191#issuecomment-602899754

kevin-bates:
I've pushed changes that switch to the approach taken in https://github.com/jupyter/jupyter_server/pull/191 after we addressed the concerns I raised in the previous comment.  This push also includes the display of the notebook server's version when starting (mentioned for the sake of transparency).

kevin-bates:
@Zsailer, @davidbrochart, @lresende could you please review this PR.  It's essentially the same as https://github.com/jupyter/jupyter_server/pull/191 except that the changes to async/await are focused on kernel management and not other services in order to minimize changes.

Once approved and merged, I think we should cut a `6.1.0` release soon after as other projects are awaiting (like that? :smile:) this functionality.

notebook/notebookapp.py
```diff
@@ -1782,7 +1801,11 @@ def cleanup_kernels(self):
         n_kernels = len(self.kernel_manager.list_kernel_ids())
         kernel_msg = trans.ngettext('Shutting down %d kernel', 'Shutting down %d kernels', n_kernels)
         self.log.info(kernel_msg % n_kernels)
-        self.kernel_manager.shutdown_all()
+        # If we're using async kernel management, we need to invoke the async method via the event loop.
+        if isinstance(self.kernel_manager, AsyncMappingKernelManager):
+            asyncio.get_event_loop().run_until_complete(self.kernel_manager.shutdown_all())
+        else:
+            self.kernel_manager.shutdown_all()
```
davidbrochart: Maybe you want to be more cautious, and use the convenient [run_sync](https://github.com/jupyter/jupyter_server/blob/a2610d33be535b42e37e3c200bcec6afeec341d8/jupyter_server/utils.py#L403-L442) we will use in jupyter_server:
```python
run_sync(self.kernel_manager.shutdown_all())
```

notebook/services/kernels/kernelmanager.py
```diff
@@ -117,7 +123,7 @@ def _default_kernel_buffers(self):
         help="The last activity on any kernel, including shutting down a kernel")
 
     def __init__(self, **kwargs):
-        super(MappingKernelManager, self).__init__(**kwargs)
+        super().__init__(**kwargs)
         self.last_kernel_activity = utcnow()
```
davidbrochart: Should be removed?

notebook/services/kernels/kernelmanager.py
```diff
@@ -130,6 +136,11 @@ def __init__(self, **kwargs):
     # Methods for managing kernels and sessions
     #-------------------------------------------------------------------------
 
+    def __init__(self, **kwargs):
+        self.super = MultiKernelManager
```
Zsailer: Clever! I was slightly uncomfortable with this when I first saw it—I thought you just wanted an alias for `super(...)`, but I now recognize this is the magic for the class hierarchy. 

That said, I have two recommendations... 

1. I think we need to add some thorough inline comments here (and below, in `AsyncMultiKernelManager`. Especially using the attr name `super`, I'm afraid someone reading this a few years from now might think this is a mistake... that you actually mean super().
2. Maybe even consider renaming this attribute? Something like `base_multikernel_manager` or something that makes it clear you're calling a class other than super().

notebook/services/kernels/kernelmanager.py
Zsailer: Haha I just added mentioned this too! :+1:

notebook/services/kernels/kernelmanager.py
davidbrochart: I agree it could be confusing to name it `super`, my bad!

kevin-bates:
@Zsailer - Any idea why Travis is not happening?  It is happening on my repo at least: https://travis-ci.org/github/kevin-bates/notebook

Zsailer:
Hm, that's weird. It's not showing up here, but I see it running on [travis-ci.org](https://travis-ci.org/github/jupyter/notebook/builds/667822515).

Zsailer:
@kevin-bates Travis is all green. This PR is good to go from my perspective.

Zsailer:
Is this ready for merge, @kevin-bates ?

kevin-bates:
Thanks for asking Zach.  I was just about to add the logic to prevent async kernel management on 3.5 - like we're doing elsewhere in the ecosystem.  I'll ping you once completed.

kevin-bates:
Hi @Zsailer - this should be ready for review/merge.  

Aside from adding the checks for 3.5, I found the test-skipping logic wasn't working correctly because the server startup occurs as part of the setup_class() method, while I had been "skipping" the test in the setUp() method - which was too late.  The python 3.5 group now shows this: `SKIP: AsyncKernelAPITest tests skipped due to Python < 3.6!`

I also needed to add an initializer to the fake AsyncMultiKernelManager class that's fabricated. This allows the initial load (of the fabricated class) to succeed in order for the server to determine that an async kernel manager is being configured when it's not available/supported.  

Lastly, the test failure appears to be transient related to `nvm` installation.  I got the same installation failure on a different group in my repo.  [I reran the build](https://travis-ci.org/github/kevin-bates/notebook/builds/669930951) and the `nvm` installation succeeded.

kevin-bates:
Trigger CI

Zsailer:
This looks good, @kevin-bates. I'm ready to merge.

kevin-bates:
@Zsailer - per your suggestion in today's server meeting, I added comments indicating when a given dependency check (jupyter_client/Python versions) can be removed.  Thanks for the tip.

Zsailer:
@kevin-bates Awesome work here! This is good to go. Merging away!

PR #4513, opened 2019-03-26
Fix regressions in 5.7.x

minrk:
Fixes some closed-socket issues with direct-checks for closed or closing sockets, which seems to change across tornado versions.

for backport to 5.7.x

debugging issues in https://github.com/jupyterlab/jupyterlab/pull/6138


minrk:
In testing  jupyterlab/jupyterlab#6138, I also discovered a regression caused by #4412, where the async restart_kernel was not properly awaited. Instead, the Future was *returned* from the coroutine instead of the result of the future.

minrk:
Tests passed prior to rebase which fixed only the selenium and docs tests. Merging this so we can get 5.7.7 out.

PR #4520, opened 2019-03-27
wait for kernel to be ready in selenium tests

minrk:
might fix recent flaky tests


minrk:
seems to have worked

minrk:
This didn't need backporting

Issue #4467, opened 2019-03-10
notebook gives blank page when start 


andrewsunst:
It used to work fine until I changed python to lower version today(3.7 to 3.6). The jupyter notebook keeps giving me blank page when started. I changed the version back and it doesn't help. No error in the terminal. The error in the browser are as follows:
`Refused to execute script from '<URL>' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:1 Refused to execute script from 'http://localhost:8888/static/components/es6-promise/promise.min.js?v=f004a16cb856e0ff11781d01ec5ca8fe' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:1 Refused to execute script from 'http://localhost:8888/static/components/preact/index.js?v=00a2fac73c670ce39ac53d26640eb542' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:1 Refused to execute script from 'http://localhost:8888/static/components/proptypes/index.js?v=c40890eb04df9811fcc4d47e53a29604' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:1 Refused to execute script from 'http://localhost:8888/static/components/preact-compat/index.js?v=aea8f6660e54b18ace8d84a9b9654c1c' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:1 Refused to execute script from 'http://localhost:8888/static/components/requirejs/require.js?v=951f856e81496aaeec2e71a1c2c0d51f' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:1 Refused to execute script from 'http://localhost:8888/static/tree/js/main.min.js?v=ab9f3a6cf8347df927864d58cfad7931' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
VM26 tree:24 Uncaught ReferenceError: require is not defined
    at VM26 tree:24`

I check some other posts which said to clean browser caches and cookies which doesn't help for my case. Any idea about how to fix this problem?

andrewsunst:
Or any idea about how to completely remove all the notebook files?

joddiy:
Me too, with python3.7 at windows10

windyslove:
I have the same issue , python3.7 ,win10

andrewsunst:
> Me too, with python3.7 at windows10.
[...]

PR #4528, opened 2019-03-28
explicitly ignore windows registry for mimetypes

minrk:
since it has proven untrustworthy for many users, causing issues like #4467


takluyver:
Seems reasonable.

PR #4531, opened 2019-03-30
Changed the default protocol

### LINKED ISSUE #4427: SSL Version
Notebook is setting an explicit version of SSL as the default here: https://github.com/jupyter/notebook/blob/bba82ca3bdb90ff6382dccef5fb3a805facf623a/notebook/notebookapp.py#L1424

This is overriding the [default version for ssl](https://docs.python.org/2/library/ssl.html#ssl.wrap_socket) of `PROTOCOL_SSLv23` (which is now named `PROTOCOL_TLS`). I think the ssl default is better as it is [documented as](https://docs.python.org/2/library/ssl.html#ssl.PROTOCOL_TLS)
> Selects the highest protocol version that both the client and server support. Despite the name, this option can select “TLS” protocols as well as “SSL”.

This comes up as TLS 1 is now considered poor, just like the thing that line was meant to protect against, and chrome has deprecated it: https://www.chromestatus.com/feature/5654791610957824.

Can we consider either removing that line and/or moving to be explicitly `ssl.PROTOCOL_TLS`?

xmnlab:
Resolves #4427  

Changed PROTOCOL_TLSv1 to PROTOCOL_TLS

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1421,10 +1421,11 @@ class NotebookApp(JupyterApp):
             # SSL may be missing, so only import it if it's to be used
             import ssl
             # PROTOCOL_TLS selects the highest ssl/tls protocol version that both the client and
-            # server support. When PROTOCOL_TLS is not available use PROTOCOL_SSLv23
+            # server support. When PROTOCOL_TLS is not available use PROTOCOL_SSLv23.
+            # PROTOCOL_TLS is new in version 2.7.13, 3.5.3 and 3.6
             ssl_options.setdefault(
                 'ssl_version',
-                getattr('ssl', 'PROTOCOL_TLS', ssl.PROTOCOL_SSLv23)
+                getattr(ssl, 'PROTOCOL_TLS', ssl.PROTOCOL_SSLv23)
             )
             if ssl_options.get('ca_certs', False):
                 ssl_options.setdefault('cert_reqs', ssl.CERT_REQUIRED)
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1420,12 +1420,12 @@ class NotebookApp(JupyterApp):
         else:
             # SSL may be missing, so only import it if it's to be used
             import ssl
-            # Disable SSLv3 by default, since its use is discouraged.
-            _ssl_protocol = (
-                ssl.PROTOCOL_TLS if hasattr(ssl, 'PROTOCOL_TLS') 
-                else ssl.PROTOCOL_SSLv23
+            # PROTOCOL_TLS selects the highest ssl/tls protocol version that both the client and
+            # server support. When PROTOCOL_TLS is not available use PROTOCOL_SSLv23
+            ssl_options.setdefault(
+                'ssl_version',
+                getattr('ssl', 'PROTOCOL_TLS', ssl.PROTOCOL_SSLv23)
             )
-            ssl_options.setdefault('ssl_version', _ssl_protocol)
             if ssl_options.get('ca_certs', False):
                 ssl_options.setdefault('cert_reqs', ssl.CERT_REQUIRED)
         
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1421,7 +1421,11 @@ class NotebookApp(JupyterApp):
             # SSL may be missing, so only import it if it's to be used
             import ssl
             # Disable SSLv3 by default, since its use is discouraged.
-            ssl_options.setdefault('ssl_version', ssl.PROTOCOL_TLSv1)
+            _ssl_protocol = (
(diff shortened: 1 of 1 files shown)


takluyver:
The idea seems good, but the name is new in Python 3.6, and we're still supporting 3.5 at present. Can you include some compatibility code so it uses `PROTOCOL_SSLv23` if `PROTOCOL_TLS` is not found?

xmnlab:
thanks @takluyver I did the changes you suggested. it is done for a review.

notebook/notebookapp.py
```diff
@@ -1421,7 +1421,10 @@ def init_webapp(self):
             # SSL may be missing, so only import it if it's to be used
             import ssl
             # Disable SSLv3 by default, since its use is discouraged.
-            ssl_options.setdefault('ssl_version', ssl.PROTOCOL_TLSv1)
+            ssl_options.setdefault(
+                'ssl_version', 
+                ssl.PROTOCOL_TLS if hasattr(ssl, PROTOCOL_TLS) else ssl.PROTOCOL_SSLv23
```
takluyver: `hasattr()` takes a string as the second argument. Though it might be neater to use `getattr()` with a default.

xmnlab: > hasattr() takes a string as the second argument. 
you're right. thanks for catch that up. I fixed that it is done for review. 

> Though it might be neater to use getattr() with a default.
sorry I didn't understand how to use this default

takluyver: Like `getattr(ssl, 'PROTOCOL_TLS', ssl.PROTOCOL_SSLv23)`. But the hasattr version is fine too.

notebook/notebookapp.py
```diff
@@ -1421,7 +1421,11 @@ def init_webapp(self):
             # SSL may be missing, so only import it if it's to be used
             import ssl
             # Disable SSLv3 by default, since its use is discouraged.
```
takluyver: Can you update this accompanying comment to describe what we're now using and why?

notebook/notebookapp.py
xmnlab: oh I see ... that is much better thanks. I forgot getattr has default argument. thanks. I will change that.

notebook/notebookapp.py
xmnlab: sure I am going to do that now

xmnlab:
@takluyver I am reviewing the documentations here and python35 already supports `ssl.PROTOCOL_TLS` and  `PROTOCOL_SSLv23` is deprecated.

```
ssl.PROTOCOL_TLS
Selects the highest protocol version that both the client and server support. Despite the name, this option can select “TLS” protocols as well as “SSL”.

New in version 3.5.3.

ssl.PROTOCOL_SSLv23
Alias for data:PROTOCOL_TLS.

Deprecated since version 3.5.3: Use data:PROTOCOL_TLS instead.
```

and as I could find here python 3.5.5 is available on conda-forge and pkgs/main

I will keep `PROTOCOL_SSLv23` (because maybe it should need a pinning on the libraries that depend on `notebook`) but if you think it is ok to remove that, let me know.

takluyver:
Ah, OK. As it's pretty easy to fall back to `PROTOCOL_SSLv23`, let's leave that in for now just in case anyone has an older Python 3.5 for some reason. It would be good to mention in the comment what version needs that, so it's clear when the code can be simplified again.

notebook/notebookapp.py
```diff
@@ -1420,8 +1420,12 @@ def init_webapp(self):
         else:
             # SSL may be missing, so only import it if it's to be used
             import ssl
-            # Disable SSLv3 by default, since its use is discouraged.
-            ssl_options.setdefault('ssl_version', ssl.PROTOCOL_TLSv1)
+            # PROTOCOL_TLS selects the highest ssl/tls protocol version that both the client and
+            # server support. When PROTOCOL_TLS is not available use PROTOCOL_SSLv23
+            ssl_options.setdefault(
+                'ssl_version',
+                getattr('ssl', 'PROTOCOL_TLS', ssl.PROTOCOL_SSLv23)
```
takluyver: Sorry, the first parameter for getattr should be the module object (i.e. `ssl` with no quotes). It's only the attribute name that needs to be a string.

xmnlab: you're totally right. sorry about that. I am still trying to have a functional dev environment here.
that is why I am pushing things without tested locally. sorry again T_T

takluyver:
Thanks, that looks good now. I'm about to head out, but I'll merge this tomorrow assuming the CI passes (I suspect this code path isn't tested, so...)

xmnlab:
thank you so much @takluyver ! yes it seems it is not tested ...

takluyver:
Thanks for your patience on getting this in.

xmnlab:
Thanks @takluyver for your patience and your guidance :)

PR #4536, opened 2019-04-01
Use separately packaged sphinxcontrib_github_alt

takluyver:
And remove the bundled version of the extension


hroncok:
I was about to ask why this code is here. Thanks for opening this issue.

PR #4576, opened 2019-04-20
Fix binary message issue when configured for gateway forwarding

SolarisYan:
If message is bytes, we should set binary=True
This is from PR https://github.com/jupyter/nb2kg/pull/33


kevin-bates:
@SolarisYan - I just noticed the PR title isn't very specific - there are multiple `handlers.py` files within notebook.  In order to help the maintainers (and general readers), would you mind updating the title to something along the lines of: _Fix binary message issue when configured for gateway forwarding_

SolarisYan:
@kevin-bates Sure, thanks for your reminding

Issue #1027, opened 2016-01-29
Not able to connect to remote Jupyter Notebook from Jupyterhub


pawannrao-ebay:
I am spawning a remote notebook on a docker container and passing the IP and port info to the spawner class, but I m not able to connect to it eventhough all the config seems right from jupyterhub and jupyter notebook stand point. I have done all the essentials like allowing all ips, trusting x-header configuration etc. Let me know if you can make sense out of this error.

The curl also works for the remote notebook:(curl http://172.X.Y.Z:PORT)

```
[I 2016-01-29 23:09:20.490 JupyterHub base:281] User test1 server took 83.033 seconds to start
[I 2016-01-29 23:09:20.491 JupyterHub orm:170] Adding user test1 to proxy /user/test1 => http://172.20.193.152:9009
[I 2016-01-29 23:09:20.511 JupyterHub log:100] 302 GET /hub/user/test1/ (test1@10.65.63.36) 83078.03ms
[I 2016-01-29 23:09:20.745 JupyterHub log:100] 302 GET /static/components/requirejs/require.js?v=6da8be361b9ee26c5e721e76c6d4afce (@10.65.63.36) 3.30ms
[I 2016-01-29 23:09:20.767 JupyterHub log:100] 302 GET /custom/custom.css (@10.65.63.37) 1.97ms
[I 2016-01-29 23:09:20.774 JupyterHub log:100] 302 GET /static/components/es6-promise/promise.min.js?v=f004a16cb856e0ff11781d01ec5ca8fe (@10.65.63.36) 0.68ms
[I 2016-01-29 23:09:20.786 JupyterHub log:100] 302 GET /static/components/jquery-ui/themes/smoothness/jquery-ui.min.css?v=9b2c8d3489227115310662a343fce11c (@10.65.63.39) 0.67ms
[I 2016-01-29 23:09:20.792 JupyterHub log:100] 302 GET /static/components/jquery-typeahead/dist/jquery.typeahead.min.css?v=7afb461de36accb1aa133a1710f5bc56 (@10.65.63.41) 0.70ms
[I 2016-01-29 23:09:20.801 JupyterHub log:100] 302 GET /static/style/style.min.css?v=807166c81ded9672f91872a576572dcd (@10.65.63.40) 0.69ms
[W 2016-01-29 23:09:20.811 JupyterHub log:100] 404 GET /hub/custom/custom.css (test1@10.65.63.37) 3.29ms
[...]

PR #4588, opened 2019-04-25
fix duplicates in download as menu

### LINKED ISSUE #4560: Duplicate Entries in Download as Menu
My apologies if this is already open here.

Duplicate entries of exporters appear in the download as menu. See [nbconvert #938 comment](https://github.com/jupyter/nbconvert/issues/938#issuecomment-482229310). In theory it can be fixed on either the nbconvert or notebook side. The consensus as nbconvert is that it is a notebook problem.

t-makaro:
fixes #4560. nbconvert 5.4.0 implemented the ```export_from_notebook``` to the built-in exporters. This means that due to the hardcoding of the built-in exporters in the menu that they are duplicated. I also included a safeguard to ensure that exporters explicitly define an ```export_from_notebook``` that is different from its superclass, so that custom exporters must be explicitly added to the menu, and don't create confusion with duplicate names.

Before:
before

After:
image

Unfortunately, currently nbconvert 5.4 does not define ```export_from_notebook``` for the html exporter and it inherits ```export_from_notebook``` meaning that it will not appear in the list after this change. This was however fixed in [nbconvert 5.5](https://github.com/jupyter/nbconvert/pull/1001) which released today.

Should you wish to put in this change without relying on nbconvert 5.5, I can remove the safeguard that checks the superclass. (the html exporter will then appear as ```custom (.html)``` until the nbconvert fix).

### DIFF
--- a/notebook/notebook/handlers.py
+++ b/notebook/notebook/handlers.py
@@ -20,18 +20,43 @@ def get_frontend_exporters():
 
     ExporterInfo = namedtuple('ExporterInfo', ['name', 'display'])
 
-    for name in sorted(get_export_names()):
+    default_exporters = [
+        ExporterInfo(name='html', display='html (.html)'),
+        ExporterInfo(name='latex', display='latex (.tex)'),
+        ExporterInfo(name='markdown', display='markdown (.md)'),
+        ExporterInfo(name='notebook', display='notebook (.ipynb)'),
+        ExporterInfo(name='pdf', display='pdf (.pdf)'),
+        ExporterInfo(name='rst', display='rst (.rst)'),
+        ExporterInfo(name='script', display='script (.txt)'),
+        ExporterInfo(name='slides', display='slides (.slides.html)')
+    ]
+
+    frontend_exporters = []
+    for name in get_export_names():
         exporter_class = get_exporter(name)
         exporter_instance = exporter_class()
         ux_name = getattr(exporter_instance, 'export_from_notebook', None)
-        # ensure export_from_notebook is explicitly defined & not inherited
         super_uxname = getattr(super(exporter_class, exporter_instance),
-                               'export_from_notebook',
-                               None)
+                               'export_from_notebook', None)
+
+        # Ensure export_from_notebook is explicitly defined & not inherited
         if ux_name is not None and ux_name != super_uxname:
             display = _('{} ({})'.format(ux_name,
                                          exporter_instance.file_extension))
-            yield ExporterInfo(name, display)
+            frontend_exporters.append(ExporterInfo(name, display))
+
+    # Ensure default_exporters are in frontend_exporters if not already
+    # This protects againts nbconvert versions lower than 5.5
+    names = set(exporter.name.lower() for exporter in frontend_exporters)
+    for exporter in default_exporters:
+        if exporter.name.lower() not in names:
+            frontend_exporters.append(exporter)
+
+    # Protect against nbconvert 5.4.x
+    template_exporter = ExporterInfo(name='custom', display='custom (.txt)')
+    if template_exporter in frontend_exporters:
+        frontend_exporters.remove(template_exporter)
+    return sorted(frontend_exporters)
 
 
 class NotebookHandler(IPythonHandler):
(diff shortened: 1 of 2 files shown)


MSeal:
I think we need to guard against nbconvert < 5.5 and make sure it at least doesn't fail, and probably write a test for the older nbconvert access pattern.

t-makaro:
The screenshot is with nbconvert 5.4.1. Anything lower than 5.4 with this method will produce an empty list. The only way that I can think to fix that is to check if ```nbconvert.__version__``` is less than ```5.5``` (or 5.4.1). And if so, then re-add the hard-coding menu options (either by added an ```if``` in the template or a second loop in the ```get_frontend_exporters``` to yield hard coded values from). Thoughts?

MSeal:
Maybe we could try the new method and only inject the old hard-coded ones if the list is empty after that?

MSeal:
That would at least be more pythonic than checking the version explicitly

t-makaro:
Yes. I'm trying to do that, but I'm still finding that I have a duplicate "Python" export option.

MSeal:
Make a best effort. So long as the safety guard has a comment about why we're using a pattern here it'll be fine.

MSeal:
LGTM when the tests pass

t-makaro:
I just pushed what I currently have. For the menu, I'm seeing this:
image

Notice duplicate python, and no script exporter. My ```get_frontend_exporters``` function is however returning the proper thing.
```
[ExporterInfo(name='asciidoc', display='asciidoc (.asciidoc)'),
ExporterInfo(name='html', display='html (.html)'),
ExporterInfo(name='latex', display='latex (.tex)'),
ExporterInfo(name='markdown', display='markdown (.md)'),
ExporterInfo(name='notebook', display='notebook (.ipynb)'),
ExporterInfo(name='pdf', display='pdf (.pdf)'),
ExporterInfo(name='python', display='python (.py)'),
ExporterInfo(name='rst', display='rst (.rst)'),
ExporterInfo(name='script', display='script (.txt)'),
ExporterInfo(name='slides', display='slides (.slides.html)')]
```

MSeal:
@takluyver would you mind merging this for the next release?

t-makaro:
Ok, if I remove ```ExporterInfo(name='script', display='script (.txt)')``` from the list that ```get_frontend_exporters``` returns, then the duplicate python disappears. It appears that the script exporter getting renamed to "Python (.py)" is an issue that was already there before this change, so it is not due to this PR. Something else appears to be renaming it.

t-makaro:
Is changing the name of the script exporter to the language name a feature? If I had another language kernel on my system I would test it out, but if that's the case then the python exporter should not have defined ```export_from_notebook```. We also need to do something about it being capitalized (or the other ones not being capitalized).

t-makaro:
Yep. It is [intentional](https://github.com/jupyter/notebook/blob/f4183bb6121d88f139fd92e1a0255b9f9e19ba22/notebook/static/notebook/js/menubar.js#L387-L397). Given this. I think that the python exporter should not define ```export_from_notebook```. 

To be consistent with the other exporters. I also think that we should not be capitalizing the language name that is replacing "script", or we could capitalize the display names for the other exporters, but if we go that route then I think that is best done on nbconvert's side (otherwise pdf might be capitalized as "Pdf" not "PDF" and similarly for "html").

Thoughts? I will update my PR here, and make a PR over at nbconvert, if needed, once I have some feedback.

t-makaro:
Ok this is currently what the menu looks like:

After [nbconvert #1027](https://github.com/jupyter/nbconvert/pull/1027) is merged:


With nbconvert 5.5.0:


With nbconvert 5.4.x:
5 4 x

With nbconvert <5.4:


Only thing to note is the with nbconvert 5.3 the asciidoc exporter does not appear in the menu. This could be fixed with added another ```ExporterInfo``` to the ```default_exporters```, but it didn't appear in the menu until 5.4 anyway, so I'd rather not hard code another exporter into the menu since it's not a regression.

MSeal:
Yep we'll get the naming capitalization fixed in 5.5.1 release then. I think this is a good solution across versions and will land everyone in the right best place we can given version differences.

I don't have write permissions to this repo atm, so @takluyver would probably need to review and merge for us.

takluyver:
Thanks, this looks OK to me.

blink1073:
@meeseeksdevbox please backport to 5.7.x

blink1073:
@meeseeksdev please backport to 5.7.x

blink1073:

PR #4604, opened 2019-05-03
Add folder selector to the move files functionality

jmg:
I added a folder selector for the move functionality. It will make a lot easier to move files/folders through different directories without having to manually type their names.

selector

As a new text was added in the code ("Or select a folder:") it will require translation for the i18n support. Could you please point me in the right direction to do this?

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -1046,9 +1046,55 @@ define([
             return;
         }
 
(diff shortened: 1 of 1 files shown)


jtpio:
Closing as such change would have to be implemented in the `nbclassic` repo if still needed: https://github.com/jupyter/nbclassic

Please open an issue or PR there if it is still relevant, thanks!

PR #4612, opened 2019-05-07
Convert dualmode_clipboard to selenium

LuisARodr:
A test for the [#3335](https://github.com/jupyter/notebook/issues/3335) issue.
A new function was created on utils.py to get the mode of the cells


LuisARodr:
Closed because I'm working on the test `dualmode_execute` and changed some more code of `utils.py` so `dualmode_clipboard` had to change to use the `utils.py` changes. But this test is going into the `dualmode_execute` PR.

PR #4615, opened 2019-05-08
Convert dualmode_execute, dualmode_clipboard & dualmode_markdown to selenium

LuisARodr:
Test conversion for the [#3335 issue.](https://github.com/jupyter/notebook/issues/3335)
The fixed version of the [#4612 PR](https://github.com/jupyter/notebook/pull/4612) is in this PR
New functions where added to `utils.py`.


LuisARodr:
Added the test dualmode_markdown, this test also makes use of the validate_notebook_state function, so that is the reason it is on the same PR as the other tests

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +394,71 @@ def trigger_keystrokes(browser, *keys):
             ac.perform()
         else:              # single key stroke. Check if modifier eg. "up"
             browser.send_keys(getattr(Keys, keys[0].upper(), keys[0]))
+
+def validate_notebook_state(notebook, mode, index):
```
takluyver: Can we have a docstring here? What exactly does the function validate? Maybe it can have a clearer name, for that matter?

Also, I think the whole thing could be simplified and cleaned up to Python code style standards. It doesn't look like it's ever called with `index=None`, so maybe it doesn't need to handle that.

notebook/tests/selenium/utils.py
```diff
@@ -222,6 +222,14 @@ def set_cell_metadata(self, index, key, value):
         JS = 'Jupyter.notebook.get_cell({}).metadata.{} = {}'.format(index, key, value)
         return self.browser.execute_script(JS)
 
+    def get_cells_mode(self):
+        JS = 'return Jupyter.notebook.get_cells().map(function(c) {return c.mode;})'
+        return self.browser.execute_script(JS)
+
+    def get_cells_length(self):
```
takluyver: Not sure it's worth adding this for one call, when code can do `len(notebook)`.

notebook/tests/selenium/utils.py
```diff
@@ -222,6 +222,14 @@ def set_cell_metadata(self, index, key, value):
         JS = 'Jupyter.notebook.get_cell({}).metadata.{} = {}'.format(index, key, value)
         return self.browser.execute_script(JS)
 
+    def get_cells_mode(self):
```
takluyver: Ditto, I'm not sure it's clearer to have a method for this when it's only called once.

notebook/tests/selenium/utils.py
LuisARodr: Solved the issue with the fccc7d7 commit

notebook/tests/selenium/utils.py
LuisARodr: Solved with the 3a0cb7b commit

notebook/tests/selenium/utils.py
LuisARodr: Worked on this on 3a0cb7b, I'm new to python so I couldn't find how to simplify the code by python standards so if there are some more specific changes I can work on them.

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
             ac.perform()
         else:              # single key stroke. Check if modifier eg. "up"
             browser.send_keys(getattr(Keys, keys[0].upper(), keys[0]))
+
+def validate_dualmode_state(notebook, mode, index):
+    '''Validate the entire dual mode state of the notebook.  
+    Make sure no more than one cell is selected, focused, in edit mode, etc...
```
takluyver: I'd rather this described exactly what the function checks, without saying "etc"

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
...
+        #None of the cells are in edit mode
+        if(index == None):
+            for mode in cells_mode:
+                if(mode == 'edit'):
+                    return False
+            return True
+        #Only the index cell is on edit mode
+        i = 0
+        for mode in cells_mode:
+            if(i == index):
```
takluyver: Python doesn't need parentheses for `if` - you can write conditions like:

```python
if i == index:
```

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
...
+def validate_dualmode_state(notebook, mode, index):
+    '''Validate the entire dual mode state of the notebook.  
+    Make sure no more than one cell is selected, focused, in edit mode, etc...
+
+    '''
+    def is_only_cell_edit(index):
+        JS = 'return Jupyter.notebook.get_cells().map(function(c) {return c.mode;})'
+        cells_mode = notebook.browser.execute_script(JS)
+        #None of the cells are in edit mode
+        if(index == None):
```
takluyver: By convention, we check for None using `index is None` rather than `index == None`.

This can matter in some situations, because objects can override the `==` operator and do strange things. But it's also just a common convention for Python code.

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
...
+
+        assert is_focused_on(None)#no focused cells
+
+        assert is_only_cell_edit(None)#no cells in edit mode
+    
+    elif(mode == 'edit'):
+        #Are the notebook and keyboard manager in edit mode?
+        assert mode == keyboard_mode #keyboard mode is correct
+
+        is_focused_on(index)#The specified cell is focused
```
takluyver: Should this be an `assert` statement?

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
...
+    cell_index = notebook.browser.execute_script(JS)
+    assert cell_index == [index] #only the index cell is selected
+
+    if(mode == 'command'):
+        #validate mode
+        assert mode == keyboard_mode #keyboard mode is correct
+
+        assert is_focused_on(None)#no focused cells
+
+        assert is_only_cell_edit(None)#no cells in edit mode
```
takluyver: Minor code style point: there should be one or two spaces between the code and the comment.

I also find comments are more readable with one space after the `#`, but that may be more of a personal preference.

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
...
+        if(index == None):
+            return focused_cells == 0
+
+        assert focused_cells == 1
+
+        JS = "return $('#notebook .CodeMirror-focused textarea')[0];"
+        focused_cell = notebook.browser.execute_script(JS)
+        JS = "return IPython.notebook.get_cell(%s).code_mirror.getInputField()"%index
+        cell = notebook.browser.execute_script(JS)
+        assert focused_cell == cell
```
takluyver: The naming here suggests that it's meant to return `True` or `False`, rather than assert internally.

But whichever way you do it, it should be consistent whether it returns or asserts.

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,77 @@ def trigger_keystrokes(browser, *keys):
...
+    elif(mode == 'edit'):
+        #Are the notebook and keyboard manager in edit mode?
+        assert mode == keyboard_mode #keyboard mode is correct
+
+        is_focused_on(index)#The specified cell is focused
+
+        assert is_only_cell_edit(index) #The specified cell is the only one in edit mode
+    
+    else:
+        assert False #An unknown mode is send
```
takluyver: If this occurs, `assert False` won't give a useful error message. It should raise an exception including the unexpected value.

LuisARodr:
Solved the comments, I'll be waiting if there are any more suggestions

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,80 @@ def trigger_keystrokes(browser, *keys):
...
+        cells_mode = notebook.browser.execute_script(JS)
+        #None of the cells are in edit mode
+        if index is None:
+            for mode in cells_mode:
+                if mode == 'edit':
+                    return False
+            return True
+        #Only the index cell is on edit mode
+        i = 0
+        for mode in cells_mode:
```
takluyver: This would be a good place to use the [enumerate()](https://docs.python.org/3/library/functions.html#enumerate) function, instead of managing i separately:

```python
for i, mode in enumerate(cells_mode):
```

notebook/tests/selenium/utils.py
```diff
@@ -382,3 +386,80 @@ def trigger_keystrokes(browser, *keys):
...
+        #validate mode
+        assert mode == keyboard_mode #keyboard mode is correct
+
+        assert is_focused_on(None) #no focused cells
+
+        assert is_only_cell_edit(None) #no cells in edit mode
+    
+    elif mode == 'edit':
+        #Are the notebook and keyboard manager in edit mode?
+        assert mode == keyboard_mode #keyboard mode is correct
```
takluyver: We do the same assertion in both branches, so we could move it outside the `if`.

LuisARodr:
Solved the suggestions again :+1:

takluyver:
Thanks!

PR #4618, opened 2019-05-09
fix default file name for make a copy

vibvibgyor:
If  a copy of notebook is created using _File>Make a Copy..._, the default filename is copied after partitioning the filename by dot (.) and appending _-Copy1_ (for the first copy) to the name of the file as shown in the picture below. In case, a filename itself contains dot (.), then it will partition from left instead of the right dot(.). I have fixed this issue by partitioning from the right side instead of the left which is more logical. I have attached the screenshots for the same.

Original file:
before

On making a copy of the notebook (before fix):
before_1

On making a copy of the notebook (after fix):
after

### DIFF
--- a/notebook/services/contents/manager.py
+++ b/notebook/services/contents/manager.py
@@ -331,7 +331,10 @@ class ContentsManager(LoggingConfigurable):
         """
         # Extract the full suffix from the filename (e.g. .tar.gz)
         path = path.strip('/')
-        basename, dot, ext = filename.partition('.')
+        basename, dot, ext = filename.rpartition('.')
+        if ext != 'ipynb':
+                basename, dot, ext = filename.partition('.')
+                
         suffix = dot + ext
 
         for i in itertools.count():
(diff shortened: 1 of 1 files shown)


vibvibgyor:
I beleive that it is more intuitive to have 'Copy' part at the last of the filename. If I have many copies of the notebook in the same folder, it is easier for me to find the right one to open if 'Copy' part is at the last of a filename.

vibvibgyor:
I have just seen that the test cases are failing on the problem where I made change. We would also need to rewrite the test cases according.

takluyver:
I think it's deliberately using the left-most dot to account for multi-part extensions like `.tar.gz`, hence why there are tests for it. The same code is used for all types of files, not just notebooks.

Unfortunately, I don't think there's any good general-way to know which part of a filename should be considered the extension.

vibvibgyor:
@takluyver Since, many of my colleagues and friends have felt similar problem, how about using separate code in case of .ipynb files only as we deal with mostly notebooks in Jupyter environment?

I can make changes, if you are interested.

takluyver:
I'd be OK with special casing .ipynb files if you want to do that. Be sure
to add a comment summarising how we got to that compromise. Thanks!

vibvibgyor:
Thanks @takluyver for your approval. I have made the suggested changes.

takluyver:
Thanks

PR #4622, opened 2019-05-10
Migrate execute code test to selenium

TeresaPartidaS:
This change migrates the test execute_code to use selenium as suggested in issue [#3335](https://github.com/jupyter/notebook/issues/3335)

Test Plan:
py.test -v notebook/tests/selenium/test_execute_code.py


notebook/tests/selenium/test_execute_code.py
```diff
@@ -0,0 +1,46 @@
...
+    clear_outputs()
+    notebook.browser.find_element_by_css_selector(
+        "button[data-jupyter-action='jupyter-notebook:run-cell-and-select-next']").click()
+    assert get_output_text() == '13', 'cell execute (cell execute (using "play" toolbar button))'
+
+    notebook.edit_cell(index=0, content='raise IOError')
+    notebook.edit_cell(index=1, content='a=14; print(a)')
+    clear_outputs()
+    notebook.execute_cell(1)
+    assert get_output_text(1) == '14', "cell execute, don't stop on error"
```
takluyver: This is failing at the moment. It's probably a race condition, because `notebook.execute_cell()` just sends Ctrl-Enter; it doesn't actually wait for execution to finish. We'll need to find some way to wait for the execution to finish to make the tests run reliably.

This fragment is also missing the point of the original test, which is to execute all the cells with the *stop on error* option set to false, to check that it continues after the error in the first cell to run the second cell.

TeresaPartidaS: Hi! I put some wait_for_selector for the race condition issue, I'll keep working on the test itself

notebook/tests/notebook/execute_code.js
```diff
@@ -1,115 +0,0 @@
...
-    });
-    
-    this.wait_for_output(0);
-
-    this.then(function () {
-        var result = this.get_output_cell(0);
-        this.test.assertEquals(result.text, '13\n', 'cell execute (using "play" toolbar button)')
-    });
-
-    // run code with skip_exception
```
TeresaPartidaS: @takluyver Hi, I was struggling to find a way to ignore the error cell, so I read this test again and notice that only the cell 1 was being executed, I replicated this with python and if you remove the `cell0.execute(false)` line, the test passes anyway
Do you know if there's any real way to ignore a cell? I couldn't find it in the UI neither

notebook/tests/notebook/execute_code.js
```diff
@@ -1,115 +0,0 @@
...
-        this.test.assertEquals(result.text, '14\n', "cell execute, don't stop on error");
-    });
-
-    this.thenEvaluate(function () {
-        var cell0 = IPython.notebook.get_cell(0);
-        cell0.set_text('raise IOError');
-        IPython.notebook.insert_cell_below('code',0);
-        var cell1 = IPython.notebook.get_cell(1);
-        cell1.set_text('a=14; print(a)');
-        cell0.execute();
```
TeresaPartidaS: @takluyver Hi, I was struggling to find a way to ignore the error cell, so I read this test again and notice that only the cell 1 was being executed, I replicated this with python and if you remove the `cell0.execute(false)` line, the test passes anyway
Do you know if there's any real way to ignore a cell? I couldn't find it in the UI neither

takluyver: I don't think this is exposed in the UI at all, but this is the idea of the test:

Normally, if you select two cells together and run them and the first one raises an exception, the execution request for the second one will be cancelled. That's what this block (from line 97) is testing.

You can override this behaviour in the Javascript API - if you call `cell.execute(false)` and it errors, later cells queued for execution will still be executed. The block above (lines 80 - 95) is testing this. So if you remove the `cell0.execute(false)` line,  it will definitely pass, but it's missing what it's designed to test.

I think the only way to test this from Python will be to execute some Javascript to get those cell objects and call their `.execute()` method.

takluyver:
I've rebased and updated this as part of a push to get the selenium migration unstuck.

PR #4623, opened 2019-05-11
Make tooltip buttons have consistent appearance

mtd91429:
Change class of `expandlink` from `ui-corner-all` to `ui-button` to be consistent with `morelink` and `closelink`

Current appearance of tooltip:

image

In the added red rectangle, the "+" sign is different in appearance compared to surrounding "^" and "x" signs.

Current html:

image

Changing the html within the browser inspector to the following:

image

Produces:

image

The changes within this commit are aimed to reproduce this appearance.


takluyver:
Thanks,  looks reasonable.

PR #4642, opened 2019-05-24
Use JavaScript to redirect users

betatim:
This is a followup to #4260.

This uses JS to do the redirection which works on Firefox (tested on FF 67.0b17 (64-bit)). The meta-refresh is still there so users without JS will be redirected as well.

### DIFF
--- a/notebook/templates/browser-open.html
+++ b/notebook/templates/browser-open.html
@@ -6,6 +6,11 @@
     <meta charset="UTF-8">
     <meta http-equiv="refresh" content="1;url={{ open_url }}" />
     <title>Opening Jupyter Notebook</title>
+    <script>
+      setTimeout(function() {
+        window.location.href = "{{ open_url }}";
+      }, 1000);
+    </script>
 </head>
 <body>
 


notebook/templates/browser-open.html
```diff
@@ -4,8 +4,13 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <meta http-equiv="refresh" content="1;url={{ open_url }}" />
+    <meta http-equiv="refresh" content="2;url={{ open_url }}" />
```
betatim: I increased this to two seconds to let the JS version do its job first. Not sure what is best out of: disable the meta-refresh (if the JS redirect will work), reduce the time of the JS based refresh to have it take precedence or increase this constant to two from 1.

betatim:
cc @minrk @takluyver

notebook/templates/browser-open.html
minrk: I don't quite understand the reason to delay this. If it works, is there a reason to wait for js? Shouldn't this be preferred to js unless it's being ignored for some reason?

betatim: For me/on Firefox the existing redirect mechanism doesn't work properly, but the JS based one does.

My thinking was that people who use the notebook have JS turned on so using JS should catch most people, then we try and get some more users by using the meta-refresh. I assumed that there was no reason to prefer meta-refresh over JS (maybe that isn't right) but the fact that meta-refresh doesn't work properly in FF is a reason to prefer JS over meta-refresh.

minrk: What does "not work properly" mean? Does it redirect wrong, or just do nothing? I was assuming that the failure mode was that nothing happened and the js redirect would take over. Does having this on a shorter time mean that the js redirect doesn't work?

betatim: You do get redirected but the page you end up with is:
Screen Shot 2019-05-27 at 15 30 54

The path shown in the location bar at the top is `file:///Users/timsusername/Library/Jupyter/runtime/nbserver-86595-open.html` instead of `http://localhost:8888`

I assumed there would be a race between the two refresh methods if set to the same timeout. However when actually trying it locally about ten times I always got the JS redirect. So I changed it back to a 1s delay for the meta-refres.

meeseeksmachine:

https://discourse.jupyter.org/t/firefox-vs-chrome/1199/2

minrk:
Great, thanks!

PR #4652, opened 2019-06-03
Update CONTRIBUTING.rst for windows users

abhinavsagar:


CONTRIBUTING.rst
```diff
@@ -39,7 +39,8 @@ If you use `Homebrew <https://brew.sh/>`_ on Mac OS X::
 Installation on Linux may vary, but be aware that the `nodejs` or `npm` packages
 included in the system package repository may be too old to work properly.
 
-You can also use the installer from the `Node.js website <https://nodejs.org>`_.
+If you are on Windows run the installer from the `Node.js website <https://nodejs.org>`_. Mac OS X and Linux users can also follow the
```
lresende: How about:
- You can also use the installer from the `Node.js website <https://nodejs.org>`_, which is the only option available if you are on Windows.

dhirschfeld: How about:

If you are on Windows and using `conda` you can install `nodejs` with:
```
conda install -c conda-forge nodejs
```
If not using `conda` then you can use the installer from the  `Node.js website <https://nodejs.org>`_

lresende: Going back and looking into the latest version of `Contributing.rst` I am inclined to leave it the way it is. Conda installation is already mentioned in the paragraphs above.... etc

abhinavsagar:
Agreed

PR #4654, opened 2019-06-04
.travis.yml: The 'sudo' tag is now deprecated in Travis CI

abhinavsagar:
`sudo: required` no longer is. "If you currently specify sudo: in your .travis.yml, we recommend removing that configuration".

[Travis are now recommending removing the sudo tag.](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,4 +101,3 @@ matrix:
 
 after_success:
     - codecov
-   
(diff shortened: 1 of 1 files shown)


takluyver:
Doesn't seem to break anything. Thanks!

PR #4658, opened 2019-06-05
altered tooltip and pager button

mtd91429:
- increased `smalltooltip` height to 85px to account for padding in `tooltiptext`
- reduced `ui-button` dimensions in tooltip and pager

### DIFF
--- a/notebook/static/notebook/less/pager.less
+++ b/notebook/static/notebook/less/pager.less
@@ -31,6 +31,13 @@ div#pager {
         right: 20px; 
     }
 
+    #pager-button-area .ui-button{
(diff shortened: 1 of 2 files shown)


takluyver:
Can I ask for some before/after screenshots?

mtd91429:
Here is a short before/after from the `master` branch (what is available on the latest release `5.7.8` is different, more detail can be found in #4659)

before:
before

after:
after

takluyver:
Thanks. That looks fine to me. I'll leave it a couple of days in case anyone else wants to have a look. If there's no other input, ping me in a couple of days and I'll merge it.

PR #4661, opened 2019-06-06
Add id for command palette button group

tmetzl:
- Added an id for the command palette button group since its the only one that doesn't have one


takluyver:
Thanks, this looks reasonable.

PR #4139, opened 2018-10-25
Fix default value method for `allow_remote_access` when ip is '*'

daf:
Fixes #3946.  When the config value for `ip` is set to `*`, meaning listen on all interfaces, it causes the default value method for the `allow_remote_access` config option to fail. This PR adds a short circuit for that default value method.

I didn't see an obvious place to put a new test, but on my local machine, this actually fixes a currently failing one.

PR #4664, opened 2019-06-07
Updated config to listen on all interfaces

matthew5025:
As per https://github.com/jupyter/notebook/pull/3714, the ip address for listening on all interfaces has changed from '*' to '0.0.0.0'


takluyver:
Thanks - I think `'*'` should work again since #4139. People are working on a new release.

Issue #4669, opened 2019-06-12
A single recursive link in a directory will make it unviewable


scanon:
If a directory contains a recursive symlink, then the file manager will return a 500 if you try to list it.  It should probably just skip the bad link.

Test
ln -s self self
and open in the file manager.

I'll put posting a PR for a fix.

kevin-bates:
Fixed by #4670

PR #4670, opened 2019-06-12
Fix for recursive symlink issue #4669

scanon:
This is a potential fix to issue #4669.  The fix simply catches the
recursive symlink error and moves on.  It is possible that the
try/except belongs in the utils but I wanted to limit the scope.

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -342,7 +342,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                             )
                 except OSError as e:
                     # Ignore recursive links and move on
-                    if e.errno==62 or e.errno==40:
+                    if e.errno==errno.ELOOP:
                         continue
                     else:
                         raise e
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -334,12 +334,18 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                     self.log.debug("%s not a regular file", os_path)
                     continue
 
-                if self.should_list(name):
-                    if self.allow_hidden or not is_file_hidden(os_path, stat_res=st):
-                        contents.append(
-                                self.get(path='%s/%s' % (path, name), content=False)
-                        )
-
+                try:
+                    if self.should_list(name):
+                        if self.allow_hidden or not is_file_hidden(os_path, stat_res=st):
+                            contents.append(
+                                    self.get(path='%s/%s' % (path, name), content=False)
+                            )
+                except OSError as e:
+                    # Ignore recursive links and move on
+                    if e.errno==62:
+                        continue
+                    else:
+                        raise e
             model['format'] = 'json'
 
         return model
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -342,7 +342,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                             )
                 except OSError as e:
                     # Ignore recursive links and move on
-                    if e.errno==62:
+                    if e.errno==62 or e.errno==40:
                         continue
                     else:
                         raise e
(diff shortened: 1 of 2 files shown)


notebook/services/contents/filemanager.py
```diff
@@ -334,12 +334,18 @@ def _dir_model(self, path, content=True):
...
-
+                try:
+                    if self.should_list(name):
+                        if self.allow_hidden or not is_file_hidden(os_path, stat_res=st):
+                            contents.append(
+                                    self.get(path='%s/%s' % (path, name), content=False)
+                            )
+                except OSError as e:
+                    # Ignore recursive links and move on
+                    if e.errno==62 or e.errno==40:
```
takluyver: Can we use the constants from the [errno module](https://docs.python.org/3/library/errno.html) rather than hardcoded numbers?

scanon: Good point.  Let me look.

notebook/services/contents/tests/test_manager.py
```diff
@@ -129,6 +129,26 @@ def test_bad_symlink(self):
...
+
+            model = cm.get(path)
+
+            contents = {
+                content['name']: content for content in model['content']
+            }
+            self.assertTrue('untitled.txt' in contents)
+            self.assertEqual(contents['untitled.txt'], file_model)
+            # recusrive symlinks should not be shown in the contents manager
+            self.assertFalse('recusrive' in contents)
```
takluyver: Spelling of 'recursive'.

Also, we should use the `assertIn` and `assertNotIn` methods rather than assertTrue and assertFalse. The more specific methods give more useful messages when things fail.

notebook/services/contents/tests/test_manager.py
```diff
@@ -129,6 +129,26 @@ def test_bad_symlink(self):
...
+
+            model = cm.get(path)
+
+            contents = {
+                content['name']: content for content in model['content']
+            }
+            self.assertIn('untitled.txt', contents)
+            self.assertEqual(contents['untitled.txt'], file_model)
+            # recursive symlinks should not be shown in the contents manager
+            self.assertNotIn('recursive', contents)
```
takluyver: I fixed the spelling, but this test is now failing on Windows:

```
======================================================================
FAIL: test_recursive_symlink (notebook.services.contents.tests.test_manager.TestFileContentsManager)
  File "C:\Miniconda36-x64\lib\site-packages\ipython_genutils\testing\decorators.py", line 186, in skipper_func
    return f(*args, **kwargs)
  File "C:\projects\notebook\notebook\services\contents\tests\test_manager.py", line 150, in test_recursive_symlink
    self.assertNotIn('recursive', contents)
AssertionError: 'recursive' unexpectedly found in {'recursive': {'name': 'recursive', 'path': 'test recursive symlink/recursive', 'last_modified': datetime.datetime(2019, 6, 25, 8, 35, 46, 553212, tzinfo=<notebook._tz.tzUTC object at 0x000000B0AB495D68>), 'created': datetime.datetime(2019, 6, 25, 8, 35, 46, 553212, tzinfo=<notebook._tz.tzUTC object at 0x000000B0AB495D68>), 'content': None, 'format': None, 'mimetype': None, 'size': 0, 'writable': True, 'type': 'file'}, 'untitled.txt': {'name': 'untitled.txt', 'path': 'test recursive symlink/untitled.txt', 'last_modified': datetime.datetime(2019, 6, 25, 8, 35, 46, 553212, tzinfo=<notebook._tz.tzUTC object at 0x000000B0AB495D68>), 'created': datetime.datetime(2019, 6, 25, 8, 35, 46, 553212, tzinfo=<notebook._tz.tzUTC object at 0x000000B0AB495D68>), 'content': None, 'format': None, 'mimetype': 'text/plain', 'size': 0, 'writable': True, 'type': 'file'}}
```

I guess Windows doesn't handle this in the same way Unix does. It's probably fine to just skip the test on Windows completely. I think it's unusual to use symlinks on Windows anyway.

takluyver:
I've got the tests passing.

I was informed of a related issue someone at EuXFEL encountered, where there was a permissions error stat-ing a symlink. So I've broadened this to catch all OSError, logging the exception details for anything besides the recursive symlink case.

Issue #4015, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_EditAttachments_ScreenReader:Narrator reads 'Close' Button as 'Times' Button.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106 
Operating System:Windows10

**Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    

2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library

3.Click on Run or right click and select Run to run the notebook.

4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Attachments. Click on Edit Attachments button appeared on page.
5.Start Narrator and verify how Narrator reads the 'Close' Button present on the PopUp.

**Actual**
Narrator reads 'Close' Button as 'Times' Button.
**NVDA reads close button as "Section".**

**Note:
Same issue observed on below pages:**
1)RenameConfirmationPopUp: Navigate to run button & hit enter --> select any file --> navigate to Rename button & hit enter.

**Expected**
Narrator should read 'Close' Button as 'Close' Button.

**User Impact**
https://microsoft.sharepoint.com/teams/msenable/mas/pages/mas-details.aspx?masid=1.3.1
**Attachment:**
[A11y_AzureNotebooks_Jupyter_EditMetada_ScreenReader21.pptx](https://github.com/jupyter/notebook/files/2401309/A11y_AzureNotebooks_Jupyter_EditMetada_ScreenReader21.pptx)

joshuazeltser:
We have solved this in #4677

takluyver:
Closed by #4677.

Issue #3951, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_EditTags_ScreenReader:Narrator is reading the "Times " for the Cross button.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Tags. Click on Edit Tags button appeared on page.
5.Use keyboard to navigate to page with the help og any of the keyboard keys.
6.Start the narrator.
**Actual:**
Narrator is reading the "Times " for the Cross button.
With NVDA it reads "Heading level 4" with Shift +TAB key
**Similar issue has been observed for the page**
1.DeleteConfirmationPopUp:
Navigate to run button & hit enter --> select any file --> navigate to delete button & hit enter.
2.EditAttachments:
Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Attachments. Click on Edit Attachments button appeared on page.
3. MoveConfirmationPopUp:
Navigate to run button & hit enter --> select any file --> navigate to delete button & hit enter

**Expected:**
Narrator should read the correct name for the control such as cancel or dismiss pop-up button. 
**UserImpact:**
User will not get wrong information regarding the cross button. 
**Attachment**

[A11y_AzureNotebooks_Jupyter_EditTags_ScreenReader.pptx](https://github.com/jupyter/notebook/files/2399704/A11y_AzureNotebooks_Jupyter_EditTags_ScreenReader.pptx)

takluyver:
Closed by #4677

PR #4677, opened 2019-06-18
Solved Issues #4015 and #3951

joshuazeltser:
Fix for issues #4015 & #3951. This issue involved the narrator saying "x button" instead of "close button". We fixed this by adding an attribute to the javascript which sets the aria-label to "Close".


takluyver:
Translation again (JS `i18n`)

takluyver:
Thanks, this looks good.

Travis tests have passed, but for some reason the status on Github hasn't updated.

Issue #4010, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_CreationAndDeployment_ColorContrastAnalyser:Ratio for the text"Azureml" is less than required(4.5:1)


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106 
Operating System:Windows10

**Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    

2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3Navigate to run button & hit enter --> Navigate to Creation and Deployment of an Azure ML Web Service & hit enter.
4.Navigate to "Azureml" text present 
![image](https://user-images.githubusercontent.com/39204909/45818317-f4d33e80-bcfe-11e8-8c94-a23aa246272d.png)

**Actual**
Ratio for the text"Azureml" is less than required(4.5:1)
**Expected**
Ratio for the text"Azureml" should be more than the ratio required(4.5:1)

User Impact
Low vision user will not be aabke to view the content .
**Attachment:**
![bugattachment5](https://user-images.githubusercontent.com/39204909/45818228-b8074780-bcfe-11e8-99d2-6bbf0e9ee418.png)

takluyver:
From a comment on #4678, this seems to have been improved somewhere along the way, and the contrast ratio between the black text and the light grey background is now ~18.

The contrast between the light grey background of inline code and the white of the larger background is much smaller, but I don't believe distinguishing the background is important for understanding, so we'll keep that as is.

PR #4678, opened 2019-06-18
Contrast of inline code snippets: issue #4010

joshuazeltser:
Fixed issue #4010. We changed line 61 #d4d4d4 to #757474 to meet the required ratio

### DIFF
--- a/notebook/static/notebook/less/renderedhtml.less
+++ b/notebook/static/notebook/less/renderedhtml.less
@@ -57,7 +57,7 @@
     }
 
     code {
-        background-color: #eff0f1;
+        background-color: #757474;
     }
 
     p code {


taohan16:
The Screenshots of changes for issue #4010
[Issue_4010_Screenshot.pptx](https://github.com/jupyter/notebook/files/3324912/Issue_4010_Screenshot.pptx)

joshuazeltser:
Before:
image

After:
image

takluyver:
This one looks wrong - making the background darker decreases the text contrast. I struggle to read the text in the 'after' screenshot with corrected-to-normal vision.

taohan16:
> This one looks wrong - making the background darker decreases the text contrast. I struggle to read the text in the 'after' screenshot with corrected-to-normal vision.

Hi, thanks for your comment. I agree with you that this does not look good, but it does reach the contrast ratio of 4.5:1 for both black and white. I think a better way to fix this is to change the color to white when using "``" in markdown. How do you feel about the below one?
image

takluyver:
My experience with situations that limit contrast (projectors, screens in bright sunlight) is that it's much easier to read dark text on a light background than the other way round.

Is there actually a problem with how it was before? I think only the contrast between the text and the background is important - the distinction between the light grey background of the inline code and the white of the page is only cosmetic, as far as I'm concerned. If an automated tool decided this was a problem, I think we can override it with a human judgment that it's not.

taohan16:
> My experience with situations that limit contrast (projectors, screens in bright sunlight) is that it's much easier to read dark text on a light background than the other way round.
> 
> Is there actually a problem with how it was before? I think only the contrast between the text and the background is important - the distinction between the light grey background of the inline code and the white of the page is only cosmetic, as far as I'm concerned. If an automated tool decided this was a problem, I think we can override it with a human judgment that it's not.

I agree with you. The contrast between the background and the text is more important to ensure the contents are readable. So maybe we can close this PR.

takluyver:
OK, thanks.

takluyver:
Actually, I think issue #4010 is talking about the ratio between the text and the background, looking at the  screenshot it shows. But the hex codes it shows are not the same as those involved here. If you've got more time to work on this, could you check the text/background contrast again, and record on #4010 whether it's still a problem or not.

taohan16:
For the inline code snippets in current Jupyter, the HEX code is #EFF0F1.
The contrast ratio between it and the white background is 1.1:1.
The contrast ratio between it and the black text is 18.4:1.

The background color for the inline code snippets after changing is #757474.
The contrast ratio between it and the white background is 4.7:1.
The contrast ratio between it and the black text is 4.5:1.

takluyver:
Thanks, I've closed #4010 on the grounds of that.

Issue #4680, opened 2019-06-18
Latest preact-compat release breaks notebook.


ssanderson:
Three of our javascript dependencies, `preact`, `preact-compat`, and `proptypes`, are currently managed via [`unpkg` links](https://github.com/jupyter/notebook/blob/master/bower.json#L19), and pinned with `^` rather than `~`. `^` appears to have the same semantics as in npm, in which it means "allow non-major version upgrades".

The most recent release of `preact-compat` appears to have broken the notebook. If I clear my `~/.cache/bower` and then run `python setup.py css js` in an editably-installed `notebook` repo, when I run `jupyter notebook` I get a blank page on the notebook page, with console errors about missing `createContext` from `preact-compat`:

![image](https://user-images.githubusercontent.com/1588357/59719365-32132100-91ea-11e9-8104-7241f5d7dd8f.png)

I think the easiest fix here is to pin via `~`, which seems to be the convention followed for most other js dependencies (MathJax seems to be a notable exception to this rule). We may also want to report this upstream as a bug to `preact-compat`? I don't really have a good sense of whether that's desirable.

ssanderson:
Switching to `~` fixes the issue for me locally: https://github.com/jupyter/notebook/pull/4681

ssanderson:
NOTE: Clearing your bower cache is required to reproduce this issue. If you've previously build the js dependencies successfully, bower appears to continue to re-use the old deps even though it resolves different versions with an empty cache. :(

leestott:
Just as update this fixes build issues in Build Pipelines in Azure DevOps Pipelines we had  the following issues and Errors 

![image](https://user-images.githubusercontent.com/2511341/59762792-ed41c580-928f-11e9-987c-37eeb3c33282.png)

Error 
 ##[section]Starting: Run base Tests
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.151.1
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
========================== Starting Command Output ===========================
[command]/bin/bash --noprofile --norc /home/vsts/work/_temp/0ec4b1f4-49ab-41ab-9b5f-a8215f60bf3e.sh
Available kernels:
  python3    /opt/hostedtoolcache/Python/3.6.8/x64/share/jupyter/kernels/python3

Test group: base
[...]

PR #4681, opened 2019-06-18
BUG: Pin preact and friends using ~ rather than ^.

ssanderson:
The most recent release of preact-compat appears to have broken notebook.

Fixes https://github.com/jupyter/notebook/issues/4680.


lresende:
Do we know what in the new release of preact is causing the issues in Notebook? Should we actually look into making it compatible for the 6.0 release or folks think that the pin might be a good option for now? @minrk @blink1073

blink1073:
I'd say pin for now, I'm not as familiar with the frontend parts of this repo.

ssanderson:
I'm not super familiar with javascript packaging generally, but it looks to me like what happened is that [the latest release of preact-compat](https://github.com/developit/preact-compat/releases/tag/3.19.0) added a new `createContext` method, which added a new dependency on [`preact-context`](https://github.com/valotas/preact-context), which we're not picking up properly. 

I'm 100% spitballing here, but my *guess* is that the failure to pick up `preact-context` is related to the fact that we're using `bower` for package management, which doesn't know about the transitive dependencies of `preact-compat` (presumably b/c they don't have a bower.json?). 

Assuming the above is correct, what this probably means is that any change that adds a new dependency to any of our bower-unaware dependencies (my guess is that's increasingly the majority of modern js projects given that bower seems to be) will end up being a breaking change for us for as long as we're using bower to manage dependencies.

Issue #3936, opened 2018-09-19
A11y_AzureNotebooks_Jupyter_AzureNotebooksWelcome_ScreenReader: In the dialog which get displayed after selecting "heading" from the dialog, Used heading markup for the text "Use markdown headings" is inaccurate.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106 
Operating System:Windows10

**Repro Steps:**
1.Login with the appropriate credentials to azure notebook account.    
2. Navigate to run button & hit enter --> Navigate to Azure Notebooks welcome & hit enter.
3. Navigate to Combobox present after the "Restart Kernal" button.
4. Select Heading list item by using arrow key.
5.Verify heading markup using screen reader.

**Actual:**
In the dialog which get displayed after selecting "heading" from the dialog, the text "Use markdown headings" is markup using H4 which is inaccurate.

**Expected:**
Heading markup should be accurate for the content on page.
It should be markup using H1.

**User Impact:**
Screen reader user will not able to understand the content structure of popup.

[A11y_AzureNotebooks_Jupyter_AzureNotebooksWelcome_ScreenReader.pptx](https://github.com/jupyter/notebook/files/2396687/A11y_AzureNotebooks_Jupyter_AzureNotebooksWelcome_ScreenReader.pptx)

takluyver:
Does this mean all dialogs should have an `<h1>` heading? Or is it something specific about the 'Use markdown headings' dialog?

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4684, opened 2019-06-19
h1 heading for dialog: issue #3936

joshuazeltser:
Fixed issue #3936. We have made the markdown popup have its title as h1 like the markdown text itself.


takluyver:
I'm having a bit of trouble understanding the issue, but should it be h1 in *all* dialogs, rather than just this one?

Also, have you checked that increasing the heading level doesn't mess up the layout visually?

joshuazeltser:
So I have checked the increased heading level on several browsers on windows and mac and the layout is fine visually. My changes are based on my understanding of the issue, which seems to say that as the markdown heading itself will be h1, the heading should be too.

takluyver:
I don't think this is anything to do with the fact that the dialog is talking about markdown headings - I think that's just a confusing coincidence.

joshuazeltser:
In that case should all dialogs have an h1 header?

takluyver:
That's what I'm wondering. I've asked for clarification on #3936.

jtpio:
Closing since there has not been any activity for a while, and the classic notebook UI has now moved to https://github.com/jupyter/nbclassic.

Please feel free to open the PR there if it is still relevant, thanks!

Issue #3941, opened 2018-09-19
A11y_AzureNotebooks_Jupyter_RenameConfirmationPopUp_ScreenReader: Narrator is not reading label relationship of text field present on Rename Notebook popup.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106
Operating System:Windows10

**Repro Steps:**
1.Enable Narrator screen reader.
2.Login with the appropriate credentials to azure notebook account.
3.Navigate to Libraries->Create clone of anyone library->Click on libraries->Click on Sample created library.
4.Navigate to run button & hit enter -->Select any one non running library->navigate to Rename option & hit enter.
5. Verify label relationship on popup for text area control. 

**Actual:**
Narrator is not reading label relationship of text field present on Rename Notebook popup.

**Expected:**
Form element (Textarea) should have meaningful deacription read by narrator.

User Impact:
Screen reader user would not get information about what type of textfield present on popup.

Attachment:
![a11y_azurenotebooks_jupyter_renameconfirmationpopup](https://user-images.githubusercontent.com/39204909/45754457-5fba4200-bc39-11e8-84d3-0c5151f390c7.jpg)

takluyver:
Closed by #4686.

PR #4686, opened 2019-06-19
Solved Issue #3941

joshuazeltser:
Fixed issue #3941. Used aria-labelledby to read the label of the textfield in popups.


takluyver:
Thanks!

Issue #3961, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader:Narrator is not reading the checked/Unchecked state of "Select All/None" CheckBox.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Select All/None" dropdown.
6.Start the Narrator.
**Actual:
Issue1:**
Narrator is not reading the checked/Unchecked state of "Select All/None" CheckBox.
**Issue2:**
Narrator is not information regarding the number of checkboxes selected present inside the box.
**Expected:
Issue1:**
Narrator should read the checked/Unchecked state of "Select All/None" CheckBox.
**Issue2:**
Narrator should give  information the regarding  number of checkboxes selected present inside the CheckBox.
**UserImpact:**
User will not get wrong information regarding the state of the control..  
**Attachment:**

[A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader5.pptx](https://github.com/jupyter/notebook/files/2399980/A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader5.pptx)

PR #4688, opened 2019-06-19
Solved Issue #3961

joshuazeltser:
Fixed Issue #3961. We used jquery to add an aria-label dynamically to ensure screen reader reads the checked/unchecked state and the number of selected items for the checkbox on the tree page.

### DIFF
--- a/notebook/static/tree/js/notebooklist.js
+++ b/notebook/static/tree/js/notebooklist.js
@@ -234,6 +234,8 @@ define([
                     that.select('select-all');
                 }
             });
+
+
             $('#button-select-all').click(function (e) {
                 // toggle checkbox if the click doesn't come from the checkbox already
                 if (!$(e.target).is('input[type=checkbox]')) {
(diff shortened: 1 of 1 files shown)


takluyver:
Search for `i18n` in this file to see how strings in JS can be translated.

joshuazeltser:
Added the translation function calls as requested

notebook/static/tree/js/notebooklist.js
```diff
@@ -815,14 +814,14 @@ define([
         //#issue 3961, update the checkbox aria-label when it changed
         if(selected.length>=1){
           if($('#select-all').prop("checked")){
-            $('#button-select-all').attr("aria-label","Selected All "+ selected.length+" items");
+            $('#button-select-all').attr("aria-label", i18n.msg._("Selected All "+ selected.length+" items"));
           }
           else{
-            $('#button-select-all').attr("aria-label","Selected, "+ selected.length+" items");
+            $('#button-select-all').attr("aria-label", i18n.msg._("Selected, "+ selected.length+" items"));
```
takluyver: I think it needs to have a string literal inside the `_()` so that the translation tools can pick it up as something that needs to be translated (they do some kind of simple static analysis). I'm not 100% sure how that works with numbers - can you hunt around the codebase to see if we do that anywhere else? Or look for info online.

joshuazeltser: Based on research through the project/online I think this I have added a way that takes into account non string literals such as numbers

takluyver:
Thanks

Issue #3998, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_MenuBar_Inspect: Role for Some of the Controls(New Notebook,Revert to Checkpoint,Download as) is incorrect.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Operating System:Windows10

**ReproSteps:**
1. Login with the appropriate credentials to azure notebook account.    
2. Navigate to any of the file names present in the samples-->>clickin the file name-->Welcome page opens with menubar.
3. Start Navigating through Menubar.
4. Navigate to "File" menu and hit spacebar/enter to expand. 
5. Open Inspect Tool.
6. Verify Role for  Controls(New Notebook,Revert to Checkpoint,Download as).

**Actual:**
Role for Controls(New Notebook,Revert to Checkpoint,Download as) is given as Hyperlink/Link which is incorrect.
 * Narrator is also reading as "New Notebook"Link which is incorrect.
 * NVDA is also reading as "New Notebook"Link which is incorrect.

**Note:**
Same issue is observed for all Dropdowns present inside all Menus(File,Edit,View,Insert,Cell...etc).

**Expected:**
Role for Controls(New Notebook,Revert to Checkpoint,Download as) Should be Dropdown as they are Expanding.
 * Narrator should read as "New Notebook"Dropdown.
 * NVDA should read as "New Notebook"Dropdown.

**UserImpact:**
The user will not be able to know the significance of the control.

**Attachment:**
![a11y_azurenotebooks_jupyter_menubar_inspect](https://user-images.githubusercontent.com/39204909/45813822-4674cc80-bcf1-11e8-8d4f-5aeb3ae5ba61.png)

takluyver:
Closed by #4690

Issue #3971, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader:Narrator is not reading descriptive tooltip for "Select......" dropdown.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Files" tab.
6.Start the Narrator.
**Actual:**
Narrator is not reading descriptive tooltip for "Select......" dropdown.
**Expected:**
Narrator should read descriptive tooltip for the toggle dropdown which gives some onformation regading the control.
**UserImpact:**
User will get confused  about the purpose of the control as it is not clear from the tooltip. 
**Attachmment**
[A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader16.pptx](https://github.com/jupyter/notebook/files/2400110/A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader16.pptx)

takluyver:
Closed by #4690

Issue #3968, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader:Narrator is only reading the tooltip for the "Duplicate","Edit","View","Refresh" buttons .


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails: 
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Files" tab.
6.Start the Narrator.
**Actual:**
Narrator is only reading the tooltip for the "Duplicate","Edit","View","Refresh" buttons in .It is not reading the control name.
**Expected:**
Narrator should read  the control name and then the tooltip for  "Duplicate","Edit","View" ,"Refresh"buttons. 
**UserImpact:**
User will not get confused between the control name and the tooltip. 
**Attachment:**
[A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader14.pptx](https://github.com/jupyter/notebook/files/2400088/A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader14.pptx)

takluyver:
Closed by #4690

PR #4690, opened 2019-06-19
Solved issues #3998, #3971 and  #3968

joshuazeltser:
Fixed issues three issues as follows:
- #3998 - added correct role for controls: "New Notebook", "Revert to Checkpoint", "Download as"
- #3971 - added more descriptive tooltip for the "Select" dropdown
- #3968 - added aria-labels for the control names of the following buttons: "Duplicate", "Edit", "View", "Refresh"


notebook/templates/notebook.html
```diff
@@ -98,7 +98,7 @@
                         <li id="save_checkpoint"><a href="#">{% trans %}Save and Checkpoint{% endtrans %}</a></li>
                         <!-- <hr/> -->
                         <li class="divider"></li>
-                        <li id="restore_checkpoint" class="dropdown-submenu"><a href="#">{% trans %}Revert to Checkpoint{% endtrans %}</a>
+                        <li id="restore_checkpoint" class="dropdown-submenu"><a href="#">{% trans %}Revert to Checkpoint{% endtrans %}<span class="sr-only">Toggle Dropdown</span></a>
```
takluyver: Should these have `{% trans %}` markers to allow them to be translated as well?

notebook/templates/tree.html
```diff
@@ -41,13 +41,13 @@
...
-                  <button title="{% trans %}Download selected{% endtrans %}" aria-label="{% trans %}Download selected{% endtrans %}" class="download-button btn btn-default btn-xs">{% trans %}Download{% endtrans %}</button>
-                  <button title="{% trans %}Shutdown selected notebook(s){% endtrans %}" aria-label="{% trans %}Shutdown selected notebook(s){% endtrans %}" class="shutdown-button btn btn-default btn-xs btn-warning">{% trans %}Shutdown{% endtrans %}</button>
-                  <button title="{% trans %}View selected{% endtrans %}" aria-label="{% trans %}View selected{% endtrans %}" class="view-button btn btn-default btn-xs">{% trans %}View{% endtrans %}</button>
-                  <button title="{% trans %}Edit selected{% endtrans %}" aria-label="{% trans %}Edit selected{% endtrans %}" class="edit-button btn btn-default btn-xs">{% trans %}Edit{% endtrans %}</button>
+                  <button title="{% trans %}Duplicate selected{% endtrans %}" aria-label="{% trans %}Duplicate{% endtrans %}" class="duplicate-button btn btn-default btn-xs">{% trans %}Duplicate{% endtrans %}</button>
```
takluyver: Is there any need for an `aria-label` if it's exactly the same as the text on the button? Presumably the screenreader can use the 'visible' text by default?

(Forgive me if I'm asking stupid questions - I know very little about web accessibility)

notebook/templates/notebook.html
joshuazeltser: Added trans tags

notebook/templates/tree.html
takluyver: (Ping)

joshuazeltser: So the button was originally just reading the tooltip (as defined by the title attribute) and not the   control name of the button as well. The issue says both need to be read and this is why we added the aria-label (so that the control name is read)

notebook/templates/tree.html
```diff
@@ -100,7 +100,8 @@
                 <button title="{% trans %}Select All / None{% endtrans %}" aria-label="{% trans %}Select All / None{% endtrans %}" type="button" class="btn btn-default btn-xs" id="button-select-all" role="checkbox">
                   <input type="checkbox" class="pull-left tree-selector" id="select-all" tabindex="-1"><span id="counter-select-all">&nbsp;</span></input>
                 </button>
-                <button title="{% trans %}Select...{% endtrans %}" class="btn btn-default btn-xs dropdown-toggle" type="button" id="tree-selector-btn" data-toggle="dropdown" aria-expanded="true">
+                <button title="{% trans %}Select Folders/All Notebooks/Running/Files {% endtrans %}" class="btn btn-default btn-xs dropdown-toggle" type="button" id="tree-selector-btn" data-toggle="dropdown" aria-expanded="true">
```
takluyver: Drop the trailing space?

notebook/templates/tree.html
takluyver: Thanks. LGTM once the tests pass, then.

notebook/templates/tree.html
joshuazeltser: removed

Issue #3973, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_Inspect:Role is incorrect for the "New" dropdown..


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "New" control.
6.Start the Narrator ot inspect.
**Actual:**
Role is incorrect for the "New" contol
**Expected:**
Correct role for the "New" should be dropdown, as it has list items and sub menu list-items associated to it.
 **UserImpact:**
User will not get correct information regarding the control type of the Drop-down
**Attachment:**
![attachment1](https://user-images.githubusercontent.com/39204909/45803385-1f110600-bcd7-11e8-9d98-9a02ef2e8f8f.jpg)

joshuazeltser:
I have been attempting to solve this problem and have been able to get the narrator to now say "New Toggle Dropdown, expanded, button". This makes what this button is a bit clearer. Does this seem clear enough?

takluyver:
Closed by #4691.

PR #4691, opened 2019-06-19
Solved Issue #3973

joshuazeltser:
Fixed issue #3973. Added span element causing the narrator to correctly state the role of the "New" button


notebook/templates/tree.html
```diff
@@ -65,6 +65,7 @@
                   <button class="dropdown-toggle btn btn-default btn-xs" id="new-dropdown-button" data-toggle="dropdown">
                   <span>{% trans %}New{% endtrans %}</span>
                   <span class="caret"></span>
+                  <span class="sr-only">Toggle Dropdown</span>
```
takluyver: Should this have a `{% trans %}` marker for translation?

joshuazeltser: Yes it should and they have been added

takluyver:
Thanks

Issue #3992, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_Inspect:Name is missing for the "Folder" Icon.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106
Operating System:Windows10
**Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Folder" Icon. 
**Actual:**
Name is missing for the "Folder" Icon.
**Expected:**
An appropriate name should be present for the "Folder" Icon.
**UserImpact:**
ScreenReader user will get no information regarding the folder present on the page.
**Attachment:**
![bugattachment4](https://user-images.githubusercontent.com/39204909/45810787-1cb7a780-bce9-11e8-9ffa-c97e6548cf56.png)

PR #4692, opened 2019-06-19
Solved Issue #3992

joshuazeltser:
Fixed issue #3992. Added appropirate title to the icon and other aria tags added to make dropdowns more understandable when read by a narrator


notebook/static/base/js/namespace.js
```diff
@@ -1,83 +1,83 @@
-// Copyright (c) Jupyter Development Team.
-// Distributed under the terms of the Modified BSD License.
-
-
```
takluyver: Why is this entire file showing up as changed? Can you do it without this?

takluyver:
Comments from #4695 also apply here.

notebook/static/base/js/namespace.js
joshuazeltser: As far as I can see there are no changes. I have no idea why this has happened

notebook/static/tree/js/notebooklist.js
```diff
@@ -386,6 +386,7 @@ define([
         var root = $('<li/>').append(
             $("<a/>")
             .attr('href', root_url)
+            .attr('title','Link to root folder'+root_url)
```
takluyver: This presumably needs integrating with the translation machinery.

Also, it doesn't seem right to put the URL in the title. Again, I'm quite ignorant of accessibility concerns, but presumably screenreader tools have a way to inspect the URL of a link if the user wants that.

joshuazeltser: This has been updated

takluyver:
I've cherry-picked the intended changes from here as PR #4719, preserving commit authorship, so I'll close this one.

Issue #3996, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_MenuBar_Inspect: Role is incorrect for "Not Trusted" Control which is present at right of the menubar.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Operating System:Windows10

**ReproSteps:**
1. Login with the appropriate credentials to azure notebook account.    
2. Navigate to any of the file names present in the samples-->>clickin the file name-->Welcome page opens with menubar.
3. Start Navigating through Menubar. 
4. Open Inspect Tool.
5. Verify "Role" or "Control Type" for "Not Trusted" control.

**Actual:**
Control Type for the "Not Trusted" button is given as Text which is incorrect.

**Expected:**
Control Type for the "Not Trusted" control should be button.

**UserImpact:**
The user will not be able to know the significance of the control.

**Attachment:**
![a11y_azurenotebooks_jupyter_menubar_inspect](https://user-images.githubusercontent.com/39204909/45813365-09f4a100-bcf0-11e8-8488-c3999e8effa8.png)

takluyver:
Closed by #4693

PR #4693, opened 2019-06-19
Solved Issue #3996 

joshuazeltser:
Fixed issue #3996. Added role attribute with value button to the trusted/not trusted control

### DIFF
--- a/notebook/static/notebook/js/notificationarea.js
+++ b/notebook/static/notebook/js/notificationarea.js
@@ -398,15 +398,14 @@ define([
                   return false;
                 }, {'title':'Javascript enabled for notebook display'});
                 // don't allow 'Trusted' button to be clicked
-                $(tnw.selector).attr('disabled', true)
-                $(tnw.selector).attr('role', 'button')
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/notificationarea.js
```diff
@@ -399,12 +399,14 @@ define([
                 }, {'title':'Javascript enabled for notebook display'});
                 // don't allow 'Trusted' button to be clicked
                 $(tnw.selector).attr('disabled', true)
+                $(tnw.selector).attr('role', 'button')
```
takluyver: Is it still appropriate to use the 'button' role when it's the 'trusted' notification which can't be clicked?

ednut15: Initial thoughts were that even if the element was not clickable (disabled), it is still a button. If the notification was created as a button, it's role would not change when it is disabled. The role attribute has been removed from the Trusted notification. Since the element is a span and is disabled there is no difference based on the presence of the attribute.

Issue #3963, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader:Narrator is not reading the Order of list items present in the "Select......" Dropdown.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Select All/None" dropdown.
6.Start the Narrator.
**Actual:**
Narrator is not reading the Order of list items present in the "Select......" Dropdown.

**Same issue is observed for NVDA also.
Similar Issue has been observed in:**
"New" dropdown.

**Note:
Same issue is observed in below pages:**
1. Menu Bar: Navigate to any of the file names present in the samples-->>clickin the file name.

**Expected:**
Narrator should read the Order of list items present in the "Select......" Dropdown.
**UserImpact:**
User will not get wrong information regarding the state of the control.. 
**Attachment:**
[A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader7.pptx](https://github.com/jupyter/notebook/files/2400004/A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader7.pptx)

takluyver:
Closed by #4695.

Issue #3960, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader:"Select All/None" Form elements should have an appropriate label associated to it.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Folder" icon.
6.Start the Narrator.
**Actual:**
"Select All/None" should have an appropriate label associated to it.
**Expected:**
"Select All/None" should have an appropriate label associated to it.
Element Path
#select-all

 
Snippet
<input type="checkbox" class="pull-left tree-selector" id="select-all">

 
**How to fix
Fix any of the following:**
- aria-label attribute does not exist or is empty
- aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty or not visible
- Form element does not have an implicit (wrapped) <label>
- Form element does not have an explicit <label>
- Element has no title attribute or the title attribute is empty
**UserImpact:**
User will not get wrong information regarding the state of the link. 
**Attachment:**
[A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader123.pptx](https://github.com/jupyter/notebook/files/2399966/A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader123.pptx)
![png_attachment](https://user-images.githubusercontent.com/39204909/45800229-d739b100-bccd-11e8-8bce-e14716baaabd.png)

takluyver:
Closed by #4695.

PR #4695, opened 2019-06-19
Solved Issues #3963 and #3960

joshuazeltser:
Fixed issues #3963 and #3960. For static html add aria-label with ordering and for dynamic elements used javascript to add aria-labels dynamically.

### DIFF
--- a/notebook/static/tree/js/newnotebook.js
+++ b/notebook/static/tree/js/newnotebook.js
@@ -63,6 +63,7 @@ define([
                 .attr("id", "kernel-" +ks.name)
                 .data('kernelspec', ks).append(
                     $('<a>')
+                        .attr("aria-label","list item 1,"+ks.name)
                         .attr('href', '#')
                         .click($.proxy(this.new_notebook, this, ks.name))
                         .text(ks.spec.display_name)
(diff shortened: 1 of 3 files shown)


takluyver:
Is it really necessary to explicitly number things which already have an order in the DOM? That seems like it's bound to get broken as people modify the menus.

notebook/static/tree/js/newnotebook.js
```diff
@@ -63,6 +63,7 @@ define([
                 .attr("id", "kernel-" +ks.name)
                 .data('kernelspec', ks).append(
                     $('<a>')
+                        .attr("aria-label","list item 1,"+ks.name)
```
takluyver: Will this not result in multiple `list item 1`?

joshuazeltser: This has been made dynamic

joshuazeltser:
The specified menus should all now be numbered dynamically by the screenreader

takluyver:
Thanks!

takluyver:
BTW, if you write your commit messages slightly differently, like "Fixes #3963, fixes #3960", the issues get automatically closed when we merge the PR. There are a few synonyms it recognises, but "Solved issues" is not among them:

https://github.blog/2013-01-22-closing-issues-via-commit-messages/

Issue #4009, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_Keyboard: User have to 'TAB' twice to move over the controls such as 'Upload' button and 'Select All/None' checkbox.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
4. Navigate to 'Upload' Button and 'Select All/None' Checkbox using TAB.
5. Verify how Focus moves using TAB when moving over these controls.

**Actual:**
User have to 'TAB' twice to move over these controls. i.e. focus remains same after first TAB.

Same issue observed throughout the application on two "header" controls.

**Expected:**
Single 'TAB' should be enough to move to the next controls present in the sequence.

**User Impact:**
Keyboard user will get confuse while moving over these controls as he would have to 'TAB' twice to move over these controls.

**Attachment:**
(https://github.com/jupyter/notebook/files/2401236/A11y_AzureNotebooks_Jupyter_FilesTab_Keyboard.pptx)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4696, opened 2019-06-19
Solved issue #4009

joshuazeltser:
Fixed issue #4009. To avoid having to tab twice over controls such as 'Upload' button, we added a tabindex=-1 for the inner input


bower.json
```diff
@@ -16,9 +16,9 @@
     "marked": "~0.5",
     "MathJax": "^2.7.4",
     "moment": "~2.19.3",
-    "preact": "https://unpkg.com/preact@^7.2.0/dist/preact.min.js",
-    "preact-compat": "https://unpkg.com/preact-compat@^3.14.3/dist/preact-compat.min.js",
-    "proptypes": "https://unpkg.com/proptypes@^0.14.4/index.js",
+    "preact": "https://unpkg.com/preact@~7.2.0/dist/preact.min.js",
+    "preact-compat": "https://unpkg.com/preact-compat@~3.14.3/dist/preact-compat.min.js",
+    "proptypes": "https://unpkg.com/proptypes@~0.14.4/index.js",
```
takluyver: I'm hazy on this, what's the difference between `^` and `~`?

takluyver:
This makes sense, but #4009 mentions two controls - this one and the upload button. Does the same change work on the upload button as well?

bower.json
takluyver: I see this change was already merged in #4681, so never mind.

jasongrout: Essentially, `~` is more restrictive. In this situation, `^` means include the latest minor release (7.x, 3.x, 0.14.x, respectively), while `~` means the latest patch release (7.2.x, 3.14.x, and 0.14.4, respectively). See examples at https://semver.npmjs.com/ (enter the package name and semver range, and it will show you matching versions)

takluyver: Thanks Jason!

joshuazeltser:
We added a fix for the "Upload" button now too

takluyver:
Thanks

Issue #4006, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_EditAttachments_Keyboard: Keyboard focus is lost when pressing 'ESC' key on the active Pop-up.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106 
Operating System:Windows10

**Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    

2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library

3.Click on Run or right click and select Run to run the notebook.

4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Attachments. Click on Edit Attachments button appeared on page.
5. Press 'ESC' key and verify the keyboard focus is lost or not.

**Actual**
Keyboard focus is lost when user presses 'ESC' key on the active Pop-up.

**Expected**
Keyboard focus should stay on the 'Edit attachment' Button when the Popup closes while pressing 'ESC' key.

**User Impact**
Keyboard user would get confuse and would have to navigate again when the keyboard focus is lost on the active pop-up.
**Attachments:**

[A11y_AzureNotebooks_Jupyter_EditAttachments_Keyboardnew.pptx](https://github.com/jupyter/notebook/files/2401180/A11y_AzureNotebooks_Jupyter_EditAttachments_Keyboardnew.pptx)

takluyver:
Closed by #4711.

PR #4711, opened 2019-06-24
Solved Issue #4006

joshuazeltser:
Fixed issue #4006. To ensure focus returns to the "Edit Metadata" button after closing the popup, we stored the button once created (in JS) and then set it to focus on the button once the popup is hidden.

### DIFF
--- a/notebook/static/base/js/dialog.js
+++ b/notebook/static/base/js/dialog.js
@@ -234,6 +234,7 @@ define(['jquery',
         });
 
         modal_obj.on('shown.bs.modal', function(){ editor.refresh(); });
+        modal_obj.on('hide.bs.modal', function(){options.edit_metadata_button.focus(); });
     };
 
     var edit_attachments = function (options) {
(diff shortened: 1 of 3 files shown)


takluyver:
NB the `edit_metadata` dialog function is used for both cell metadata and notebook metadata. Does this do anything strange when it's triggered by notebook metadata?

Also, are there any other ways that editing cell metadata can be triggered? E.g. can it be bound to a keyboard shortcut?

joshuazeltser:
I have added a check to see whether the edit metadata is present before trying to focus on it. This should completely now solve the problem that this issue is addressing

notebook/static/base/js/dialog.js
```diff
@@ -234,7 +234,8 @@ define(['jquery',
         });
 
         modal_obj.on('shown.bs.modal', function(){ editor.refresh(); });
-        modal_obj.on('hide.bs.modal', function(){options.edit_metadata_button.focus(); });
+        modal_obj.on('hide.bs.modal', function(){
+            options.edit_metadata_button ? options.edit_metadata_button.focus() : ""});
```
takluyver: Should there be a `;` inside the anonymous function?

(Nitpicking,  I know)

joshuazeltser: Done

Issue #3962, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader:Narrator is not reading any alert message regarding the  action performed on the refresh button.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
5.Navigate to the "Select All/None" dropdown.
6.Start the Narrator.
**Actual:**
Narrator is not reading any information about the  action performed on the refresh button.

**Note:
Same issue is observed in below pages:**
1. RunningTab: Navigate to run button & hit enter --> Navigate to running tab & hit enter.

**Expected:**
Narrator should read any information about the  action performed on the refresh button.
**UserImpact:**
User will not get any information regarding the refreshing of theb page. 
**Attachment:**
[A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader6.pptx](https://github.com/jupyter/notebook/files/2399996/A11y_AzureNotebooks_Jupyter_FilesTab_ScreenReader6.pptx)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4716, opened 2019-06-25
Solved Issue #3962: aria-live attribute for dynamic refresh button

taohan16:
Solved Issue #3962
Add "aria-live" attribute to allow the NVDA to read "pressing refresh button" while pressing the button.

### DIFF
--- a/notebook/static/tree/js/main.js
+++ b/notebook/static/tree/js/main.js
@@ -116,8 +116,16 @@ requirejs([
     );
 
     $("#refresh_notebook_list").click(function(){
-        $(this).attr("aria-pressed","true");
-    })
+        if($(this).attr("aria-label")=="pressing refresh button"){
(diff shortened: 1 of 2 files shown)


notebook/static/tree/js/main.js
```diff
@@ -115,6 +115,18 @@ requirejs([
         )
     );
 
+    $("#refresh_notebook_list").click(function(){
+        if($(this).attr("aria-label")=="pressing refresh button"){
+            $(this).attr("aria-label","pressing refresh button.");
+        }
+        else if($(this).attr("aria-label")=="pressing refresh button."){
+            $(this).attr("aria-label","pressing refresh button");
```
takluyver: What's this for? It seems to toggle the "." at the end, but I don't understand the aim of that.

It should be integrated with the translation, and I don't like the check against a human-readable string.

Issue #4005, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_EditMetada_Keyboard: Keyboard Trap occurs inside the text area of the Popup.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106 
Operating System:Windows10

**Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    

2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library

3.Click on Run or right click and select Run to run the notebook.
4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Edit Metadata. Click on Edit Metadata button appeared on page.
5. Try to navigate throughout the text box using Keyboard.

**Actual**
Keyboard trap occurs inside the text area of the Popup. User can't get out of the text area using TAB or 'Down' arrow key.

This issue exist for 'Chrome + NVDA' configuration also.

**Expected**
Keyboard User should be able to navigate through text-box using TAB or any other keys.

**User Impact**
Keyboard only user will get struck inside the text box and will not be able to reach to other controls of the popup using the keyboard.
**Attachment:**
[A11y_AzureNotebooks_Jupyter_EditMetada_Keyboard18.pptx](https://github.com/jupyter/notebook/files/2401152/A11y_AzureNotebooks_Jupyter_EditMetada_Keyboard18.pptx)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

Issue #4012, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_EditMetada_Keyboard: Keyboard focus does not go to 'Close' button of the Popup.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Chrome: Version 68.0.3440.106 
Operating System:Windows10

**Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Edit Metadata. Click on Edit Metadata button appeared on page.
5. Try to navigate to 'Close' button of the pop-up using keyboard.

**Actual:**
Keyboard focus does not go to 'Close' button of the Popup.

This issue exist for '**Chrome + NVDA**' configuration also.

**Note:**
Same issue is observed in below pages:
1. MoveConfirmationPopup: Navigate to run button & hit enter --> select any file --> navigate to Move button & hit enter.

**Expected:**
Keyboard focus should go to 'Close' button of the Popup.

**User Impact:**
Keyboard only user will not be access the 'Close' button present on the Popup.

**Attachment:**
(https://github.com/jupyter/notebook/files/2401275/A11y_AzureNotebooks_Jupyter_EditMetada_Keyboard.pptx)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4728, opened 2019-06-27
Solved Issue #4012: Keyboad focus does not go to "Close" button in "Edit metadata" dialog

taohan16:
Configure CodeMirror to disable the key "Tab", "Shift" and "Shift-Tab" hot-key.

### DIFF
--- a/Untitled.ipynb
+++ b/Untitled.ipynb
@@ -1,0 +1,81 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Create and deploy an Azure Machine Learning web service\n",
+    "\n",
+    "Although web services can be created directly within [Azure ML Studio](https://studio.azureml.net), in some cases it may be more straightforward to develop and deploy a web service entirely within a Python notebook. Python notebooks allow rapid prototyping and allow developers to include commentary in Markdown. This tutorial demonstrates how the [`azureml`](https://github.com/Azure/Azure-MachineLearning-ClientLibrary-Python) package can be used to deploy Azure ML web services directly from within a Python notebook (or other Python environment).\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


joshuazeltser:
This also fixes issue #4005, where the focus is trapped inside the textarea in the "edit metadata" dialog

Untitled.ipynb
```diff
@@ -0,0 +1,81 @@
+{
```
takluyver: I'm guessing this was committed by mistake - can you remove it again?

takluyver:
I have some reservations about this, because I think that in a large code-editor area, other people will expect tab and shift-tab to work for changing indentation. Is there any way to allow keyboard navigation without blocking that? Or are the two absolutely in conflict?

taohan16:
> I have some reservations about this, because I think that in a large code-editor area, other people will expect tab and shift-tab to work for changing indentation. Is there any way to allow keyboard navigation without blocking that? Or are the two absolutely in conflict?

I was also thinking about the original indentation function, but I think they are in conflict.
I tried to set the "tabindex" attribute to make the CodeMirror editor and other elements to different levels. But the focus still going to the editor directly.
The disable of the "Tab", "Shift" and "Shift+Tab" keymap will only affect the CodeMirror editor in "Edit Metadata" modal dialog.

Here is a link from CodeMirror repo I think is similar to our issue #4012: https://github.com/codemirror/CodeMirror/issues/5444

blink1073:
I agree with @takluyver, we should scope this behavior to those dialogs, and not affect the behavior of all CodeMirror editors.

jtpio:
Closing as such change would have to be implemented in the `nbclassic` repo if still needed: https://github.com/jupyter/nbclassic

Please open an issue or PR there if it is still relevant, thanks!

PR #4729, opened 2019-06-27
Tooltips visible through keyboard navigation for specified buttons: issue #3964

joshuazeltser:
Fixed issue #3964. We have fixed the tooltips for the "Duplicate", "View", "Edit", "Delete" buttons, so that they are visible when navigated to using the keyboard.


takluyver:
I don't pretend to be an expert, but I find it hard to believe we need all this. Tooltips have been a standard browser feature for ages, and the buttons have `aria-label` to expose them to screenreaders. Surely there's some way to make this information visible for keyboard navigation without reimplementing tooltips in a pile of CSS?

joshuazeltser:
From the research we have done just using the title attribute does not work as a tooltip when using focus (as explained here: [https://developer.paciellogroup.com/blog/2013/01/using-the-html-title-attribute-updated/](title attribute). We, therefore, had to implement it in a different way in order for this to work

Issue #3940, opened 2018-09-19
A11y_AzureNotebooks_Jupyter_EditTags_Keyboard:Unable to Edit tags page with the help of keyboard.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 
 **Steps to reproduce:**
1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Tags. Click on Edit Tags button appeared on page.
5.Use keyboard to navigate to page with the help of any of the keyboard keys.
**Actual:**
Unable to navigate to the page according to repro steps with the help of keyboard.
Expected:
We should be able to navigate to the page with the help of keybaord.
 
**UserImpact:**
User will not able to navigate to the page and do unnecessary traversing if focus does not remain intact within the pop-up.
 **Attachment**:

[A11y_AzureNotebooks_Jupyter_EditTags_Keyboard1.pptx](https://github.com/jupyter/notebook/files/2397081/A11y_AzureNotebooks_Jupyter_EditTags_Keyboard1.pptx)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4732, opened 2019-06-27
Solve issue #3940: The dropdown submenu at notebook page is not keyboard accessible

taohan16:
Add "dropdown-toggle" and "dropdown-submenu" classes to link element in html.

Addresses #3940


takluyver:
Are all the CSS & JS changes necessary for this as well, or did they just get mixed up with the HTML changes you mentioned? If they're necessary, can they have some comments to briefly explain what they are?

taohan16:
> Are all the CSS & JS changes necessary for this as well, or did they just get mixed up with the HTML changes you mentioned? If they're necessary, can they have some comments to briefly explain what they are?

For issue#3940, the change in menubar.js is to prevent the dropdown menu is closed when clicking.
The changes in notebook.html are adding "dropdown-toggle" class and data-toggle attribute to <a> tag of submenu because of the when keyboard focus on the dropdown submenu option is actually on the <a> element.
The changes in acc_overwrites.less is to prevent the focus highlight border on submenu options lost after changing.

Issue #3958, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_MenuBar_ScreenReader: Narrator is not announcing "state of links"(enable or disable) under Edit tab.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Operating System:Windows10

**ReproSteps:**
1. Login with the appropriate credentials to azure notebook account.    
2. Navigate to any of the file names present in the samples-->>click on the file name-->Welcome page opens with menubar.                                           
3. Start Narrator.                                                                                 
4.Navigate to Edit tab Links.                                                                              

**Actual:**
Narrator in not announcing "state of links"(enable or disable) under Edit tab.

**Note:**
Same issue is observed in **NVDA** also.
Same issues observed for all the links which are in Enable or Disable State present under all Menus.

**Expected:**
Narrator should announce "state of links"(enable or disable) under Edit tab.

**User Impact:**
Difficult to understand the current state of the link for Screenreader users.

**Attachment:**
(https://github.com/jupyter/notebook/files/2399942/A11y_AzureNotebooks_Jupyter_MenuBar_ScreenReader.pptx)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4733, opened 2019-06-27
Narrator is not announcing "state of links"(enable or disable) under Edit tab. #3958

fabihaahmed:
I have fixed the issue #3958. So now the state of the links(enable or disable) under Edit tab is announced by the narrator.


notebook/templates/notebook.html
```diff
@@ -136,11 +136,11 @@
...
+                        <li id="paste_cell_above" class="disabled" role="none"><a href="#" class="paste_above" role="menuitem" aria-disabled="true">{% trans %}Paste Cells Above{% endtrans %}</a></li>
+                        <li id="paste_cell_below" class="disabled" role="none"><a href="#" class="paste_below" role="menuitem" aria-disabled="true">{% trans %}Paste Cells Below{% endtrans %}</a></li>
+                        <li id="paste_cell_replace" class="disabled" role="none"><a href="#" class="paste_replace" role="menuitem" aria-disabled="true">{% trans %}Paste Cells &amp; Replace{% endtrans %}</a></li>
                         <li id="delete_cell" role="none"><a href="#" role="menuitem">{% trans %}Delete Cells{% endtrans %}</a></li>
-                        <li id="undelete_cell" class="disabled" role="none"><a href="#" role="menuitem">{% trans %}Undo Delete Cells{% endtrans %}</a></li>
+                        <li id="undelete_cell" class="disabled" role="none"><a href="#" class="undelete_link" role="menuitem" aria-disabled="true">{% trans %}Undo Delete Cells{% endtrans %}</a></li>
```
takluyver: We shouldn't need to add classes to all these. We can select them based on the parent's ID, e.g. `$('#undelete_cell > a')` - https://api.jquery.com/child-selector/ .

fabihaahmed: Thanks for the suggestion.

notebook/static/notebook/js/notebook.js
```diff
@@ -1944,6 +1948,7 @@ define([
     Notebook.prototype.set_insert_image_enabled = function(enabled) {
         if (enabled) {
             $('#insert_image').removeClass('disabled');
+            $('.image_link').removeAttr("aria-disabled");
         } else {
             $('#insert_image').addClass('disabled');
```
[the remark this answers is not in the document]
fabihaahmed: I have made the necessary changes(changed the attribute value to true if they are disabled again) however, instead of adding and deleting the aria-disabled attribute I used the attribute value(true/false) to keep track of the state. https://www.w3.org/WAI/PF/aria/states_and_properties#aria-disabled

notebook/static/notebook/js/notebook.js
```diff
@@ -1617,13 +1619,16 @@ define([
         if (!this.paste_enabled) {
             $('#paste_cell_replace').removeClass('disabled')
                 .on('click', function () {that.keyboard_manager.actions.call(
-                    'jupyter-notebook:paste-cell-replace');});
+                    'jupyter-notebook:paste-cell-replace');})
```
takluyver: I think the `;` at the end here should remain.

notebook/static/notebook/js/notebook.js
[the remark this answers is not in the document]
takluyver: Thanks, switching it to true and false is good.

Issue #3217, opened 2018-01-16
Accessibility issues on page.html


snidersd:
Also see: [jupyter/notebook#1801](https://github.com/jupyter/notebook/issues/1801)
After running the Dynamic Assessment Plugin for Chrome and the Digital Content Checker on /localhost/tree the following issues were reported missing and can be corrected as indicated:

1. Indicate the `<main>` region of the page.
2. Add a `<lang>` attribute to the tag.

takluyver:
Thanks for identifying all these things! Unlike some of the other issues, though, it's not clear to me what these mean. Does 'region' refer to geography or part of the screen? How do we indicate it? And which tag needs what attribute added?

takluyver:
I just spotted the HTML tags in the original post. I've edited it to make them visible. You need to put backticks around them, or they vanish:

```
`<main>`
```

I'm still not sure what the `<main>` region is, but that's probably google-able now.

snidersd:
Adding the main element to a web page allows a screenreader user to skip to the main content. Here is a link to additional information on the use of main. https://www.w3.org/TR/html-main-element/

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4734, opened 2019-06-27
Added main/navigation tags to meet DAP requirements: issue #3217

joshuazeltser:
Fixed issue #3217. Ran the Dynamic Assessment Plugin and fixed some of the errors caused by missing regions by adding main and navigation roles to the tree.html page. The lang attribute error does not seem to appear for us.

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -122,7 +122,7 @@ dir="ltr">
   </div>
 </noscript>
 
-<div id="header" role="navigation" aria-label="Top Menu">
+<div id="header" role="navigation" aria-label="{% trans %}Top Menu{% endtrans %}">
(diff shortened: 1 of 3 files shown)


notebook/templates/page.html
```diff
@@ -122,7 +122,7 @@
   </div>
 </noscript>
 
-<div id="header">
+<div id="header" role="navigation" aria-label="Top Menu">
```
yuvipanda:
```suggestion
<div id="header" role="navigation" aria-label="{% trans %}Top Menu{% endtrans %}">
```

yuvipanda: This will ensure localization of this message to users' native languages.

joshuazeltser:
I have added the relevant tags

Issue #3939, opened 2018-09-19
A11y_AzureNotebooks_Jupyter_AzureNotebooksWelcome_Inspect: Label is missing for form textarea element on page.


AccessibilityTestingTeam-TCS:
**Environment Details:**
Application Name: AzureNotebooks_Jupyter
Login URL: notebooks.azure.com
Browser Details:
Edge: Microsoft Edge 42.17134.1.0
Operating System:Windows10

**Repro Steps:**
1.Login with the appropriate credentials to azure notebook account.
2.Navigate to run button & hit enter --> Navigate to Azure Notebooks welcome & hit enter.
3.Verify label relationship of controls on page(i.e. for markdown combobox).

**Actual:**
Label is missing for form textarea element on page.

**Expected:**
Form element or textarea should have meaningful label.

**User Impact:**
Screen reader user would not get information about what type of textbox present on page.

**Recommendation:**
Fix any of the following:
- aria-label attribute does not exist or is empty
- aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty or not visible

**Attachment:**
![a11y_azurenotebooks_jupyter_azurenotebookswelcome_inspect_textarea](https://user-images.githubusercontent.com/39204909/45753025-17992080-bc35-11e8-9563-0a2af423c684.jpg)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4735, opened 2019-06-27
Label is missing for form textarea element on notebook page: issue #3939

fabihaahmed:
I have fixed the issue #3939 by using the aria-label attribute. So now the type of the textarea on notebook page is announced by narrator.


takluyver:
Can you add localisation, like in #4744?

fabihaahmed:
Hi, I have made the necessary changes.

takluyver:
Thanks

Issue #3978, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_RunningTab_ScreenReader: Narrator focus is not in logical order after links present under Notebooks dropdown.


AccessibilityTestingTeam-TCS:
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Operating System:Windows10

**ReproSteps:**
1. Login with the appropriate credentials to azure notebook account.    
2. Navigate to run button & hit enter --> Navigate to running tab & hit enter.
3. Start Narrator.
4. Start Navigating through "Notebooks" dropdown using "Caps+Arrow" keys.
5. Verify Narrator focus is not in logical order after "Azure Notebooks - Welcome.ipynb" link.

**Actual:**
1. While navigating using "Caps+Arrow" keys inside "Notebooks" dropdown Narrator focus is not in logical order after links(Azure Notebooks - Welcome.ipynb).
2. After "Azure Notebooks - Welcome.ipynb" link Narrator focus first goes to (Some Plane text where there is no text or controls present then -->Seconds ago then -->Python 2 then-->Shutdown button) which is incorrect. 

**Note:** 
Same Narrator focus order is observed for all the links which are present under "Notebooks" dropdown.

Same issue is observed for "**Chrome+NVDA**" configuration.

**Expected:**
1. While navigating using "Caps+Arrow" keys inside "Notebooks" dropdown Narrator focus should be in logical order after links(Azure Notebooks - Welcome.ipynb) i.e from "Top to Bottom" and from "Left to Right".
2. After "Azure Notebooks - Welcome.ipynb" link Narrator focus Should not go to Some Plane text where there is no text or controls present. 
After "Azure Notebooks - Welcome.ipynb" link Narrator Focus Order should be
 ("Python 2 text"-->"Shutdown button"-->"Seconds ago text").

**UserImpact:**
Screen reader users face difficulty if Narrator focus order is not in logical order.

**Attachment:**
(https://github.com/jupyter/notebook/files/2400265/A11y_AzureNotebooks_Jupyter_RunningTab_ScreenReader.pptx)

takluyver:
Closed by #4740

PR #4740, opened 2019-06-28
Narrator focus is not in logical order for running notebook links: issue #3978

joshuazeltser:
Fixed issue #3978 by adding a div around the links to change the focus order. This means that focus visits the links in the correct order for each running notebook

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ config.rst
 
 package-lock.json
 geckodriver.log
+*.iml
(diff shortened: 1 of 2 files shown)


takluyver:
- Do we need `div2` at all? Can the spans not just go in `div` after `buttons`?
- Do we need `pull-left`? Things will start from the left inside the div anyway, won't they?

joshuazeltser:
Div2 was used to ensure the correct tab order of the buttons as some buttons exist on multiple tabs within the tree page.  Pull-left ensures the span in aligned with the header for that section. Without these additions, the tab order will be wrong.

takluyver:
OK, I'll take your word for that.

Issue #3977, opened 2018-09-20
A11y_AzureNotebooks_Jupyter_EditTags_Inspect:Name of the "EditTextbox" is missing


AccessibilityTestingTeam-TCS:
EnvironmentDetails:
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Chrome:Version 68.0.3440.106 
Operating System:Windows10
 1.Login with the appropriate credentials to azure notebook account.    
2.Go to Get started à libraries à Select any of the libraries à click on clone to clone the library
3.Click on Run or right click and select Run to run the notebook.
4.Navigate to any file present under Files tab-->Hit enter-->View Menu-->Cell Toolbar-->Tags. Click on Edit Tags button appeared on page.

****Actual:**
Name of the "EditTextbox" is missing
**Expected:****
Proper name of the textbox i.e edit textbox should be provided.
**UserImapct:**
User will not know the correct name of textbox.
**Attachment:**
![attachment2](https://user-images.githubusercontent.com/39204909/45803836-3270a100-bcd8-11e8-97db-6b6524d67eba.png)

jtpio:
Closing as there are now different issues tracking accessibility improvements for Jupyter Notebook 7, for example https://github.com/jupyter/notebook/issues/6800

Thanks!

PR #4744, opened 2019-07-01
Name of the "EditTextbox" is missing: #3977

fabihaahmed:
I have fixed the issue #3977 by adding the aria-label attribute. So now the name of the edit textbox is announced by the narrator.

### DIFF
--- a/notebook/static/notebook/js/celltoolbarpresets/tags.js
+++ b/notebook/static/notebook/js/celltoolbarpresets/tags.js
@@ -166,6 +166,7 @@ define([
             "is treated as tag separators.");
 
         var textarea = $('<textarea/>')
+            .attr('aria-label', 'Edit the tags in the text area')
             .attr('rows', '13')
             .attr('cols', '80')
             .attr('name', 'tags')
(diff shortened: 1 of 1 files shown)


notebook/static/notebook/js/celltoolbarpresets/tags.js
```diff
@@ -166,6 +166,7 @@ define([
             "is treated as tag separators.");
 
         var textarea = $('<textarea/>')
+            .attr('aria-label', 'Edit the tags in the text area')
```
yuvipanda:
```suggestion
            .attr('aria-label', i18n.msg._('Edit the tags in the text area'))
```

yuvipanda: I think this will make it localizable, so users in other languages will hear it in their own language

fabihaahmed: Thanks for the suggestion. I have made the necessary changes.

takluyver:
Thanks!

PR #4745, opened 2019-07-01
Fixed issue #3983: Return focus to Not trusted notification when modal is closed

### LINKED ISSUE #3983: A11y_AzureNotebooks_Jupyter_MenuBar_Keyboard: Keyboard focus is not Shifting to "Not Trusted" button.
**EnvironmentDetails:**
Application Name:AzureNotebooks_Jyupter
Login Url: notebooks.azure.com
BrowserDetails:
Edge:Microsoft Edge 41.16299.611.0
Operating System:Windows10

**ReproSteps:**
1. Login with the appropriate credentials to azure notebook account.    
2. Navigate to any of the file names present in the samples-->>clickin the file name-->Welcome page opens with menubar.
3. Start Navigating through Menubar using keyboard.
4. Verify Keyboard focus is not Shifting to "Not Trusted" button which is present on right of menubar.

**Actual:**
Keyboard focus is not Shifting to "Not Trusted" button.

**Note:**
1. Same issue is observed in "**Chrome+NVDA**" Configuration.
2. And also **Narrator** and **NVDA** focus is also not shifting to "Not Trusted" button.

**Expected:**
Keyboard focus should shifts to all actionable controls or buttons.
In this MenuBar Keyboard focus should shifts to "Not Trusted" button as it is actionable.

**UserImpact:**
Keyboard Only Users face difficulty if Keyboard focus is not Shifting to actionable controls.

**Attachment:**
(https://github.com/jupyter/notebook/files/2400344/A11y_AzureNotebooks_Jupyter_MenuBar_Keyboard.pptx)

ednut15:
#3983 Keyboard focus was not returning to Not Trusted notification button if the pop was closed. Javascript is used to set focus to the button when the modal is closed.


notebook/static/notebook/js/notebook.js
```diff
@@ -2982,13 +2982,17 @@ define([
                 .text(i18n.msg._("here"))
             )
         );
-
+        var fb = '';
+        if(from_notification === 'Notification_button') {
```
takluyver: We've got two stages in this where we compare something against a hardcoded string and then use another hardcoded string:

- `'Notification_button'` -> `'notification_trusted'`
- `'notification_trusted'` -> `'#notification_trusted'`

It's probably better if we pass in the final selector string in the first place, and avoid the comparisons. Just if there's any `focus_button` set at the end, use it as a selector to focus.

.gitignore
```diff
@@ -41,3 +41,7 @@ config.rst
 
 package-lock.json
 geckodriver.log
+
+*.ipynb
+*.iml
+Part\ IA\ Computing\ (Michaelmas)
```
takluyver: I hope your course is going well, but we should keep the project's gitignore file clear of your personal files. ;-)

Also, we don't want to ignore `.ipynb` files in this repository, because we use some of them in docs and tests, e.g. https://github.com/jupyter/notebook/tree/master/docs/source/examples/Notebook

Not sure about .iml files.

notebook/static/base/js/dialog.js
```diff
@@ -130,6 +130,7 @@ define(['jquery',
                 options.keyboard_manager.enable();
                 options.keyboard_manager.command_mode();
             }
+            $(options.focus_button).focus();
```
takluyver: It probably doesn't make much difference to what actually happens, but I think this should still be inside some `if` to check that `focus_button` is defined.

That makes it clear when you're reading the code that it might not be set. And it avoids the possibility of calling some random jQuery method which might not be what we want to do.

Issue #4206, opened 2018-11-13
MAS 1.4.3: Elements must have sufficient color contrast (.c1)


AccessibilityTestingTeam-TCS:
Azure Notebooks - Welcome has an accessibility issue that needs investigation.

Issue Details
Elements must have sufficient color contrast
color-contrast

Element Path
#iframe;.cell.code_cell.rendered:nth-child(4) > .input > .inner_cell > .input_area > .highlight.hl-ipython3 > pre > .c1

Snippet
<span class="c1"># hyperdrive run with child runs having metrics as series</span>

How to fix
Fix any of the following:
- Element has insufficient color contrast of 4.24 (foreground color: #408080, background color: #f7f7f7, font size: 10.5pt, font weight: normal). Expected contrast ratio of 4.5:1

To reproduce
Use K4W to investigate the issue details

Environment
Chrome version 69.0.3497.100

This accessibility issue was found using Keros for Web, a tool that helps debug and find accessibility issues earlier. Get more information & download this tool at http://aka.ms/K4Web.

[MAS 1.4.3 Elements must have sufficient color contrast (.c1)_1.zip](https://github.com/jupyter/notebook/files/2576198/MAS.1.4.3.Elements.must.have.sufficient.color.contrast.c1._1.zip)

takluyver:
Fixed by #4746.

PR #4746, opened 2019-07-01
Solve issue #4206: The contrast of font color of code comment in code cell is lower than 4.5:1

taohan16:
Overwrite the highlight-comment style class with color #007979


taohan16:
Before
color contrast: 4.2:1
After
color contrast: 4.8:1

takluyver:
Is it possible you've used the wrong 'after' screenshot?

Untitled.ipynb
```diff
@@ -0,0 +1,81 @@
+{
```
takluyver: I guess this was committed accidentally. :-)

notebook/static/notebook/less/highlight.less
```diff
@@ -64,6 +64,12 @@ Adapted from GitHub theme
   color: #AA22FF;
 }
 
+//fix issue #4206, overwrite .highlight-comment class
+.highlight-comment{
+  color:#007979;
```
takluyver: Do we need to add a new rule? Can we not modify the existing one at line 32?

taohan16: Sorry for the mistakes made, I had updated the screenshots and code now.

taohan16: > Do we need to add a new rule? Can we not modify the existing one at line 32?

Sorry for the mistakes made, I had updated the screenshots and code now.

PR #4752, opened 2019-07-07
[WIP] [Proof of Concept] Add eventlogging telemetry

yuvipanda:
Very very early WIP PR to add Event Logging telemetry to
notebook server. 

Mostly here so we can run stuff against Travis :)


yuvipanda:
This is ready for review now!

notebook/notebookapp.py
```diff
@@ -1667,10 +1672,23 @@ def init_shutdown_no_activity(self):
             pc = ioloop.PeriodicCallback(self.shutdown_no_activity, 60000)
             pc.start()
 
+    def init_eventlog(self):
+        self.eventlog = EventLog(parent=self)
+
+        yaml = YAML(typ='safe')
```
Zsailer: Can we drop this call if it's already inside jupyter/telemetry (see [here](https://github.com/jupyter/telemetry/blob/0022131cefac065909fcbb85a5c2b2bf3a6adc8a/jupyter_telemetry/eventlog.py#L16))?

notebook/services/contents/handlers.py
```diff
@@ -122,9 +126,17 @@ def patch(self, path=''):
...
             raise web.HTTPError(400, u'JSON body missing')
+        self.log.info(model)
         model = yield maybe_future(cm.update(model, path))
         validate_model(model, expect_content=False)
         self._finish_model(model)
+        self.log.info(model)
+        self.eventlog.record_event(
+            eventlogging_schema_fqn('contentsmanager-actions'), 1,
+            # FIXME: 'path' always has a leading slash, while model['path'] does not.
+            # What to do here for source_path? path munge manually? Eww
```
Zsailer: Ew is right. 

It looks like we call `path.strip('/')` [in many places](https://github.com/jupyter/notebook/search?q=path.strip%28%27%2F%27%29&unscoped_q=path.strip%28%27%2F%27%29) across the notebook codebase to strip leading slash.

notebook/services/eventlog/handlers.py
```diff
@@ -0,0 +1,42 @@
...
+        for rf in required_fields:
+            if rf not in raw_event:
+                raise web.HTTPError(400, '{} is a required field'.format(rf))
+
+        schema_name = raw_event['schema'] 
+        version = raw_event['version']
+        event = raw_event['event']
+        
+        # Profile, and move to a background thread if this is problematic
+        # FIXME: Return a more appropriate error response if validation fails
```
Zsailer: Flagging this here, so we don't forget to catch this error.

Zsailer:
This is awesome, @yuvipanda. Great work! I'm really excited to see this land. I'll do a more careful review when I get some time, but I think this is mostly ready!

notebook/notebookapp.py
yuvipanda: Yep, we do that in jupyter_telemetry now, this PR needs updating

meeseeksmachine:

https://discourse.jupyter.org/t/jupyter-telemetry-system/1217/1

Zsailer:
@yuvipanda and @jaipreet-s, I think this is ready for review. 

I've added documentation, autogeneration of event schema documentation, and fixed some small code related things. 

When I get your approval, I think we can merge this.

@lresende, do you have any objections to merging this into notebook? (We'll move over to jupyter_server if this gets merged.)

Zsailer:
Also, @kevin-bates if you have thoughts/opinions on this, I'd love to hear them (I don't want to add extra work to your plate 😃).

notebook/services/contents/handlers.py
```diff
@@ -237,6 +277,10 @@ def delete(self, path=''):
         yield maybe_future(cm.delete(path))
         self.set_status(204)
         self.finish()
+        self.eventlog.record_event(
+            eventlogging_schema_fqn('contentsmanager-actions'), 1,
+            { 'action': 'delete', 'path': path }
```
Zsailer: Need to `lstrip` path here.

docs/source/eventlog.rst
```diff
@@ -0,0 +1,61 @@
...
+
+The output is a file, ``"event.log"``, with events recorded as JSON data.
+
+`eventlog` endpoint
+-------------------
+
+The Notebook Server provides a public REST endpoint for external applications to validate and log events
+through the Server's Event Log. 
+
+To log events, send a `POST` request to the `/api/eventlog` endpoint. The body of the request should be a
```
jaipreet-s: Do we need to add documentation around this REST API in https://github.com/jupyter/jupyter/wiki/Jupyter-Notebook-Server-API or where-ever is the authoritative source>

notebook/services/contents/handlers.py
```diff
@@ -122,9 +127,20 @@ def patch(self, path=''):
         model = self.get_json_body()
         if model is None:
             raise web.HTTPError(400, u'JSON body missing')
+        self.log.info(model)
```
jaipreet-s: Left-over debug logs

notebook/services/contents/handlers.py
```diff
@@ -122,9 +127,20 @@ def patch(self, path=''):
...
         model = yield maybe_future(cm.update(model, path))
         validate_model(model, expect_content=False)
         self._finish_model(model)
+        self.log.info(model)
```
jaipreet-s: Left-over debug logs?

notebook/services/eventlog/handlers.py
```diff
@@ -0,0 +1,42 @@
...
+    A handler that receives and stores telemetry data from the client.
+    """
+    @json_errors
+    @web.authenticated
+    def post(self, *args, **kwargs):
+        try:
+            # Parse the data from the request body
+            raw_event = json.loads(self.request.body.strip().decode())
+        except Exception as e:
+            raise web.HTTPError(400, str(e))
```
jaipreet-s: We're doing str(e) here and just passing e in line 35 - Lets keep it consistent?

notebook/services/eventlog/handlers.py
```diff
@@ -0,0 +1,42 @@
...
+        version = raw_event['version']
+        event = raw_event['event']
+        
+        # Profile, may need to move to a background thread if this is problematic
+        try: 
+            self.eventlog.record_event(schema_name, version, event)
+        except Exception as e:
+            raise web.HTTPError(400, e)
+            
+        self.set_status(204)
```
jaipreet-s: nitpick: 204 is more common for PUTs, but do we have a reason for 200 vs 204?

yuvipanda: My original thinking was that we'd want `record_event` to be in a background thread, so we can return immediately after the request, without waiting for the sequence of events from `record_event` (which might include further HTTP requests to the ultimate backend!) to complete. This means we won't actually know if the request was a success - since we returned early. This is why I had this return 204 rather than 200.

I also considered making this a PUT but realized early on it isn't a good match since we're posting to `/api/eventlog` rather than `/api/eventlog/schema`

PR #4756, opened 2019-07-10
Update jupyter_client

SylvainCorlay:


SylvainCorlay:
This is just bumping jupyter_client requirement by a patch version to prevent installing a broken version.

PR #4785, opened 2019-07-26
Prevent tests from hanging

dleen:
When the NotebookApp raises an exception on initialization, a test can
hang forever. This can happen when passing configuration when
customizing NotebookTestBase:
```
class TestMissingExtension(NotebookTestBase):
    @classmethod
    def get_argv(cls):
        argv = super(TestMissingExtension, cls).get_argv()
        argv.extend(
            [
                "--NotebookApp.session_manager_class=doesnt_exist",
            ]
        )
        return argv

    def test_this_will_hang_forever(self):
        pass
```

Since the exception happens before the try/finally the `Event` will
never be triggered. By including the construction and initialization of
the notebook in the try/finally we can handle situations like this.


jaipreet-s:
CI failure looks un-related(Frontend/browser)
https://travis-ci.org/jupyter/notebook/jobs/564220954

takluyver:
Looks good, thanks. If you're interested in improving this further, I think it might be useful to re-raise exceptions from the thread so the error is clearer if it fails to start.

PR #4803, opened 2019-08-02
exit status for serverextension validation failure

amangarg96:
I have a use case where I need to enable certain jupyter serverextensions before launching a notebook. 
To ensure that the extensions were successfully enabled, I looked at the exit status and realized that the [validate_serverextension()](https://github.com/jupyter/notebook/blob/master/notebook/serverextensions.py#L80) does not handle failure.

I'm raising this PR for fixing that, where it sets exit status 1 when serverextension enable fails.

### DIFF
--- a/notebook/serverextensions.py
+++ b/notebook/serverextensions.py
@@ -106,20 +106,21 @@ def validate_serverextension(import_name, logger=None):
         version = getattr(mod, '__version__', '')
     except Exception:
         logger.warning("Error loading server extension %s", import_name)
-
-    import_msg = u"     {} is {} importable?"
-    if func is not None:
-        infos.append(import_msg.format(GREEN_OK, import_name))
-    else:
-        warnings.append(import_msg.format(RED_X, import_name))
-
-    post_mortem = u"      {} {} {}"
-    if logger:
-        if warnings:
-            [logger.info(info) for info in infos]
-            [logger.warn(warning) for warning in warnings]
+        sys.exit(1)
+    finally:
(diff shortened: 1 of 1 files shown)


amangarg96:
@kevin-bates Since you have worked on nb2kg which is a server extension, do you have any inputs on this?

notebook/serverextensions.py
```diff
@@ -106,20 +106,21 @@ def validate_serverextension(import_name, logger=None):
...
-        infos.append(import_msg.format(GREEN_OK, import_name))
-    else:
-        warnings.append(import_msg.format(RED_X, import_name))
-
-    post_mortem = u"      {} {} {}"
-    if logger:
-        if warnings:
-            [logger.info(info) for info in infos]
-            [logger.warn(warning) for warning in warnings]
+        sys.exit(1)
```
kevin-bates: We shouldn't exit the process since this method is called multiple times from [`ListServerExtensionsApp`](https://github.com/jupyter/notebook/blob/master/notebook/serverextensions.py#L240) and an early warning will prevent listing subsequent extensions.

jtpio:
Hey @amangarg96, thanks to opening this PR :+1: 

It looks like there has not been any activity for a while. Also Jupyter Server is now developed in https://github.com/jupyter-server/jupyter_server. So closing this PR since such change is now unlikely to be implemented in the notebook repo.

Please feel free to re-open the PR on https://github.com/jupyter-server/jupyter_server if it is still relevant. Thanks all!

PR #4826, opened 2019-08-13
Ctrl-A to select all cells

### LINKED ISSUE #4820: Use Ctrl-A to select all cells so then they can be run with Shift-Enter or Ctrl-Enter
I think this would be a cool feature and serve as an easy way to run all cells on the keyboard. We should be in command mode to do this so it doesn't interfere with selecting all the text when we're working inside of a cell.

benthayer:
The code to do this was very simple and uses the same methods as shift-clicking on a range. Now, when in command mode, Ctrl-A selects all cells

closes #4820

### DIFF
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -298,6 +298,15 @@ define([
                 env.notebook.get_selected_cell().element.focus();
             }
         },
+        'select-all' : {
+            cmd: i18n.msg._('select all'),
+            help: i18n.msg._('select all cells'),
+            help_index : 'de',
+            handler : function (env) {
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks. Normally we avoid Ctrl- shortcuts because they clash with browser shortcuts and people complain, but I'm inclined to make this an exception, like Ctrl-S, because AFAIK it always means 'select all', and the default browser behaviour for that is not very useful in a notebook.

I'll leave it a couple more days in case anyone sees this and wants to point out that their browser does something different, or argue that selecting all text on the page is actually useful.

benthayer:
Thanks for merging!

PR #4856, opened 2019-09-02
Update contents API docs

vidartf:
Now the contents API docs better match the default notebook implementation:

https://github.com/jupyter/notebook/blob/c2d4561ba6b25ec4e0f1f9ec617c3338f4de1ef4/notebook/services/contents/filemanager.py#L362

### DIFF
--- a/docs/source/extending/contents.rst
+++ b/docs/source/extending/contents.rst
@@ -100,9 +100,9 @@ model. There are three model types: **notebook**, **file**, and **directory**.
    In certain circumstances, we don't need the full content of an entity to
    complete a Contents API request.  In such cases, we omit the ``content``, and
    ``format`` keys from the model, and the default values for the ``mimetype``
-   field. This most commonly occurs    when listing a directory, in which
-   circumstance we represent files within    the directory as content-less models
-   to avoid having to recursively traverse    and serialize the entire filesystem.
+   field. This most commonly occurs when listing a directory, in which
+   circumstance we represent files within the directory as content-less models
+   to avoid having to recursively traverse and serialize the entire filesystem.
 
 **Sample Models**
 
--- a/docs/source/extending/contents.rst
+++ b/docs/source/extending/contents.rst
@@ -81,7 +81,9 @@ model. There are three model types: **notebook**, **file**, and **directory**.
     - The ``format`` field is either ``"text"`` or ``"base64"``.
     - The ``mimetype`` field can be any mimetype string, but defaults to 
       ``text/plain`` for text-format models and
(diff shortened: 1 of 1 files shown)


docs/source/extending/contents.rst
```diff
@@ -79,7 +79,8 @@ model. There are three model types: **notebook**, **file**, and **directory**.
 
 - ``file`` models
     - The ``format`` field is either ``"text"`` or ``"base64"``.
-    - The ``mimetype`` field is ``text/plain`` for text-format models and
+    - The ``mimetype`` field is can be any mimetype string, but defaults to
```
Zsailer:
```suggestion
    - The ``mimetype`` field can be any mimetype string, but defaults to 
```

docs/source/extending/contents.rst
```diff
@@ -97,11 +98,11 @@ model. There are three model types: **notebook**, **file**, and **directory**.
...
 
    In certain circumstances, we don't need the full content of an entity to
-   complete a Contents API request.  In such cases, we omit the ``mimetype``,
-   ``content``, and ``format`` keys from the model. This most commonly occurs
-   when listing a directory, in which circumstance we represent files within
-   the directory as content-less models to avoid having to recursively traverse
-   and serialize the entire filesystem.
+   complete a Contents API request.  In such cases, we omit the ``content``, and
+   ``format`` keys from the model, and the default values for the ``mimetype``
+   field. This most commonly occurs when listing a directory, in which
```
takluyver: > we omit... and the default values for the `mimetype` field.

This wasn't very clear to me without looking at the extension. Maybe "...and the `mimetype` field may be `None` if the file extension is not recognised."

vidartf: Thanks! I'll add it as a separate sentence to make it clear.

vidartf: Commits added to clarify. The logic is rather convoluted, so the description ends up somewhat unclear, but at least it makes it clear that the value for files might be `None` when doing dir listings.

takluyver:
Thanks :)

docs/source/extending/contents.rst
takluyver: I wonder if we should actually change the logic so it either omits the mimetype entirely when there's no content or uses `application/octet-stream` as the default instead of None. AFAIK, 'octet-stream' is applicable to any regular file including text (but presumably not directories etc.).

vidartf: I'm OK with the way it is currently working, I was just making some assumptions based on the docs that didn't hold up in practice. :)

PR #4858, opened 2019-09-03
update CodeMirror

### LINKED ISSUE #4854: Looks like Python f-formatted strings display incorrectly in edit mode
I haven't fully diagnosed what the issue is but it seems that when you use a string like `'A'` inside a Python format string, it seems to confuse the display of the text after that.

Here is a screenshot from a notebook for example:

Jupyter_Notebook_screenshot_190901

Problem:
 - When you look at the line `print(f"Lookup the {things['A']} and the remainder of the text print(looks_like(code + 1))")` in the edit cell, the text to the right of  `{things['A']}` is formatted as if it where code.

Expected result:
 - Expected the viewer to recognize that the text after `{things['A']}` is part of a string not embedded code.

Further information:
 - This problem only seems to occur when you use a string literal like `'A'` or `"A"` inside a `f""` type format string.  The last line in the screenshot shows that if you replace this with an integer key the problem does not occur.

System info:
 - The version of the notebook server is: 5.7.8
 - Safari Version 12.1.2
 - Python 3.7.1 (default, Dec 14 2018, 13:28:58)

bermani:
The current CodeMirror version incorrectly highlights f-formatted strings, upgrading solves the problem

Fixes #4854

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "backbone": "components/backbone#~1.2",
     "bootstrap": "bootstrap#~3.4",
     "bootstrap-tour": "0.9.0",
-    "codemirror": "components/codemirror#~5.37",
+    "codemirror": "components/codemirror#~5.48.4",
     "create-react-class": "https://cdn.jsdelivr.net/npm/create-react-class@15.6.3/create-react-class.min.js",
     "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.7.0",


t-makaro:
Also fixes [jupyter_contrib_nbextensions #1149](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1149).

lokeshh:
This is a very annoying bug. Lets merge this ASAP?

PR #4874, opened 2019-09-11
Allow a session to persist beyond its kernel's lifetime.

jasongrout:
This allows sessions to persist beyond a kernel’s lifetime, allowing us to have a more persistent reference to shared kernel that can die, be restarted, and be changed.

This is somewhat still a work-in-progress. I'm still experimenting with a frontend that can take advantage of such persistent sessions.

### DIFF
--- a/notebook/services/sessions/sessionmanager.py
+++ b/notebook/services/sessions/sessionmanager.py
@@ -15,16 +15,34 @@ from tornado import gen, web
 
 from traitlets.config.configurable import LoggingConfigurable
 from ipython_genutils.py3compat import unicode_type
-from traitlets import Instance
+from traitlets import Instance, CBool
 
 from notebook.utils import maybe_future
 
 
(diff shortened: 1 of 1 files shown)


kevin-bates:
Hi @jasongrout - these server-side changes look consistent with what you were describing at last week's open studio. I'll be happy to review once ready.

vidartf:
Just for reference, what will we be able to do after this change, that isn't currently possible (or, what will become easier)?

jasongrout:
> Just for reference, what will we be able to do after this change, that isn't currently possible (or, what will become easier)?

It will mean that a session can exist independent of its kernel, i.e., if a kernel is shut down, the session still lives. Another consequence should be that a session can be created without a kernel, and then a kernel started later, though I haven't tried that yet.

jtpio:
Closing as stale.

Please feel free to continue in https://github.com/jupyter-server/jupyter_server if still needed, thanks!

PR #4893, opened 2019-09-16
Add some minimal support for increasing the soft limit on open file handles, default to at least 4096

jasongrout:
This hopefully fixes https://github.com/jupyterlab/jupyterlab/issues/6727

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -22,6 +22,7 @@ import mimetypes
 import os
 import random
 import re
+import resource
 import select
 import signal
 import socket
@@ -31,6 +32,12 @@ import threading
 import time
 import warnings
 import webbrowser
+
+try:
+    import resource
+except ImportError:
+    # Windows
+    resource = None
 
 try: #PY3
     from base64 import encodebytes
@@ -802,6 +809,14 @@ class NotebookApp(JupyterApp):
         """
     )
 
+    min_open_files_limit = Integer(4096, config=True,
+        help="""
+        Gets or sets a lower bound on the open file handles process resource
+        limit. This may need to be increased if you run into an
+        OSError: [Errno 24] Too many open files.
+        This is not applicable when running on Windows.
+        """)
+
     @observe('token')
     def _token_changed(self, change):
         self._token_generated = False
@@ -1379,6 +1394,22 @@ class NotebookApp(JupyterApp):
         logger.parent = self.log
         logger.setLevel(self.log.level)
     
+    def init_resources(self):
+        if resource is None:
+            self.log.debug('Ignoring min_open_files_limit because the limit cannot be adjusted (for example, on Windows)')
+            return
+
+        old_soft, old_hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+        soft = self.min_open_files_limit
+        hard = old_hard
+        if old_soft < soft:
+            if hard < soft:
+                hard = soft
+            resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
+            self.log.debug(
+                'Raised open file limit: soft {}->{}; hard {}->{}'.format(old_soft, soft, old_hard, hard)
+            )
+
     def init_webapp(self):
         """initialize tornado webapp and httpserver"""
         self.tornado_settings['allow_origin'] = self.allow_origin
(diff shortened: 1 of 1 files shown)


jasongrout:
WIP - let's test this with some of the people that have been having problems on jupyterlab/jupyterlab#6727 before merging.

notebook/notebookapp.py
```diff
@@ -1379,6 +1393,22 @@ def init_logging(self):
...
+        soft = self.min_open_files_limit
+        hard = old_hard
+        if old_soft < soft:
+            if hard < soft:
+                hard = soft
+            resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
+            self.log.debug(
+                'Raised open file limit: soft {}->{}; hard {}->{}'.format(old_soft, soft, old_hard, hard)
+            )
+
```
kevin-bates: This strikes me as more of an informational message than debug.  I would also suggest taking an optimistic approach and logging 'Raising open file...' prior to the actual call.  This way, the log will contain the values that were presumably bad and could prove helpful in troubleshooting.

jasongrout: Very good point!

jasongrout:
> I believe this is an excellent example of a traitlet that would be useful for dynamic configuration - where the server auto-detects config changes and reloads. In this case, we'd probably want the code that performs the update in an @observe decorator - but that's future stuff (see [ipython/traitlets#521](https://github.com/ipython/traitlets/pull/521) and [ipython/traitlets#522](https://github.com/ipython/traitlets/pull/522)). This would allow admins encountering 'Too Many Open Files' to at least extend the server's life while addressing the (probable) underlying issue.

Good point too!

jasongrout:
Feedback at https://github.com/jupyterlab/jupyterlab/issues/6727#issuecomment-536682037 was that this fixed the issue we were seeing. I'll clean things up like you mention, @kevin-bates.

notebook/notebookapp.py
jasongrout: I went ahead and left it a debug message, but moved it before the action. My thinking is that for most people, this should never be a concern, so don't clutter up their terminal. However, if you are trying to track down an issue, likely you are running with --debug.

jasongrout:
@kevin-bates, @lresende, and whoever else - I think this is ready for review.

Zsailer:
LGTM, too.

Zsailer:
I'll port this to jupyter_server, too.

notebook/notebookapp.py
```diff
@@ -802,6 +808,14 @@ def _token_default(self):
         """
     )
 
+    min_open_files_limit = Integer(4096, config=True,
```
mpacer: Feel free to ignore this suggestion, but I think the traitlet name be clearer if it were `soft_open_files_limit` or `lower_open_files_limit`. The reason I suggest this is that this is a soft bound on the maximum number of files that can be created, rather than a minimum number of files that must be opened.

mpacer: To be fair, you could call it simply `open_files_limit` as well, since in practice we wouldn't be allowing people to increase the hard limit imposed by the operating system.

jasongrout: I called it `min_open_files_limit` because I only set it if the current open files limit is lower than this number (i.e., I will never decrease an open files limit to this number, only raise lower open file limits to at least this.)

https://github.com/jupyter/notebook/pull/4893/files#diff-bbd307804bb6b8836da9f06b8d8a06e9R1405

That said, I also think the name is confusing, but descriptive, and would love an easier to understand name.

notebook/notebookapp.py
```diff
@@ -1371,6 +1384,23 @@ def init_logging(self):
...
+        old_soft, old_hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+        soft = self.min_open_files_limit
+        hard = old_hard
+        if old_soft < soft:
+            if hard < soft:
+                hard = soft
+            self.log.debug(
+                'Raising open file limit: soft {}->{}; hard {}->{}'.format(old_soft, soft, old_hard, hard)
+            )
+            resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
```
shevelevs: According to the documentation (https://docs.python.org/3/library/resource.html): 
> Raises ValueError if an invalid resource is specified, if the new soft limit exceeds the hard limit, or (*if a process tries to raise its hard limit*. ... A process with the effective UID of *super-user* can request any valid limit value

So it seems this will fail for non-root

PR #4927, opened 2019-10-01
Readme

santosh2702:
Added System Requirement for using Jupyter notebook. Please review


minrk:
Can I ask why you've added these and where they came from? They are not necessarily accurate system requirements, especially the "Included Python packages." I'd rather leave it as "Python >=3.5" because it's very unlikely that anywhere that has Python can't run the notebook.

santosh2702:
It is just a detail because documentation must include these things as we try any new software we check system requirement.

takluyver:
README space is super valuable - we're not going to clutter it up with irrelevant lists of processor models.

Jupyter should run on essentially any computer from the last ~10 years or so. It's what you do inside Jupyter that might need RAM, CPU cores, GPUs...

PR #4928, opened 2019-10-01
5.2.2

mmaithani:
Still they don't provide dark mode though i am using my custom made dark mode by my neuroscientist friend


minrk:
also a feature request opened as a pull request. Please open Issues instead. Also, JupyterLab has a dark mode, so I'd recommend you try that:

Screen Shot 2019-10-03 at 09 43 58

PR #4933, opened 2019-10-01
fix doc about pip install with test extra require for zsh

00Kai0:
When I install dev requirement with `pip install -e .[test]`, I get a error:
```
zsh: no matches found: .[test]
```
My `pip` version is `19.2.3`.

A solution is use '.[test]' in cmd.


minrk:
thanks!

Issue #4922, opened 2019-09-30
Shell we remove py2 dependence in jupyter notebook?


00Kai0:
I saw the new `setup.py` in notebook shows:
```python
    error = """
Notebook 6.0+ supports Python 3.5 and above.
When using Python 3.4 or earlier (including 2.7), please install Notebook 5.x.

Python {py} detected.
{pip}
""".format(py=sys.version_info, pip=pip_message )

```

So the new notebook just supports `py>=3.5`.

And I saw many codes like:
```python
try: #PY3
    from base64 import encodebytes
except ImportError: #PY2
    from base64 import encodestring as encodebytes

```

Is it time to remove them?
I wanne try to do this.

takluyver:
Fixed by #4936 (as far as I know - please feel free to open further PRs if there's more compatibility code that can be removed)

PR #4936, opened 2019-10-02
Rm dependences of py2 in base path.

00Kai0:
like #4922

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -32,10 +32,7 @@ import time
 import warnings
 import webbrowser
 
-try: #PY3
-    from base64 import encodebytes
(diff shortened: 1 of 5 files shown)


notebook/notebookapp.py
```diff
@@ -71,10 +68,7 @@
 )
 
 # py23 compatibility
-try:
-    raw_input = raw_input
-except NameError:
-    raw_input = input
+raw_input = input
```
takluyver: It looks like this is no longer used, so it should just go away entirely.

00Kai0: Hi, @takluyver. You are right. I have removed it.

takluyver:
Thanks!

PR #4945, opened 2019-10-04
Compress and download folder

hadim:
The goal to this PR is to integrate https://github.com/hadim/jupyter-archive to Jupyter. 

Two others to JLab and `jupyter_server` will also be submitted.

For now, there is nothing in this commit because I need some guidance on how to integrate the Python code to the notebook.

My idea was to create a class called `ArchiveHandler` next to `FilesHandler` in `files/handlers.py` but it appears that `FilesHandler.get` is never called when downloading a file from notebook or Jlab.

Instead `web.StaticFileHandler` is used `base.handlers.AuthenticatedFileHandler.get`.

Two questions:

- What is the point of `FilesHandler` then?
- Where should I implement the `ArchiveHandler`? The code is here: https://github.com/hadim/jupyter-archive/blob/master/jupyter_archive/handlers.py#L45

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ config.rst
 package-lock.json
 geckodriver.log
 *.iml
+.vscode/


hadim:
JLab related PR: https://github.com/jupyterlab/jupyterlab/pull/7294

hadim:
I am not sure I have implemented the way you guys want but this PR + the JLab PR is fully working.

meeseeksmachine:

https://discourse.jupyter.org/t/jupyter-archive-make-download-and-extract-archive-files/2363/2

PR #4961, opened 2019-10-14
ordering of the drop down menus based on severity

yonashub:
I think it is better to group (and order) the dropdown menu options based on their effect (and severity) for this I propose to put reconnect and interrupt at first and adding a separator, then the rest of the options.


blink1073:
Thanks @yonashub!  Sorry for the delay.  We're catching up on reviews.  Are you available to pick this back up?  If so, please rebase and include a screenshot of the new dialog.

PR #4971, opened 2019-10-15
Fix default value for display_url function with fqdn instead of hostname

glegoux:
A fully qualified domain name (FQDN) is the complete domain name for a specific computer.
A hostname is the name of a computer.

The DNS should contain an A record of the FQDN and not of the hostname.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1458,7 +1458,7 @@ class NotebookApp(JupyterApp):
                 url += '/'
         else:
             if self.ip in ('', '0.0.0.0'):
-                ip = "%s" % socket.gethostname()
+                ip = "%s" % socket.getfqdn()
             else:
                 ip = self.ip
             url = self._url(ip)


notebook/notebookapp.py
```diff
@@ -1458,7 +1458,7 @@ def display_url(self):
                 url += '/'
         else:
             if self.ip in ('', '0.0.0.0'):
-                ip = "%s" % socket.gethostname()
+                ip = "%s" % socket.getfqdn()
```
kevin-bates, line 1461: I think more needs to happen here.  On my mac, `socket.getfqdn()` returns: `'1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa'` vs. `socket.gethostname()`: `'Kevins-MacBook-Pro.local'` and, while it might be perfectly valid, the former just doesn't make for a good user presentation.  

Also, your title states that you're fixing the value for `custom_display_url`, but this block only occurs when that option has not been configured.

glegoux: I updated the title

glegoux: First, I think a working link is better than a beautiful link not working in other contexts. 

But yes, your display is not user friendly. I don't know what is your local configuration in MAC OS, I am a Linux user, I guess it is the same thing for default configuration?

My vision is that by default, the link should be always valid in local or remote server context, 2 solutions for you:
1. You can configure your `hosts` file to have a better display  for the default value displayed. 
2. Your use case enters in the usage of setting custom_display_url with `Kevins-MacBook-Pro.local` directly.

kevin-bates: > First, I think a working link is better than a beautiful link not working in other contexts.

I agree. And my Linux machine produces the correct result as well - which also matches the result from `gethostname()`.

I polled a few colleagues and all of their macs return expected values.  I also completely agree with the spirit of this change and had my computer been working like my colleagues, this would have been approved.  

However, while I'm sure I can adjust my configuration to get my scenario working,  the fact that a change is required at all makes me hesitant to accept this PR since I'd be on the hook for answering the onslaught of issues that could potentially arise. Based on the number of google hits I see others encountering, this could be significant - and I suspect most that might have this issue on their computers don't even know it.  (I had never seen an IP like that before testing this out.)

I'm sorry, but I'll need to defer to another maintainer because I'm not interested in signing up for the support burden this could introduce, and the demand for such a change, as best as I can determine, just hasn't been there.

lresende:
@glegoux Is there an issue associated with this change? Otherwise, could you explain a little more on the PR details what is the overall issue you are trying to resolve (e.g. what is broken, sample URL versus expected URL, etc).

glegoux:
Hello @lresende, concretely I launch a Jupyter Notebook on any remote machine, the displayed link is not working. Because the hostname is displayed, that is not a valid domain name (locally it works because the `hosts` file gives the hostname as an alias name to localhost, 127.0.0.1 or ::1 to do the DNS resolution). For example, I get 'server01' instead of 'server01.example.com'.

For the users that run Jupyter Notebook locally, some ones have configured their fqdn (it works), but a lambda user don't have configured that, it works again (because socket.getfqdn() shows their hostname, in this case nothing change for them).

kevin-bates:
HI @glegoux - because there's a non-zero chance that this breaks existing users, would you be open to introducing an option that allows users to opt-in to this and we tie that option to a flag for ease of enablement?

I'm thinking the option would be `use_fqdn` (default = False) and we'd tie it to a flag of `--use-fqdn` (equating to a True value for the option).   The code would then become...
```python
if self.ip in ('', '0.0.0.0'):
     ip = "%s" % (socket.getfqdn() if self.use_fqdn else socket.gethostname())
```
Your users could either set `--use-fqdn` when launching notebook/lab or configure `c.NotebookApp.use_fqdn = True`

glegoux:
Good idea, yes I can do that 😄 !

jtpio:
Based on the comments above, it looks like there is no real action to be taken here, and the PR can be closed.

Also the server part of the notebook application now lives in https://github.com/jupyter-server/jupyter_server.

Feel free to open a new issue in https://github.com/jupyter-server/jupyter_server if it is still relevant, thanks!

PR #4987, opened 2019-10-18
Fix checking if kernel is running through Selenium

takluyver:
Avoid error "Jupyter.notebook.kernel is null", as seen in e.g. [this recent CI failure on master](https://travis-ci.org/jupyter/notebook/jobs/592932509).

Presumably this is a race condition - if it tries to check the status before `Jupyter.notebook.kernel` is created, it throws this error.


takluyver:
Running the Selenium tests locally, I've got the error 2/2 times on master,  and 0/2 times on this branch.

takluyver:
Thanks Kevin :)

Issue #4517, opened 2019-03-27
Safari monospace font issue


pcr910303:
I can't understand why this is happening, but on Safari 12.0.3 in macOS High Sierra 10.13.6, fonts are not monospaced.

<img width="1240" alt="non-monospace font" src="https://user-images.githubusercontent.com/37990858/55047244-6feb2800-5087-11e9-814e-e979b5956c0b.png">

Could also reproduce same issue in macOS Mojave 10.14.3 with Safari 12.0.3.

Strangely, when I change the CSS code of `.CodeMirror` from `font-family: monospace` to `font-family: [any monospace font name here]` it displays well. 

E.g. something like this

```diff
.CodeMirror {
-   font-family: monospace;
+   font-family: SF Mono;
    height: 300px;
    color: black;
    direction: ltr;
}
```

I'm not a CSS guru, but this issue doesn't happen in plain CodeMirror editor. Maybe Jupyter notebook's CSS is overriding the default `monospace` font? (Is this even possible?)

minrk:
I don't believe we can override the default monospace font. This should be browser configuration. What do you see [on this page](https://jsfiddle.net/zpr16ou2/1/)?

pcr910303:
@minrk That works fine... It's using Courier New.
As I said, the plain CodeMirror editor also works fine, so I'm not sure what's happening.
While I was searching for this, I found [this SO question](https://stackoverflow.com/questions/52529025/jupyter-notebook-monospace-font-doesnt-render-properly-in-safari), but there is no answer yet :-(

zcqian:
I am seeing the exact same issue here. Perhaps it is related to user's locale settings? I will test this if it would be helpful.

I can confirm that on Firefox, text areas in notebooks use monospace fonts correctly.

pcr910303:
Well, I found that when I delete the attribute `lang` of tag `html` in DevTools, monospace fonts come back.
So, something like this:

```diff
- <html lang="ko-kr">
+ <html>
```

makes from this:

<img width="366" alt="무제" src="https://user-images.githubusercontent.com/37990858/57071319-065cda00-6d15-11e9-87ee-1de2c11f4d42.png">

to this:
<img width="384" alt="무제" src="https://user-images.githubusercontent.com/37990858/57071358-2d1b1080-6d15-11e9-9b66-de4eeeee4642.png">

I tested this only in BinderHub (not in local), but I'm fairly sure this is the problem. 

[...]

PR #4989, opened 2019-10-19
Fix Safari monospace font issue (bug #4517).

adah1972:
As can be seen in bug #4517, commit 6b79838 made Notebook more or less unusable for Safari users. This PR reverts the line that triggered the Safari bug.

Yes, it does seem to be a Safari bug. However, reverting (part of) the change does not seem to cause any other problems. I have tested the change in various Notebook version, and everything seems to be fine.

Thanks to @pcr910303 for finding out the cause of the problem.

### DIFF
--- a/notebook/templates/page.html
+++ b/notebook/templates/page.html
@@ -99,7 +99,6 @@
       })
 
     document.nbjs_translations = {{ nbjs_translations|safe }};
-    document.documentElement.lang = navigator.language.toLowerCase();
     </script>
 
     {% block meta %}


blink1073:
Thanks @adah1972!  I'm hesitant to remove this for all browsers, would you mind adding a check for whether the browser is Safari?  https://www.geeksforgeeks.org/how-to-detect-the-user-browser-safari-chrome-ie-firefox-and-opera-using-javascript/

adah1972:
OK. Done.

notebook/templates/page.html
```diff
@@ -98,8 +98,16 @@
           }
       })
 
+    // Detect Safari
+    let chromeAgent = navigator.userAgent.indexOf("Chrome") > -1;
+    let safariAgent = navigator.userAgent.indexOf("Safari") > -1;
+    if (chromeAgent && safariAgent) safariAgent = false;
```
blink1073, line 104: Ah, I think this needs to be put into a block with `{ }`

adah1972: Please check again whether the new change is what you want.

notebook/templates/page.html
```diff
@@ -98,8 +98,18 @@
           }
       })
 
-    document.nbjs_translations = {{ nbjs_translations|safe }};
-    document.documentElement.lang = navigator.language.toLowerCase();
+      {
+          // Detect Safari
+          let chromeAgent = navigator.userAgent.indexOf("Chrome") > -1;
+          let safariAgent = navigator.userAgent.indexOf("Safari") > -1;
+          if (chromeAgent && safariAgent) safariAgent = false;
```
blink1073, line 105: Sorry I meant that this line needs `{ }` around `safariAgent = false; `.  I believe that is what is causing the build failures.

adah1972: Oops. Sorry for the misunderstanding. Done.

blink1073: It looks like the build is still failing, do the tests run successfully on your mac?

adah1972: Sorry, I am not a JavaScript developer. I know just enough of the language to fix such small problems, but have no idea about the failures. I do not see why my change could possibly trigger such errors...

blink1073: Understood, I will revisit when I have some more bandwidth.

venkatasg:
Will this change be merged into the base branch soon? Safari is usable right now, but any custom fonts or styling that I may set for my Jupiter notebook using `custom.css` doesn't work.

goranmoomin:
This probably isn't the bug you're finding for. The existence of the `lang` attribute doesn't block your custom CSS, and this is more of a workaround for a very specific macOS Mojave bug. It's irrelevant to Catalina and higher.

(For context, I'm the original big reporter.)

venkatasg:
@goranmoomin Thanks for the update! Any idea what could be causing this issue specifically with Safari, even on Big Sur? Firefox and Chrome seem to have no issues loading the custom css styles like fonts.

JLTastet:
> this is more of a workaround for a very specific macOS Mojave bug. It's irrelevant to Catalina and higher.

Interestingly, I am hitting this very same bug on Monterey with Safari 15.1 (17612.2.9.1.2) and Jupyter **Lab** version 3.2.4 (installed from conda-forge, ARM version). Should I open a dedicated issue on the Jupyter Lab tracker?

PR #4991, opened 2019-10-20
remove py2 dependence

00Kai0:


00Kai0:
@lresende ,thank you for your approval.
I think I have removed most of py2 dependence.
And I find other problems that should be solved:
* Some code imported but unused.
* Replace `gen.coroutine` to `async/await`, because we just want to support `>=py3.5`.
* Some code don't follow `pep 8`code style (if we want to follow this style).

takluyver:
Feel free to open PRs to clean up unused imports, and we can probably convert coroutines to async/await syntax. I'd say try to work piecewise where practical - small to medium PRs are easier to review than large ones.

Follow PEP 8 where you're adding or changing code anyway, but please don't reformat huge amounts of code that you're not otherwise touching - that causes merge conflicts unnecessarily and makes git blame less useful.

takluyver:
PS: When I say "follow PEP 8", I mean roughly. As you can probably tell, we're not strict about this - if leaving a line a couple of characters over the limit is more readable, for example, that's OK.

00Kai0:
@takluyver I agree with you.

PR #4995, opened 2019-10-21
fix: respect --no-browser

maartenbreddels:
When I run:
```
$ jupyter notebook ~/src/QuantStack/voila/notebooks/bqplot.ipynb --no-browser
```
I still get a browser window opened, which I don't want. One might argue that "what is the point of giving the path on the command line?".

I often run the notebook of voila like this, and keep 1 browser tab open, kill the server, and start voila or notebook again (after a fix, or reinstallation of some other package). Having it open a browser each time is annoying and unexpected.

I wanted to fix this in jupyter server, but it makes more sense to have this consistent, and first discuss it here.

cc @minrk this is introduced in 26a1cc7ee21d8569e09349167f66ef9260b1429b

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1825,7 +1825,7 @@ class NotebookApp(JupyterApp):
         self.write_server_info_file()
         self.write_browser_open_file()
 
-        if self.open_browser or self.file_to_run:
+        if self.open_browser:
             self.launch_browser()
 
         if self.token and self._token_generated:


minrk:
There is inherent ambiguity in this command-line, since `--no-browser` means "don't launch the browser" but `jupyter notebook .../bqplot.ipynb` explicitly means "launch the browser with bqplot.ipynb open." The current behavior I think is based on the fact that `c.NotebookApp.open_browser = False` is more likely to be permanent configuration in a config file (fairly common configuration, I think), while `file_to_run` always comes from the command-line.

```
jupyter notebook ~/src/QuantStack/voila/notebooks --no-browser
```

has precisely the desired effect (set the notebook directory, but don't open a particular file), right? But the current behavior is harder to achieve with this change (don't open browser by default, but still easy to open a single file when desired). Is there something in voila that changes the interpretation of file_to_run to make these assumptions different? If so, it may make sense for it to be a deviation from the classic notebook server.

lresende:
Closing based on @minrk comment, please reopen if there are still issues.

PR #4999, opened 2019-10-23
Setting to disable redirect file browser launch

betteridiot:
When it was originally added, the PR for launching a browser via a redirect file (jupyter#4260) created some issues for some environments-namely WSL, Chromebook, and Android (jupyter#4346 (comment), jupyter/help#496). The reason for the break in these environments is due to the file structure/path differences between the runtime and the browser. 

This commit adds a setting to the `jupyter_notebook_config.py` that allows users to disable the use of redirect file for launching a browser in favor of the original, yet visible, URL + token approach. This setting: `c.NotebookApp.use_redirect_file` will be set to True by default.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -223,7 +223,6 @@ class NotebookWebApplication(web.Application):
             base_url=base_url,
             default_url=default_url,
             template_path=template_path,
-            use_redirect_file=jupyter_app.use_redirect_file,
             static_path=jupyter_app.static_file_path,
             static_custom_path=jupyter_app.static_custom_path,
             static_handler_class = FileFindHandler,
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -540,20 +540,6 @@ flags['allow-root']=(
     _("Allow the notebook to be run from root user.")
 )
 
-flags['use_redirect_file']=(
-    {'NotebookApp' : {'use_redirect_file' : True}},
-    """Disable launching browser by redirect file
-
-     For some environments (like Windows Subsystem for Linux (WSL) and Chromebooks),
-     launching a browser using a redirect file (introduced in notebook versions 
-     > 5.7.2) can lead the browser failing to load. This is because of the 
-     difference in file structures/paths between the runtime and the browser. 
-     
-     Disabling this setting will  by setting to False will disable this behavior, 
-     allowing the browser to launch by using a URL and visible token (as before).
-     """
-)
-
 # Add notebook manager flags
 flags.update(boolean_flag('script', 'FileContentsManager.save_script',
                'DEPRECATED, IGNORED',
@@ -663,13 +649,18 @@ class NotebookApp(JupyterApp):
     use_redirect_file = Bool(True, config=True,
         help="""Disable launching browser by redirect file
 
-     For some environments (like Windows Subsystem for Linux (WSL) and Chromebooks),
-     launching a browser using a redirect file (introduced in notebook versions 
-     > 5.7.2) can lead the browser failing to load. This is because of the 
-     difference in file structures/paths between the runtime and the browser. 
+     For versions of notebook > 5.7.2, a security feature measure was added that 
+     prevented the authentication token used to launch the browser from being visible.
+     This feature makes it difficult for other users on a multi-user system from
+     running code in your Jupyter session as you.
+
+     However, some environments (like Windows Subsystem for Linux (WSL) and Chromebooks),
+     launching a browser using a redirect file can lead the browser failing to load. 
+     This is because of the difference in file structures/paths between the runtime and 
+     the browser. 
      
-     Disabling this setting will  by setting to False will disable this behavior, 
-     allowing the browser to launch by using a URL and visible token (as before).
+     Disabling this setting to False will disable this behavior, allowing the browser 
+     to launch by using a URL and visible token (as before).
      """
     )
 
@@ -1839,7 +1830,7 @@ class NotebookApp(JupyterApp):
             if self.token:
                 uri = url_concat(uri, {'token': self.token})
 
-        elif self.file_to_run:
+        if self.file_to_run:
             if not os.path.exists(self.file_to_run):
                 self.log.critical(_("%s does not exist") % self.file_to_run)
                 self.exit(1)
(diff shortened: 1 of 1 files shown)


betteridiot:
@takluyver, this PR solves https://github.com/jupyter/notebook/pull/4346 as well

notebook/notebookapp.py
```diff
@@ -1820,10 +1854,17 @@ def launch_browser(self):
         else:
             open_file = self.browser_open_file
 
-        b = lambda: browser.open(
-            urljoin('file:', pathname2url(open_file)),
-            new=self.webbrowser_open_new)
-        threading.Thread(target=b).start()
+        if not self.use_redirect_file:
+            if browser:
```
takluyver: You don't need to check `if browser:` here - it's already checked for an early return at the start of the function.

notebook/notebookapp.py
```diff
@@ -539,6 +540,20 @@ def start(self):
...
+    """Disable launching browser by redirect file
+
+     For some environments (like Windows Subsystem for Linux (WSL) and Chromebooks),
+     launching a browser using a redirect file (introduced in notebook versions 
+     > 5.7.2) can lead the browser failing to load. This is because of the 
+     difference in file structures/paths between the runtime and the browser. 
+     
+     Disabling this setting will  by setting to False will disable this behavior, 
+     allowing the browser to launch by using a URL and visible token (as before).
+     """
```
takluyver: The redirect file was adopted as a security measure: without it, it's possible for another user on the same computer to steal your Jupyter session and run code as you. So the help about this option should be clear about that risk. Undoubtedly some people will turn it off without reading the help, but we can't do anything about that.

notebook/notebookapp.py
```diff
@@ -1805,7 +1833,13 @@ def launch_browser(self):
...
             return
 
-        if self.file_to_run:
+        if not self.use_redirect_file:
+            uri = self.default_url[len(self.base_url):]
+
+            if self.token:
+                uri = url_concat(uri, {'token': self.token})
+
+        elif self.file_to_run:
```
takluyver: `use_redirect_file` and `file_to_run` are orthogonal, so they can't be in an if/else. This would mean that `jupyter notebook foo.ipynb` wouldn't open to `foo.ipynb` if you've set `use_redirect_file = False`.

notebook/notebookapp.py
```diff
@@ -1820,10 +1854,17 @@ def launch_browser(self):
         else:
             open_file = self.browser_open_file
 
-        b = lambda: browser.open(
-            urljoin('file:', pathname2url(open_file)),
-            new=self.webbrowser_open_new)
-        threading.Thread(target=b).start()
+        if not self.use_redirect_file:
```
takluyver: If practical, can we have the if/else the other way round, so it doesn't need `if not`?

notebook/notebookapp.py
```diff
@@ -1820,10 +1854,17 @@ def launch_browser(self):
...
+            if browser:
+                b = lambda: browser.open(url_path_join(self.connection_url, uri),
+                                        new=self.webbrowser_open_new)
+                threading.Thread(target=b).start()
+
+        else:
+            b = lambda: browser.open(
+                urljoin('file:', pathname2url(open_file)),
+                new=self.webbrowser_open_new)
+            threading.Thread(target=b).start()
```
takluyver: The `browser.open()` and the thread call can be shared between the code paths - only assembling the URL needs to be different.

notebook/notebookapp.py
```diff
@@ -539,6 +540,20 @@ def start(self):
     _("Allow the notebook to be run from root user.")
 )
 
+flags['use_redirect_file']=(
+    {'NotebookApp' : {'use_redirect_file' : True}},
```
takluyver: Having a flag for this doesn't really make sense, because use redirect file is already the default, so `--use-redirect-file` wouldn't do anything.

You could make a flag to *disable* it, like `--no-use-redirect-file`, but I suspect it's the sort of thing you'd usually set in a config file anyway, so maybe it doesn't need a flag.

notebook/notebookapp.py
betteridiot: This makes sense now. I was still pretty unfamiliar with the code base when I created this and was making some assumptions.

notebook/notebookapp.py
betteridiot: Okay. I was thinking this was the redirect `file_to_run`, not an argument file.

notebook/notebookapp.py
betteridiot: Acknowledged and removed.

notebook/notebookapp.py
```diff
@@ -645,6 +646,24 @@ def _default_log_format(self):
...
+     However, some environments (like Windows Subsystem for Linux (WSL) and Chromebooks),
+     launching a browser using a redirect file can lead the browser failing to load. 
+     This is because of the difference in file structures/paths between the runtime and 
+     the browser. 
+     
+     Disabling this setting to False will disable this behavior, allowing the browser 
+     to launch by using a URL and visible token (as before).
+     """
+    )
+
```
betteridiot: Updates to the documentation regarding the potential security risk introduced if the setting is disabled

notebook/notebookapp.py
```diff
@@ -663,13 +649,18 @@ def _default_log_format(self):
...
+     However, some environments (like Windows Subsystem for Linux (WSL) and Chromebooks),
+     launching a browser using a redirect file can lead the browser failing to load. 
+     This is because of the difference in file structures/paths between the runtime and 
+     the browser. 
      
-     Disabling this setting will  by setting to False will disable this behavior, 
-     allowing the browser to launch by using a URL and visible token (as before).
+     Disabling this setting to False will disable this behavior, allowing the browser 
+     to launch by using a URL and visible token (as before).
```
betteridiot: Changed language regarding the security risk if the setting is disabled

notebook/notebookapp.py
```diff
@@ -1854,17 +1845,13 @@ def launch_browser(self):
...
+            assembled_url = urljoin('file:', pathname2url(open_file))
         else:
-            b = lambda: browser.open(
-                urljoin('file:', pathname2url(open_file)),
-                new=self.webbrowser_open_new)
-            threading.Thread(target=b).start()
+            assembled_url = url_path_join(self.connection_url, uri)
+        
+        b = lambda: browser.open(assembled_url, new=self.webbrowser_open_new)                            
+        threading.Thread(target=b).start()
```
betteridiot: Created an intermediate variable (`assembled_url`) that is handled depending on `use_redirect_file`. This is now sent to the same `browser.open` call, removing the boilerplate previously written.

betteridiot: Also fixed the `if not/else` to be `if/else`

betteridiot:
I believe these changes address all of your review notes @takluyver

notebook/notebookapp.py
```diff
@@ -223,6 +223,7 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
             base_url=base_url,
             default_url=default_url,
             template_path=template_path,
+            use_redirect_file=jupyter_app.use_redirect_file,
```
takluyver: I don't think the tornado web application object needs this - it's only used in the NotebookApp class. So this line can go.

betteridiot: Thanks for the quick review. The change should be reflected in the newest file version now.

takluyver:
Thanks :-)

betteridiot:
No. Thank you!

Any input as to the release schedule to conda-forge on this?

takluyver:
@lresende has done the last couple of releases - I'll let him answer about if/when he's planning another one.

PR #5021, opened 2019-10-31
Fixes #5002

### LINKED ISSUE #5002: `root_dir` for Contents Manager not loaded from config file
I added the lines
```
c.ContentsManager.root_dir = '~'
c.FileContentsManager.root_dir = '~'
```
to `~/.jupyter/jupyter_notebook_config.py`. 

I then ran the Jupyter notebook server in debug mode, as follows (the setting of `notebook_dir` is deliberate):
```
jupyter lab --notebook-dir='/' --debug
```
**EDIT:** the exact same things happen when using the command
```
jupyter notebook --notebook-dir='/' --debug
```
and as far as I can tell, JupyterLab uses Notebook's contents manager anyway **/EDIT**
and it confirmed that the config file was loaded:
```
[D 16:50:53.349 LabApp] Loaded config file: ~/.jupyter/jupyter_notebook_config.py
```
My notebook server extension, which uses [the `.save` method of the contents manager](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L453) to save a notebook to the user's home directory (well in practice, to the [`root_dir` of the contents manager](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L72)), reports a 403 permissions error when trying to run, and in its log output ([see code on GitHub](https://github.com/krinsman/clonenotebooks/blob/master/clonenotebooks/cloners/cloners.py#L47)) shows that the value of the contents manager's root directory is the value of `notebook_dir`, instead of the value specified in the config file.
```
[D 16:51:48.310 LabApp] Contents manager root_dir: /
```
This _would_ be expected if no value of `root_dir` had been specified in the config file, since according to the code, [the value of `notebook_dir` is to be loaded as the default](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L74) if no value is specified in the configuration. However, a value _was_ specified, and [the value of `root_dir` _is_ supposed to be configurable](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L72), so this seems to be a bug.

For further evidence that the value of `root_dir` specified in the config file is being ignored, I changed the config file to say
```
c.FileContentsManager.root_dir = 42
```
[Because the value of `root_dir` is validated](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L140), if this configuration is being loaded, [then a `TraitError` should be thrown](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L148) as a result of this configuration setting. However, no errors are raised when starting the notebook server.

**Random Thoughts:** 
All I can think of at the moment is that perhaps JupyterLab uses a different contents manager than the Jupyter notebook, maybe a subclass, and that for this reason the configuration is being effectively ignored? But I also changed the value of `c.ContentsManager.root_dir` in the config file, as well as `c.LargeFileContentsManager.root_dir` (since as far as I can tell the `LargeFileContentsManager` subclass of `FileContentsManager` is actually what's loaded/used), but it still does nothing. [I initially thought that this was an issue with JupyterHub](https://github.com/jupyterhub/jupyterhub/issues/2793), but am now fairly convinced that this is not the case, especially since it's possible to reproduce the problem without running JupyterHub.

krinsman:
[Deleting line 1278](https://github.com/jupyter/notebook/compare/master...krinsman:master#diff-bbd307804bb6b8836da9f06b8d8a06e9L1278) fixes https://github.com/jupyter/notebook/issues/5002

The rest of the deletions are code which was supposed to be removed with notebook 5.0, and it seems like notebook has advanced beyond that version.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1270,14 +1270,6 @@ class NotebookApp(JupyterApp):
             raise TraitError(trans.gettext("No such notebook dir: '%r'") % value)
         return value
 
-    @observe('notebook_dir')
-    def _update_notebook_dir(self, change):
-        """Do a bit of validation of the notebook dir."""
-        # setting App.notebook_dir implies setting notebook and kernel dirs as well
-        new = change['new']
-        self.config.FileContentsManager.root_dir = new
-        self.config.MappingKernelManager.root_dir = new
-
     # TODO: Remove me in notebook 5.0
     server_extensions = List(Unicode(), config=True,
         help=(_("DEPRECATED use the nbserver_extensions dict instead"))


takluyver:
I suspect removing that line also breaks using `--notebook-dir` without setting the contents manager root dir in config. Interacting config options are hard.

takluyver:
I suspect removing that line also breaks using `--notebook-dir` without setting the contents manager root dir in config. Interacting config options are hard.

krinsman:
@takluyver To clarify, what is the intended behavior? And how do these changes break that behavior? 

To me everything seems to work with the changes, which suggests I don't understand what the intended behavior is, and thus what is broken.

By default `FileContents.root_dir` takes its value from `notebook_dir`: https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L70

Right now, if both are specified by the user, only one of those specifications is listened to, which is rather confusing and frustrating (it took me 1-2 weeks of digging through the source code to figure out what was going on).

Being able to specify both independently is important e.g. when running `jupyterhub-singleuser` with a value of `c.Spawner.notebook_dir` different from what the user's content manager root directory is. Ideally what we want is for users to be able to _view_ inside of notebook all files on the system, thus setting `notebook_dir` equal to root, but at the same time have notebook only attempt to _write_ files inside of their home directory, thus the setting for `ContentsManager.root_dir`. Right now however the latter setting is ignored.

If it is actually intended that `NotebookApp.notebook_dir == ContentsManager.root_dir` _always_, no ifs, buts, etc., then it seems like `ContentsManager.root_dir` shouldn't be a configurable.

takluyver:
Yuck, this is the kind of thing that makes me hate traitlets: the relationship between `NotebookApp.notebook_dir` and `FileContents.root_dir` is written in two completely separate bits of code, one of them without even mentioning NotebookApp directly. :disappointed: 

I suspect that the fact that setting them to different values gives you read-only access to files in one but not the other is a coincidence of how the implementation works, not something that's designed or guaranteed for the future.

krinsman:
@takluyver So would it make sense to change this PR to instead make `FileContents.root_dir` non-configurable, since the intended behavior seems to be to always have `NotebookApp.notebook_dir == ContentsManager.root_dir`?

krinsman:
@takluyver I updated the PR to just make the contents manager and kernel manager root directories non-configurable, since those settings are already silently overwritten by `NotebookApp.notebook_dir` anyway.

That being said it would probably be better if a deprecation warning for their non-configurability were inserted somehow -- I don't know what the best way would be to do that.

`NotebookApp.notebook_dir`, under the new name `ServerApp.root_dir`, as the "one root directory to rule them all" seems to be the direction the Jupyter Server project is already going anyway, at least to the best of my crude outsider understanding: https://github.com/jupyter/jupyter_server/pull/36

takluyver:
That would be my preferred direction - fewer config options that can interfere with one another. But as it's possible to set them separately at the moment, there's a good chance that someone has a use case for that, so it would be tricky to switch config off for those attributes now. :-(

krinsman:
OK, so what if this PR was changed to create deprecation warnings saying that configuration for these traits is deprecated and will be removed in Notebook 7 or Notebook 8? As well as changes/additions to the Docs and docstrings saying that these will be deprecated? Would that be better?

Again, these traits are already de facto un-configurable, this is the gist of #5002, so it seems like it would be impossible for anyone to have a use case that depends on configuring them.

My main interest is in fixing #5002, because I lost a lot of time puzzling over that one before figuring out what the issue was, and it was an incredibly frustrating experience which I would prefer to ensure that future users can avoid. I really like Jupyter a lot and prefer that other people avoid frustrating experiences with it so that they can like Jupyter a lot too.

takluyver:
Well, they are configurable when you don't use the `--notebook-dir` option.

I suspect that your original fix was probably the best idea - let the contents manager get the application option as part of calculating the default. It's just annoyingly hard to reason about these kind of interactions to verify that that will do the right thing in all situations - e.g. that it couldn't initialise the default before the config option is set on the application.

krinsman:
OK @takluyver that makes sense to me too and I updated the PR with that strategy in mind. 

The kernel manager and contents manager already automatically check the notebook app's `notebook_dir` for their default values of `root_dir`, so it seemed to me that the only code that needed to be changed to implement this was the code for the notebook app.

takluyver:
Thanks for your patience; let's try it like that and see if anyone complains.

PR #5022, opened 2019-10-31
Modified python3 to python

schadalapaka:
python2 is going to be deprecated in Jan 2020 and so changed it to python (and also so it is consistent with the other documentation)


schadalapaka:
Updated README file from python3 to python (since python2 is about to be deprecated soon).

PR #5023, opened 2019-10-31
Updating python3 to python

schadalapaka:
changed python3 to python because python2 is deprecating soon (and also so the documentation is consistent)


takluyver:
Thanks, but I think it's better to leave this as `python3` for quite a while to make it clear that Python 3 is needed. This shouldn't stop working as distributors point `python` to Python 3 as well.

lresende:
Closing per @takluyver comments.

PR #5043, opened 2019-11-11
Selenium test fixture to open a notebook with prefilled content

takluyver:
Many of the Selenium tests start by opening a new notebook and writing text in some cells through the interface, just to set up the conditions for what they're really testing. This is slightly awkward, because you need a different method to change the initial empty cell versus adding new cells. It also takes a little bit of time, as it involves a number of Selenium calls.

This adds a `prefill_notebook` fixture which can create a notebook file with some content, and then opens that in the browser. The content can be a list of nbformat cell objects or strings (which are treated as code cells). I believe this makes the test code more readable by condensing boilerplate to set up notebook state.

### DIFF
--- a/notebook/tests/selenium/conftest.py
+++ b/notebook/tests/selenium/conftest.py
@@ -1,9 +1,12 @@
 import json
+import nbformat
+from nbformat.v4 import new_notebook, new_code_cell
 import os
 import pytest
 import requests
 from subprocess import Popen
 import sys
+from tempfile import mkstemp
 from testpath.tempdir import TemporaryDirectory
 import time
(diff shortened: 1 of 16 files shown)


takluyver:
Speeding up the tests is not a primary goal here, but it is a nice side effect.

- Travis CI job: [7 min 17 s](https://travis-ci.org/jupyter/notebook/jobs/609648192) -> [5 min 43 s](https://travis-ci.org/jupyter/notebook/jobs/610317145) 
- pytest run: 4 min 16 s -> 2 min 53 s

takluyver:
I'll merge this in a day or two unless anyone objects.

PR #5052, opened 2019-11-12
Add a metadata tag to override notebook direction (ltr/rtl)

m2-farzan:
Currently, the direction of text flow in a notebook (left-to-right / right-to-left) is determined by browser language settings. If the browser is set to use a language with right-to-left script, all notebooks are displayed in RTL direction and vice versa.

This behavior could be improved, IMHO (I'm a student from the middle east). I write my homework in my mother language, but I also download stuff from the internet which are mostly in English. This means that I have to change my browser language settings frequently, which is not user friendly.

This PR makes it possible to override this default behavior *per notebook* using a metadata tag. Here's how to use this tag in order to make a notebook RTL:

`"direction": "rtl"`

or to force notebook to use LTR:

`"direction": "ltr"`

The end user could either manually set this tag or use the `toggle rtl layout` command from the command palette to change the direction of a notebook permanently, without needing to change the browser settings.

That's it from the end user point of view.

### DIFF
--- a/notebook/static/base/js/dialog.js
+++ b/notebook/static/base/js/dialog.js
@@ -230,6 +230,7 @@ define(['jquery',
                             return false;
                         }
                         options.callback(new_md);
+                        options.notebook.apply_directionality();
                     }
                 }
             },
--- a/notebook/static/notebook/js/actions.js
+++ b/notebook/static/notebook/js/actions.js
@@ -65,11 +65,27 @@ define([
      *
      **/
     var _actions = {
+        'toggle-cell-rtl-layout': {
+            cmd: i18n.msg._('toggle current cell ltr/rtl direction'),
+            help: i18n.msg._('Toggle current cell directionality between left-to-right and right-to-left'),
+            handler: function (env) {
+              var notebook_direction = document.body.getAttribute('dir') == 'rtl' ? 'rtl' : 'ltr';
+              var current_cell_default_direction = env.notebook.get_selected_cell().cell_type == 'code' ? 'ltr' : notebook_direction;
+              var current_cell_direction = env.notebook.get_selected_cell().metadata.direction || current_cell_default_direction;
+              var new_direction = current_cell_direction == 'rtl' ? 'ltr' : 'rtl';
+              env.notebook.get_selected_cells().forEach(cell => cell.metadata.direction = new_direction);
+              env.notebook.set_dirty(true);
+              env.notebook.apply_directionality();
+            }
+        },
         'toggle-rtl-layout': {
-            cmd: i18n.msg._('toggle rtl layout'),
-            help: i18n.msg._('Toggle the screen directionality between left-to-right and right-to-left'),
-            handler: function () {
-              (document.body.getAttribute('dir')=='rtl') ? document.body.setAttribute('dir','ltr') : document.body.setAttribute('dir','rtl');
+            cmd: i18n.msg._('toggle notebook ltr/rtl direction'),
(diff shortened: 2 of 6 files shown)


notebook/static/notebook/js/main.js
```diff
@@ -77,14 +77,6 @@ requirejs([
...
     // Pull typeahead from the global jquery object
     var typeahead = $.typeahead;
-    
-    try{
-        requirejs(['custom/custom'], function() {});
-        bidi.loadLocale();
-    } catch(err) {
-        console.log("Error processing custom.js. Logging and continuing");
-        console.warn(err);
-    }
```
m2-farzan: Important note:

I've moved these lines to the very end of the file, because the function `bidi.loadLocale` now reads metadata from the notebook and must be called after loading it. The problem is that I don't know the side effects of this change, especially because there's a line that loads a custom script and I don't know what it is at all.

Please just make sure I'm not breaking things here.

takluyver:
Thanks! There's some relevant discussion at https://github.com/ipython/ipython/issues/10980 . Someone has suggested ignoring the browser locale settings and always using LTR layout by default. That would still be compatible with having a direction stored in the notebook metadata, I think.

(I can't read any RTL languages, so I'm hoping that a consensus will appear among people who can about the best ways to handle them)

notebook/static/notebook/js/main.js
takluyver: `custom.js` is meant as a place for users to put arbitrary javascript to customise the notebook interface. Which probably means any change to how it's loaded is going to break something for someone, somewhere, but that can't really be helped.

I'm not sure why loading it is tied to bidi support, though. :-/

blink1073: Hi @m2-farzan, I suggest we break this up: leave the `custom.js` loading where it is and move the `bidi.locale()` handling to the end of the file.  This PR also needs a rebase.

m2-farzan:
So I've been working on this recently, and I hope you like the end result.

30ed681a8f9b75d0a1d9334bcd550904c0b9380f will allow the users to control the direction of the notebook by the following means:

* a notebook metadata tag, `direction`, which could be either `ltr`or `rtl`

* a cell metadata tag, again `direction`, that overrides notebook settings. Either `rtl` or `ltr`.
  (Why this one? A common example is that when writing in RTL languages, users often need to write latex equation blocks for which you need an LTR cell. So cell-level control is essential.)

The behavior is described below:

* A notebook without a `direction` tag will be LTR. That is, notebook will ignore browser locale settings from now on. This behavior was suggested by two users in ipython/ipython#10980. It also keeps a notebook from looking different on different computers. However, please note that we're breaking the previous behavior here.

* A markdown cell without a `direction` tag will inherit direction from notebook. A `direction` tag overrides it. I've also added an action in the command palette "toggle current cell ltr/rtl layout direction" which sets the tag for the user. (How about a shortcut for it?)

* In a code cell, the code itself (code mirror) will never become RTL—no one ever writes code in RTL mode. However, a `direction` tag with value `rtl` will cause the output to become RTL. The code cell always defaults in LTR (that is, it doesn't inherit from notebook settings), because IMHO RTL output from code is rarely required and it would be a pain change it back every time. Most people are just dealing with math or getting simple English logs/errors.

blink1073:
I kicked the CI jobs

blink1073:
Thanks for fixing the CI!

zareami10:
It would be great if we could also have this in Lab.

m2-farzan:
@zareami10 I've been working on it recently. It will be ready soon.

PR #5055, opened 2019-11-15
Fix packages of exceptions catched by gateway manager

shuichiro-makigaki:
`gateway_request` in `gateway.manager` raises `tornado.web.HTTPError` exceptions, but the callers, such as `GatewayKernelManager.get_kernel`, catch `tornado.httpclient.HTTPError`, instead of `tornado.web.HTTPError`. Therefore, the callers can not handle exceptions during gateway interactions.

This causes that, for example, when Jupyter Enterprise Gateway culled a kernel by idle timeout, the gateway manager can not handle the kernel's absent appropriately. As a result, notebook users see ambiguous "Kernel Error" and can not restart the kernel or start a new kernel.

### DIFF
--- a/notebook/gateway/managers.py
+++ b/notebook/gateway/managers.py
@@ -283,12 +283,12 @@ def gateway_request(endpoint, **kwargs):
     except ConnectionRefusedError:
         raise web.HTTPError(503, "Connection refused from Gateway server url '{}'.  " 
               "Check to be sure the Gateway instance is running.".format(GatewayClient.instance().url))
-    except HTTPError:
+    except HTTPError as e:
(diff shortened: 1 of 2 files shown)


kevin-bates:
@shuichiro-makigaki - would you mind submitting essentially this same PR against [jupyter_server](https://github.com/jupyter/jupyter_server)?

shuichiro-makigaki:
I added some fixes to tests also.

> would you mind submitting essentially this same PR against [jupyter_server](https://github.com/jupyter/jupyter_server)?

Sure! #jupyter/jupyter_server#139

Issue #5070, opened 2019-11-20
init_resources fails for non-root user


shevelevs:
When running as non-privileged user, we get an error

...
  File "/.../lib/python3.5/site-packages/notebook/notebookapp.py", line 1420, in init_resources
    resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
ValueError: not allowed to raise maximum limit

I think this is how setrlimit behaves according to the doc (https://docs.python.org/3/library/resource.html) when process is trying to change hard limit as a non-root user.

I think the reasonable behavior would be not to fail unless the config was explicitly specified and is above the default limits for the effective user.

shevelevs:
Related PR: https://github.com/jupyter/notebook/pull/4893

takluyver:
Yes, that sounds reasonable. Do you want to make a PR?

It should probably still warn in that case (as the configured minimum must be higher than the hard limit), and maybe still raise the soft limit to the hard limit if they differ.

takluyver:
Closed by #5075.

PR #5075, opened 2019-11-22
Making default value for min_open_files_limit dynamic

shevelevs:
Making default value for min_open_files_limit dependent on user's
ulimit values so trying to increase them above hard limit while
running as non-privileged user is not causing error on start-up.

See https://github.com/jupyter/notebook/issues/5070

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -817,13 +817,29 @@ class NotebookApp(JupyterApp):
         """
     )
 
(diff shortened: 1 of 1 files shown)


jasongrout:
Thanks! This logic looks good at first glance on my phone. Can you change the name of the function to be aligned with the other default functions? `_default_parameter_name_`, I think.

PR #5077, opened 2019-11-24
Allow contents manager get to return future

michalc:
This is a follow up to https://github.com/jupyter/notebook/pull/4099, wrapping other uses of contents manager get with `maybe_future` to allow it to yield the event loop.


takluyver:
Seems reasonable, thanks.

PR #5078, opened 2019-11-24
Ensure jupyter config dir exist

haobibo:
The bug: when
 - the user login for the first file with a token and want to change his/her password
 - and if the folder `~/.jupyter` (`jupyter_config_dir()`) does not exist

then the jupyter server will raise a `FileNotFoundError` exception and return a 500 error.

This change tries to fix this bug by ensuring the folder exists so that the `config_file` can be opened.

### DIFF
--- a/notebook/auth/security.py
+++ b/notebook/auth/security.py
@@ -118,6 +118,7 @@ def persist_config(config_file=None, mode=0o600):
     """
 
     if config_file is None:
+        os.makedirs(jupyter_config_dir(), exist_ok=True)
         config_file = os.path.join(jupyter_config_dir(), 'jupyter_notebook_config.json')
 
     loader = JSONFileConfigLoader(os.path.basename(config_file), os.path.dirname(config_file))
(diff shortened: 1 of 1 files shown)


notebook/auth/security.py
```diff
@@ -118,6 +118,7 @@ def persist_config(config_file=None, mode=0o600):
     """
 
     if config_file is None:
+        os.makedirs(jupyter_config_dir(), exist_ok=True)
```
takluyver: I think this should be outside the `if` block, and use `os.path.dirname()`, so it works the same way if a path is passed in explicitly.

haobibo: fixed

takluyver:
An alpha release of nbconvert has broken notebook's tests, so we're stalled until someone gets to figuring that out and fixing it (I'm not sure yet if it needs changes in the tests or in nbconvert). Other than that, I think this looks good.

haobibo:
sure, let's wait for that.

takluyver:
Close/reopen to rerun tests with master.

takluyver:
Thanks!

PR #5099, opened 2019-12-04
preserve term "Markdown" untranslated

forest0:
Hi guys,

I am Chinese user and have used Markdown for several years,
I've never heard any Chinese say "Markdown" as "标记"(as did by the current translation),
we just keep the term "Markdown" even in our daily communication in Chinese,
so I believe just keep "Markdown" untranslated is more intuitive.


minrk:
Thanks!

PR #5107, opened 2019-12-07
Toolbar buttons tooltip: show help instead of label

manics:
Currently the tooltip for a toolbar button with just an icon is the help text for the action/command.
If a toolbar button has a label the label text is shown alongside the button icon as expected, but it also replaces the tooltip help.

With this PR the help text is always shown as the tooltip.

### DIFF
--- a/notebook/static/notebook/js/toolbar.js
+++ b/notebook/static/notebook/js/toolbar.js
@@ -94,7 +94,8 @@ define(['jquery','base/js/i18n'], function($, i18n) {
                 }
                 var button  = $('<button/>')
                     .addClass('btn btn-default')
-                    .attr("title", i18n.msg._(action.help))
+                    .attr("aria-label", el.label)
+                    .attr("title", i18n.msg._(action.help)||el.label)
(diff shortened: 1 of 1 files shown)


lresende:
What is the accessibility side effect of this change? @jasongrout @takluyver any thoughts here?

kevin-bates:
@jasongrout - can you please comment on Luciano's accessibility question?  Otherwise, I think this looks good.

kevin-bates:
@tgeorgeux - I believe you deal in the realm of UX.  Do these changes introduce any accessibility issues and, if so, might there be a way they can be addressed while still providing this improved behavior?

tgeorgeux:
@kevin-bates Sorry I missed this ping. I don't really know enough about the mechanics of what's going on to know if this change will negatively impact accessibility. It looks like a net improvement to me, with @afshin's changes mentioned above.

manics:
Thanks for looking, I've made the requested change.

afshin:
Thank you!

PR #5120, opened 2019-12-17
use cmdtrl-enter to run a cell

madhums:
Instead of ctrl-enter to run a cell, use cmdtrl-enter - this will provide better ux for mac users.


rkohli3:
Why would this be changed after several years of using notebook. I was used to control+enter. I've been using notebook for several years and didn't have issues using control+enter. I don't understand how you could say it for everyone? I would have recommended this as a command mode entry, but since edit mode shortcuts can't be changed, it's an arbitrary decision to change this.

kevin-bates:
I suspect this side-effect wasn't intended at the time this PR was merged.  As others are also having some issues with this change, I suggest we move the discussion to #5794 for how this can be best addressed.  Thanks.

rkohli3:
Thanks!! I hope it's addressed!

Issue #1914, opened 2016-11-21
Tries random ports, even if --port is given


ProgVal:
Using `--port` makes the notebook try other ports of the provided one is not available.

I understand this behavior if `--port` is provided, but is it on purpose to make it try other ports in that case?

takluyver:
I think you're probably right that if `--port` is specified but unavailable, the server should fail rather than picking another port. The way it currently works is probably just because the code that discovers that a port is in use doesn't know whether that port was specified at the command line or not.

pallavibharadwaj:
Hello. Can I try fixing this? I understand what the issue is and It would be great if you could throw some light on where I should start looking to get this fixed. Thank you!

kevin-bates:
@pallavibharadwaj - thank you for your interest in contributing.

Rather than trying to figure out if the port was set on the command line, it seems like leveraging `--port-retries=0` is cleaner - since that's precisely what that option is for.  If this route is taken I would suggest the info and error messages in [this block](https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L1487-L1507) be made sensitive to port-retries == 0 (which should probably be the case anyway).

Making `--port n` imply `--port-retries=0` could also introduce a compatibility issue for some admins that, say, are required to "segment" their Notebook servers via port ranges.

pallavibharadwaj:
@kevin-bates - thank you for your response.
If I got you right, it is rather better to specify --port-retries=0 along with --port n, if random ports are not to be picked when the specified port fails? The info and error messages are already sensitive to port-retries==0. Please do let me know if you could think of an alternate fix for this, thank you.

kevin-bates:
> If I got you right, it is rather better to specify --port-retries=0 along with --port n, if random ports are not to be picked when the specified port fails?

Correct, I believe so. The `port-retries` configurable option _is_ what instructs the server to not attempt to locate a different port so we might as well use it. 

> The info and error messages are already sensitive to port-retries==0.

By this I meant that the _messaging_ would be different when port-retries==0.  For example, we probably shouldn't indicate that we're trying another port when retries are disabled and we should probably produce a different error message that say's we failed to start because "port X was not available" rather than implying we couldn't find any port.

pallavibharadwaj:
@kevin-bates - thank you for clearing it out. I will work on it!

pallavibharadwaj:
[...]

PR #5136, opened 2020-01-04
descriptive log for port unavailable and port-retries=0

pallavibharadwaj:
when `--port-retries=0` and the launch fails on the port, the error log`ERROR: the notebook server could not be started because no available port could be found.` is not very appropriate. Fixes: #1914

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1490,10 +1490,13 @@ class NotebookApp(JupyterApp):
                 self.http_server.listen(port, self.ip)
             except socket.error as e:
                 if e.errno == errno.EADDRINUSE:
-                    self.log.info(_('The port %i is already in use, trying another port.') % port)
+                    if self.port_retries:
+                        self.log.info(_('The port %i is already in use, trying another port.') % port)
+                    else:
+                        self.log.info(_('The port %i is already in use.') % port)
                     continue
                 elif e.errno in (errno.EACCES, getattr(errno, 'WSAEACCES', errno.EACCES)):
-                    self.log.warning(_("Permission to listen on port %i denied") % port)
+                    self.log.warning(_("Permission to listen on port %i denied.") % port)
                     continue
                 else:
                     raise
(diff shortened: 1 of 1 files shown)


kevin-bates:
Also please edit your initial comment above to include `Fixes: #1914` as the last line in the comment.  This instructs `git` to close that Issue upon the merge of this PR. Thanks.

pallavibharadwaj:
Edited the comment, thank you.

Is removing the `trying another port` part of the log a good idea, since that would be consistent with `Permission to listen on port 403 denied`? Or would you suggest having this clause in both the cases?

kevin-bates:
I'm only suggesting the clause's removal when `self.port_retries == 0`.  When non-zero we should indicate other ports will be attempted, but not when retries are disabled.  I believe the permission message is fine as-is - although I suppose one could argue that it too should include the 'trying another port` clause (only when retries are enabled).   That said, you typically want to provide only the necessary information in security-related messaging, so perhaps leaving that one alone is probably best given it didn't include the clause in the first place.

pallavibharadwaj:
@kevin-bates - I have added the changes you suggested, in the PR. I made a new commit instead of rebasing, in order to highlight my new changes. I will rebase the commits if needed after you approve it. Thank you.

notebook/notebookapp.py
```diff
@@ -1490,7 +1490,10 @@ def init_webapp(self):
                 self.http_server.listen(port, self.ip)
             except socket.error as e:
                 if e.errno == errno.EADDRINUSE:
-                    self.log.info(_('The port %i is already in use, trying another port.') % port)
+                    if self.port_retries:
+                        self.log.info(_('The port %i is already in use, trying another port.') % port)
+                    else:
+                        self.log.info(_('The port %i is already in use') % port)
```
kevin-bates: nit: Let's add the period.  Since the permission warning doesn't have one, yet the other messages do, please add a period to that warning message as well.  Thanks.

pallavibharadwaj: Sure. I've updated the commit, thank you.

PR #5193, opened 2020-01-28
Check for invalid characters prior to rename

### LINKED ISSUE #5190: All notebook data is lost after renaming
So after trying to rename notebook like some franchise movie "Notebook: Return of the Jedi.ipynb" jupyter rename it just "Notebook" and made it empty file, basically deleting it. Same happened in .ipynb_checkpoints. Apparently it happened because name contained ":" symbol, that is prohibited in file names in windows. But still, I lost several hours of work and i'm sad.

kevin-bates:
On Windows platforms, rename operations containing ':' are problematic
since the operation succeeds, but creates an unintended "alternate data
stream" that gives the user the impression their file contents have been
deleted.  This change performs cursory validation of the destination
name on platforms in which issues can occur.  Currently, non-Windows
systems don't require this level of validation.

When an invalid character is encountered during the rename operation, a message similar to the following will occur:
Screen Shot 2020-01-28 at 1 55 50 PM

Fixes #5190

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -600,12 +600,13 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
             # be created and confusion will reign! (See https://github.com/jupyter/notebook/issues/5190)
             # Go ahead and add other invalid (and non-path-separator) characters here as well so there's
             # consistent behavior - although all others will result in '[Errno 22]Invalid Argument' errors.
-            invalid_chars = (':', '>', '<', '*', '?', '|', '"')
+            invalid_chars = '?:><*"|'
         else:
             # On non-windows systems, allow the underlying file creation to perform enforcement when appropriate
-            invalid_chars = ()
+            invalid_chars = ''
 
         for char in invalid_chars:
             if char in path:
-                raise web.HTTPError(400, u"Path '{}' contains invalid characters {} relative to its platform.  "
-                                         u"Please edit the path and re-submit the request.".format(path, invalid_chars))
+                raise web.HTTPError(400, "Path '{}' contains characters that are invalid for the filesystem. "
+                                         "Path names on this filesystem cannot contain any of the following "
+                                         "characters: {}".format(path, invalid_chars))
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -558,6 +558,10 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         if new_path == old_path:
             return
 
+        # Perform path validation prior to converting to os-specific value since this
+        # is still relative to root_dir.
+        self._validate_path(new_path)
+
         new_os_path = self._get_os_path(new_path)
         old_os_path = self._get_os_path(old_path)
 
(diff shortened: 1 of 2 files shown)


notebook/services/contents/filemanager.py
```diff
@@ -586,3 +590,22 @@ def get_kernel_path(self, path, model=None):
...
+            # Go ahead and add other invalid (and non-path-separator) characters here as well so there's
+            # consistent behavior - although all others will result in '[Errno 22]Invalid Argument' errors.
+            invalid_chars = (':', '>', '<', '*', '?', '|', '"')
+        else:
+            # On non-windows systems, allow the underlying file creation to perform enforcement when appropriate
+            invalid_chars = ()
+
+        for char in invalid_chars:
+            if char in path:
+                raise web.HTTPError(400, u"Path '{}' contains invalid characters {} relative to its platform.  "
```
jasongrout: How about `Path '{}' contains characters that are invalid for the filesystem. Path names on this filesystem cannot contain any of the following characters: {}.`? I'm trying to make the message simpler, though it seems a little less actionable too (though in the validation method, it's not clear exactly who is calling us and what the recovery procedure should be...)

kevin-bates: Thank you - that reads much more cleanly.  Here's a screen shot with the update:
Screen Shot 2020-01-28 at 3 25 12 PM

Previously, I was taking advantage of the parenthesis that sets print with.  That could now be argued as unnecessary.  If there's an easy pythonic way to remove those, I can do that - otherwise it seems okay to me.

jasongrout: You could just define it as a string instead of a tuple.

jasongrout: `invalid_chars = ':><*?|"'`

notebook/services/contents/filemanager.py
```diff
@@ -586,3 +590,23 @@ def get_kernel_path(self, path, model=None):
...
+    @staticmethod
+    def _validate_path(path):
+        """Checks if the path contains invalid characters relative to the current platform"""
+
+        if sys.platform == 'win32':
+            # On Windows systems, we MUST disallow colons otherwise an Alternative Data Stream will
+            # be created and confusion will reign! (See https://github.com/jupyter/notebook/issues/5190)
+            # Go ahead and add other invalid (and non-path-separator) characters here as well so there's
+            # consistent behavior - although all others will result in '[Errno 22]Invalid Argument' errors.
+            invalid_chars = (':', '>', '<', '*', '?', '|', '"')
```
jasongrout:
```suggestion
            invalid_chars = ':><*?|"'
```

notebook/services/contents/filemanager.py
```diff
@@ -586,3 +590,23 @@ def get_kernel_path(self, path, model=None):
...
+        else:
+            # On non-windows systems, allow the underlying file creation to perform enforcement when appropriate
+            invalid_chars = ()
```
jasongrout:
```suggestion
            invalid_chars = ''
```

notebook/services/contents/filemanager.py
```diff
@@ -586,3 +590,23 @@ def get_kernel_path(self, path, model=None):
...
+
+        for char in invalid_chars:
+            if char in path:
+                raise web.HTTPError(400, u"Path '{}' contains characters that are invalid for the filesystem. "
```
jasongrout: The notebook is python 3 only now, right? So we don't need the `u` unicode identifier? `u"..."`

notebook/services/contents/filemanager.py
kevin-bates: :+1: I forget strings are arrays in Python - thanks for the suggestion.

notebook/services/contents/filemanager.py
kevin-bates: Yes - good point.  Left other occurrences alone for now though.

PR #5221, opened 2020-02-11
Add env variable support for port options

### LINKED ISSUE #5212: Use environment variable to set port to run on
So i get this is how you would set the port via cli:

```
jupyter notebook --port 9999
```

Is there currently a way to set this and other run options with an environment variable eg:

```
export JUPYTER_PORT=9999
jupyter notebook
```

kevin-bates:
In order to better support use cases relating to containerized environments,
this change adds environment variable support for the `--port` option.  Since
anyone setting a specific port probably doesn't want port retry logic enabled,
the `--port-retries` option has also been backed by an env.

Option `--port` will be backed by env `JUPYTER_PORT` and still defaults to `8888`.
Option `--port-retries` will be backed by env `JUPYTER_PORT_RETRIES` and still
defaults to `50`.

The CLI options will override those set via the environment, but environment
values override those set via configuration files.

Closes #5212


kevin-bates:
Thanks for reviewing this Steve.  Updated with master.

PR #5278, opened 2020-03-11
Reduce code duplication

mikebeijen:
Removes duplicated code

### DIFF
--- a/notebook/services/contents/fileio.py
+++ b/notebook/services/contents/fileio.py
@@ -99,14 +99,8 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
     if os.path.isfile(path):
         copy2_safe(path, tmp_path, log=log)
 
-    if text:
-        # Make sure that text files have Unix linefeeds by default
-        kwargs.setdefault('newline', '\n')
-        fileobj = io.open(path, 'w', encoding=encoding, **kwargs)
-    else:
-        fileobj = io.open(path, 'wb', **kwargs)
-
     try:
+        fileobj = open_file(path, text, encoding, log, **kwargs)
         yield fileobj
     except:
         # Failed! Move the backup file back to the real path to avoid corruption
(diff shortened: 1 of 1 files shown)


mikebeijen:
Closed and reopened pull request to trigger CI rebuild

notebook/services/contents/fileio.py
```diff
@@ -99,14 +99,8 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
...
 
-    if text:
-        # Make sure that text files have Unix linefeeds by default
-        kwargs.setdefault('newline', '\n')
-        fileobj = io.open(path, 'w', encoding=encoding, **kwargs)
-    else:
-        fileobj = io.open(path, 'wb', **kwargs)
-
     try:
+        fileobj = open_file(path, text, encoding, log, **kwargs)
```
kevin-bates, line 103: To preserve the original integrity, this should occur prior to the `try` block, otherwise you need to protect the `fileobj.close()` call since `fileobj` will be `None` if `open_file()` throws an exception.

notebook/services/contents/fileio.py
```diff
@@ -148,17 +170,9 @@ def _simple_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
...
-
-    if text:
-        # Make sure that text files have Unix linefeeds by default
-        kwargs.setdefault('newline', '\n')
-        fileobj = io.open(path, 'w', encoding=encoding, **kwargs)
-    else:
-        fileobj = io.open(path, 'wb', **kwargs)
 
     try:
+        fileobj = open_file(path, text, encoding, log, **kwargs)
```
kevin-bates, line 175: Same comment as above regarding `fileobj` being `None` in exception block.

notebook/services/contents/fileio.py
```diff
@@ -124,6 +118,34 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
...
+      True.
+
+    encoding : str, optional
+      The encoding to use for files opened in text mode. Default is UTF-8.
+
+    **kwargs
+      Passed to :func:`io.open`."""
+
+    if os.path.islink(path):
+        path = os.path.join(os.path.dirname(path), os.readlink(path))
```
kevin-bates, line 140: This `if` block shouldn't be part of the reduction since this same code exists in [`atomic_writting()`](https://github.com/jupyter/notebook/pull/5278/files#diff-81cc8fe92165832b275b72c7a4df05b0L94-L95) with dependent code in between.

PR #5288, opened 2020-03-16
More conventional way to retrieve user's default shell on Linux/MacOS

dmikushin:
The SHELL value could be inherited from a user who starts Jupyter Not…ebook, such as root or nginx, with possibly non-standard SHELL setting. So the script should explicitly query the /etc/passwd instead. Checked to work on Linux (Ubuntu) and MacOS (Catalina).

### DIFF
--- a/notebook/terminal/__init__.py
+++ b/notebook/terminal/__init__.py
@@ -17,7 +17,10 @@ def initialize(webapp, notebook_dir, connection_url, settings):
     if os.name == 'nt':
         default_shell = 'powershell.exe'
     else:
-        default_shell = which('sh')
+        # The SHELL value could be inherited from a user who starts Jupyter Notebook, such as root or nginx,
+        # with possibly non-standard SHELL setting. So the script should explicitly query the /etc/passwd instead
+        import pwd
+        default_shell = pwd.getpwuid(os.getuid()).pw_shell
     shell = settings.get('shell_command',
         [os.environ.get('SHELL') or default_shell]
     )


notebook/terminal/__init__.py
```diff
@@ -17,7 +17,10 @@ def initialize(webapp, notebook_dir, connection_url, settings):
     if os.name == 'nt':
         default_shell = 'powershell.exe'
     else:
-        default_shell = which('sh')
+        # The SHELL value could be inherited from a user who starts Jupyter Notebook, such as root or nginx,
+        # with possibly non-standard SHELL setting. So the script should explicitly query the /etc/passwd instead
+        import pwd
+        default_shell = pwd.getpwuid(os.getuid()).pw_shell
```
kevin-bates, line 23: I have a couple of comments on this...
1. I don't believe `pw_shell` necessarily has a value, so adding `or which('sh')` would probably be useful.
2. When deployed as a service, the user account Notebook is running as shouldn't have a login shell, but instead, some placeholder (e.g., `/sbin/nologin`).  In this case, the service startup would need to ensure a `SHELL` environment variable is configured.

dmikushin: Dear @kevin-bates thank you for your review! So, the pseudocode should look like:
```
set default_shell to pw_shell;
if (default_shell is empty) then
    set default_shell to env("SHELL");
if (default_shell is empty) then
    default_shell = which('sh')
```

Does it make sense to you? If yes, I will reformulate Python patch to the same.

kevin-bates: I think it is simply: 
```python
default_shell = pwd.getpwuid(os.getuid()).pw_shell or which('sh')
```

There's nothing we can do about /etc/passwd entries configured with non-login shells other than rely on the `SHELL` env because we don't know what the set of non-login shells are.

The `or` statement above addresses the (rare) case when `pw_shell` is `None` or the empty string.

jtpio:
Thanks @dmikushin and @kevin-bates.

Since this logic has now moved to the `jupyter_server_terminals` Jupyter Server extension, it would have to be implemented there: https://github.com/jupyter-server/jupyter_server_terminals

Would you like to open an issue or PR on this repo if you think it would still be useful? Thanks!

PR #5355, opened 2020-04-08
Culling: ensure last_activity attr exists before use

### LINKED ISSUE #5345: Kernel manager object has no attribute 'last_activity' during culling check
Culling runs independently of a kernel's startup and uses the corresponding kernel manager's `last_activity` attribute to help determine if it should be culled or not.  However, because the `last_activity` attribute is added to the kernel manager instance dynamically, there's a small (typically < 10ms) window where the culler sees that the `KernelManager` instance exists sans its `last_activity` attribute.  This can result in the following error on the console:
```
[E 2020-04-07 16:15:08.965 EnterpriseGatewayApp] The following exception was encountered while checking the idle duration of kernel 0070571c-4f92-4027-b9e1-1f98cf62421d: 'RemoteKernelManager' object has no attribute 'last_activity'
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/kernelmanager.py", line 451, in cull_kernels
        await self.cull_kernel_if_idle(kernel_id)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/kernelmanager.py", line 463, in cull_kernel_if_idle
        kernel_id, kernel.kernel_name, kernel.last_activity)
    AttributeError: 'RemoteKernelManager' object has no attribute 'last_activity'
[I 2020-04-07 16:15:08.969 EnterpriseGatewayApp] Kernel started: 0070571c-4f92-4027-b9e1-1f98cf62421d
```
Notebook probably should have subclassed juptyer_client's `KernelManager` to host `last_activity`, but didn't (and it's too late for that), so we should check its existence before checking its cull status.  Since this should only occur milliseconds after a kernel's startup, it won't affect the integrity of the cull.

kevin-bates:
KernelManager's last_activity attribute is added following the instance's
creation.  Culling (independently) uses this attribute and should ensure
it exists prior to its use, else skip the culling check for that instance.

Fixes #5345

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -459,9 +459,9 @@ class MappingKernelManager(MultiKernelManager):
         except KeyError:
             return  # KeyErrors are somewhat expected since the kernel can be shutdown as the culling check is made.
 
-        self.log.debug("kernel_id=%s, kernel_name=%s, last_activity=%s",
-                       kernel_id, kernel.kernel_name, kernel.last_activity)
-        if kernel.last_activity is not None:
+        if hasattr(kernel, 'last_activity'):  # last_activity is monkey-patched, so ensure that has occurred
+            self.log.debug("kernel_id=%s, kernel_name=%s, last_activity=%s",
+                           kernel_id, kernel.kernel_name, kernel.last_activity)
             dt_now = utcnow()
             dt_idle = dt_now - kernel.last_activity
             # Compute idle properties
--- a/notebook/services/kernels/tests/test_kernels_api.py
+++ b/notebook/services/kernels/tests/test_kernels_api.py
@@ -4,6 +4,7 @@ import json
 import sys
 import time
 
+from requests import HTTPError
 from traitlets.config import Config
 
 from tornado.httpclient import HTTPRequest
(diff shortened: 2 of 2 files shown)


notebook/services/kernels/kernelmanager.py
```diff
@@ -459,9 +459,9 @@ async def cull_kernel_if_idle(self, kernel_id):
         except KeyError:
             return  # KeyErrors are somewhat expected since the kernel can be shutdown as the culling check is made.
```
akkim31, line 460: #

notebook/services/kernels/tests/test_kernels_api.py
```diff
@@ -234,3 +235,49 @@ class KernelFilterTest(NotebookTestBase):
     # Sanity check verifying that the configurable was properly set.
     def test_config(self):
```
akkim31, line 236: 236

notebook/services/kernels/tests/test_kernels_api.py
```diff
@@ -234,3 +235,49 @@ class KernelFilterTest(NotebookTestBase):
     # Sanity check verifying that the configurable was properly set.
```
akkim31, line 235: 235

notebook/services/kernels/kernelmanager.py
```diff
@@ -459,9 +459,9 @@ async def cull_kernel_if_idle(self, kernel_id):
...
 
-        self.log.debug("kernel_id=%s, kernel_name=%s, last_activity=%s",
-                       kernel_id, kernel.kernel_name, kernel.last_activity)
-        if kernel.last_activity is not None:
```
akkim31, line 464: if

akkim31 (approved):

notebook/services/kernels/tests/test_kernels_api.py
```diff
@@ -234,3 +235,49 @@ class KernelFilterTest(NotebookTestBase):
...
+                                  )
+
+    def tearDown(self):
+        for k in self.kern_api.list().json():
+            self.kern_api.shutdown(k['id'])
+
+    def test_culling(self):
+        kid = self.kern_api.start().json()['id']
+        ws = self.kern_api.websocket(kid)
+        model = self.kern_api.get(kid).json()
```
urbas, line 266: Heads-up: this test has a race condition. On slower hardware the kernel is already culled by the time the model is fetched and the assertion below fails. Example build failure: https://hydra.nixos.org/build/134833130

kevin-bates: Thank you Matej.  Yeah, I raised this suspicion in the jupyter_server meeting the other day and believe 2 seconds is a bit short.  I think extending the idle timeout to 5 seconds - given this is the only test like this - will be sufficient.

Issue #3868, opened 2018-08-21
Cull idle terminals


clkao:
#2963 added `shutdown_no_activity_timeout` and `terminal_last_activity`, but it seems there's no way to cull idle terminal right now, which [prevent the notebook from being shutdown automatically](https://github.com/jupyter/notebook/blob/98085dcb8420ecc0effaae240bdd8e85bb888105/notebook/notebookapp.py#L1587).

Should we have another pair of `cull_idle_timeout` `cull_interval` for terminals?

foxel:
Is there any update on this feature? Maybe implemented somewhere else?

kevin-bates:
I just noticed @clkao's linked comment in the opening description.  Since #5372 creates an application-level terminal manager, we can easily shut down terminals when the Notebook server is stopped (similar to running kernels).

I'll go ahead and add that to the PR.  If anyone disagrees, please let me know.

RaghavendraSingh:
Hi, Any update on this feature? 
Also, until this feature becomes available, is there a workaround for culling terminal sessions?

kevin-bates:
PR #5372 was created a few weeks back.  I'm hoping to have it reviewed shortly.  If you'd like to take it for a spin and report back on that PR, it would be appreciated.

Issue #1834, opened 2016-10-14
can we find the Jupyter terminals which is running for long ?


mkaja:
can we find the Jupyter terminals which is running for long or idle ? it takes system resources and leads to poor performance

gnestor:
Are you aware of the Home > Running > Terminals tab? It lists all terminal instances. Are you proposing that terminal instances display _age_ like the notebook instances?

![image](https://cloud.githubusercontent.com/assets/512354/19405151/53e0069a-9229-11e6-9759-6e00d22404c4.png)

mkaja:
yes Gnestor - Are you proposing that terminal instances display age like the notebook instances? yes that would be good . what we are trying to achieve is 100s of ppl  open the terminal and not using it for days the linux server is getting killed with high cpu and mem. if we can list the open terminals in linux command line -we can notify folks to close it properly.

gnestor:
@Carreau @minrk I took a look through https://github.com/jupyter/notebook/blob/179bb24fbf79d153812858126127a91431da3319/notebook/static/tree/js/terminallist.js and it looks like a `created` property is not available on terminal items returned from the server. I also looked through https://github.com/jupyter/jupyter_server/blob/dd4b052403cc0ce70dbde9bfc6c71d6b09e814ec/notebook/terminal/api_handlers.py and it looks like terminado doesn't provide anything like a `created` property. Do you have an idea about how we could provide this info for terminal instances?

minrk:
I'm not sure this information is present, but we can easily record a timestamp when a terminal is created. It might make more sense to follow #1827, though, and track activity on the websocket, so the stamp is last-activity, not created.

gnestor:
@minrk Last activity is more similar to last modified (than created) which is what the "Running" tab in the dashboard exposes for notebooks. 

A couple questions regarding #1827: Does creating a terminal instance from the dashboard launch a kernel instance? If so, then would I just query the kernel model and use the `last_activity` value as the "last modified" value for a terminal instance?

takluyver:
No, terminals do not use kernels.

tuliocasagrande:
I added `/api/terminals/` to the documentation: https://github.com/jupyter/jupyter/wiki/Jupyter-Notebook-Server-API#terminals-api, but it doesn't return the state indeed.

PR #5372, opened 2020-04-14
Add ability to cull terminals and track last activity

kevin-bates:
This adds functionality to track a terminal's last activity and optionally cull terminals that have been inactive for some specified duration.  Similar to culling kernels, culling terminals is configured by setting the parameter `--TerminalManager.cull_inactive_timeout` to a value greater than 0 (default is 0 - not enabled).  

The implementation introduces a `TerminalManager class` which is a subclass of terminado's `NamedTermManager` class.  The handlers have been updated to work with this terminal manager.

Here are the configurable options of the TerminalManager class...
```
TerminalManager options
-----------------------
--TerminalManager.cull_inactive_timeout=<Int>
    Default: 0
    Timeout (in seconds) in which a terminal has been inactive and ready to be
    culled. Values of 0 or lower disable culling.
--TerminalManager.cull_interval=<Int>
    Default: 300
    The interval (in seconds) on which to check for terminals exceeding the
    inactive timeout value.
```
This change also updates the REST API such that both name and last_activity are returned in the JSON response.  As a result, this may be helpful in addressing #1834.

I don't see any issues in Notebook's display and handling of terminals with the additional field and I'm assuming it's just ignored.  However, this kind of thing may be considered an API break, although, given its only in responses, most applications can tolerate this.

Resolves: #3868


kevin-bates:
I should also note that I didn't introduce any kind of `execution_state` because it seems the code is more of an echo-er and I believe you'd need to introduce a more formal protocol - like that of kernels.  As a result, if a given operation were performed within a terminal session and that operation produced no output for the more than the duration of the `cull_inactive_timeout`, that terminal session would be culled mid-operation.  This behavior is analogous to kernel culling where both `cull_connected` and `cull_busy` are True.  Caveat emptor.

kevin-bates:
Moving to WIP to add auto-shutdown capabilities similar to kernels when the server is requested to shutdown.

notebook/terminal/terminalmanager.py
```diff
@@ -0,0 +1,150 @@
...
+        # because a terminal has been shutdown
+        TERMINAL_CURRENTLY_RUNNING_TOTAL.dec()
+
+    async def terminate_all(self):
+        """Terminate all terminals."""
+        terms = [name for name in self.terminals]
+        for term in terms:
+            await self.terminate(term, force=True)
+
+    def terminal_model(self, name):
```
Zsailer, line 87: *nitpick*, I think we should name this `get_terminal_model(...)`; otherwise this looks like a property and not a method.

notebook/terminal/terminalmanager.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _initialize_culler(self):
+        """Start culler if 'cull_inactive_timeout' is greater than zero.
+        Regardless of that value, set flag that we've been here.
+        """
+        if not self._initialized_culler and self.cull_inactive_timeout > 0:
+            if self._culler_callback is None:
+                loop = IOLoop.current()
+                if self.cull_interval <= 0:  # handle case where user set invalid value
+                    self.log.warning("Invalid value for 'cull_interval' detected (%s) - using default value (%s).",
```
Zsailer, line 112: Hm, this looks like a type issue that we should catch at the trait definition, no? Should we catch this with a `@validate` call at the top of the class, rather than wait for a terminal to be initialized.

kevin-bates:
Thanks for the review @Zsailer.  You're right about the test (or lack of).  Since there weren't any previously I didn't go down that road for this but would be willing to look into this in jupyter_server (if that's okay for the time being).

I'll address the other issues in an update.

notebook/terminal/terminalmanager.py
kevin-bates: Good idea (same should probably happen with the kernel culling stuff as well - from which this was derived). I wasn't able to get logging to work from the validation presumably because it's not initialized(??), so I used warnings.warn.  What's the correct way to do that from validate methods?

kevin-bates:
@Zsailer - I've gone ahead and added some basic testing of the new terminal manager and culling.  I then noticed (and should have remembered) that jupyter_server has some terminal tests already (awesome) - so those tests can be updated when this is ported to JS.

I suspect the reason there weren't tests in notebook is because this functionality is getting tested in terminado, so I think it's sufficient with just exercising the terminal manager functionality right now.

notebook/terminal/terminalmanager.py
```diff
@@ -0,0 +1,156 @@
...
+    cull_interval = Integer(cull_interval_default, config=True,
+        help="""The interval (in seconds) on which to check for terminals exceeding the inactive timeout value."""
+                            )
+
+    @validate('cull_interval')
+    def _cull_interval_validate(self, proposal):
+        value = proposal['value']
+        if value <= 0:
+            warnings.warn("Invalid value for 'cull_interval' detected ({}) - using default value ({}).".
+                          format(value, self.cull_interval_default))
```
Zsailer, line 43: @kevin-bates, I saw your comment on logging here. `self.log.warn` doesn't work here? The class should be initialized already. I'm diving into traitlets now to see if I can get it to work.

Zsailer:
@kevin-bates, the tests for the TerminalManager look great! Thank you for crafting those.

(Jupyter Server needed some terminal REST API tests, since were passing some extra params through the REST API.)

Zsailer:
> I don't see any issues in Notebook's display and handling of terminals with the additional field and I'm assuming it's just ignored. However, this kind of thing may be considered an API break, although, given its only in responses, most applications can tolerate this.

I think a short (separate) section in the changelog mentioning that a REST API change happened should be enough to cover this "breaking" change. (Like you) I assume that most Jupyter client/frontend applications don't validate every field, so the extra fields will be ignored.

notebook/terminal/terminalmanager.py
Zsailer: In the example below, I'm able to trigger logs from validate methods. This suggests that you should be able log within this validate method. Are you seeing an error? Or just no logs being emitted?

```python
import io
import logging
from traitlets.config import LoggingConfigurable
from traitlets import validate, Unicode


class A(LoggingConfigurable):

    a = Unicode()

    @validate('a')
    def _validate_a(self, proposal):
        # Fire a log warning in a validate method.
        self.log.warning("Warning seen!")
        return proposal



if __name__ == "__main__":
    # Setup logger and catch log output as io stream
    logger = logging.getLogger()
    out = io.StringIO()
    handler = logging.StreamHandler(out)
    logger.addHandler(handler)

    # Initialize the Logging Configurabel
    a = A(log=logger)

    # Setting the trait should fire the validate
    # method and trigger the warning to the log.
    a.a = "test"

    print(f"Log output: {out.getvalue()}")
```

```bash
>>> python traitlog.py                                                                                                                                   
Log output: Warning seen!
```

kevin-bates: I don't see anything logged.  Here's the output using `warnings.warn`...
```
[I 09:11:22.883 NotebookApp] It looks like you're running the notebook from source.
        If you're working on the Javascript of the notebook, try running
    
        npm run build:watch
    
        in another terminal window to have the system incrementally
        watch and build the notebook's JavaScript for you, as you make changes.
/Users/kbates/repos/oss/jupyter/notebook/notebook/terminal/terminalmanager.py:43: UserWarning: Invalid value for 'cull_interval' detected (-10) - using default value (300).
  format(value, self.cull_interval_default))
[I 09:11:23.255 NotebookApp] Serving notebooks from local directory: /Users/kbates/notebooks/alice
```
And here's that same sequence when switched to `self.log.warning`...
```
[I 09:13:39.518 NotebookApp] It looks like you're running the notebook from source.
        If you're working on the Javascript of the notebook, try running
    
        npm run build:watch
    
        in another terminal window to have the system incrementally
        watch and build the notebook's JavaScript for you, as you make changes.
[I 09:13:39.768 NotebookApp] Serving notebooks from local directory: /Users/kbates/notebooks/alice
```
One issue is that TerminalManager isn't derived from `LoggingConfigurable` - which I will update.  However, even after that change, still the same behavior (nothing logged).

notebook/terminal/__init__.py
```diff
@@ -24,11 +25,12 @@ def initialize(webapp, notebook_dir, connection_url, settings):
...
         shell.append('-l')
-    terminal_manager = webapp.settings['terminal_manager'] = NamedTermManager(
+    terminal_manager = webapp.settings['terminal_manager'] = TerminalManager(
         shell_command=shell,
         extra_env={'JUPYTER_SERVER_ROOT': notebook_dir,
                    'JUPYTER_SERVER_URL': connection_url,
                    },
+        parent=parent,
     )
     terminal_manager.log = app_log
```
Zsailer, line 35: I think this should be `terminal_manager.log = parent.log`

notebook/terminal/terminalmanager.py
Zsailer: See my message below. I *think* this is grabbing the wrong logger object (python logging is always confusing to me).

notebook/notebookapp.py
```diff
@@ -1742,7 +1743,7 @@ def init_terminals(self):
 
         try:
             from .terminal import initialize
-            initialize(self.web_app, self.notebook_dir, self.connection_url, self.terminado_settings)
+            initialize(self.web_app, self.notebook_dir, self.connection_url, self.terminado_settings, self)
```
Zsailer, line 1746: Looking at this closer, this function signature bothers me. This is a server extension, so this method signature *should be* equivalent to `load_jupyter_server_extension(nbapp)`.  All of these arguments are attributes of `self`, so why pass them individually.

kevin-bates: Yep - I agree.  This was probably during a quick POC and never revisited. Thank you.

notebook/terminal/__init__.py
```diff
@@ -7,36 +7,38 @@
...
-from terminado import NamedTermManager
 from tornado.log import app_log
 from notebook.utils import url_path_join as ujoin
+from .terminalmanager import TerminalManager
 from .handlers import TerminalHandler, TermSocket
 from . import api_handlers
 
-def initialize(webapp, notebook_dir, connection_url, settings):
+
+def initialize(parent):
```
Zsailer, line 17: This is nitpicky, but I think this argument should be `nbapp` or something similar. It is expected to be *specifically* a NotebookApp instance—all the attributes accessed below are unique to notebookapps.

notebook/terminal/terminalmanager.py
Zsailer: Addressed in later commits. We decided to keep the logging outside of traitlets validation due to odd behavior in class construction from terminado.

Zsailer:
This looks good to me! Thanks, @kevin-bates.

takluyver:
BTW, was it deliberate that this made `terminado` a hard requirement? Previously, it would catch an `ImportError` and the UI would show that terminals were unavailable. Now `notebookapp` can't be imported without terminado available. See #5462.

kevin-bates:
hi Thomas - it's great to hear from you.

No, it wasn't deliberate by any means - that should be evident by looking at the changeset.  The act of wrapping terminado in a local manager class, coupled with providing users help options for enabling culling, inadvertently (and unknowingly) affected the runtime handling of teminado's presence.  I will make this my focus today.

Zsailer:
Oops, thanks, @takluyver. I didn’t know we made terminado an optional dependency either. Apologies for missing this in the review, Kevin.

meeseeksmachine:

https://discourse.jupyter.org/t/terminal-manager-cull-properties-not-being-applied-to-notebook-config/6222/2

PR #5394, opened 2020-04-21
Convert display_id JS test to selenium (#3335)

nfelger:
Hi @takluyver! This is my first time on this project, so please do let me know if anything about this PR isn't in line with how things are done here - and apologies in advance if so!

I was looking for a way to help out and found issue #3335, where I picked a test that seems not to have been converted yet.

The selenium test is passing and all assertions from the JS original are there, but I still have some questions, which I'll post as comments.


notebook/tests/selenium/utils.py
```diff
@@ -256,14 +260,15 @@ def edit_cell(self, cell=None, index=0, content="", render=False):
         for line_no, line in enumerate(content.splitlines()):
             if line_no != 0:
                 self.current_cell.send_keys(Keys.ENTER, "\n")
+            self.current_cell.send_keys(Keys.ENTER, Keys.HOME)  # Avoid auto-indent.
```
nfelger, line 263: Before I discovered the `prefill_notebook` fixture, I was having trouble with auto-indentation when setting cell content to a multiline string that ended in a colon. This resets the cursor to the start of the line before sending the line content.

It is now redundant because the display_id test uses `prefill_notebook`, which doesn't have that problem, but I figured it's a useful fix regardless.

nfelger:
Ugh, apologies for the whitespace spam in the commits for `utils.py` - I've recently switched editors and wasn't aware that I had set it up to do this!

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+            cell.iopub_messages.push(msg.content);
+        };
+        kernel.set_callbacks_for_msg(callback_id, {
+            iopub: {
+                output: add_msg,
+                clear_output: add_msg,
+            }
+        }, false);
+        kernel.output_callback_overrides_push(msg_id, callback_id);
+        """))
```
nfelger, line 108: I ended up keeping most of this test in JavaScript. Why?

- Starting with the second half (L. 92-108), seems to me like these callbacks have to be set from JS, but this is the first time I've come across iopub, so maybe I'm missing something?
- As for the first bit (L. 84-90) - adding a cell and setting its contents - this could be done in selenium, via the UI. However, when I tried that, I started getting intermittent test failures. I played around with this in the browser and it turns out that adding just a slight delay (tens of milliseconds) on L. 91 causes the cell's `iopub_messages` array to remain empty.

It should be obvious from my comments that I don't know what I'm doing or what's going here! :)
If there's a better way to approach this or some documentation I could be reading to understand it better, I'd be all ears.

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+        script = f"return IPython.notebook.get_cell({cell_idx}).iopub_messages"
+        return notebook.browser.execute_script(script)
+    return _get_iopub_messages
+
+
+def test_display_id(prefill_notebook, get_outputs, get_iopub_messages):
+    notebook = prefill_notebook(INITIAL_CELLS)
+    notebook.execute_cell(0)
+
+    #TODO describe test 1
```
nfelger, line 49: Since this is all one big, long test, it'd be nice to have a short description of what each part is testing. I'm not really sure what it _is_ that is being tested here. Could someone who knows more maybe suggest a short summary?

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+        display('below')
+        """))
+    notebook.wait_for_cell_output(1)
+
+    cell1_outputs = get_outputs(1)
+    assert len(cell1_outputs) == 3, 'cell 1 has the right number of outputs'
+    assert cell1_outputs[1]['transient']['display_id'] == 'here', 'has transient display_id'
+    assert cell1_outputs[1]['data']['text/plain'] == '1', 'display_with_id produces output'
+
+    #TODO describe test 2
```
nfelger, line 62: Same as above

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+    assert len(cell3_outputs) == 0, "correct number of cell outputs"
+    assert len(cell3_iopub_messages) == 2, "correct number of callback outputs"
+    assert cell3_iopub_messages[0]['output_type'] == 'display_data', 'check output_type 0'
+    assert cell3_iopub_messages[0]['transient']['display_id'] == 'here', 'check display id 0'
+    assert cell3_iopub_messages[0]['data']['text/plain'] == '5', 'value'
+    assert cell3_iopub_messages[1]['output_type'] == 'update_display_data', 'check output_type 1'
+    assert cell3_iopub_messages[1]['data']['text/plain'] == '6', 'value'
+    assert cell3_iopub_messages[1]['transient']['display_id'] == 'here', 'display id 1'
+
+    #TODO describe test 4
```
nfelger, line 127: Same as above

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+        session.send(iopub, msg_type, content, parent=ip.parent_header)
+    """)]
+
+
+@pytest.fixture
+def get_outputs(notebook):
+    def _get_outputs(cell_idx):
+        return notebook.browser.execute_script(f"""\
+            var cell = Jupyter.notebook.get_cell({cell_idx});
+            return cell.output_area.outputs;
```
nfelger, line 32: The JS version converts the `outputs` first to JSON and then loads them back, I'm not sure why. I skipped this in the Python version (also because selenium probably does something similar anyway).

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+    notebook.execute_cell(0)
+
+    #TODO describe test 1
+    notebook.add_and_execute_cell(content=textwrap.dedent("""\
+        display('above')
+        display_with_id(1, 'here')
+        display('below')
+        """))
+    notebook.wait_for_cell_output(1)
+
```
nfelger, line 56: Here and at various other points in the test, the JS version calls `wait_for_idle`, which checks if `IPython._status == "idle"`. I tried replicating this in the Python version, but it fails because there's no `_status` on  the IPython object? I'm not sure what's going on here, but the tests don't seem affected by this.

notebook/tests/selenium/test_display_id.py
```diff
@@ -0,0 +1,160 @@
...
+import textwrap
+
+from .utils import wait_for_script_to_return_true
+
+INITIAL_CELLS = [textwrap.dedent("""\
+    ip = get_ipython()
+    from IPython.display import display
+    def display_with_id(obj, display_id, update=False, execute_result=False):
+        iopub = ip.kernel.iopub_socket
+        session = get_ipython().kernel.session
```
nfelger, line 14: I'm guessing it's just an oversight by the original author that this line uses `get_ipython()` instead of the `ip` variable (like the other lines)?

```suggestion
        session = ip.kernel.session
```

notebook/tests/selenium/utils.py
nfelger: Hm, looks like this doesn't quite work as I intended, so I'll remove it for now.

PR #5443, opened 2020-05-12
Configuration alias for open_browser

consideRatio:
JupyterHub creates a class called SingleUserNotebookApp that configure
NotebookApp. This commit is meant to remove the commonly seen warning
from configuring open_browser through that class.

```
[W 2020-05-12 14:34:25.041 SingleUserNotebookApp configurable:168] Config option `open_browser` not recognized by `SingleUserNotebookApp`.  Did you mean `browser`?
```

I've suggested a change here but I'm not confident about it as I don't fully understand the configuration setup, haven't test it, or knows if this is the way to go about it.

## My understanding

- We use traitlets, and traitlets are typically configure like `c.ClassNameWithTraitlets.configuration_name = "some value"`.
- NotebookApp is a class derived from JupyterApp which both has traitlet configurations
- SingleUserNotebookApp is a NotebookApp derived class that wants to change the default values of traitlets in NotebookApp. It is done [here](https://github.com/jupyterhub/jupyterhub/blob/5dee864afd86e9dfdc4c12814683412194b363d9/jupyterhub/singleuser.py#L329-L338).
- Aliases are maintained, allowing for example `browser` on SingleUserNotebookApp to refer to the configuration of NotebookApp.
  - [Applications aliases](https://github.com/ipython/traitlets/blob/32ced29b0ab06914ff92e9bdcd8806536633d701/traitlets/config/application.py#L239-L242)
  - [JupyterApp's appended aliases](https://github.com/jupyter/jupyter_core/blob/ac712354ec2c9377a290acca03a91712f991c889/jupyter_core/application.py#L45-L68)
  - [NotebookApp's appended aliases](https://github.com/jupyter/notebook/blob/0090315f6e61e469bb3a353787504ea6407c4c2e/notebook/notebookapp.py#L557-L569)
  - [SingleUserNotebookApp appended aliases](https://github.com/jupyterhub/jupyterhub/blob/5dee864afd86e9dfdc4c12814683412194b363d9/jupyterhub/singleuser.py#L133-L145)

#### Conclusion
So, perhaps open_browser is added as a configuration in NotebookApp, but isn't added to the aliases dictionary, which makes derivative classes unable to configure it on them, because it becomes for example `c.SingleUserNotebookApp.open_browser` instead of `c.NotebookApp.open_browser` due to a missing alias, somehow.

This PR adds this alias, does it make sense? Perhaps we should add more aliases as well to fix multiple issues?

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -563,6 +563,7 @@ aliases.update({
     'certfile': 'NotebookApp.certfile',
     'client-ca': 'NotebookApp.client_ca',
     'notebook-dir': 'NotebookApp.notebook_dir',
+    'open_browser': 'NotebookApp.open_browser',
     'browser': 'NotebookApp.browser',
     'pylab': 'NotebookApp.pylab',
     'gateway-url': 'GatewayClient.url',


kevin-bates:
Hi Erik - I completely agree about configuration system that is provided by traitlets being a bit mysterious.

I think the correct approach here is for `SingleUserNotebookApp` to override the default values for a few of the referenced traitlets (and probably `allow_remote_access`?) - which appears to be your original intent.

The configurable classes have built into them knowledge about the class hierarchy - such that the "configuration point" for derived classes is the class themselves (e.g., SingleUserNotebookApp) even though the configurable option originates on a superclass.  This is evident when you run `jupyter-singleuser --help-all`.  This will display all of the configurable traits and their associated classes (or, what I referred to previously as "configuration points" - for lack of a better term).  

What you'll also notice relative to `--SingleUserNotebookApp` is that none of the traitlets whose default values were manually overridden appear as configurable options in that output.  This is because the override of `open_browser = False` probably overrode the traitlet itself.  As a result, attempts to reference that as a configurable item result in the warning message.  (At least that's my suspicion.)

For a derived class to override an inherited configurable option, its built-in "default" function is required.  You can see this in [`JupyterApp` where it adjusts the default for `log_level`](https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/application.py#L70-L71) from its inherited value (`WARN`) to its desired value (`INFO`).  Likewise, I _could_ change NotebookApp's default `log_level` to `DEBUG` by implementing:
```python
    @default('log_level')
    def _default_log_level(self):
        return logging.DEBUG
```
Note: it's not clear to me when/if the `@default` decorator is necessary since `JupyterApp` doesn't use such a thing to change log level from `WARN` to `INFO`.  Hmm, it might be something as simple as appending `_default` to the function name while prepending it with `_` to eliminate the need for the decorator.  Boy, that's traitlets for you! :smile:

So, I think what should happen here is for `SingleUserNotebookApp` to add default methods for each of the inherited configurable options it wants to override.  I'm hoping that will eliminate the warning and restore the ability to use those options in a configurable manner.

consideRatio:
@kevin-bates thank you so much!! I'll use these insights as a foundation to figure out what I think makes sense which I'm not sure about yet.

The origin of this PR is an itchy feeling of seeing the JupyterHub logged warning about open_browser for two-three years. I bet many others have considered if that is the cause of some issue, so I figure it would be good to remove that distraction.

kevin-bates:
You're welcome @consideRatio.  I think if you replace the current initialization with applicable default methods, you'll find the configurable aspect of those options restored and the behavior (via the defaulting) you desire.

I'm going to close this for now.  Should something unforeseen come up that warrants this PR's reopening, we can do so then.  Thanks.

PR #5451, opened 2020-05-16
Remove unnecessary import of requests_unixsocket

blink1073:
Follow up to #4835.   This import gave a hard dependency on `requests_unixsocket`, which was caught by a failing docs build: https://readthedocs.org/projects/jupyter-notebook/builds/11048908/.


blink1073:
cc @kwlzn, heads up for when you port to `jupyter_server`.

kwlzn:
ah, thanks - noted. remnant of a prior impl that was replaced with socket.socket.

PR #5462, opened 2020-05-20
Install terminado for docs build

takluyver:
RTD is currently failing with an ImportError (e.g. https://readthedocs.org/projects/jupyter-notebook/builds/11068903/ )

Also move to a recent Python, because I don't know how long 3.5 will be available on conda.


takluyver:
(Appears to be introduced by #5372)

PR #5465, opened 2020-05-20
Restore detection of missing terminado package

kevin-bates:
Because the change to add [culling of terminals](#5372) derived from terminado's terminal manager in a local class, it inadvertently broke Notebook server's functionality to tolerate a missing terminado package.  These changes restore that behavior.


kevin-bates:
Support for the command line disablement still isn't right (sigh).  Will remove [WIP] once resolved.

notebook/notebookapp.py
```diff
@@ -300,7 +306,7 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
             allow_password_change=jupyter_app.allow_password_change,
             server_root_dir=root_dir,
             jinja2_env=env,
-            terminals_available=False,  # Set later if terminals are available
+            terminals_available=terminals_available,
```
takluyver, line 309: There are three related things, which I think might be getting a bit confused here:

a. The `terminals_enabled` config option, which is True by default.
b. Whether terminado can be imported
c. Whether we are exposing terminals (i.e. enabled *and* importable)

Currently (before this PR), we use `terminals_enabled` for a and `terminals_available` for c (i.e. available to the user). In this PR,`terminals_available` refers to b, and `terminals_enabled` to c.

kevin-bates: Agreed - see latest commit.

notebook/notebookapp.py
```diff
@@ -300,7 +306,7 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
             allow_password_change=jupyter_app.allow_password_change,
             server_root_dir=root_dir,
             jinja2_env=env,
-            terminals_available=False,  # Set later if terminals are available
+            terminals_available=terminado_available and terminals_enabled,
```
takluyver, line 309: It might be neater to use `jupyter_app.terminals_enabled` to save having to pass the extra parameter through.

notebook/notebookapp.py
```diff
@@ -1759,8 +1768,8 @@ def init_terminals(self):
         try:
             from .terminal import initialize
             initialize(nb_app=self)
-            self.web_app.settings['terminals_available'] = True
         except ImportError as e:
+            self.terminals_enabled = False
```
takluyver, line 1772: Modifying this still seems like it's inviting confusion, because what `.terminals_enabled` means depends on the point at which you use it: before this, it means a config option, and after this it means whether the terminal machinery is used.

Can I suggest that we keep clearly separate terms for the three things we might want to know, e.g. `terminals_available`, `terminals_enabled` and `terminals_in_use`?

kevin-bates: +1, although I believe `terminado_available` as an indication of the package's presence is more clear than `terminals_available` if that's okay.

I also think it's important to issue the warning message when terminals are enabled but the terminado is not present even though we know that answer prior to `init_terminals`.  I've retained that try-catch in case there might be other initialization failures down the road.

notebook/notebookapp.py
```diff
@@ -1908,13 +1923,12 @@ def shutdown_no_activity(self):
...
 
-        try:
-            term_mgr = self.web_app.settings['terminal_manager']
-        except KeyError:
-            pass  # Terminals not enabled
-        else:
-            if term_mgr.terminals:
-                return   # Terminals still running
+        if not self.terminals_in_use:
+            return
```
takluyver, line 1927: This doesn't look quite right. This method is to shut down the server after a certain time with no kernels/terminals. Terminals not in use means no terminals running, but this early return prevents it from ever shutting down.

notebook/notebookapp.py
takluyver: No problem - I wasn't tied to those specific names, just keen to distinguish the concepts.

Now `terminals_available` on the webapp is the equivalent of `terminals_in_use` on the Jupyter app class. It's possibly worth aligning those, but I don't think it's a big deal either.

takluyver:
Other than my latest comment about shutting down on a timeout, I think this looks good now.

notebook/notebookapp.py
kevin-bates: :smile: Looks like we could do with another pass at naming, as you alluded to in another comment.  I was coming to same conclusion about [`terminals_in_use`](https://github.com/kevin-bates/notebook/blob/detect-missing-terminado/notebook/notebookapp.py#L1481-L1485) yesterday (while doing yard work).  I think a switch (alignment) to `terminals_available` might just be the best name:
a. `terminado_available` = Terminado package is installed
b. `terminals_enabled` = User has condoned the use of terminals (independent of terminado's presence)
c. `terminals_available` = User has condoned the use of terminals and terminado is installed

takluyver: To be clear, I think there's an actual logic bug here, regardless of which name we're using. ;-)

kevin-bates: You're absolutely correct.  Thank you for spotting that.

takluyver: Thanks. I think this is in good shape now. :-)

PR #5484, opened 2020-05-27
Added functionality to allow filter kernels by Jupyter Enterprise Gat…

lucabem:
We come from [here](https://github.com/jupyter/enterprise_gateway/pull/800)

We have allowed Jupyter Enterprise Gateway to filter kernels by a query param.
This query param is env var KERNEL_USERNAME 

@kevin-bates

### DIFF
--- a/notebook/gateway/managers.py
+++ b/notebook/gateway/managers.py
@@ -498,14 +498,24 @@ class GatewayKernelManager(MappingKernelManager):
             self.remove_kernel(kernel_id)
 
 
+
 class GatewayKernelSpecManager(KernelSpecManager):
 
     def __init__(self, **kwargs):
         super(GatewayKernelSpecManager, self).__init__(**kwargs)
-        self.base_endpoint = url_path_join(GatewayClient.instance().url,
-                                           GatewayClient.instance().kernelspecs_endpoint)
+        base_endpoint = url_path_join(GatewayClient.instance().url,
+                                      GatewayClient.instance().kernelspecs_endpoint)
+
+        self.base_endpoint = GatewayKernelSpecManager._get_endpoint_for_user_filter(base_endpoint)
         self.base_resource_endpoint = url_path_join(GatewayClient.instance().url,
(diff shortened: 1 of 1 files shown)


notebook/gateway/managers.py
```diff
@@ -543,6 +549,9 @@ def get_all_specs(self):
     def list_kernel_specs(self):
         """Get a list of kernel specs."""
         kernel_spec_url = self._get_kernelspecs_endpoint_url()
+
+        kernel_spec_url = self._get_endpoint_for_user_filter(kernel_spec_url)
```
kevin-bates, line 553: Rather than redo the url calculation, I'm inclined to bake this into `self.base_endpoint` directly within the constructor.  Since env values are static there's really no reason to not run with the query parameter baked in.

notebook/gateway/managers.py
```diff
@@ -519,6 +519,12 @@ def _get_kernelspecs_endpoint_url(self, kernel_name=None):
 
         return self.base_endpoint
 
+    def _get_endpoint_for_user_filter(self, default_endpoint):
```
kevin-bates, line 522: We should probably make this a static method - particularly since it will be called from the constructor, per previous comment.

notebook/gateway/managers.py
```diff
@@ -586,6 +598,7 @@ def get_kernel_spec_resource(self, kernel_name, path):
             The name of the desired resource
         """
         kernel_spec_resource_url = url_path_join(self.base_resource_endpoint, str(kernel_name), str(path))
+        kernel_spec_resource_url = self._get_endpoint_for_user_filter(kernel_spec_resource_url)
```
kevin-bates, line 601: We're not applying filtering to the resource files (typically icons).  These are fetched only after the set of (filtered) kernelspecs have been returned, so are essentially filtered already.

lucabem: Ok, I have deleted it

PR #5497, opened 2020-05-31
remove old JS markdown tests, add a new one in selenium

dmpe:
The new test add a check for a correct rendering of markdown headings in a notebook cell. 

addresses https://github.com/jupyter/notebook/issues/3335


dmpe:
In case you would have an idea how to further improve or test for instance mathjax, then let me know.

PR #5508, opened 2020-06-04
Document contents chunks

vidartf:
Add a documentation entry for the contents API regarding use of "chunk" in save.

Adds an `extra_requires` value in setup.py for installing documentation dependencies.

### DIFF
--- a/docs/source/extending/contents.rst
+++ b/docs/source/extending/contents.rst
@@ -200,6 +200,25 @@ You may be required to specify a Checkpoints object, as the default one,
 ``FileCheckpoints``, could be incompatible with your custom 
 ContentsManager.
 
+
+Chunked Saving
+~~~~~~~~~~~~~~
+
+The contents API allows for "chunked" saving of files, i.e.
+saving/transmitting in partial pieces:
+
+* This can only be used when the ``type`` of the model is ``file``.
+* The model should be as otherwise expected for
+  :meth:`~manager.ContentsManager.save`, with an added field ``chunk``.
+* The value of ``chunk`` should be an integer starting at ``1``, and incrementing
+  for each subsequent chunk, except for the final chunk, which should be
+  indicated with a value of ``-1``.
+* The model returned from using :meth:`~manager.ContentsManager.save` with
+  ``chunk`` should be treated as unreliable for all chunks except the final one.
+* Any interaction with a file being saved in a chunked manner is unreliable
+  until the final chunk has been saved. This includes directory listings.
+
+
 Customizing Checkpoints
 -----------------------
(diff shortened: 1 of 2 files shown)


docs/source/extending/contents.rst
```diff
@@ -200,6 +200,25 @@ You may be required to specify a Checkpoints object, as the default one,
...
+* This can only be used when the ``type`` of the model is ``file``.
+* The model should be as otherwise expected for
+  :meth:`~manager.ContentsManager.save`, with an added field ``chunk``.
+* The value of ``chunk`` should be an integer starting at ``1``, and incrementing
+  for each subsequent chunk, except for the final chunk, which should be
+  indicated with a value of ``-1``.
+* The model returned from using :meth:`~manager.ContentsManager.save` with
+  ``chunk`` should be treated as unreliable for all chunks except the final one.
+* Any interaction with a file being saved in a chunked manner is unreliable
+  until the final chunk has been saved. This includes directory listings.
```
kevin-bates, line 219: Thanks Vidar.

Looking at this area of the code, I see the pre and post save hooks are invoked on the first and last chunks.  While that should be obvious, I wonder if it's worth an explicit statement that the hook invocations are relative to the overall file and not per chunk/request.

vidartf: As these hooks are not mentioned in the current docs at all, I'm not entirely sure what to write, or what context to add it in (it would be weird to only mention them in passing when discussing something else). Any suggestions for what to write are welcome (either as a comment or as a commit on the PR branch).

vidartf: I added an attempt below for now.

kevin-bates: Ah - I didn't realize the hooks weren't documented - so I apologize for bringing this up at all.

docs/source/extending/contents.rst
```diff
@@ -200,6 +200,29 @@ You may be required to specify a Checkpoints object, as the default one,
...
+* For contents managers that fire the pre-/post-save hooks as defined in 
+  :class:`~manager.ContentsManager`, the pre-save hook should be invoked before the
+  first chunk is processed, and the post-save hook should be invoked after the last chunk
+  is received and stored.
```
kevin-bates, line 223: I also didn't realize that the hooks spanned classes - pre-save on the base class and post-save first introduced on the FileContentsManager (which seems like a bug to me).   As a result, I'm not sure we should "go here" as was your initial concern.

If we do decide to take this, I would phrase this paragraph more from a user's perspective than a developer's perspective and change "should be invoked" to "will be invoked".

vidartf:
> Thanks Vidar - sorry for the "hook thrash".

No worries 😄

PR #5530, opened 2020-06-12
Fix exception causes in handlers.py

cool-RR:
I recently went over [Matplotlib](https://github.com/matplotlib/matplotlib/pull/16706), [Pandas](https://github.com/pandas-dev/pandas/pull/32322) and [NumPy](https://github.com/numpy/numpy/pull/15731), fixing a small mistake in the way that Python 3's exception chaining is used. If you're interested, I can do it here too. I've done it on just one file right now. 

The mistake is this: In some parts of the code, an exception is being caught and replaced with a more user-friendly error. In these cases the syntax `raise new_error from old_error` needs to be used.

Python 3's exception chaining means it shows not only the traceback of the current exception, but that of the original exception (and possibly more.) This is regardless of `raise from`. The usage of `raise from` tells Python to put a more accurate message between the tracebacks. Instead of this: 

You'll get this: 

    The above exception was the direct cause of the following exception:

The first is inaccurate, because it signifies a bug in the exception-handling code itself, which is a separate situation than wrapping an exception.

Let me know what you think!

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -452,7 +452,7 @@ class IPythonHandler(AuthenticatedHandler):
                         msg = "Blocking Cross Origin request from {}.".format(referer)
                     else:
                         msg = "Blocking request from unknown origin"
-                    raise web.HTTPError(403, msg)
+                    raise web.HTTPError(403, msg) from e
             else:
                 raise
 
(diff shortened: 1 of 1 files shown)


kevin-bates:
Hi @cool-RR - this seems like a useful set of changes that might better assist in troubleshooting efforts - thank you.

Would you also be interested in contributing this exercise to [`jupyter_server`](https://github.com/jupyter/jupyter_server) as well?  The reason I bring that up is because if you were to make these changes in one of notebook or jupyter_server repos, it would probably be of higher value applied only to the latter (jupyter_server) - since that's more in the long-range plans of the ecosystem.  However, if you'd like to apply to both repositories, these changes are more than welcome!

cool-RR:
I put jupyter/jupyter_server on my list of repos to look at.

cool-RR:
@Zsailer Would you like this change?

kevin-bates:
@cool-RR - this seems like a worthwhile endeavor for this repo.  Would you mind cranking out the rest of the changes?

cool-RR:
I prefer to do it after the initial PR is accepted, so if there are unexpected difficulties, they come up before I've done that work.

cool-RR:
Awesome, I'll open a PR next week with the rest of the fixes.

Issue #5400, opened 2020-04-24
custom_display_url changes only one URL


liamhuber:
First, sorry if I'm posting this in the wrong sub-repo.

My issue is that after setting `c.NotebookApp.custom_display_url = 'http://mycustomurl/'`, it changes only the *first* url in the user output, e.g.

```
    Or copy and paste one of these URLs:
        http://mycustomurl/?token=[a token string]
     or http://127.0.0.1:8888/?token=[a token string]
```

In my particular case I'm just changing the port to 8009, and the result is `localhost:8009` and `127.0.0.1:8888` both appearing.

Would it be possible to show only one link whenever `custom_display_url` is set to something non-empty?

I tracked down [the issue](https://github.com/jupyter/docker-stacks/issues/709) where this url display was decided on, but couldn't find any more recent related issues.

kevin-bates:
Hi @liamhuber - I apologize for the delayed response - this is the correct place to raise this issue.

Your issue makes sense to me.  Did you want to submit a pull request for this?

kevin-bates:
Closed via #5544

liamhuber:
@kevin-bates @LaurMerl,

Thanks for the fix and merge! Thanks for your patience on the gratitude too -- I'm on parental leave atm so my timelines for work stuff are super stretched out.

kevin-bates:
No worries @liamhuber - thanks for raising the issue.  I'm sure there will be other opportunities to contribute to Jupyter.
 
Congratulations on your new addition! :tada:

PR #5544, opened 2020-06-20
FIX: removed double link when custom_display_url (#5400)

LaurMerl:
Hi @kevin-bates, @liamhuber, I had a look at the issue in question and have a suggestion on how to fix it.

This is my first contribution so please do let me know if anything about this PR isn't in line with how things are done - and apologies in advance if so!
Thanks in advance!


LaurMerl:
Hey @kevin-bates, good catch, completely overlooked that!
That also has been taken care of, please let me know what you think of it.

liamhuber:
@LaurMerl Thanks!!!

PR #5545, opened 2020-06-20
Respect cell.is_editable during find-and-replace

adamblake:
Find and replace (searchandreplace.js) will overwrite the contents of cells even if they are marked as non-editable. Add a check against the cell's is_editable() method to ensure this only happens for editable cells.


adamblake:
I see there is some work on find-and-replace in #558, but this wasn't covered

kevin-bates:
Thank you for the contribution @adamblake!

kevin-bates:
@blink1073 - is there a reason we shouldn't include this in 6.1?  Seems like a low-risk bug fix, but I'm not familiar with what impact this may have on the front-end.

blink1073:
6.1 is fine by me

PR #5587, opened 2020-07-06
Added support for JUPYTER_TOKEN_FILE

NemesisFLX:
## Motivation

Currently, we use `JUPYTER_TOKEN` as an environment variable to pass a token to the jupyter notebook with which we can authenticate us later. This is a good practice, for a local environment, but in a docker environment the best practice is to use `docker secrets`. Docker secrets uses instead of a string directly in the env variable, a file location where the string is stored. This file is stored in memory and only accessible by the container that has the correct access right. 

## Changes

Added an env variable `JUPYTER_TOKEN_FILE`, if it is available it will read the token from the given location and return it. Priority has `JUPYTER_TOKEN`. 

## Further information

Docker-compose with docker secret, best practices: https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -931,6 +931,9 @@ class NotebookApp(JupyterApp):
     token = Unicode('<generated>',
         help=_("""Token used for authenticating first-time connections to the server.
 
+        The token can be read from the file referenced by JUPYTER_TOKEN_FILE or set directly
+        with the JUPYTER_TOKEN environment variable.
+
         When no password is enabled,
         the default is to generate a new, random token.
 
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -947,7 +947,7 @@ class NotebookApp(JupyterApp):
             return os.getenv('JUPYTER_TOKEN')
         if os.getenv('JUPYTER_TOKEN_FILE'):
             self._token_generated = False
-            with io.open(os.getenv('JUPYTER_TOKEN_FILE'), "r") as token_file
+            with io.open(os.getenv('JUPYTER_TOKEN_FILE'), "r") as token_file:
                 return token_file.read()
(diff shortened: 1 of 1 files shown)


kevin-bates:
Hi @NemesisFLX - I'm wondering if we should use this opportunity to extend the help test to reference both `JUPYTER_TOKEN` (for by-value) and `JUPYTER_TOKEN_FILE` (for by-reference) support options via the env variables?  Otherwise, it seems folks would need to peruse the code to make that determination.

NemesisFLX:
@kevin-bates I am unsure what you mean by your request. Both `JUPYTER_TOKEN` and `JUPYTER_TOKEN_FILE` can still be used and `JUPYTER_TOKEN_FILE ` has the priority. I'll try my best to add your request, but I am unsure of what to do? I don't see when people need to look inside the code to find something out. Should I write documentation for it?

NemesisFLX:
Oh wait. Do you want me to check if `JUPYTER_TOKEN` is a file reference and if it is pull the token out of the file and if not use the token provided? 

I am not sure if that is best practice. I could do that, but most of the docker images have a special env file for that purpose described in here. https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images

kevin-bates:
I'm referring to the help text relative to the `token` attribute the resides just above your change: https://github.com/jupyter/notebook/pull/5587/files#diff-bbd307804bb6b8836da9f06b8d8a06e9R931-R939

I think it would be helpful to note that `token` can be effectively set via (either) env variable and what the differences are between the two envs (one is the token value, the other is a filename containing the token value (i.e., "by reference")).

NemesisFLX:
@kevin-bates I updated the description. What do you think?

PR #5588, opened 2020-07-06
Improve login shell heuristics

dlukes:
A follow-up on #5565, with better heuristics and respecting overrides via `NotebookApp.terminado_settings`. Cf. https://github.com/jupyter/notebook/pull/5565#issuecomment-653544660 for details.

I'm wondering whether to also update the documentation for `NotebookApp.terminado_settings`, so that it makes it clearer what the shell defaults are and how they can be overridden. Something like:

> Supply overrides for terminado. Currently only supports "shell_command". On Unix systems, a non-login shell is run by default when the notebook server is launched from a terminal, a login shell otherwise.

But AFAICS, this would also require updating translations, so I'm not sure it's worth the trouble.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1160,7 +1160,10 @@ class NotebookApp(JupyterApp):
         """)
     )
     terminado_settings = Dict(config=True,
-            help=_('Supply overrides for terminado. Currently only supports "shell_command".'))
+            help=_('Supply overrides for terminado. Currently only supports "shell_command". '
+                 'On Unix, if "shell_command" is not provided, a non-login shell is launched '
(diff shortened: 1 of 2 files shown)


kevin-bates:
I think updating the help text would be helpful.  I don't think we can worry about translations.  The nature of those is that they'll need to be revisited from time to time to reflect updates and we can't let that jeopardize doing the right thing.

The source changes look good and behave as expected, btw.

Thanks David!

dlukes:
Good point, agreed :) I updated the help text, feel free to suggest edits, or even perform them directly if you have the bandwidth.

Thanks for the quick response!

Issue #3182, opened 2018-01-05
Can't paste two images from clipboard in markdown cell


benjaminabel:
Hi, I recently discovered the awesome paste image feature from clipboard in a markdown cell. However, as images are inserted as markdown code `![image.png](attachment:image.png)`, **when inserting two images in the same cell, the first image is replaced by the second one and the first one is overwritten.**

telamonian:
Found this issue after having the exact same problem. I think the overwrite occurs due to every pasted image having the same name (ie `image.png`) in the notebook's metadata.

Does anyone know if there's a way to rename the inline image immediately after pasting?

telamonian:
One easy fix would be for the notebook to automatically name each image sequentially: `image00.png`, `image01.png`, etc. I bet they didn't implement it that way in order to prevent images from becoming orphaned amongst the cell metadata over the course of multiple cycles of copy-paste. If the user had to manually set extra names, though, they would at least be aware of which images were being preserved.

ihnorton:
Proposed fix: #3821, use a uuid for each attachment.

A more sophisticated thing would be to hash the image content to avoid storing duplicate attachments if an image is pasted twice (at least in the same cell). I didn't see a hashing library dependency already, so I didn't do that.

> prevent images from becoming orphaned amongst the cell metadata over the course of multiple cycles of copy-paste.

I was worried about this too, but found per #621, attachments will be scrubbed automatically when the notebook is saved.

zimmermant:
This still seems to be a problem.  Anyone have a fix for this?

nlahri:
I am also looking for a solution to the same problem.

cyapur:
I am also looking for a solution to the same problem.

evyasonov:
Same problem. Please fix

prampey:
This issue is still relevant now 2 years since it was opened. A fix would be greatly appreciated.

DanielGoldfarb:
This was working fine for me yesterday.  As I pasted each image in the same cell, and number was automatically attached to the name of the image: image.png, image-1.png, image-2.png, etc.

Now today it seems to be behaving as described above.  I have not changed anything.  Any ideas?  Is there some configuration on my machine that could have changed?

PR #5598, opened 2020-07-10
Fixes 3182

withrocks:
Issue #3182 "Can't paste two images from clipboard in markdown cell"
happens when the user tries to paste a second image into a text
cell and the image gets the same attachment key as the previous one.

This patch fixes that issue by adding an index to each picture that's
pasted after the previous one. Example:

* User pastes an image blob A. The text ``
  is added to the text cell and the blob is saved as attachment `image.png`
* The user pastes a second image blob B. The same happens as with blob
  A, except the key is now `image-2.png` rather than `image.png`

Special case:

If the user pastes the exact same blob A twice, it will not be added to
the attachments, but the textcell will get a second entry of
``.

### DIFF
--- a/notebook/static/notebook/js/textcell.js
+++ b/notebook/static/notebook/js/textcell.js
@@ -33,6 +33,58 @@ define([
     "use strict";
     function encodeURIandParens(uri){return encodeURI(uri).replace('(','%28').replace(')','%29')}
 
+    /**
+     * Given a file name and a list of existing file names, returns a new file name
+     * that is not in the existing list. If the file name already exists, a new one with
+     * an incremented index is returned instead.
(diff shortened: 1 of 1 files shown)


withrocks:
I wasn't able to read the general Jupyter general contribution guidelines because the link is broken: https://jupyter.readthedocs.io/en/latest/contributor/content-contributor.html

I hope that won't be an issue.

kevin-bates:
Hi @withrocks - your PR content seems fine although I'm not qualified to review it. 

The broken link should point here: https://jupyter.readthedocs.io/en/latest/contributing/content-contributor.html

I've opened issue #5599 for this.

PR #5616, opened 2020-07-18
bump CodeMirror to 5.55.0

### LINKED ISSUE #5615: Reproducible browser hang in Julia cell
Hey all,
I managed to track down ~~the~~ a ( :smile: ) nasty full-browser hang I was seeing when investigating #5469   .
To reproduce:
 - Create a new Julia notebook.
 - Create a cell, paste in 
  ```jl
  f(
      [i for i in list])
  )
  ```
  ( note the extra close bracket )
  - Wait for CodeMirror to kick in (on my local there appears to be a delay until your Julia kernel starts, which takes a while) and highlight your code.
  - Position the cursor at the start of "`list`", and type "`f2(a`".
  - Browser will hang shortly after due to what looks like infinite parser loop.

I can't repro this on the CodeMirror Julia demo which is updated to CM 5.55. However it does happen reproducibly on Jupyter Notebook 6.1 pre, which is on CM 5.51. Tested in Firefox.
So, any chance of an upgrade to 5.55?

Cheers

akdor1154:
Fixes #5615 .

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "backbone": "components/backbone#~1.2",
     "bootstrap": "bootstrap#~3.4",
     "bootstrap-tour": "0.9.0",
-    "codemirror": "components/codemirror#~5.51.0",
+    "codemirror": "components/codemirror#~5.55.0",
     "create-react-class": "https://cdn.jsdelivr.net/npm/create-react-class@15.6.3/create-react-class.min.js",
     "es6-promise": "~1.0",
     "font-awesome": "components/font-awesome#~4.7.0",


kevin-bates:
@blink1073 - Hmm. If I'm reading things correctly, I see that [Lab uses codemirror >= 5.53.2](https://github.com/jupyterlab/jupyterlab/blob/master/packages/codemirror/package.json#L54).  Given there's an explicit specificity happening there, is it too much of a leap to go to `5.55` here (now that I think about it)?  I'm sorry, but I'm just not in-tune with these packages and how disruptive different minor versions might be.

akdor1154:
thanks guys!

PR #5649, opened 2020-08-03
Lazily import requests_unixsocket

### LINKED ISSUE #5644: Testing using notebook>=6.1 requires `requests_unixsocket`
Update to provide a quick answer:

If you are relying on `notebook.tests` for testing your own code (like a jupyter server extension), be sure to install `notebook[test]` dependencies in your test environment. Starting with `notebook>=6.1`, `requests-unixsocket` was added as test dependency.

---

That import makes `notebook.test` incompatible with Windows:

https://github.com/jupyter/notebook/blob/94af525a2c0aaa99311647c6cccf9add07a63033/notebook/tests/launchnotebook.py#L19

Is it even used as that package was removed in https://github.com/jupyter/notebook/pull/5451?

If it is needed what about using [`pytest.importorskip`](https://docs.pytest.org/en/stable/skipping.html#skipping-on-a-missing-import-dependency) ?

blink1073:
Fixes #5644

### DIFF
--- a/notebook/tests/launchnotebook.py
+++ b/notebook/tests/launchnotebook.py
@@ -16,7 +16,6 @@ pjoin = os.path.join
 from unittest.mock import patch
 
 import requests
-import requests_unixsocket
 from tornado.ioloop import IOLoop
 import zmq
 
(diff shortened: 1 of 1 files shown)


blink1073:
cc @kevin-bates.  I'm happy to cut the release with this patch.

blink1073:
Ha, I think we need a combination of this and #5650

kevin-bates:
Looks like we just duplicated some efforts. :smile: 

Thanks for looking at this.  I guess I addressed this slightly differently in that I also prevent the inclusion of the requests_unixsocket package, although I like the idea of deferring its import as you did.

blink1073:
Sounds like a plan.  Closing in favor of #5650

PR #5650, opened 2020-08-03
Prevent inclusion of requests_unixsocket on Windows

### LINKED ISSUE #5644: Testing using notebook>=6.1 requires `requests_unixsocket`

kevin-bates:
Prevent the need to install `requests_unixsocket` on Windows so that downstream projects that rely on notebook's test framework are not side-affected by a package that is not used.

Fixes #5644

### DIFF
--- a/notebook/tests/launchnotebook.py
+++ b/notebook/tests/launchnotebook.py
@@ -16,7 +16,8 @@ pjoin = os.path.join
 from unittest.mock import patch
 
 import requests
-import requests_unixsocket
+if not sys.platform.startswith('win'):
+    import requests_unixsocket
 from tornado.ioloop import IOLoop
 import zmq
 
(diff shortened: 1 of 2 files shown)


blink1073:
Released https://pypi.org/project/notebook/6.1.1/#files

Issue #5656, opened 2020-08-06
File manager cannot remove invalid soft link


franckchen:
I thought this maybe a bug.the problem is quite simple. If I create a file or directory, then create a soft link and link it to that file we just mentioned.next step, remove the file to make the soft link invalid. 
Then, If i delete the soft link by api. i will get an 404 error, tell me File or directory does not exist.
so seems like this invalid soft link cant remove forever.

PR #5657, opened 2020-08-06
Fix cant delete invalid soft link #5656

franckchen:


Zsailer:
@franckchen I'm not sure this fixes your original issue. This still won't allow you to delete the soft link... It will just throw an error that the directory doesn't exist.

I think there are two possible directions to go here:

1. If it is a soft link, but it's broken, return an error saying the link can't be deleted by the contents API. Try removing the link directly on the server.
2. Enable the contents API to remove the symbolic link even if it's broken. 

We can check if the link is broken with some chunk like:
```python 
if not os.path.exists(os.readlink(path)):
    ...
    # Raise error or remove.
```

PR #5662, opened 2020-08-07
Fix russian message format for delete/duplicate actions

rails-to-cosmos:
In Russian environment delete and duplicate methods are broken when perform on exactly one file/folder.

In this pull request I've fixed russian message format for signular prompts.


kevin-bates:
@rails-to-cosmos  - This has been delivered in Notebook 6.1.2 which is now available - thank you for your contribution!

PR #5672, opened 2020-08-12
Tooltip style scope fix

### LINKED ISSUE #5670: Global [role=tooltip] styles impact notebook extensions
In https://github.com/jupyter/notebook/commit/fb08a1bda3ec03a820891cb10434f128734b4fb8#diff-321a1fb8c95f4257fbede791b7856b7cR438-R459 some global styles regarding elements with `role="tooltip"` were introduced. They were meant to fix issues specific to the `.dynamic-buttons`-div element, and wouldn't need to be put in global scope.

Setting every element with `role="tooltip"` `visibility` to `hidden` impacts notebook extensions and makes it tedious to use this attribute correctly. 

I suggest changing the scope of the above mentioned style to `.dynamic-buttons [role=tooltip]` in notebook/static/tree/less/tree.less line 438.

magdapoppins:
- Make styles of `.dynamic-buttons` class scoped
- Drop some unused styles

Fixes #5670

### DIFF
--- a/notebook/static/tree/less/tree.less
+++ b/notebook/static/tree/less/tree.less
@@ -414,46 +414,46 @@ ul#new-menu {
 }
 
 //TO show tooltips via keyboard for"Duplicate","View","Rename", "Download", "Shutdown", "Edit","Move", "Delete" buttons.
+.dynamic-buttons {
+    .visually-hidden {
+        clip-path: inset(100%);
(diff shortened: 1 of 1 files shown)


kevin-bates:
@magdapoppins - This has been delivered in Notebook 6.1.2 which is now available - thank you for your contribution!

PR #5676, opened 2020-08-12
Title new buttons with label if action undefined

rgbkrk:
On a recent release of the notebook js that included https://github.com/jupyter/notebook/pull/5107, some extensions would fail because they either did not have an `action` _or_ their `action.help` was not part of the i18n lookup (most extensions). This provides a fallback to `el.label`.

### DIFF
--- a/notebook/static/notebook/js/toolbar.js
+++ b/notebook/static/notebook/js/toolbar.js
@@ -92,10 +92,14 @@ define(['jquery','base/js/i18n'], function($, i18n) {
                     action = that.actions.get(el.action);
                     action_name = el.action
                 }
+                var title = el.label;
+                if(action && action.help) {
+                    title = i18n.msg._(action.help) || el.label;
+                }
                 var button  = $('<button/>')
                     .addClass('btn btn-default')
                     .attr("aria-label", el.label)
-                    .attr("title", i18n.msg._(action.help)||el.label)
+                    .attr("title", title)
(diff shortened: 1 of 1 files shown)


rgbkrk:
@drillan -- when people use https://github.com/drillan/jupyter-black with a more recent Jupyter release they run into this error:

image

Since this would affect more than just the black extension, I went after fixing it in Jupyter directly. Just a heads up in case you get any issues!

rgbkrk:
@Zsailer @kevin-bates How should extension developers adapt to i18n?

kevin-bates:
Shoot - I just cut NB 6.1.2 this morning!  Since this was introduced in 6.1, this is a regression and worthy of a patch.  Seems like the previous change also intended for el.label to be the fallback but didn't anticipate a null action.

> How should extension developers adapt to i18n?

@rgbkrk - Sorry, I have no idea.  Seems like a question for when this was first developed, but the [i18n README](https://github.com/jupyter/notebook/blob/master/notebook/i18n/README.md) doesn't touch on that.

rgbkrk:
> Shoot - I just cut NB 6.1.2 this morning!

Love to keep those patches coming. 😄

kevin-bates:
@rgbkrk et al - Notebook 6.1.3 is available on pypi

rgbkrk:
Thank you Kevin!

PR #5703, opened 2020-08-27
Revert change in page alignment

Zsailer:
This reverts #3173, which introduces some new alignment issues for browsers in on Arch Linux.

The new issues were slightly worse the original issue that #3173 tried to fix, so we're reverting it here.


Zsailer:
Pinging @kevin-bates

PR #5716, opened 2020-08-31
Reliable list_running_servers(runtime_dir) design

abaelhe:
active check whether that process is really available via real HTTP request;
process pid maybe recycle and reused by host Operating System that i encountered;
and we must use real HTTP request other than the past check_pid(info['pid']).

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -34,6 +34,7 @@ from notebook._tz import utcnow
 from notebook.i18n import combine_translations
 from notebook.utils import is_hidden, url_path_join, url_is_absolute, url_escape, urldecode_unix_socket_path
 from notebook.services.security import csp_report_uri
+from notebook import JUPYTER_NOTEBOOK_TAG
 
 #-----------------------------------------------------------------------------
 # Top-level handlers
@@ -849,7 +850,7 @@ class APIVersionHandler(APIHandler):
 
     def get(self):
         # not authenticated, so give as few info as possible
-        self.finish(json.dumps({"version":notebook.__version__}))
+        self.finish(json.dumps({"version": notebook.__version__, "module": JUPYTER_NOTEBOOK_TAG}))
 
 
 class TrailingSlashHandler(web.RequestHandler):
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -430,7 +430,7 @@ class NotebookPasswordApp(JupyterApp):
 
 
 def kernel_request(server_info, path='/login', method='GET', body=None, headers=None, timeout=5, log=None):
-    """query a notebook server in a separate process."""
+    """Query a notebook server in a separate process."""
     from tornado import gen
     from tornado.httpclient import AsyncHTTPClient, HTTPClient, HTTPRequest, HTTPError
     from tornado.netutil import Resolver
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1424,7 +1424,7 @@ class NotebookApp(JupyterApp):
 
     @default('file_id')
     def _default_file_id(self):
-        int_hash = "%s" % (hash(self.sock) if self.sock else self.port)
+        return str(hash(self.sock) if self.sock else self.port)
         return str(int_hash)
 
     info_file = Unicode()
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -435,7 +435,6 @@ def kernel_request(server_info, path='/login', method='GET', body=None, headers=
     from tornado.httpclient import AsyncHTTPClient, HTTPClient, HTTPRequest, HTTPError
     from tornado.netutil import Resolver
     url = server_info['url']
-    pid = server_info['pid']
     resolver = None
 
     # UNIX Socket handling.
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -494,7 +494,6 @@ def shutdown_server(server_info, timeout=5, log=None):
     Returns True if the server was stopped by any means, False if stopping it
     failed (on Windows).
     """
-    url = server_info['url']
     pid = server_info['pid']
     kernel_request(server_info, path='api/shutdown', method='POST', body=b'', timeout=timeout, log=log)
 
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2244,7 +2244,8 @@ class NotebookApp(JupyterApp):
                  "resources section at https://jupyter.org/community.html."))
 
         self.write_server_info_file()
-        self.write_browser_open_file()
+        if not self.sock:
+            self.write_browser_open_file()
 
         if (self.open_browser or self.file_to_run) and not self.sock:
             self.launch_browser()
@@ -2326,7 +2327,7 @@ def list_running_servers(runtime_dir=None):
 
             # Actively check whether that process is really available via an HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
-            response = kernel_request(info, path='/api')
+            response = kernel_request(info, path=url_path_join(info.get('base_url') or '/','/api'))
             if response.body == api_version_json_bytes:
                 yield info
             else:
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -107,6 +107,7 @@ from traitlets import (
     Any, Dict, Unicode, Integer, List, Bool, Bytes, Instance,
     TraitError, Type, Float, observe, default, validate
 )
+from traitlets.config.configurable import MultipleInstanceError
 from ipython_genutils import py3compat
 from jupyter_core.paths import jupyter_runtime_dir, jupyter_path
 from notebook._sysinfo import get_sys_info
@@ -2342,12 +2343,11 @@ def list_running_servers(runtime_dir=None):
                 received_version = version_dict.get('version', '0.0.0')
                 msg_special = "Jupyter Notebook HIGHER VERSION detected: %s, current:%s" %(
                     received_version, __version__)
-                msg_upgrade = "!!! Jupyter Notebook Upgrade REQUIRED: >=%s, got:%s(info file removed) !!!" % (
+                msg_upgrade = "Jupyter Notebook Upgrade REQUIRED: >=%s, got:%s(info file removed)" % (
                     __version__, received_version)
 
                 if JUPYTER_NOTEBOOK_TAG != version_dict.get('module'):
-                    print(msg_upgrade)
-                    flush_info_file()
+                    raise MultipleInstanceError(msg_upgrade)
                 else:
                     if __version__ < received_version:
                         print(msg_special)
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2320,7 +2320,8 @@ def list_running_servers(runtime_dir=None):
             # active check whether that process is really available via real HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
             response = kernel_request(info, path='/login')
-            if response.body.find(b'<title>Jupyter Notebook</title>') > 0:
+            if response.body.find(b'<title>Jupyter Notebook</title>') > 0 or \
+                    response.body.find(b'<title>Home Page - Select or create a notebook</title>') > 0:
                 yield info
             else:
                 # If the process has died, try to delete its info file
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2317,7 +2317,7 @@ def list_running_servers(runtime_dir=None):
             with io.open(os.path.join(runtime_dir, file_name), encoding='utf-8') as f:
                 info = json.load(f)
 
-            # active check whether that process is really available via real HTTP request
+            # Actively check whether that process is really available via an HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
             response = kernel_request(info, path='/login')
             if response.body.find(b'<title>Jupyter Notebook</title>') > 0 or \
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1421,18 +1421,25 @@ class NotebookApp(JupyterApp):
               "sent by the upstream reverse proxy. Necessary if the proxy handles SSL"))
     )
 
+    file_id = Unicode()
+
+    @default('file_id')
+    def _default_file_id(self):
+        int_hash = "%s" % (hash(self.sock) if self.sock else self.port)
+        return str(int_hash)
+
     info_file = Unicode()
 
     @default('info_file')
     def _default_info_file(self):
-        info_file = "nbserver-%s.json" % os.getpid()
+        info_file = "nbserver-%s.json" % self.file_id
         return os.path.join(self.runtime_dir, info_file)
 
     browser_open_file = Unicode()
 
     @default('browser_open_file')
     def _default_browser_open_file(self):
-        basename = "nbserver-%s-open.html" % os.getpid()
+        basename = "nbserver-%s-open.html" % self.file_id
         return os.path.join(self.runtime_dir, basename)
 
     pylab = Unicode('disabled', config=True,
@@ -2311,6 +2318,7 @@ def list_running_servers(runtime_dir=None):
     if not os.path.isdir(runtime_dir):
         return
 
+    api_version_json_bytes = json.dumps({"version":__version__}).encode()
     for file_name in os.listdir(runtime_dir):
         if re.match('nbserver-(.+).json', file_name):
             with io.open(os.path.join(runtime_dir, file_name), encoding='utf-8') as f:
@@ -2318,9 +2326,8 @@ def list_running_servers(runtime_dir=None):
 
             # Actively check whether that process is really available via an HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
-            response = kernel_request(info, path='/login')
-            if response.body.find(b'<title>Jupyter Notebook</title>') > 0 or \
-                    response.body.find(b'<title>Home Page - Select or create a notebook</title>') > 0:
+            response = kernel_request(info, path='/api')
+            if response.body == api_version_json_bytes:
                 yield info
             else:
                 # If the process has died, try to delete its info file
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -71,6 +71,7 @@ if not sys.platform.startswith('win'):
     from tornado.netutil import bind_unix_socket
 
 from notebook import (
+    JUPYTER_NOTEBOOK_TAG,
     DEFAULT_NOTEBOOK_PORT,
     DEFAULT_STATIC_FILES_PATH,
     DEFAULT_TEMPLATE_PATH_LIST,
@@ -2327,9 +2328,10 @@ def list_running_servers(runtime_dir=None):
             # Actively check whether that process is really available via an HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
             response = kernel_request(info, path=url_path_join(info.get('base_url') or '/','/api'))
-            if response.body == api_version_json_bytes:
+            try:
+                assert JUPYTER_NOTEBOOK_TAG == json.loads(response.body)["module"]
                 yield info
-            else:
+            except:
                 # If the process has died, try to delete its info file
                 try:
                     os.unlink(os.path.join(runtime_dir, file_name))
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2329,7 +2329,7 @@ def list_running_servers(runtime_dir=None):
             # Also remove leftover files from IPython 2.x without a pid field
             response = kernel_request(info, path=url_path_join(info.get('base_url') or '/','/api'))
             try:
-                assert JUPYTER_NOTEBOOK_TAG == json.loads(response.body)["module"]
+                assert JUPYTER_NOTEBOOK_TAG == json.loads(response.body.decode())["module"]
                 yield info
             except:
                 # If the process has died, try to delete its info file
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2341,7 +2341,7 @@ def list_running_servers(runtime_dir=None):
                 flush_info_file()
             else:
                 received_version = version_dict.get('version', '0.0.0')
-                msg_special = "Jupyter Notebook HIGHER VERSION detected: %s, current:%s" %(
+                msg_special = "WARNING: Jupyter Notebook HIGHER VERSION was detected: %s, current:%s" %(
                     received_version, __version__)
                 msg_upgrade = "Jupyter Notebook Upgrade REQUIRED: >=%s, got:%s(info file removed)" % (
                     __version__, received_version)
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2319,7 +2319,6 @@ def list_running_servers(runtime_dir=None):
     if not os.path.isdir(runtime_dir):
         return
 
-    api_version_json_bytes = json.dumps({"version":__version__}).encode()
     for file_name in os.listdir(runtime_dir):
         if re.match('nbserver-(.+).json', file_name):
             with io.open(os.path.join(runtime_dir, file_name), encoding='utf-8') as f:
@@ -2328,15 +2327,36 @@ def list_running_servers(runtime_dir=None):
             # Actively check whether that process is really available via an HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
             response = kernel_request(info, path=url_path_join(info.get('base_url') or '/','/api'))
-            try:
-                assert JUPYTER_NOTEBOOK_TAG == json.loads(response.body.decode())["module"]
-                yield info
-            except:
-                # If the process has died, try to delete its info file
+            def flush_info_file():
                 try:
                     os.unlink(os.path.join(runtime_dir, file_name))
                 except OSError:
                     pass  # TODO: This should warn or log or something
+
+            version_dict = {}
+            erro = False
+            try:
+                version_dict.update(json.loads(response.body.decode()))
+            except:
+                flush_info_file()
+                erro = True
+            else:
+                received_version = version_dict.get('version', '0.0.0')
+                msg_special = "Jupyter Notebook HIGHER VERSION detected: %s, current:%s" %(
+                    received_version, __version__)
+                msg_upgrade = "!!! Jupyter Notebook Upgrade REQUIRED: >=%s, got:%s(info file removed) !!!" % (
+                    __version__, received_version)
+
+                if JUPYTER_NOTEBOOK_TAG != version_dict.get('module'):
+                    print(msg_upgrade)
+                    flush_info_file()
+                else:
+                    if __version__ < received_version:
+                        print(msg_special)
+                    yield info
+
+
+
 #-----------------------------------------------------------------------------
 # Main entry point
 #-----------------------------------------------------------------------------
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -429,20 +429,10 @@ class NotebookPasswordApp(JupyterApp):
         self.log.info("Wrote hashed password to %s" % self.config_file)
 
 
-def shutdown_server(server_info, timeout=5, log=None):
-    """Shutdown a notebook server in a separate process.
-
-    *server_info* should be a dictionary as produced by list_running_servers().
-
-    Will first try to request shutdown using /api/shutdown .
-    On Unix, if the server is still running after *timeout* seconds, it will
-    send SIGTERM. After another timeout, it escalates to SIGKILL.
-
-    Returns True if the server was stopped by any means, False if stopping it
-    failed (on Windows).
-    """
+def kernel_request(server_info, path='/login', method='GET', body=None, headers=None, timeout=5, log=None):
+    """query a notebook server in a separate process."""
     from tornado import gen
-    from tornado.httpclient import AsyncHTTPClient, HTTPClient, HTTPRequest
+    from tornado.httpclient import AsyncHTTPClient, HTTPClient, HTTPRequest, HTTPError
     from tornado.netutil import Resolver
     url = server_info['url']
     pid = server_info['pid']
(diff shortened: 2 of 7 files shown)


kevin-bates (commented):
Thanks for opening this pull request.  I understand why this is being done, although I'm curious how often you're finding the PIDs getting recycled (I know some OSs recycle much more frequently than others) and wonder why you're encountering "stale" nbserver files in the first place.  How are you terminating the Notebook server normally?  Is there some interaction that is using `SIGKILL` instead of `SIGTERM` or a larger problem where the server is improperly terminating?  I suppose it takes just one bad apple (i.e., stale PID) and things can go wrong, and, I agree, this approach helps mitigate that - I'm just not sure about the trade-off.

I've provided a few comments and observations that I hope you find helpful.  I didn't comment on the test code and it appears some socket-based tests are failing which rely on the list and stop functionality.

abaelhe:
> Thanks for opening this pull request. I understand why this is being done, although I'm curious how often you're finding the PIDs getting recycled (I know some OSs recycle much more frequently than others) and wonder why you're encountering "stale" nbserver files in the first place. How are you terminating the Notebook server normally? Is there some interaction that is using `SIGKILL` instead of `SIGTERM` or a larger problem where the server is improperly terminating? I suppose it takes just one bad apple (i.e., stale PID) and things can go wrong, and, I agree, this approach helps mitigate that - I'm just not sure about the trade-off.

i am building a Dockerfile on  RedHat's quay.io, to deploy my Cloud based Amazon.com Crawler on Mac/Linux/...;
in the meantime; god or the Operating System selected my dear "jupyter-notebook" process;
and i found "jupyter-notebook" previous check_pid(pid) design flaw at first glance;
meanwhile, i am making my IPython Kernel for Apple Swift Language(Python/C++/Swift/LLVM/CLang/LLDB... );
and i choose this reliable and better design and implement as my first milestone;


> 
> I've provided a few comments and observations that I hope you find helpful. I didn't comment on the test code and it appears some socket-based tests are failing which rely on the list and stop functionality.

your sparks are welcome.

notebook/notebookapp.py
```diff
@@ -429,20 +429,10 @@ def start(self):
...
-    *server_info* should be a dictionary as produced by list_running_servers().
-
-    Will first try to request shutdown using /api/shutdown .
-    On Unix, if the server is still running after *timeout* seconds, it will
-    send SIGTERM. After another timeout, it escalates to SIGKILL.
-
-    Returns True if the server was stopped by any means, False if stopping it
-    failed (on Windows).
-    """
+def kernel_request(server_info, path='/login', method='GET', body=None, headers=None, timeout=5, log=None):
```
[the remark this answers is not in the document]
abaelhe: > This method should be renamed since its unrelated to kernel. Perhaps `server_request`?
we creator have this naming right;
with respect to Guido name Python; and their first creator name IPython and Jupyter also Anaconda;
i am working on my IPython Swift Kernel, kernel in my brain;
don't mind, different think lead sparks.

> Also, please remove the default value from `path` since that should be explicitly provided by the caller (which is already the case anyway).
caller will provide their; this default value does not conflict to different `path`;
we use '/login' as 'default value' to remind caller authentication first.

notebook/notebookapp.py
```diff
@@ -2294,9 +2317,10 @@ def list_running_servers(runtime_dir=None):
             with io.open(os.path.join(runtime_dir, file_name), encoding='utf-8') as f:
                 info = json.load(f)
 
-            # Simple check whether that process is really still running
+            # active check whether that process is really available via real HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
-            if ('pid' in info) and check_pid(info['pid']):
+            response = kernel_request(info, path='/login')
+            if response.body.find(b'<title>Jupyter Notebook</title>') > 0:
```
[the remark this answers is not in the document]
abaelhe: > When running on my system, I find a title of: `<title>Home Page - Select or create a notebook</title>`. Not sure if there can be others, but both of these should be tolerated.

i have assigned two kind of values:
'<title>Home Page - Select or create a notebook</title' for UNIX socket: "jupyter-notebook --sock=... ";
'<title>Jupyter Notebook</title>' for TCP socket;

> 1. Can these titles be localized?  If so, that will be problematic.  Perhaps there's another indicator that can be used?

till now, all tests passed;

> 2. When invoked when no server is running, it can take a while to get the exception back (or perhaps `None` based on previous comment).  If there are multiple files to process, this can take a while.

one step one footprint;
i can make multiple commits;
you think is useful. we can works together on more commits;

> 3. Since there can be multiple files referencing the same port, but different PIDs, results can be a little misleading.  For example, should the subsequent `/api/shutdown` request not shut the server down (for whatever reason), the fallback logic to send `SIGTERM` to the pid, could result in the termination of the wrong application, and we're back to where we started.

this commit only one footprint;
you consideration is really useful.

we can define server_info_file naming rule; for example: hash(str(self.sock) + str(self.port));
then reuse port will overwrite previous connection file in "self.write_server_info_file()":
further commits on persistent file naming rule; mark here. 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
    info_file = Unicode()                                                                                                                                                   
                                                                                                                                                                            
    @default('info_file')                                                                                                                                                   
    def _default_info_file(self):                                                                                                                                           
        info_file = "nbserver-%s.json" % os.getpid()                                                                                                                        
        return os.path.join(self.runtime_dir, info_file)                                                                                                                    
                                                                                                                                                                            
    browser_open_file = Unicode()                                                                                                                                           
                                                                                                                                                                            
    @default('browser_open_file')                                                                                                                                           
    def _default_browser_open_file(self):                                                                                                                                   
        basename = "nbserver-%s-open.html" % os.getpid()                                                                                                                    
        return os.path.join(self.runtime_dir, basename)

kevin-bates:
I just now found that Lab returns yet a different title.  This implies any front-end extension will have a different "title", making this approach too fragile.  Since those applications could also rely on `jupyter notebook list` and `jupyter notebook stop <port/socket>` functionality, I believe a different approach should be adopted and, again, question the utility when, in practice, PID reuse would be rare in normal circumstances (i.e., when the notebook server is stopped gracefully).

notebook/notebookapp.py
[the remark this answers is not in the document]
abaelhe: [master a4aae8fe6] consistent persistent file naming rule and same version check rule
+            response = kernel_request(info, path='/api')
+            if response.body == api_version_json_bytes:
                 yield info

abaelhe:
> Thanks for opening this pull request. I understand why this is being done, although I'm curious how often you're finding the PIDs getting recycled (I know some OSs recycle much more frequently than others) and wonder why you're encountering "stale" nbserver files in the first place. How are you terminating the Notebook server normally? Is there some interaction that is using `SIGKILL` instead of `SIGTERM` or a larger problem where the server is improperly terminating? I suppose it takes just one bad apple (i.e., stale PID) and things can go wrong, and, I agree, this approach helps mitigate that - I'm just not sure about the trade-off.
> 
> I've provided a few comments and observations that I hope you find helpful. I didn't comment on the test code and it appears some socket-based tests are failing which rely on the list and stop functionality.

abaelhe:
reopen

notebook/notebookapp.py
kevin-bates, line 432: Per the previous conversation, this method name does not reflect its function and `path` should not have a default since it's used in exactly two locations, neither of which use `'/login'`.
```suggestion
def server_request(server_info, path, method='GET', body=None, headers=None, timeout=5, log=None):
```

notebook/notebookapp.py
```diff
@@ -429,20 +429,10 @@ def start(self):
...
-    failed (on Windows).
-    """
+def kernel_request(server_info, path='/login', method='GET', body=None, headers=None, timeout=5, log=None):
+    """Query a notebook server in a separate process."""
     from tornado import gen
-    from tornado.httpclient import AsyncHTTPClient, HTTPClient, HTTPRequest
+    from tornado.httpclient import AsyncHTTPClient, HTTPClient, HTTPRequest, HTTPError
     from tornado.netutil import Resolver
     url = server_info['url']
     pid = server_info['pid']
```
kevin-bates, line 438: `pid` is not used in the request method and can be removed.

notebook/notebookapp.py
```diff
@@ -468,12 +458,44 @@ def resolve(self, host, port, *args, **kwargs):
...
+
+    savedAsyncHTTPClient = AsyncHTTPClient._save_configuration()
+    try:
+        AsyncHTTPClient.configure(None, resolver=resolver)
+        response = HTTPClient(AsyncHTTPClient).fetch(req)
+    except HTTPError as e:
+        if e.response is not None:
+            response = e.response
+        else:
+            raise
```
kevin-bates, line 481: Again, exceptions need to be handled within this method or each of its two callers.  Neither caller (`jupyter notebook list` and `jupyter notebook stop <port>`) should terminate when encountering an exception while issuing a request to a potentially non-running endpoint.

notebook/notebookapp.py
```diff
@@ -468,12 +458,44 @@ def resolve(self, host, port, *args, **kwargs):
...
+    On Unix, if the server is still running after *timeout* seconds, it will
+    send SIGTERM. After another timeout, it escalates to SIGKILL.
+    Returns True if the server was stopped by any means, False if stopping it
+    failed (on Windows).
+    """
+    pid = server_info['pid']
+    kernel_request(server_info, path='api/shutdown', method='POST', body=b'', timeout=timeout, log=log)
 
     # Poll to see if it shut down.
     for _ in range(timeout*10):
```
kevin-bates, line 501: Since the entire purpose of this PR is to not send a termination signal to an invalid PID, I think we should probably remove this logic and replace with something like the following...

1. Issue a request against `/api` to see if the server that was just shutdown is shutdown.  This may need to happen a couple times.
2. If the server is still considered running, issue a warning message indicating that the server does not appear to be terminated.  This message could also include the PID, but we should probably hold off killing the PID.  We should _suggest_ the user do that once they confirm the PID is something they want to terminate.

notebook/notebookapp.py
```diff
@@ -1399,18 +1421,25 @@ def _update_mathjax_config(self, change):
               "sent by the upstream reverse proxy. Necessary if the proxy handles SSL"))
     )
 
+    file_id = Unicode()
+
+    @default('file_id')
+    def _default_file_id(self):
+        int_hash = "%s" % (hash(self.sock) if self.sock else self.port)
```
kevin-bates, line 1428: This is redundant or the cast using `str()` is - take your pick.
```suggestion
        return str(hash(self.sock) if self.sock else self.port)
```

notebook/notebookapp.py
```diff
@@ -2289,14 +2319,16 @@ def list_running_servers(runtime_dir=None):
     if not os.path.isdir(runtime_dir):
         return
 
+    api_version_json_bytes = json.dumps({"version":__version__}).encode()
```
kevin-bates, line 2322: This implies this will only work for like-versioned notebook servers.  For example, if I'm hosting multiple environments that span different NB server versions, any versions running that do not match the version performing the list operation will have their `nbserver.json` files removed due to the mismatch and essentially be "zombied".

notebook/notebookapp.py
```diff
@@ -2289,14 +2319,16 @@ def list_running_servers(runtime_dir=None):
...
     for file_name in os.listdir(runtime_dir):
         if re.match('nbserver-(.+).json', file_name):
```
kevin-bates, line 2324: In testing, I happened to rename the file to `nbserver-8889.jsonxx` and it still matched.  We should probably make the match a bit more robust.
```suggestion
        if re.match('^nbserver-(.+).json$', file_name):
```

notebook/notebookapp.py
```diff
@@ -2289,14 +2319,16 @@ def list_running_servers(runtime_dir=None):
...
             with io.open(os.path.join(runtime_dir, file_name), encoding='utf-8') as f:
                 info = json.load(f)
 
-            # Simple check whether that process is really still running
+            # Actively check whether that process is really available via an HTTP request
             # Also remove leftover files from IPython 2.x without a pid field
-            if ('pid' in info) and check_pid(info['pid']):
+            response = kernel_request(info, path=url_path_join(info.get('base_url') or '/','/api'))
```
kevin-bates, line 2330: Is the base-url stuff necessary?  It's not used here: https://github.com/jupyter/notebook/pull/5716/files#diff-bbd307804bb6b8836da9f06b8d8a06e9R498
```suggestion
            response = kernel_request(info, path='api')
```

kevin-bates, line 2330: As noted previously, `try/except` should surround this call or have the method handle it.  

Since you might want to distinguish two different behaviors for `GET 'api'` and `POST 'api/shutdown'` and will likely need to call `GET 'api'` multiple times (per my previous comment), I would suggest a `ping_server()` method that just takes the `server_info` and issues the `GET` request, returning `True` or `False` to indicate the server is running or not.

kevin-bates (commented):
Thanks @abaelhe - this is getting much closer.  I'm still slightly concerned about requiring the use of the same notebook version, but I'm not sure that's a very big deal.  It's certainly better than terminating the wrong process!  To that point, and as I noted in the review, we should probably eliminate the PID-based termination altogether for this PR.

abaelhe (commented):
> Thanks @abaelhe - this is getting much closer. I'm still slightly concerned about requiring the use of the same notebook version, but I'm not sure that's a very big deal. It's certainly better than terminating the wrong process! To that point, and as I noted in the review, we should probably eliminate the PID-based termination altogether for this PR.

really useful key point;
i will change the "/api" protocol, define a better policy for jupyter determination;
and i still invite you as the reviewer.

notebook/notebookapp.py
```diff
@@ -2289,16 +2319,19 @@ def list_running_servers(runtime_dir=None):
...
+            try:
+                assert JUPYTER_NOTEBOOK_TAG == json.loads(response.body)["module"]
                 yield info
-            else:
+            except:
```
kevin-bates, line 2334: This is not sufficient for then the request returns successfully, but doesn't match.

notebook/base/handlers.py
```diff
@@ -849,7 +850,7 @@ class APIVersionHandler(APIHandler):
 
     def get(self):
         # not authenticated, so give as few info as possible
-        self.finish(json.dumps({"version":notebook.__version__}))
+        self.finish(json.dumps({"version": notebook.__version__, "module": JUPYTER_NOTEBOOK_TAG}))
```
kevin-bates, line 853: I don't think we can do this as it will break clients.  This is something that would need to be reserved for a major release.

Instead of hitting `/api`, we should hit either `/api/status` - which returns something like the following:
```
{"connections": 0, "kernels": 0, "last_activity": "2020-09-03T14:11:03.989420Z", "started": "2020-09-03T14:10:45.100779Z"}
```
or `/api/spec.yaml` to get the swagger/OpenAPI doc, the first portion of which looks as follows:
```
swagger: '2.0'
info:
  title: Jupyter Notebook API
  description: Notebook API
  version: "5"
  contact:
    name: Jupyter Project
    url: https://jupyter.org
```
I think interpreting either of these results would be a definitive approach that spans versions.

Seems like hitting `/api/status` and confirming "connections" and "kernels" fields are present is probably the most appropriate for this exercise.

abaelhe: clients feel good; you see all tests passed.
also i feel good you can ask so comprehensive consideration.

notebook/notebookapp.py
abaelhe: @kevin-bates , Approved. really good job!

commit 57330ff68ddaac913c615263fb1240acfd74d1c6 (HEAD -> master, origin/master, origin/HEAD)
Author: git <abaelhe@icloud.com>
Date:   Fri Sep 4 00:37:45 2020 +0800

    let user make choices to their data, preference or environment.

commit 908a630761aa849e0824b2710d61560d7df278d3
Author: git <abaelhe@icloud.com>
Date:   Fri Sep 4 00:20:37 2020 +0800

User Notifications for policy change of list_running_servers()

commit a305e0d57f5cced26201a27ec69a892e8107487a
Author: git <abaelhe@icloud.com>
Date:   Fri Sep 4 00:18:59 2020 +0800

    User Notifications for policy change of list_running_servers()

commit 958151a7b8df4f1a0cd13d96c751a961f416ce38
Author: git <abaelhe@icloud.com>
Date:   Thu Sep 3 22:43:55 2020 +0800

abaelhe:
> > Thanks @abaelhe - this is getting much closer. I'm still slightly concerned about requiring the use of the same notebook version, but I'm not sure that's a very big deal. It's certainly better than terminating the wrong process! To that point, and as I noted in the review, we should probably eliminate the PID-based termination altogether for this PR.
> 
> really useful key point;
> i will change the "/api" protocol, define a better policy for jupyter determination;
> and i still invite you as the reviewer.

notebook/notebookapp.py
abaelhe: first stage, i change the core of list_running_servers() to reliable live HTTP request; and keep their behaviors the same;
and previous check_pid() let caller make decisions to exceptions;

suggestion of a ping_server() implemented.

notebook/notebookapp.py
abaelhe: nod

notebook/notebookapp.py
abaelhe: commit 978d883c8cc3a9e7410eafd0b1cf0c5f2f558944
Author: git <abaelhe@icloud.com>
Date:   Thu Sep 3 21:19:26 2020 +0800

    Multi-instance and multi-version of Jupyter Notebook

commit 07570c67bb6c84194298907aea9253199da2ae3f
Author: Abael He <abaelhe@icloud.com>

notebook/notebookapp.py
abaelhe: reliable list_running_servers() is our this step;
better shutdown design is another step;
let's make our first footprint.

kevin-bates:
Hi @abaelhe - thank you for your patience.

I believe this is getting too complicated for the original goals.  
1. List running servers without using an HTTP request instead of PID.
2. Do not terminate any process that may have reused the PID.

Instead of adding a version to the `/api` request and force upgrades, could we just check that the response succeeded?  I think we can change to use `/api/status` later so useful information can be displayed when listing the servers, but for now, I think we just need to check the success status.

Let's try to keep these changes as simple as possible.  Does that work for you?

abaelhe:
> Hi @abaelhe - thank you for your patience.
> 
> I believe this is getting too complicated for the original goals.

we are on the way to build a reliable ipython/jupyter;

> 1. List running servers without using an HTTP request instead of PID.

this is a right way requires a bit effort;

> 2. Do not terminate any process that may have reused the PID.

no matter reused process is processing a billions transaction or just hello world;
we MUST respect to user's ownner right. especially ipython/jupyter are used to process valuable massive data;

> 
> Instead of adding a version to the `/api` request and force upgrades, could we just check that the response succeeded? I think we can change to use `/api/status` later so useful information can be displayed when listing the servers, but for now, I think we just need to check the success status.
> Let's try to keep these changes as simple as possible. Does that work for you?

the response status is success if response.body marked with JUPYTER_NOTEBOOK_TAG;
we already did that in a reliable and simple "/api" way;
i can't achieve it without you.

notebook/base/handlers.py
kevin-bates, line 853: There's no need to include the tag.  Having a successful response from `/api` and a check of the version is sufficient.  Please revert the changes in this file.

notebook/notebookapp.py
```diff
@@ -71,6 +71,7 @@
     from tornado.netutil import bind_unix_socket
 
 from notebook import (
+    JUPYTER_NOTEBOOK_TAG,
```
kevin-bates, line 74: Per previous comment, this can be removed.

notebook/notebookapp.py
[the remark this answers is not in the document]
kevin-bates: This method needs to change and the default for `path` either removed or updated to `/api`

notebook/notebookapp.py
kevin-bates: Ok. I'm okay removing the kill using the PID in another PR.

notebook/notebookapp.py
```diff
@@ -2294,16 +2325,36 @@ def list_running_servers(runtime_dir=None):
...
+                flush_info_file()
+            else:
+                received_version = version_dict.get('version', '0.0.0')
+                msg_special = "Jupyter Notebook HIGHER VERSION detected: %s, current:%s" %(
+                    received_version, __version__)
+                msg_upgrade = "Jupyter Notebook Upgrade REQUIRED: >=%s, got:%s(info file removed)" % (
+                    __version__, received_version)
+
+                if JUPYTER_NOTEBOOK_TAG != version_dict.get('module'):
+                    raise MultipleInstanceError(msg_upgrade)
```
kevin-bates, line 2350: Because we're not extending the `/api` response to include module, this is no longer necessary.

notebook/notebookapp.py
```diff
@@ -2294,16 +2325,36 @@ def list_running_servers(runtime_dir=None):
...
+                else:
+                    if __version__ < received_version:
```
kevin-bates, line 2352: Let's just check if the versions are different and issue a warning message.  I think its much easier to let the user know that there are multiple versions in use and let them decide which version is what they want.
```suggestion
                    if __version__ != received_version:
```

notebook/notebookapp.py
```diff
@@ -2294,16 +2325,36 @@ def list_running_servers(runtime_dir=None):
...
+                    pass
+
+            version_dict = {}
+            try:
+                version_dict.update(json.loads(response.body.decode()))
+            except:
+                flush_info_file()
+            else:
+                received_version = version_dict.get('version', '0.0.0')
+                msg_special = "Jupyter Notebook HIGHER VERSION detected: %s, current:%s" %(
```
kevin-bates, line 2344:
```suggestion
                msg_special = "WARNING: A different Jupyter Notebook version was detected: %s, current: %s" %(
```

kevin-bates (commented):
Thank you @abaelhe.  There is no need to alter the REST API for this change.  Let's just use the fact that a response was successful and issue a message if the version strings don't match.

I still ask that the name and signature of `kernel_request()` be changed.  Thank you.

notebook/base/handlers.py
abaelhe: actually, "/api" response includes a "version" field is general design;
we MUST choose a feature with higher significance, JUPYTER_NOTEBOOK_TAG;

notebook/notebookapp.py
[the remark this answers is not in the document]
Zsailer: Yeah, I agree with @kevin-bates here. This method **needs** to be renamed to something like `api_request` before we can accept it. It's not requesting a response from a "kernel" (as we define it in the Jupyter ecosystem). It's just a general server API ping.

I also agree, the default path should probably be `/api`.

notebook/base/handlers.py
Zsailer: > actually, "/api" response includes a "version" field is general design;

That's a fair point. Returning a "version" key from an `/api` endpoint might be fairly common. 

Is there a reason we wouldn't default to hitting the `/api/status` endpoint? There is some rich, Jupyter-specific information there...

Zsailer: The reason we're avoiding adding keys to the response model is that changes to the REST API affect many other applications (nteract, jupyterlab, voila, etc.). We should only make these kinds of changes when absolutely necessary. We shouldn't add a key just for this command.

notebook/notebookapp.py
abaelhe: JUPYTER_NOTEBOOK_TAG is positive identification tag;
there are ton's of "Kevin" over the world;
there are many "Kevin Bates" over my Facebook search;
we ONLY wanted you "Kevin Bates";
remember reused pid process may also response "/api" with "version" field.

notebook/notebookapp.py
abaelhe: we MUST conduct users by make policies:
1. we encourage users to keep update with latest version of Jupyter Notebook;
```
     if JUPYTER_NOTEBOOK_TAG != version_dict.get('module'):
         raise MultipleInstanceError(msg_upgrade)
```

2. we encourage users to hack our design and mark with greater version number than official latest version;
```
    if __version__ < received_version:
        print(msg_special)
        #hacked by users ;-)
```

3. forward compatibility(if possible):
```
        __version__ >= received_version and JUPYTER_NOTEBOOK_TAG == version_dict.get('module')
    #users can run multi-version of Jupyter Notebook if it marked with JUPYTER_NOTEBOOK_TAG
```

notebook/notebookapp.py
abaelhe: JUPYTER_NOTEBOOK_TAG required as identification tag

notebook/notebookapp.py
abaelhe: we encourage user's hack version ;-)
"higher" is better that "different" for suck kind of users;
higher version !

notebook/notebookapp.py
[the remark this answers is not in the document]
abaelhe: who create new who own naming rights.
answered.

notebook/notebookapp.py
abaelhe: "api/status" is an authenticated api.
kernel_request() MUST retrieve Jupyter Notebook identification tag before or after authentication.

notebook/notebookapp.py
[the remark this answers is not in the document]
Zsailer: The name matters here. It obscures this library's API if function names are not clear. 

`kernel_request` is the wrong name for this function. In the Jupyter ecosystem, kernel refers specifically to the underlying language kernel being executed by calls to the server's `kernel` API. This method is not making a kernel request. It's making a generic request the server's REST API. Let's generalize this function name to something like `server_request` or `api_request`. 


particularly because Jupyter has a notion of kernels that is specific.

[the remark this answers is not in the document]
Zsailer: I'm setting this conversation to unresolved, since this change is necessary.

Zsailer:
> kernel_request() MUST retrieve Jupyter Notebook identification tag before or after authentication.
and "api/status" is an AUTHENTICATED ONLY call.

Ah, I see. You're right. That is an issue. I agree with you, the `/api` endpoint should probably return more specific information about the application (namely, Jupyter Notebook).

Here's the problem... changing the response model is a significant REST API **breaking** change. Such changes take longer to review for us (at Jupyter). As I mentioned before, this type of change affects multiple other applications. In order to accept it, we would have to get buy-in from the larger community. That could take weeks (or even months). 

What you're trying to achieve in this PR is a rather small change (though much appreciate!!). To move forward, I see two possible choices:
1) Drop the API breaking change to the response model in this PR and merge the `api_request` function that pings the server to verify it's running when `list_running_servers` is called. Then, sometime in a later PR, we can lobby to change the `/api` endpoint and improve this function further.
2) Pause any further review on this PR, propose the change to the REST API through the [Jupyter Enhancement Proposals](https://github.com/jupyter/enhancement-proposals) repository, wait for approval there, then return to this PR once everything is approved. 

Unfortunately, I (and likely @kevin-bates) don't have the bandwidth to shepherd (2) through the process, but I'd be happy to review a proposal over there. In full disclosure, (2) will likely take much longer to happen. 

If you decide (1) is the direction you want to go, I'm happy to merge this after the final review. 

Thanks again, @abaelhe, for all your work here. It's great stuff!

notebook/notebookapp.py
Zsailer: Thanks, @abaelhe. See my comment [here](https://github.com/jupyter/notebook/pull/5716#issuecomment-687305368).

notebook/notebookapp.py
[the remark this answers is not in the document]
Zsailer:
```suggestion
def api_request(server_info, path='/api', method='GET', body=None, headers=None, timeout=5, log=None):
```

Zsailer (changes requested):

abaelhe (commented):
> > kernel_request() MUST retrieve Jupyter Notebook identification tag before or after authentication.
> > and "api/status" is an AUTHENTICATED ONLY call.
> 
> Ah, I see. You're right. That is an issue. I agree with you, the `/api` endpoint should probably return more specific information about the application (namely, Jupyter Notebook).
> 
> Here's the problem... changing the response model is a significant REST API **breaking** change. Such changes take longer to review for us (at Jupyter). As I mentioned before, this type of change affects multiple other applications. In order to accept it, we would have to get buy-in from the larger community. That could take weeks (or even months).

add one JUPYTER_NOTEBOOK_TAG feature is transparent to those who using "verson" field from "/api";
i don't see that is a break to REST API  response model; they are compatible.

> What you're trying to achieve in this PR is a rather small change (though much appreciate!!). To move forward, I see two possible choices:

less code, better product; more user.so,
focus on our main road;

> 
> 1. Drop the API breaking change to the response model in this PR and merge the `api_request` function that pings the server to verify it's running when `list_running_servers` is called. Then, sometime in a later PR, we can lobby to change the `/api` endpoint and improve this function further.
> 2. Pause any further review on this PR, propose the change to the REST API through the [Jupyter Enhancement Proposals](https://github.com/jupyter/enhancement-proposals) repository, wait for approval there, then return to this PR once everything is approved.
> 
> Unfortunately, I (and likely @kevin-bates) don't have the bandwidth to shepherd (2) through the process, but I'd be happy to review a proposal over there. In full disclosure, (2) will likely take much longer to happen.
> 
> If you decide (1) is the direction you want to go, I'm happy to merge this after the final review.
> 
> Thanks again, @abaelhe, for all your work here. It's great stuff!

thanks to customers tolerantly using Jupyter Notebook with previous check_pid().

systematic design principle to flexible and diverse methods; shield to spear; they are all unity.
Apple's Logo keeps one bite in history;
why not place JUPYTER_NOTEBOOK_TAG at "/api" or somewhere required; for us, for users.

we break DOS into Windows; RPC into REST API;  and many things.
we keep break REST API if that is reasonable;

merge into master branch first;
JUST DO IT!

Zsailer:
>we break DOS into Windows; RPC into REST API; and many things.
we keep break REST API if that is reasonable;
> 
> merge into master branch first;
> JUST DO IT!

This isn't how we do things in the Jupyter ecosystem, for better or worse. Jupyter's REST API is an old protocol that many applications depend on. Changes require formal review through the enhancement proposal process.

In this case, we need to follow Jupyter's process. If you want to propose such a change, we'd appreciate if you'd open the discussion on the [enhancement-proposals](https://github.com/jupyter/enhancement-proposals) repo. I am guessing you'll get plenty of support for this idea and we'll be able to move forward fairly quickly. (I just don't have the bandwidth to champion this change myself.)

As this PR stands, we can't accept it until that process is complete **or** we drop the changes to the REST API.

jtpio:
Closing since this targets the `6.4.x` branch, which is not supported anymore.

If you think this is still needed and would like to continue the discussion, please open a new issue in https://github.com/jupyter/nbclassic or https://github.com/jupyter-server/jupyter_server.

Thanks all!

PR #5730, opened 2020-09-04
Fix typo for Check in zh_CN

bdbai:
`检查`(check) and `检出`(check out) sound similar, but their meaning is different. In this context, it should be `检查`.

### DIFF
--- a/notebook/i18n/zh_CN/LC_MESSAGES/nbjs.po
+++ b/notebook/i18n/zh_CN/LC_MESSAGES/nbjs.po
@@ -1076,7 +1076,7 @@ msgstr "没有到服务的连接"
 
 #: notebook/static/notebook/js/notificationarea.js:160
 msgid "A connection to the notebook server could not be established. The notebook will continue trying to reconnect. Check your network connection or notebook server configuration."
-msgstr "到后台服务的连接没能建立, 我们会继续尝试重连, 请检查网络连接...还有服务配置."
+msgstr "到后台服务的连接没能建立, 我们会继续尝试重连, 请检查网络连接还有服务配置."
 
 #: notebook/static/notebook/js/notificationarea.js:165
 msgid "Connection failed"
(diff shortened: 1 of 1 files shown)


bdbai:
@kevin-bates thanks for approval!

> I'm curious if the ellipses ('`...`') are necessary? I'm approving anyway, just curious.

I don't think the ellipses here are necessary. Actually they give me a sense of uncertainty or hesitation, just like how we feel when they appear in English.

kevin-bates:
@bdbai - would it make sense to go ahead and remove them since that hesitation seems unnecessary here?

bdbai:
@kevin-bates done. There are many more ellipses in this translation. Do you think it is necessary to remove them as well?

kevin-bates:
Thank you for the update and offer.  It looks like many of the remaining ellipses are trailing - where hesitation is appropriate (e.g., "Restoring to checkpoint..."), but I listed the two additional translations where the English string does not have ellipses while the translation does.  If you could review these, that would be greatly appreciated...

- [msgid "Large file size warning"](https://github.com/jupyter/notebook/blob/master/notebook/i18n/zh_CN/LC_MESSAGES/nbjs.po#L1832-L1834)
msgstr "请注意文件大小..."
- [msgid "File names must be at least one character and not start with a period"](https://github.com/jupyter/notebook/blob/master/notebook/i18n/zh_CN/LC_MESSAGES/nbjs.po#L1961-L1963)
msgstr "文件名不能为空...并且不能以句号开始,除下划线以外的符号都不能开头..."

bdbai:
Sorry for the late reply because I am currently on a trip.

While reviewing other phrases, I also found some flaws in this translate, such as wrong usage of half-width punctuation (should be full width for Chinese sentences) and lack of spaces between Chinese characters and English or numbers. However, they are out of the scope of this issue. Would you mind my opening another issue after fixing them?

kevin-bates:
That sounds like the right approach - thank you.

PR #5733, opened 2020-09-04
Require a file name in the "Save As" dialog

### LINKED ISSUE #5732: Require a file name in the "Save As" dialog
I just had an issue today where I clicked through the "Save As" dialog without typing in a file name. Jupyter helpfully takes that empty string, attaches its extension, and so creates the file `.ipynb` in the named directory. That file is hidden, so you have a hell of a time finding it!

I believe the fix has to happen here:
https://github.com/jupyter/notebook/blob/7978c0f763391a3336cc22273c684dcb00860008/notebook/static/notebook/js/notebook.js#L2920

Another option is to add the "required" attribute to the input element used in that dialog, so the browser takes care of it for you, here:
https://github.com/jupyter/notebook/blob/7978c0f763391a3336cc22273c684dcb00860008/notebook/static/notebook/js/notebook.js#L2906

pavpanchekha:
This PR fixes  #5732 by checking for empty file names in the "Save As" dialog and refusing to save the file. It also helpfully highlights the instructions to the user in red.

It would be better if the error message were clearer, or even better if the save button were disabled until a file name was entered, but that would require a much larger and more complex change to the code. (Maintainers, feel free to take this patch and run with it!)

### DIFF
--- a/notebook/static/notebook/js/notebook.js
+++ b/notebook/static/notebook/js/notebook.js
@@ -2920,6 +2920,14 @@ define([
                     click: function() {
                         var nb_path = d.find('input').val();
                         var nb_name = nb_path.split('/').slice(-1).pop();
+                        if (!nb_name) {
+                            $(".save-message").html(
+                                    $("<span>")
+                                        .attr("style", "color:red;")
+                                        .text($(".save-message").text())
+                                );
+                            return false;
+                        }
                         // If notebook name does not contain extension '.ipynb' add it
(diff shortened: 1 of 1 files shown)


kevin-bates (approved):
This tests out great - thank you.  I'm going to ask another maintainer to review the changes.

kevin-bates:
Steve, I'm hoping you can take a look at the changes.  They test out fine on my end. Thank you.

PR #5743, opened 2020-09-10
Make sure notebook is pip installed when building docs

### LINKED ISSUE #5741: Instructions on building the docs don't mention that `notebook` needs to be installed as well
I was following the instructions on building the documentation [1] using the pip install version (not conda). I was doing it within a virtualenv called `.docs`. I pip installed `docs/doc-requirements.txt` per the instructions and then ran `make html` which failed because `notebook` isn't installed:

```console
(.docs) osboxes@osboxes:~/jupyter/notebook/docs$ make html
python3 autogen_config.py
  File "autogen_config.py", line 4, in <module>
    from notebook.notebookapp import NotebookApp
ModuleNotFoundError: No module named 'notebook'
Makefile:62: recipe for target 'source/config.rst' failed
make: *** [source/config.rst] Error 1
```

The fix is simple, I just did `pip install .` from the root of the notebook project directory and that got me past that issue, but assume it should be called out in the docs as one of the steps in building the documentation.

[1] https://github.com/jupyter/notebook/blob/master/CONTRIBUTING.rst#building-the-documentation

mriedem:
The instructions on building the docs fail to mention that
the notebook package itself needs to be installed.

This updates the instructions to pip install the local notebook
package include the `docs` dependencies using the `docs` extra
install package list.

Closes #5741

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -176,7 +176,8 @@ containing all the necessary packages (except pandoc), use::
 
     conda env create -f docs/environment.yml
     conda activate notebook_docs  # Linux and OS X
-    activate notebook_docs         # Windows
+    activate notebook_docs        # Windows
+    pip install .
 
 .. _conda environment:
     https://conda.io/docs/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -174,17 +174,17 @@ To build the documentation you'll need `Sphinx <http://www.sphinx-doc.org/>`_,
 To install (and activate) a `conda environment`_ named ``notebook_docs``
 containing all the necessary packages (except pandoc), use::
 
-    conda env create -f docs/environment.yml
+    conda create -n notebook_docs pip
     conda activate notebook_docs  # Linux and OS X
     activate notebook_docs        # Windows
-    pip install .
+    pip install .[docs]
 
 .. _conda environment:
     https://conda.io/docs/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file
 
 If you want to install the necessary packages with ``pip``, use the following instead::
 
-    pip install . -r docs/doc-requirements.txt
+    pip install .[docs]
 
 Once you have installed the required packages, you can build the docs with::
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -171,16 +171,13 @@ Building the Documentation
 To build the documentation you'll need `Sphinx <http://www.sphinx-doc.org/>`_,
 `pandoc <http://pandoc.org/>`_ and a few other packages.
 
-To install (and activate) a `conda environment`_ named ``notebook_docs``
+To install (and activate) a conda environment named ``notebook_docs``
 containing all the necessary packages (except pandoc), use::
 
     conda create -n notebook_docs pip
     conda activate notebook_docs  # Linux and OS X
     activate notebook_docs        # Windows
     pip install .[docs]
-
-.. _conda environment:
-    https://conda.io/docs/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file
 
 If you want to install the necessary packages with ``pip``, use the following instead::
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -183,7 +183,7 @@ containing all the necessary packages (except pandoc), use::
 
 If you want to install the necessary packages with ``pip``, use the following instead::
 
-    pip install -r docs/doc-requirements.txt
+    pip install . -r docs/doc-requirements.txt
 
 Once you have installed the required packages, you can build the docs with::
 
--- a/setup.py
+++ b/setup.py
@@ -118,7 +118,8 @@ for more information.
     extras_require = {
         'test': ['nose', 'coverage', 'requests', 'nose_warnings_filters',
                  'nbval', 'nose-exclude', 'selenium', 'pytest', 'pytest-cov'],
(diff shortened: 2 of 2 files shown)


mriedem:
@kevin-bates this fixes the pip-based instructions but doesn't resolve the issue for conda, which I don't have setup locally to test. However, looking at the `conda install` docs it looks like this might be the fix for that?

```diff
diff --git a/CONTRIBUTING.rst b/CONTRIBUTING.rst
index 5d9a2ecd3..3ee8a49fa 100644
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -176,6 +176,7 @@ containing all the necessary packages (except pandoc), use::
 
     conda env create -f docs/environment.yml
     conda activate notebook_docs  # Linux and OS X
+    conda install -n notebook_docs .
     activate notebook_docs         # Windows
 
 .. _conda environment:
```

I'm not sure if `-n notebook_docs` is necessary if you've already activated the `notebook_docs` environment (it wouldn't be with something like an activated virtualenv which I'm used to), so it's just a guess. Feel free to update this PR if you can verify the conda instruction fix.

kevin-bates:
Thanks Matt - I'm not very familiar with conda (and pip for that matter) myself.  In looking into the equivalent of `pip install .` in conda I found there isn't really any other than to be sure to `conda install pip` first, then you can issue `pip install .`.

Since `pip` is installed by virtue of creating the environment with the specified file, I think the instructions for conda could be listed as follows...
```
     conda env create -f docs/environment.yml
     conda activate notebook_docs  # Linux and OS X
     activate notebook_docs        # Windows
     pip install .
```
At first, I could not reproduce the issue relative to conda but that was because I had `notebook` installed in my base conda env - which acts as a fallback if my understanding is correct.  Once I removed `notebook` from my base env, then I could reproduce the issue and confirmed that `pip install .` allowed for the docs build to proceed.

If you don't mind adding this to the PR, that would be great.  I'll probably ask for an additional reviewer relative to the conda/pip stuff to be sure no additional notes are necessary or anything like that.

> I'm not sure if -n notebook_docs is necessary if you've already activated the notebook_docs environment

Once activated, all actions are performed against the activated env, no need to name it explicitly.

mriedem:
> If you don't mind adding this to the PR, that would be great.

Done, thanks for investigating.

kevin-bates (approved):
These changes look good and test out for me - thank you!

@blink1073 - I'm asking for your review with the hope you can sanity-check my [comment](https://github.com/jupyter/notebook/pull/5743#issuecomment-690760146) to ensure we don't need additional text regarding the use of `pip` within a conda env.  Thank you.

Zsailer:
With the new `extra_requires` arg in the `setup.py`, you should be able to just call:
```
pip install .[docs]
```
This will install `notebook`, its dependencies, and the docs dependencies.

It's a bit cleaner than installing from the `requirements.txt` files. We might consider dropping these files altogether.

Zsailer:
> It's a bit cleaner than installing from the requirements.txt files. We might consider dropping these files altogether.

We might consider doing the same for `docs/environment.yml`.

Instead, we could suggest the following chunk:
```bash
conda create -n notebook_docs pip
conda activate notebook_docs         # Linux and OS X
activate notebook_docs               # Windows
pip install .[docs]
```

mriedem:
> With the new `extras_require` arg in the `setup.py`, you should be able to just call:

Cool, I didn't realize that existed. That would make the pip-based setup instructions for building docs simpler.

One question, I noticed that doc-requirements.txt includes `sphinx-rtd-theme` but the `docs` extra in setup.py doesn't. Is it missing from setup.py or irrelevant in doc-requirements.txt? I'm assuming maybe the former since it's also listed in the conda docs/environment.yml.

Let me know if you'd like me to change this PR to use `pip install .[docs]` and remove docs/requirements.txt. Note that travis config would also need to be updated:

https://github.com/jupyter/notebook/blob/master/.travis.yml#L35

mriedem:
> Let me know if you'd like me to change this PR to use `pip install .[docs]` and remove docs/requirements.txt.

Alternatively, we could go with the fix here and follow up with a larger PR to remove docs/environment.yml and docs/requirements.txt.

Zsailer:
We'll need to add the `sphinx-rtd-theme` dependency in `setup.py`. Would you mind adding that in this PR?

Then, let's update the docs to use the `pip install .[docs]` command here in this PR.

We'll save removing the requirements files for a separate PR.

CONTRIBUTING.rst
```diff
@@ -174,16 +174,17 @@ To build the documentation you'll need `Sphinx <http://www.sphinx-doc.org/>`_,
...
 
-    conda env create -f docs/environment.yml
+    conda create -n notebook_docs pip
     conda activate notebook_docs  # Linux and OS X
-    activate notebook_docs         # Windows
+    activate notebook_docs        # Windows
+    pip install .[docs]
 
 .. _conda environment:
     https://conda.io/docs/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file
```
mriedem, line 183: This could go away since we're no longer using it.

kevin-bates:
Thank you @Zsailer - I agree that's much cleaner and essentially alleviates the need for Steve's review, so I'm going to remove that request.  Thanks for your help!

CONTRIBUTING.rst
Zsailer: :+1:

Zsailer (approved):

kevin-bates:
Hi Matt - sorry about the delay here.  Would you mind resolving the conflict in setup.py, then we'll get this merged?  Thanks.

mriedem:
> 
> 
> Hi Matt - sorry about the delay here. Would you mind resolving the conflict in setup.py, then we'll get this merged? Thanks.

Done, thanks for the heads up.

kevin-bates (approved):
LGTM - thank you.

Issue #4862, opened 2019-09-04
notebook.utils._check_pid_win32() is leaking handles and returns True for an already terminated process


garu57:
notebook  6.0.1

MSDN doc for [OpenProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess) says:

> When you are finished with the handle, be sure to close it using the [CloseHandle ](https://docs.microsoft.com/windows/desktop/api/handleapi/nf-handleapi-closehandle) function.

MSDN doc for [CloseHandle](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle) says:

> To remove a process object, you must terminate the process, then close all handles to the process. For more information, see [Terminating a Process](https://docs.microsoft.com/windows/desktop/ProcThread/terminating-a-process).

This means that each time that `_check_pid_win32()` is called it leaks an opened handle on the kernel process object that thus can't be freed even if the executing process is terminated.

This also means that since the kernel process object will still be around, `_check_pid_win32()` allocates a new handle for each run, failing to detect that the queried process is already terminated and `jupyter notebook stop` will always return the message `Could not stop server`.

Closing the handle allows the kernel process object to be freed and solves the problem:

```
def _check_pid_win32(pid):
    import ctypes
    # OpenProcess returns 0 if no such process (of ours) exists
    # positive int otherwise
    handle = ctypes.windll.kernel32.OpenProcess(1,0,pid)
    if handle:
        # the handle must be closed or the kernel process object won't be freed
        ctypes.windll.kernel32.CloseHandle( handle )
    return bool(handle)

```

PR #5789, opened 2020-10-02
Fix _check_pid_win32

mueslo:
Previous implementation was broken and would return True even for ended processes.

However, I think `notebook` should just use https://github.com/giampaolo/psutil instead of rolling its own.

### DIFF
--- a/notebook/utils.py
+++ b/notebook/utils.py
@@ -284,13 +284,15 @@ def check_version(v, check):
         return True
 
 
-# Copy of IPython.utils.process.check_pid:
-
 def _check_pid_win32(pid):
     import ctypes
     # OpenProcess returns 0 if no such process (of ours) exists
     # positive int otherwise
-    return bool(ctypes.windll.kernel32.OpenProcess(1,0,pid))
+    handle = ctypes.windll.kernel32.OpenProcess(1,0,pid)
+    if handle:
+        # the handle must be closed or the kernel process object won't be freed
+        ctypes.windll.kernel32.CloseHandle( handle )
+    return bool(handle)
 
 def _check_pid_posix(pid):
     """Copy of IPython.utils.process.check_pid"""


mueslo:
I don't think the failed test has anything to do with my change, as the code does not call _check_pid_win32

kevin-bates:
Hi @mueslo - thank you for contributing.  I agree that the test failure is likely not related - let's not worry about at the moment.

I must say that I'm not completely enamored with this approach as it does seem to lead to COM monikers and co-initialization, etc. (This has triggered flashbacks to 25 years ago. :smile: )  I also ran into this [SO question](https://stackoverflow.com/questions/38860185/python-cant-import-wmi-under-special-circumstance) and am a little uneasy about this biting us in various scenarios that will inevitably creep up.

I'm wondering if you might be getting bit by this [leaked handle issue](https://github.com/jupyter/notebook/issues/4862) that was posted a while back.  Would you be willing to try out that suggested change first?

mueslo:
Sadly, this seems to be difficult to reproduce.

> 
> 
> Hi @mueslo - thank you for contributing. I agree that the test failure is likely not related - let's not worry about at the moment.
> 
> I must say that I'm not completely enamored with this approach as it does seem to lead to COM monikers and co-initialization, etc. (This has triggered flashbacks to 25 years ago. 😄 ) I also ran into this [SO question](https://stackoverflow.com/questions/38860185/python-cant-import-wmi-under-special-circumstance) and am a little uneasy about this biting us in various scenarios that will inevitably creep up.
> 
> I'm wondering if you might be getting bit by this [leaked handle issue](https://github.com/jupyter/notebook/issues/4862) that was posted a while back. Would you be willing to try out that suggested change first?

I tried it, but the problem is that I currently cannot reproduce the issue I was facing where `ctypes.windll.kernel32.OpenProcess(1,0,pid)` would return a handle despite no such process existing.

kevin-bates:
I see - thank you.

I think I'd feel better about this if we take the approach you propose but fall back to the current implementation (along with the change to close the handle) within an `except` block.  Does that sound okay?

mueslo:
I think simply adding the handle check should be good enough. `psutil` also makes sure to close the handle in its usage: https://github.com/giampaolo/psutil/blob/54de8823918fc0484223a077eafc917755334b5b/psutil/_psutil_windows.c#L151-L186

I will continue to run an unpatched version for now. If it occurs the next time, I will try to close the handle I receive, and then re-run OpenProcess. However, if some *other* process is leaking handles except jupyter, you would have to 1. get the handle, 2. close the handle, 3. try to get the handle again (and 4. close it if it succeeds)...

mueslo:
https://github.com/giampaolo/psutil/blob/994c429c010049a6a0556ce3b0d1af1f86f27867/psutil/arch/windows/process_utils.c#L77-L79 very relevant

mueslo:
Okay, so I've managed to partially reproduce it.

1. `nbopen /path1/notebook1.ipynb`
2. `nbopen /path2/notebook2.ipynb`
3. kill notebook server 1
4. `jupyter notebook list` shows both notebooks running

However, this does not fully reproduce my problem. I had *no* running notebooks left (which could be holding onto the handle). In the above case, it's obviously the Python process from step 2 which has a reference to the handle. In my case, I had killed all Python processes and it still would not work correctly.

Anyway, this specific case *is* fixed by releasing the handle properly.

mueslo:
See also #5791 for a possible psutil change

kevin-bates (approved):
This change looks good - thank you!

kevin-bates:
Hi @mueslo - thank you for the contribution.  I think this change is sufficient over the psutil-based change.

PR #5791, opened 2020-10-02
replace check_pid and variants with psutil library

mueslo:
Improvements:

* a lot more tested and resilient (e.g. _check_pid_win32 has bugs)
* ensures not just that nbserver has the same PID but also was created at the same time (PID may be reused)

Drawback:

* adds a new dependency

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -40,6 +40,8 @@ except ImportError:
     # Windows
     resource = None
 
+import psutil
+
 from base64 import encodebytes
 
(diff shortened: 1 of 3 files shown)


kevin-bates:
Hi @mueslo - thank you for all the work on this and #5789.  I believe the latter is sufficient.   This approach changes the server-info structure (which I'm not sure will have side-affects) and introduces another package.  Given you've found closing the handle resolves the known issue, let's stick with that for now. Thank you.

I'm going ahead with this closure.  Should we find something more than closing the handle we can reopen at that time.

PR #5805, opened 2020-10-10
Remove unnecessary future imports

rahulporuri:
This PR removes unnecessary `__future__` imports specifically `print_function` and `absolute_import` imports which are unnecessary on Python 3.


kevin-bates (approved):
LGTM - thank you.

PR #5806, opened 2020-10-10
Update super usage

rahulporuri:
This commit updates the super usage. Because Python 2 is not supported anymore, super usage can be updated such that super is called without any arguments in the default case where super is called with the class name and self e.g. `super(NotebookApp, self)` called within `class NotebookApp(...):`.

The changes in this PR were done using automated regex-based search and replace. The regex used was - `(super)+\(+(\w+)(,)+\s(self)+\)+`. After the automated search and replace happened, each change was checked individually before being staged in git.

Note that all usage of super has not been updated - a few cases which smelled funny have been ignored. The two uses of `super` that smelled funny can be found below 

<details>

```
diff --git a/notebook/base/zmqhandlers.py b/notebook/base/zmqhandlers.py
index dc3a7b69e..7b002c960 100644
--- a/notebook/base/zmqhandlers.py
+++ b/notebook/base/zmqhandlers.py
@@ -165,7 +165,7 @@ class WebSocketMixin(object):
                 self.send_ping, self.ping_interval,
             )
             self.ping_callback.start()
-        return super(WebSocketMixin, self).open(*args, **kwargs)
+        return super().open(*args, **kwargs)

     def send_ping(self):
         """send a ping to keep the websocket alive"""
@@ -201,7 +201,7 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
         """Backport send_error from tornado 4.1 to 4.0"""
         def send_error(self, *args, **kwargs):
             if self.stream is None:
-                super(WebSocketHandler, self).send_error(*args, **kwargs)
+                super().send_error(*args, **kwargs)
             else:
                 # If we get an uncaught exception during the handshake,
                 # we have no choice but to abruptly close the connection.
```

</details>

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -301,7 +301,7 @@ class IPythonHandler(AuthenticatedHandler):
     
     def set_default_headers(self):
         """Add CORS headers, if defined"""
-        super(IPythonHandler, self).set_default_headers()
+        super().set_default_headers()
         if self.allow_origin:
             self.set_header("Access-Control-Allow-Origin", self.allow_origin)
         elif self.allow_origin_pat:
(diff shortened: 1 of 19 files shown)


kevin-bates (approved):
These changes look good.  Thank you for the great write-up.  I agree with your assessment and approach for the two areas that were left unchanged.

PR #5807, opened 2020-10-10
Remove code for unsupported python versions

rahulporuri:
This PR removes dead code, specifically code which was handling python 2.x and < 3.4.

Python 3.5 is also past EoL now but I am not sure if the package dropped support for Python 3.5 already so I have made no 3.5 related changes in this PR.


kevin-bates (approved):
Looks good - thank you.  Yeah, although 3.5 is EOL, I think it would be good to hold off on its complete removal given the amount of folks using notebook.  Thanks for your contribution.

PR #5808, opened 2020-10-11
Regression Models

Lakshaypahuja21:
The following file contains regression models of google colab file. The same code can be run on jupyter notebook.


kevin-bates:
There is no precedent for this in this repo to which I'm aware.  Working notebooks, as well as examples, tend to be kept in external repositories.  Please check with the Google Colab team for how you can contribute these relative to that project.

PR #5812, opened 2020-10-13
Fix build issue with npm upgrade

lresende:


kevin-bates (approved):
Thanks Luciano.  Not sure why this suddenly started failing, but this update appears to resolve the issues.

PR #5813, opened 2020-10-14
Add support for creating terminals via GET

### LINKED ISSUE #5790: Starting with 6.1, I can't start terminals directly with a GET request
Hello. I am teaching a beginners Operating Systems class (explaining the very basics of Linux to young students). Up to version `6.0.3`, I was able to direct my students to `https://myserver.com:8888/terminals/1` and that'd immediately start a terminal for them.

Starting with `6.1` I can't do that, now it throws a 404. Did the API change and now it's necessary to somehow "create" the terminal before issuing the GET request? Is there any way to circumvent this?

Thanks!

kevin-bates:
This pull request adds back support for creating new terminal sessions via an HTTP `GET` request against `/terminals/new/<name>`.  Upon its creation, the user is redirected to `/terminals/<name>`.  Attempts to issue a `GET` request to `/terminals/new/<name>` when terminal `<name>` is already created will result in a `409` (Conflict) status code.

I based these changes on the comments provided by @rgbkrk and @blink1073 on #4180.

Although this feature was previously "hidden", it sounds like it was quite useful and I believe this should be considered somewhat of a regression issue.  That said, this "fix" results in a change in user behavior, but I'd prefer we allow for that since the feature was not explicit previously.

This seems to address the issue, but this is relatively unfamiliar territory for me, so I'm hoping others can take it for a spin and/or let me know what else should be done.

Resolves #5790

### DIFF
--- a/notebook/terminal/__init__.py
+++ b/notebook/terminal/__init__.py
@@ -10,7 +10,7 @@ if not check_version(terminado.__version__, '0.8.3'):
 from ipython_genutils.py3compat import which
 from notebook.utils import url_path_join as ujoin
 from .terminalmanager import TerminalManager
-from .handlers import TerminalHandler, TermSocket
+from .handlers import TerminalHandler, TermSocket, NewTerminalHandler
 from . import api_handlers
 
 
@@ -45,6 +45,7 @@ def initialize(nb_app):
         (ujoin(base_url, r"/terminals/(\w+)"), TerminalHandler),
         (ujoin(base_url, r"/terminals/websocket/(\w+)"), TermSocket,
              {'term_manager': terminal_manager}),
+        (ujoin(base_url, r"/terminals/new/(\w+)"), NewTerminalHandler),
         (ujoin(base_url, r"/api/terminals"), api_handlers.TerminalRootHandler),
         (ujoin(base_url, r"/api/terminals/(\w+)"), api_handlers.TerminalHandler),
     ]
--- a/notebook/terminal/__init__.py
+++ b/notebook/terminal/__init__.py
@@ -10,7 +10,7 @@ if not check_version(terminado.__version__, '0.8.3'):
 from ipython_genutils.py3compat import which
 from notebook.utils import url_path_join as ujoin
 from .terminalmanager import TerminalManager
-from .handlers import TerminalHandler, TermSocket, NewTerminalHandler
+from .handlers import TerminalHandler, TermSocket, NewTerminalHandler, NamedTerminalHandler
 from . import api_handlers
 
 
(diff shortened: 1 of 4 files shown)


santiagobasulto:
Thanks for looking into this @kevin-bates 🙏! Could it be that, instead of returning 409, it returns a 302 with Location `/terminals/<name>`. In our case, we just put the link to the terminal in Markdown (hardcoded we could say), so for our students it'd work on the first try, but not after that.

kevin-bates:
Good point. This improves compatibility with the previous functionality.  I've pushed an update.

notebook/terminal/handlers.py
```diff
@@ -18,6 +19,21 @@ def get(self, term_name):
                    ws_path="terminals/websocket/%s" % term_name))
 
 
+class NewTerminalHandler(IPythonHandler):
+    """Renders a new terminal interface using the named argument."""
+    @web.authenticated
+    def get(self, term_name):
+        new_path = self.request.path.replace("new/{}".format(term_name), term_name)
```
rgbkrk, line 26: If the `term_name` isn't set (i.e. they visit `/new`), can we have it create a new named one?

notebook/terminal/terminalmanager.py
```diff
@@ -44,6 +44,16 @@ def __init__(self, *args, **kwargs):
...
         """Create a new terminal."""
         name, term = self.new_named_terminal()
+        return self._finish_create(name, term)
+
+    def create_with_name(self, name):
+        """Create a new terminal."""
+        if name in self.terminals:
+            raise web.HTTPError(409, "A terminal with name '{}' already exists.".format(name))
+        term = self.get_terminal(name)
+        return self._finish_create(name, term)
```
rgbkrk, line 54: 👍

notebook/terminal/handlers.py
kevin-bates: Good point.  I added a commit supporting this.  This does preclude someone from creating their own named terminal of 'new' (via `GET /terminals/new/new`), but that's probably okay.

notebook/terminal/handlers.py
```diff
@@ -14,8 +15,31 @@ class TerminalHandler(IPythonHandler):
     """Render the terminal interface."""
     @web.authenticated
     def get(self, term_name):
-        self.write(self.render_template('terminal.html',
-                   ws_path="terminals/websocket/%s" % term_name))
+        if term_name == 'new':
+            model = self.terminal_manager.create()
+            term_name = model['name']
+            new_path = self.request.path.replace("terminals/new", "terminals/" + term_name)
+            self.redirect(new_path)
```
rgbkrk, line 22: Good approach. It might be good to make this an explicit path in the handlers above with `(ujoin(base_url, r"/terminals/new"), ReallyActuallyNewTerminalHandler),`, noting that you may have to make it the first entry in the list (and figuring out a better name for that).

rgbkrk (approved):

rgbkrk:
I'm going to try rebuilding the PR, as it doesn't seem like your work should be causing the failure AppVeyor is reporting.

kevin-bates:
Thank you. I'm also working on a commit to address your last comment - which does clean things up a bit.

kevin-bates:
Hmm - odd. AppVeyor is successful in my fork - except for one of the commits - but not the one that failed above:
image

I'm not seeing any obvious side effects that I may have introduced as the two additional tests ensure all terminals have been stopped - but it is Windows!

notebook/terminal/__init__.py
```diff
@@ -42,6 +42,8 @@ def initialize(nb_app):
     terminal_manager.log = nb_app.log
     base_url = nb_app.web_app.settings['base_url']
     handlers = [
+        (ujoin(base_url, r"/terminals/new"), NamedTerminalHandler),
```
rgbkrk, line 45: ❤️

rgbkrk (approved):

rgbkrk:
Thank you so much @kevin-bates!

PR #5818, opened 2020-10-18
Error 404!

singh728om:


kevin-bates:
Hi @singh728om - thank you for your pull request.  However, this doesn't appear to add any value since the current 404 page already indicates the error:

Screen Shot 2020-10-19 at 6 59 41 AM

and adding similar information as an emphasis of such a common error code, isn't an improvement:

Screen Shot 2020-10-19 at 7 02 46 AM

As a result, I'm closing this pull request.

PR #5826, opened 2020-10-22
do not require nose for testing

pgajdos:
`nose` is deprecated and will not work in future python versions.

### DIFF
--- a/notebook/tests/test_paths.py
+++ b/notebook/tests/test_paths.py
@@ -1,6 +1,5 @@
 
 import re
-from nose.tools import assert_regex, assert_not_regex
 
 from notebook.base.handlers import path_regex
 
@@ -16,7 +15,9 @@ def test_path_regex():
         '/x/foo/bar',
         '/x/foo/bar.txt',
     ):
-        assert_regex(path, path_pat)
+        print(type(path))
+        print(type(path_pat))
+        assert re.match(path_pat, path)
 
 def test_path_regex_bad():
     for path in (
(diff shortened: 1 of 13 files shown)


Zsailer (approved):

notebook/tests/test_paths.py
```diff
@@ -16,7 +15,9 @@ def test_path_regex():
         '/x/foo/bar',
         '/x/foo/bar.txt',
     ):
-        assert_regex(path, path_pat)
+        print(type(path))
+        print(type(path_pat))
```
kevin-bates, line 19: We should probably remove these `print` statements.

kevin-bates (commented):
These changes look good - thank you!  Only had the one comment regarding the print statements.

kevin-bates (approved):
Thanks @pgajdos.

hroncok:
Note that nose appears still required via https://github.com/ipython/ipython_genutils/blob/fbcb174d06d6df1f8633014dd80fad8c45533c2e/ipython_genutils/testing/decorators.py#L162

kevin-bates:
Good catch @hroncok - thanks.  Looks like the updates to `notebook/tests/test_notebookapp_integration.py` were missed.  Any idea how `nose` is getting into the CI?  I'm not seeing anything immediately, but it's definitely present in the `pip freeze` output.

At any rate, PR forthcoming to address the remaining references that pull in `nose`.

hroncok:
AFAIK nose comes pre-installed in Travis Python environments (I guess historically this was a good idea, not sure).

PR #5838, opened 2020-10-31
Update README.md

sarveswar1:


README.md
```diff
@@ -46,6 +46,10 @@ For a local installation, make sure you have
 [pip installed](https://pip.readthedocs.io/en/stable/installing/) and run:
 
     $ pip install notebook
+             or
+    $ pip install jupyter
+             or
+    $ python -m pip install jupyter
```
kevin-bates, line 52: I don't think the `python -m` variant is necessary since it would also apply to the original `pip` statement as well.

kevin-bates (commented):
Thank you @sarveswar1.  I think adding the one variant is sufficient (and helpful) - thank you.

README.md
kevin-bates: After reading up on when `python -m` should be used, and, again, since it would also apply to `pip install notebook`, I think converting the latter example into more of a comment, might be preferred, leaving something like:
```
$ pip install notebook
    or
$ pip install jupyter
```    
`Note: if multiple python environments are present, it is recommended that each of the above commands 
be prefixed with `python -m` (e.g., `python -m pip install notebook`) to ensure the python used for `pip` 
is the currently activated python.`

Nimplex (approved):

Zsailer:
Hm, I think this addition is out-of-scope for the README of this repository. 

Using `pip install jupyter` installs a handful of other Jupyter packages beyond `notebook`, namely `qtconsole` and `jupyter-console`. The README of this package should only focus on installing `notebook`.

Zsailer:
Closing, given the previous comment.

PR #5845, opened 2020-11-04
Add wasm mimetype override

bollwyvl:
Some older pythons' (all 3.6, at least 3.7.0, but not the latest 3.7.8) `mimetypes` don't guess `.wasm` (WebAssembly) files correctly, which causes problems for extensions that wish to use this technology, as browsers will not interpret them if not served with the correct `Content-Type`. Since we're already adding types for other common files, it seems reasonable to do so here.

Ref: https://github.com/jupyter/jupyter_server/pull/328

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2000,6 +2000,8 @@ class NotebookApp(JupyterApp):
         # ensure css, js are correct, which are required for pages to function
         mimetypes.add_type('text/css', '.css')
         mimetypes.add_type('application/javascript', '.js')
+        # for python <3.8
+        mimetypes.add_type('application/wasm', '.wasm')
 
     def shutdown_no_activity(self):
         """Shutdown server on timeout when there are no kernels or terminals."""


kevin-bates (approved):
Thanks Nick.  I saw the same change merged in server yesterday.  We appreciate your update here as well.

bollwyvl:
Thanks for the speedy review!

I definitely see a future where wasm plays a _major_ role in Jupyter. While there's still some [HR shuffling](https://bytecodealliance.org/articles/1-year-update#the-lucet-and-wasmtime-teams-join-forces) going on, there's some weight by some significant players to do some really cool stuff.

Waxing a bit more poetic than my flippant postmortem about browser-based stuff on that other PR: I feel like wasm may actually deliver some of the things we've dreamed about, and have had to realize suboptimally with  increasing complexity (read: docker)... a good example is [enarx](https://github.com/enarx/enarx/wiki/Enarx-Introduction) which doesn't take too much squinting to see as a fantastic way to achieve the dream of secure-by-default kernels.

The most ludicrous end state of reproducible, literate computing I see:
- fart around in a jupyter client, get your environment/ui where you want it to be
  - use wasm in your browser, or computer, or deploy to HPC
- hit the "archive" button 
  - dump the entire experience (UI, need wasm runtimes and libraries, data sets)
  - generate a pixel-perfect, archival-grade PDF/A2, and embed the entire experience inside that
- host on dumb web server
  - or put on P2P networks (dat/ipfs) that can deduplicate these enormous things
- profit? 
  - at least SCIENCE!

PR #5869, opened 2020-11-17
Prevent redirect loop if query params are present in url

dubovinszky:
TrailingSlashHandler can cause redirect loop if the URL contains query params. This can happen in Firefox, it appends a redirects=1 after the URL.


Zsailer:
Fixed previously in a security patch [here](https://github.com/jupyter/notebook/security/advisories/GHSA-c7vm-f5p4-8fqh).

PR #5870, opened 2020-11-17
Allow toggling auth for prometheus metrics

yuvipanda:
Equivalent to https://github.com/jupyterhub/jupyterhub/pull/2224

Prometheus metrics can potentially leak information about
the user, so they should be kept behind auth by default.
However, for many JupyterHub deployments, they would need
to be scraped by a centralized Prometheus instance that can not
really authenticate separately to each user notebook without
a lot of work. Admins can use this setting to allow unauthenticated
access to the /metrics endpoint.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -914,8 +914,10 @@ class PrometheusMetricsHandler(IPythonHandler):
     """
     Return prometheus metrics for this notebook server
     """
-    @web.authenticated
     def get(self):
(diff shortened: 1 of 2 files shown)


yuvipanda:
Happy to make an equivalent PR in jupyter_server repo as well - but would really love to see this land here :)

Zsailer (approved):
Makes sense to me. Thanks, @yuvipanda.

PR #5876, opened 2020-11-19
feat: set RLIMIT_AS based on cgroups

caarlos0:
This should help avoid having the kernel oomkilled when running in a docker container with memory limits set.

I'm not well-versed on python, please let me know if I didn't follow any good patterns or if there is a better way of doing this.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1667,6 +1667,16 @@ class NotebookApp(JupyterApp):
             )
             resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
 
+        if os.path.isfile('/sys/fs/cgroup/memory/memory.limit_in_bytes'):
+            old_soft, old_hard = resource.getrlimit(resource.RLIMIT_AS)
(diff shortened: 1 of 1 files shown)


notebook/notebookapp.py
```diff
@@ -1667,6 +1667,16 @@ def init_resources(self):
             )
             resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
 
+        if os.path.isfile('/sys/fs/cgroup/memory/memory.limit_in_bytes'):
```
kevin-bates, line 1670: I've got a couple of questions (and am not very familiar with the workings of container groups).
1. My Linux system also has the file '/sys/fs/cgroup/memory/memory.soft_limit_in_bytes'.  Should that be used to acquire the new soft limit?
2. On my system, the current values for `RLIMIT_AS` are `(-1, -1)` while the file-based values are `(9223372036854771712, 9223372036854771712)`.  I understand wanting to apply the cgroup values when running in a container, but won't this impose a limit for non-containerized configurations that, today, isn't imposed?
3. Should these limits only get applied if we can determine we're running within a container?

PR #5880, opened 2020-11-23
Update handlers.py Remove or correct this useless self-assignment

mofanke:
Remove or correct this useless self-assignment


kevin-bates (approved):
LGTM - thank you.

PR #5889, opened 2020-11-29
Added section for file editor configuration

jmandreoli:
The doc shows how to update codemirror options for the notebook cell editor, but not for the whole file editor. I have added lines 51-62 to give that information. The information comes from Stack-Overflow (and has been tested), but it has been frustratingly long to find, so I think it is worth sharing ...


kevin-bates (approved):
This seems useful to me although I'd prefer to get another review from someone with frontend dev experience (prior to merging).

@jasongrout or @blink1073 could one of you please give this a look?

PR #5893, opened 2020-12-01
example notebook: don't recommend using $$...$$ for display math

mgeier:


kevin-bates (approved):
This looks good Matthias - thank you.

Issue #3374, opened 2018-02-25
send2trash.exceptions.TrashPermissionError


nick-youngblut:
I'm getting the following error when trying to trash (delete) a notebook (Untitled.ipynb):

```
Traceback (most recent call last): 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/send2trash/plat_other.py", 
line 141, in find_ext_volume_fallback_trash check_create(trash_dir) File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/send2trash/plat_other.py", 
line 86, in check_create os.makedirs(dir, 0o700) File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/os.py", 
line 220, in makedirs mkdir(name, mode) PermissionError: [Errno 13] Permission denied: b'/ebio/abt3_projects/.Trash-6354' 

During handling of the above exception, another exception occurred: Traceback (most recent call last): 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/tornado/web.py", line 1512, in _execute result = yield result 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run value = future.result() 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result raise_exc_info(self._exc_info) File "<string>", line 4, in raise_exc_info 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/tornado/gen.py", line 307, in wrapper yielded = next(result) 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/notebook/services/contents/handlers.py", line 235, in delete yield gen.maybe_future(cm.delete(path)) 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/notebook/services/contents/manager.py", line 278, in delete self.delete_file(path) 
File "/ebio/abt3_projects/software/miniconda3_gt4.4/lib/python3.6/site-packages/notebook/services/contents/filemanager.py", line 499, in delete_file send2trash(os_path) 
[...]

PR #5894, opened 2020-12-01
Handle TrashPermissionError, now that it exists

stefanor:
In Ubuntu's automated test environment [full log example](https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-hirsute/hirsute/amd64/j/jupyter-notebook/20201110_184225_68a35@/log.gz), both HOME and the temporary directories are likely to be on a tmpfs. That will cause
`gi.repository.GLib.GError: g-io-error-quark: Trashing on system internal mounts is not supported (15)` which will raise a `TrashPermissionError`.

As the comments in the code indicate, `_check_trash()` was added (in #3304) because `TrashPermissionError` didn't
exist, yet. We asked send2trash to implement such an exception, so that we wouldn't have to guess whether a file was trash-able or not, and they did. So, let's take advantage of it.

Because we don't permanently delete, recursively, `test_delete_non_empty_dir` will now fail in environments where the test data isn't trash-able. So, handle that, and skip the test gracefully.

Closes: #3374

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -15,6 +15,7 @@ import mimetypes
 import nbformat
 
 from send2trash import send2trash
+from send2trash.exceptions import TrashPermissionError
 from tornado import web
 
 from .filecheckpoints import FileCheckpoints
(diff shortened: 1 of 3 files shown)


kevin-bates (approved):
Nice cleanup - thank you.

hendis85 (approved):

PR #5896, opened 2020-12-02
Add an issue template and issue landing page.

Zsailer:
Here's a snapshot of the landing page.

Screen Shot 2020-12-02 at 11 19 16 AM

The goal is to redirect user questions to discourse, feature requests to jupyterlab, and slow down issues posted here that aren't thorough bug reports.


kevin-bates (approved):
Thank you Zach.  This will be really helpful to both our users and maintainers!

PR #5898, opened 2020-12-04
handle allow_origin='*' in check_referer

minrk:
allow_origin can be the wildcard '*' to allow any host

applies logic already in check_origin to check_referer

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -404,6 +404,10 @@ class IPythonHandler(AuthenticatedHandler):
         Used on GET for api endpoints and /files/
         to block cross-site inclusion (XSSI).
         """
+
+        if self.allow_origin == "*" or self.skip_check_origin():
+            return True
+
         host = self.request.headers.get("Host")
         referer = self.request.headers.get("Referer")
 


kevin-bates (approved):
Thanks Min!

notebook/base/handlers.py
```diff
@@ -380,7 +380,7 @@ def check_origin(self, origin_to_satisfy_tornado=""):
 
         # Check CORS headers
         if self.allow_origin:
-            allow = self.allow_origin == origin
+            allow = self.allow_origin == "*" or self.allow_origin == origin
```
kevin-bates, line 383: Is this change more for synergy with the change below?  Because it seems like [this code above](https://github.com/jupyter/notebook/blob/master/notebook/base/handlers.py#L359-L360) is addressing the wildcard.

kevin-bates:
Hmm.  In checking to make sure this issue would also be present in jupyter server, I found that `check_origin()` already addressed the wildcard - thus the comment after approval - sorry about that.

minrk:
@kevin-bates good catch! I updated to apply the same logic already in check_origin to check_referer

PR #5904, opened 2020-12-10
Prevent missing IOPub upon restart

SylvainCorlay:
Upon restart, nedge the kernel with info requests until we get something on IOPub, to ensure it is properly connected.

CF https://github.com/jupyter/jupyter_client/issues/593

### DIFF
--- a/notebook/services/kernels/kernelmanager.py
+++ b/notebook/services/kernels/kernelmanager.py
@@ -310,38 +310,70 @@ class MappingKernelManager(MultiKernelManager):
         await maybe_future(self.pinned_superclass.restart_kernel(self, kernel_id, now=now))
         kernel = self.get_kernel(kernel_id)
         # return a Future that will resolve when the kernel has successfully restarted
-        channel = kernel.connect_shell()
+        shell_channel = kernel.connect_shell()
+        iopub_channel = kernel.connect_iopub()
         future = Future()
+        info_future = Future()
+        iopub_future = Future()
 
         def finish():
             """Common cleanup when restart finishes/fails for any reason."""
-            if not channel.closed():
-                channel.close()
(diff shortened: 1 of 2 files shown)


notebook/services/kernels/kernelmanager.py
```diff
@@ -310,37 +310,62 @@ async def restart_kernel(self, kernel_id, now=False):
         await maybe_future(self.pinned_superclass.restart_kernel(self, kernel_id, now=now))
         kernel = self.get_kernel(kernel_id)
         # return a Future that will resolve when the kernel has successfully restarted
-        channel = kernel.connect_shell()
+        shell_channel = kernel.connect_shell()
+        iopub_channel = kernel.connect_iopub()
```
SylvainCorlay, line 314: Really, we should use the iopub channel from `zmqhandler.channels`, which we don't have access to here.

kevin-bates:
Hi Sylvain - thanks for submitting this pull request.  Using these changes, I am unable to complete a general restart of the kernel.  The state stays at `restarting` and no commands can be performed.  It does appear that the kernel manager is completing its restart of the process, but the setup of the kernel messaging is having issues.

I also notice that `SIGINT` (multiple times) is not terminating the application (`SIGQUIT` does) implying that the process is probably tied up in lower levels.

I've left a similar comment on the jupyter-server PR as well (once the method name is changed).

kevin-bates:
I'm sorry.  It just dawned on me that this PR requires https://github.com/jupyter/jupyter_client/issues/593, so please disregard the restart failure behavior.  I'll need to pull that PR as well.

SylvainCorlay:
Closing. This is not the way to go.

We need to do this in `pre_get` / `open` when opening the web socket connection.

PR #5907, opened 2020-12-11
do not apply asyncio patch for tornado >=6.1

bollwyvl:
Tornado 6.1 now no longer requires the asyncio patch on windows for python >=3.8. :tada: 

On https://github.com/jupyter-server/jupyter_server/pull/339, we went with a hard cut to _require_ 6.1, and remove the patch, but this PR applies the lighter touch from https://github.com/ipython/ipykernel/pull/564 of just not applying the patch.

Fixing this is good, as the error manifests itself as unkillable python hang on windows, timing out CI.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2041,7 +2041,7 @@ class NotebookApp(JupyterApp):
     def _init_asyncio_patch(self):
         """set default asyncio policy to be compatible with tornado
 
-        Tornado 6 (at least) is not compatible with the default
+        Tornado <6.1 is not compatible with the default
         asyncio implementation on Windows
 
         Pick the older SelectorEventLoopPolicy on Windows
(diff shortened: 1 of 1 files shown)


bollwyvl:
Ha, thanks @minrk! How you doing?

I made some progress on #5909 so this could be exercised a little harder (especially getting the selenium tests to run on windows under a representative deployment) but looks like nobody's ever done it before, so it's slow going. Will be doing some more looking!

bollwyvl:
Also, incidentally: can you think of any other places we had to do this stuff? I think i hit most of them...

minrk:
> can you think of any other places we had to do this stuff?

I did [a search](https://github.com/search?q=WindowsSelectorEventLoopPolicy+user%3Ajupyter+user%3Ajupyterhub+user%3Ajupyter-server+user%3Aipython+user%3Ajupyter-widgets&type=code) and it looks like:

- [x] ipykernel
- [x] notebook
- [ ] qtconsole
- [ ] jupyterhub 
- [ ] nbconvert
- [ ] nbdime

minrk:
> How you doing?

I'm okay! Super busy and spread a little thin, but trying to keep up. How are you? :)

bollwyvl:
Same... it'll be great when all the stuff that's in the air lands, but it does seem like a lot right now!

PR #5908, opened 2020-12-11
ensure iopub subscriptions propagate prior to accepting websocket connections

SylvainCorlay:
Nudge kernel with info request upon opening websocket until some IOPub message is received.

- repeat info requests every 500ms until iopub is received
- busy kernels are not nudged, which would block connections to busy kernels indefinitely

cf jupyter/jupyter_client#593

### DIFF
--- a/notebook/services/kernels/handlers.py
+++ b/notebook/services/kernels/handlers.py
@@ -123,60 +123,76 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
     def create_stream(self):
         km = self.kernel_manager
         identity = self.session.bsession
-        for channel in ('shell', 'control', 'iopub', 'stdin'):
-            meth = getattr(km, 'connect_' + channel)
+        for channel in ("iopub", "shell", "control", "stdin"):
+            meth = getattr(km, "connect_" + channel)
             self.channels[channel] = stream = meth(self.kernel_id, identity=identity)
             stream.channel = channel
     
-    def nudge(self): 
+    def nudge(self):
+        """Nudge the zmq connections with kernel_info_requests
+
+        Returns a Future that will resolve when we have received
+        a shell reply and at least one iopub message,
+        ensuring that zmq subscriptions are established,
+        sockets are fully connected, and kernel is responsive.
+
+        Keeps retrying kernel_info_request until these are both received.
+        """
+        kernel = self.kernel_manager.get_kernel(self.kernel_id)
+
+        # Do not nudge busy kernels as kernel info requests sent to shell are
+        # queued behind execution requests.
+        # nudging in this case would cause a potentially very long wait
+        # before connections are opened,
+        # plus it is *very* unlikely that a busy kernel will not finish
+        # establishing its zmq subscriptions before processing the next request.
+        if getattr(kernel, "execution_state") == "busy":
+            self.log.debug("Nudge: not nudging busy kernel %s", self.kernel_id)
+            f = Future()
+            f.set_result(None)
+            return f
+
         # Use a transient shell channel to prevent leaking
         # shell responses to the front-end.
-        kernel = self.kernel_manager.get_kernel(self.kernel_id)
         shell_channel = kernel.connect_shell()
+        # The IOPub used by the client, whose subscriptions we are verifying.
+        iopub_channel = self.channels["iopub"]
 
-        # The IOPub used by the client.
-        iopub_channel = self.channels['iopub']
-        
-        future = Future()
         info_future = Future()
         iopub_future = Future()
+        both_done = gen.multi([info_future, iopub_future])
 
-        def finish():
+        def finish(f=None):
+            """Ensure all futures are resolved
+
+            which in turn triggers cleanup
+            """
+            for f in (info_future, iopub_future):
+                if not f.done():
+                    f.set_result(None)
+
+        def cleanup(f=None):
             """Common cleanup"""
-            loop.remove_timeout(timeout)
             loop.remove_timeout(nudge_handle)
             iopub_channel.stop_on_recv()
             if not shell_channel.closed():
                 shell_channel.close()
 
+        # trigger cleanup when both message futures are resolved
+        both_done.add_done_callback(cleanup)
+
         def on_shell_reply(msg):
+            self.log.debug("Nudge: shell info reply received: %s", self.kernel_id)
             if not info_future.done():
-                self.log.debug("Nudge: shell info reply received: %s", self.kernel_id)
-                if not shell_channel.closed():
-                    shell_channel.close()
-                self.log.debug("Nudge: resolving shell future")
+                self.log.debug("Nudge: resolving shell future: %s", self.kernel_id)
                 info_future.set_result(None)
-                if iopub_future.done():
-                    finish()
-                    self.log.debug("Nudge: resolving main future in shell handler")
-                    future.set_result(None)
 
         def on_iopub(msg):
+            self.log.debug("Nudge: IOPub received: %s", self.kernel_id)
             if not iopub_future.done():
-                self.log.debug("Nudge: first IOPub received: %s", self.kernel_id)
                 iopub_channel.stop_on_recv()
-                self.log.debug("Nudge: resolving iopub future")
+                self.log.debug("Nudge: resolving iopub future: %s", self.kernel_id)
                 iopub_future.set_result(None)
-                if info_future.done():
-                    finish()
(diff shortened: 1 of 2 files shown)


jasongrout:
I thought that there was one set of zmq channels (shell, iopub, etc.) shared across all websocket connections (but I haven't looked at the code carefully recently). Is that wrong, or is that changing in this PR?

SylvainCorlay:
> I thought that there was one set of zmq channels (shell, iopub, etc.) shared across all websocket connections (but I haven't looked at the code carefully recently). Is that wrong, or is that changing in this PR?

I think that it is a correct statement.

minrk:
> I thought that there was one set of zmq channels

Each websocket client gets its own zmq connection to the kernel, created [here](https://github.com/jupyter/notebook/blob/e946154112daa5ef997d7f549953cce91b336d3d/notebook/services/kernels/handlers.py#L123-L129) using `KernelManager.connect_iopub()`, etc.

This is what allows the notebook server to not need client-identifying logic, since reply-routing on e.g. stdin, stream channels is handled in the kernel. IOPub messages are indeed wastefully duplicated, however.

notebook/services/kernels/handlers.py
```diff
@@ -128,6 +128,65 @@ def create_stream(self):
...
+            if not future.done():
+                future.set_exception(TimeoutError("Timeout waiting for nudge"))
+
+        iopub_channel.on_recv(on_iopub)
+        shell_channel.on_recv(on_shell_reply)
+        loop = IOLoop.current()
+
+        # Nudge the kernel with kernel info requests until we get an IOPub message
+        def nudge():
+            self.log.debug("Nudge")
```
minrk, line 180: If this is getting retried more than once or twice, it may be worth logging a more informative message.

Maybe a counter and log with debug if count <= 1, and warning with the count itself if count > 1?

notebook/services/kernels/handlers.py
```diff
@@ -263,15 +323,21 @@ def open(self, kernel_id):
         if buffer_info and buffer_info['session_key'] == self.session_key:
             self.log.info("Restoring connection for %s", self.session_key)
             self.channels = buffer_info['channels']
-            replay_buffer = buffer_info['buffer']
-            if replay_buffer:
-                self.log.info("Replaying %s buffered messages", len(replay_buffer))
-                for channel, msg_list in replay_buffer:
-                    stream = self.channels[channel]
-                    self._on_zmq_reply(stream, msg_list)
+            connected = self.nudge()
```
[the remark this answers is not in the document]
SylvainCorlay: > I don't think nudge on reconnect is appropriate. 

This is the branch that is used when restarting a kernel, and which causes the issue with "restart and run all", so I think we should nudge on reconnect...

(The trivial `connected` Future is what I had originally).

notebook/services/kernels/handlers.py
```diff
@@ -253,6 +312,7 @@ def _register_session(self):
             yield stale_handler.close()
         self._open_sessions[self.session_key] = self
 
+    @gen.coroutine
```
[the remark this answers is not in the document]
SylvainCorlay: > This method is not being turned into a coroutine. 

Thanks! This is a leftover from earlier versions of the PR. Making it *not* a coroutine now.

notebook/services/kernels/handlers.py
SylvainCorlay: I have added a counter for the calls to nudge, and am now logging the nudge attempts.

notebook/services/kernels/handlers.py
```diff
@@ -128,6 +128,67 @@ def create_stream(self):
...
+
+        iopub_channel.on_recv(on_iopub)
+        shell_channel.on_recv(on_shell_reply)
+        loop = IOLoop.current()
+
+        # Nudge the kernel with kernel info requests until we get an IOPub message
+        def nudge(count):
+            count += 1
+            if not future.done():
+                self.log.debug("Nudging attempt %s or kernel %s" % (count, self.kernel_id))
```
minrk, line 182: If this is taking a while, it's probably worth logging at a higher level:

```suggestion
                log = self.log.warning if nudge_count % 10 == 0 else self.log.debug
                log("Nudging attempt %s on kernel %s" % (nudge_count, self.kernel_id))
```

notebook/services/kernels/handlers.py
```diff
@@ -128,6 +128,67 @@ def create_stream(self):
...
+            if not future.done():
+                future.set_exception(TimeoutError("Timeout waiting for nudge"))
+
+        iopub_channel.on_recv(on_iopub)
+        shell_channel.on_recv(on_shell_reply)
+        loop = IOLoop.current()
+
+        # Nudge the kernel with kernel info requests until we get an IOPub message
+        def nudge(count):
+            count += 1
```
minrk, line 180:
```suggestion
            nonlocal nudge_count
            nudge_count += 1
```

SylvainCorlay: I am not sure about this since we are passing cout to the callback. You want to also remove the count parameter?

minrk: Oops! No, I didn't notice the closure. Don't mind me!

minrk:
Having poked and prodded quite a bit now, I understand how both branches need a nudge in case of restart:

- zmq sockets are not recreated when the kernel is restarted (they shouldn't need to be, as zmq does its own underlying auto reconnect).
- There is no similar nudge operation triggered on restart for 'existing' connections
- To prod connections, clients tend to recreate websocket connection on restart (this is not strictly required, but it is a standard practice and may be needed in some cases where the zmq ports change)
- a 'reconnecting' websocket may not actually mean a new zmq socket (the 'restoring connection' branch is taken), if the same session id is re-used

So it's possible for a 'new' websocket connection, as part of the client's restart steps, to re-use zmq sockets and thus never nudging the new kernel's sockets.

The downside of doing this nudge on every ws connect, even on established, already-nudged zmq sockets, is that websocket connections will not be accepted while the kernel is busy, even though they should be. Currently, it is valid to connect and start sending requests without waiting for iopub, but this can result in the issues folks are seeing. The most robust and efficient nudge would be on:

1. new zmq connections, and
2. kernel restarts for all existing sockets

where requests on *existing, open sockets* are blocked until the nudge resolves, not just connection-opening. That's more complicated, and I think we can go with this solution as first pass, and refine it later, especially if the delay is bothering folks.

The last thing I want to look at before merging is what happens if nudge is still outstanding when the connections are closed, then I think this is okay to merge. Thanks @SylvainCorlay!

minrk:
I noticed that there are test failures caused by a change in status messages delivered. The nudge waits for any iopub message, then resolves, which means it will tend to resolve on the 'busy' status message, and the 'idle' status message will end up propagating to the client after resolving the connection. It may be better to wait for specifically the idle message associated with the info request to preserve message expectations, even though this is not strictly required.

SylvainCorlay:
@minrk I think this is ready to go.

When this gets in I will update the `jupyter_server` PR.

minrk (approved):

Zsailer:
Thanks, @SylvainCorlay, @minrk, and @jasongrout!

notebook/services/kernels/handlers.py
```diff
@@ -123,11 +123,125 @@ def __repr__(self):
...
+        info_future = Future()
+        iopub_future = Future()
+        both_done = gen.multi([info_future, iopub_future])
+
+        def finish(f=None):
+            """Ensure all futures are resolved
+
+            which in turn triggers cleanup
+            """
+            for f in (info_future, iopub_future):
```
jasongrout, line 170: Just checking - the function argument is f and the loop variable is f. Is that going to cause a problem?

jasongrout:
@kevin-bates - what is the plan for a release with this PR merged? We're seeing consistent problems that this PR is designed to fix, and would love to have a release fixing the problem. I'm happy to help with the release process if that is helpful.

Zsailer:
I can work on a patch release, since @kevin-bates is on vacation this month.

jasongrout:
Awesome, thanks @Zsailer! Again, I'm happy to help if needed.

Zsailer:
Oops, I thought I was a maintainer on PyPI for this repo, but it looks like I was mistaken.  🤦

jasongrout:
I thought I was too. Looks like @blink1073 and @minrk are owners of the `notebook` package on pypi, though - maybe they could make both of us maintainers too?

blink1073:
I am not an owner, only a maintainer.

Zsailer:
Notebook 6.1.6 released with this PR included.

PR #5938, opened 2021-01-13
Replace Travis and Appveyor with Github Actions

kevin-bates:
Removes `.travis.yml` and `appveyor.yml` in favor of using Github Actions.

A couple of notes:
* This separates out the nbconvert handlers tests so that we can install pandoc separately for those workflows
* Windows 3.9 Python tests will fail because `ptyprocess` is not built for this version yet.

cc: @Zsailer


Zsailer (approved):
🎉  thank you, Kevin! 🎉

PR #5941, opened 2021-01-14
Re-enable support for answer_yes flag

afshin:
This fixes a bug. When user want's to shutdown the notebook server without confirmation, they should be able to call:

```
jupyter notebook -y
```


Zsailer (approved):
Thanks, @afshin! 

The failing test is unrelated. Merging away!

Zsailer:
We'll want to port this to jupyter_server, too, to address jupyter-server/jupyter_server#383

PR #5954, opened 2021-01-18
Russian translation fixes

insolor:


kevin-bates (approved):
These changes look good to me (well, return reasonable translations via google translate :smile:  ) - thanks!

PR #5957, opened 2021-01-21
Allow jupyter_server-based contents managers in notebook

afshin:
notebook's `contents_manager_class` trait doesn't accept subclasses of jupyter_server's `ContentsManager`. This is a problem for anyone looking to write custom, forward-looking contents managers. We should relax this constraint here (albeit, raise a warning).

This PR relaxes the `contents_manager_class` to allow these managers. To do this, new `traittypes` are defined and used that check the incoming types/instances against a list of classes.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1407,7 +1407,7 @@ class NotebookApp(JupyterApp):
             new = new.__module__
 
         if new.startswith('jupyter_server'):
-            self.log.warn(
+            self.log.warning(
                 "The specified 'contents_manager_class' class inherits a manager from the "
                 "'jupyter_server' package. These (future-looking) managers are not "
                 "guaranteed to work with the 'notebook' package. For longer term support "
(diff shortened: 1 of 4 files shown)


notebook/notebookapp.py
```diff
@@ -1379,13 +1380,41 @@ def _update_mathjax_config(self, change):
...
+    @observe('contents_manager_class')
+    def _observe_contents_manager_class(self, change):
+        new = change['new']
+        # If 'new' is a class, get a string representing the import
+        # module path.
+        if inspect.isclass(new):
+            new = new.__module__
+
+        if new.startswith('jupyter_server'):
+            self.log.warn(
```
kevin-bates, line 1410: We should avoid a deprecation warning and update to `warning()`.
```suggestion
            self.log.warning(
```

blink1073 (approved):
Thanks!

kevin-bates (approved):

PR #5962, opened 2021-01-27
Drop support for Python 3.5

kevin-bates:
This change drops support for Python 3.5 (preferably in the next minor release: 6.3).  Python 3.5 was EOL last September and hasn't been included in the CI runs since moving to GH actions.  Since Python 3.9 has been included in recent CI runs, I've gone ahead and added it to the set of supported Python releases.

I've also updated the _dev version_ to be based on the next _minor_ release rather than the next major release.


blink1073 (approved):
Thanks!

Zsailer (approved):

Issue #5967, opened 2021-01-29
Terminals on Windows/Python 3.9+ fail


kevin-bates:
This issue is being opened to track https://github.com/spyder-ide/pywinpty/issues/134 that was first identified in https://github.com/jupyter/notebook/issues/5909 which is causing the terminal tests (and terminal functionality in general) to fail on Windows with Python 3.9 platforms. 

The tests fail during collection with the following traceback:
<details><summary>traceback</summary>
<p>

```
_______ ERROR collecting notebook/terminal/tests/test_terminals_api.py ________
ImportError while importing test module 'D:\a\notebook\notebook\notebook\terminal\tests\test_terminals_api.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
c:\hostedtoolcache\windows\python\3.9.1\x64\lib\site-packages\terminado\management.py:23: in <module>
    from ptyprocess import PtyProcessUnicode
E   ModuleNotFoundError: No module named 'ptyprocess'

During handling of the above exception, another exception occurred:
c:\hostedtoolcache\windows\python\3.9.1\x64\lib\site-packages\_pytest\python.py:578: in _importtestmodule
    mod = import_path(self.fspath, mode=importmode)
c:\hostedtoolcache\windows\python\3.9.1\x64\lib\site-packages\_pytest\pathlib.py:531: in import_path
    importlib.import_module(module_name)
c:\hostedtoolcache\windows\python\3.9.1\x64\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:972: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:228: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:972: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:228: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
[...]

PR #5968, opened 2021-01-29
Skip terminal tests on Windows 3.9+ (temporary)

kevin-bates:
This change adds a pytest collection hook that ignores the terminals tests when encountered on Windows platforms running Python >= 3.9.  Once issue #5967 has been resolved, this PR should be rolled back.  As a result, each change identifies the issue to assist tracking.


Zsailer (approved):

Zsailer:
Thanks!

PR #5974, opened 2021-02-04
Allow /metrics by default if auth is off

### LINKED ISSUE #5973: /metrics endpoint requires auth even if token and password are empty
**Describe the bug**

When using the standard `--NotebookApp.token="" --NotebookApp.password=""` to disable auth, the fact that auth is turned off is not reflected in `authenticate_prometheus`. As a result the `/metrics` endpoint stops working (by default) after this commit

https://github.com/jupyter/notebook/commit/c0ab18a53472372e31dee293fefd66f218107c98

On systems like Kubeflow and JupyterHub, this is unfortunate (and a breaking change), though it is solved by adding

`--NotebookApp.authenticate_prometheus=False`.

Perhaps `authenticate_prometheus=False` should be default if other authentication is also disabled?

**To Reproduce**

```bash
# Two commits to compare the behavior
POST_COMMIT=5cb007f03275
PRE_COMMIT=feacdbfc2e89

COMMIT=$PRE_COMMIT
docker run -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB=yes \
    jupyter/minimal-notebook:$COMMIT \
    jupyter notebook --no-browser --ip=0.0.0.0 --ServerApp.allow_origin='*' \
    --NotebookApp.token='' --ServerApp.password=''

# works
curl http://0.0.0.0:8888/metrics

COMMIT=$POST_COMMIT
docker run -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB=yes \
    jupyter/minimal-notebook:$COMMIT \
    jupyter notebook --no-browser --ip=0.0.0.0 --ServerApp.allow_origin='*' \
    --NotebookApp.token='' --ServerApp.password=''
    
# Does not work!!!
curl -s -o /dev/null -w "%{http_code}"  http://0.0.0.0:8888/metrics
# > 403%                                            
```

Both commands work if the flag `--NotebookApp.authenticate_prometheus=False` is added.

**Expected behavior**

If auth is disabled (empty token and password), then set `authenticate_prometheus=False`

blairdrummond:
Resolves #5973 

How to test:

Use this dockerfile to compare

```Dockerfile
FROM jupyter/minimal-notebook:5cb007f03275  
RUN pip3 install -U git+git://github.com/blairdrummond/notebook.git@prometheus-auth-default
```

And run `docker build . -t jupyter-prom-test`

# Using the old image, `/metrics` is blocked even if auth is off

```bash
# Auth is ENABLED
docker run -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB=yes  \
     jupyter/minimal-notebook:5cb007f03275 jupyter notebook \
    --no-browser --ip=0.0.0.0  --ServerApp.allow_origin='*'

# Doesn't work, which is GOOD
curl http://0.0.0.0:8888/metrics

# Works
TOKEN=XXXXXXXXX
curl "http://0.0.0.0:8888/metrics?token=$TOKEN"

# Auth is OFF
docker run -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB=yes  \
     jupyter/minimal-notebook:5cb007f03275 jupyter notebook \
    --no-browser --ip=0.0.0.0  --ServerApp.allow_origin='*' \
    --NotebookApp.token='' --ServerApp.password=''

##############################
# Doesn't work, which is BAD!!!!!!!!!!!!
curl http://0.0.0.0:8888/metrics
```

# Using the new image

```bash
# Auth is ENABLED
docker run -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB=yes  \
    jupyter-prom-test jupyter notebook \
    --no-browser --ip=0.0.0.0  --ServerApp.allow_origin='*'

# Doesn't work, which is GOOD
curl http://0.0.0.0:8888/metrics

# Works
TOKEN=XXXXXXXXX
curl "http://0.0.0.0:8888/metrics?token=$TOKEN"

# Auth is OFF
docker run -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB=yes  \
    jupyter-prom-test jupyter notebook \
    --no-browser --ip=0.0.0.0  --ServerApp.allow_origin='*' \
    --NotebookApp.token='' --ServerApp.password=''

##############################
# Works!!! Which is GOOD!!!!!!!!!!!!
curl http://0.0.0.0:8888/metrics
```

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -1584,6 +1584,21 @@ class NotebookApp(JupyterApp):
         """
     ).tag(config=True)
 
+    @default('authenticate_prometheus')
+    def _default_authenticate_prometheus(self):
+        """ Authenticate Prometheus by default, unless auth is disabled. """
+        auth = bool(self.password) or bool(self.token)
+        if auth is False:
+            self.log.info(_("Authentication of /metrics is OFF, since other authentication is disabled."))
+        return auth
+
+    @observe('authenticate_prometheus')
+    def _update_authenticate_prometheus(self, change):
+        newauth = change['new']
+        if self.authenticate_prometheus is True and newauth is False:
+            self.log.info(_("Authentication of /metrics is being turned OFF."))
+        self.authenticate_prometheus = newauth
+
     # Since use of terminals is also a function of whether the terminado package is
     # available, this variable holds the "final indication" of whether terminal functionality
     # should be considered (particularly during shutdown/cleanup).  It is enabled only


kevin-bates:
Hi @blairdrummond - thank you for the pull request.  

While this approach will certainly work, I was envisioning the initialization of `authenticate_prometheus`'s default value (which is currently hard-coded to `True`) as the means of addressing this.  This way, references to the other two trait values can remain in the application module.  

A great example of this happens to be for the [`token` trait's default-value function in which it checks the `password`'s value](https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L999).  I think adding a `_default_authenticate_prometheus()` function that checks the `token` and `password` values would be the appropriate approach with the ultimate default value being `True`. 

Another example is the [`mathjax_url` default checking the `enable_mathjax` trait](https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L1350-L1353).

I hope that helps.  Thank again.

kevin-bates:
Excellent update!  I really like the idea of an informational message, but since authentication is on by default, I think it would be best to only indicate when it's disabled, particularly now that it can be disabled implicitly.  Thanks!

blairdrummond:
@kevin-bates how's this?

kevin-bates (approved):
Thanks Blair!

PR #5980, opened 2021-02-09
[Gateway] Track only this server's kernels

kevin-bates:
While working on Enterprise Gateway, I found that if a gateway server is hosting kernels from multiple lab servers (via either jupyter_server or notebook), all kernels managed by the gateway server could be shut down when terminating _one_ of the lab servers.  This change ensures that kernel-based operations of a given lab/notebook server only operate against its own kernels.

Also:
- Fixed some faulty log statements.
- Found the KernelHandlers `get` method hadn't transitioned to a coroutine so I made that change.

Analogous changes have been made in https://github.com/jupyter-server/jupyter_server/pull/407.

### DIFF
--- a/notebook/gateway/managers.py
+++ b/notebook/gateway/managers.py
@@ -435,8 +435,13 @@ class GatewayKernelManager(AsyncMappingKernelManager):
                 raise
         else:
             kernel = json_decode(response.body)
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

PR #5986, opened 2021-02-18
Remove configuration for nosetests from setup.cfg

frenzymadness:
After #5826 , this seems to be the last bit of nosetests here.


Zsailer (approved):

Zsailer:
Thanks, @frenzymadness

Issue #5758, opened 2020-09-17
Upgrade MathJax from v2 to v3


eeopd:
Currently MathJax 2.7.4 is used, but the latest version is 3.1.2. In v3, a complete rewrite of the project, some nice improvements [were made](https://docs.mathjax.org/en/latest/upgrading/whats-new-3.0.html), including much faster rendering and some new packages. However, because MathJax 3.0 is a complete rewrite it includes quite a few [incompatible changes](https://docs.mathjax.org/en/latest/upgrading/v2.html), so this upgrade would require quite a bit of work to get working, and in the end result in some differing behaviours/broken TeX blocks, and also cause several extensions to break.

Would you be open to a patch? Or would working on integrating KaTeX seem preferable (#3094,  #4066)?

jtpio:
FYI the next version of the Jupyter Notebook (Notebook 7) is based on JupyterLab 4 and will use MathJax 3.

For reference this was implemented in https://github.com/jupyter/notebook/pull/6777 and https://github.com/jupyterlab/jupyterlab/pull/13877.

You can already try the beta with: `pip install --pre notebook`.

jtpio:
Closing as this has now been implemented in Notebook 7.

jtpio:
> i am wondering if we can use jupyter whitout Mathjax , my Browser crash all time when Mathjax load page whit safe.js ???

If you are using JupyterLab or Notebook 7, you could look into switching to KaTeX via an extension: https://github.com/jupyterlab/jupyter-renderers

malouke2020:
Hi,
I use the folowing version of jupyter notebook
notebook                          6.5.4
notebook_shim                     0.2.3
-and i try different Browser  in Incognito mode  and regular mode
- i have a big data Frame  that i can't split in diffrente sub part 
- when i start processing after clear all notebook and restare from scratch  two things can Hapen;
- 1 crash of google chrome or fireforx  
- 2 i lose the connection to my  jupyter server (all process are in local no ssh no proy .....)
i suspect Mathjax and Safe.js to triger that error beceasu the majorty of time i can see the paths to mathjax or safe.js loading when crash happen.
i can't use jupyterlab for the reason i have to change css  to dispaly dataframe side by side and this it's too simple with 
CSS = """
.output {    
    flex-direction: row;
    flex-flow: row wrap;   
    font-size: 10pt;
}
"""
HTML('<style>{}</style>'.format(CSS))
[...]

Issue #2536, opened 2017-05-31
Please add TeX fonts for MathJax, not just STIX-web (or make it very easy for end-users to do so)


jason-s:
Maybe it's just a matter of opinion but I think the STIX-web Mathjax fonts look awful and I can't for the life of me figure out how to use the regular MathJax TeX fonts.  I've read #1037 but can't figure out what to do.

Is there a way to easily install them?

jason-s:
...or this https://stackoverflow.com/a/44290246/44330

Lutalli:
I agree with you.

I followed this [answer](https://stackoverflow.com/a/42446768) to solve the problem - Notebook uses its "own" MathJax and the font `TeX` is missing. Thus we can copy the `TeX` fonts from the standard MathJax (on Github) and paste them into Notebook's MathJax. We can then modify the configuration (as the answer said, in `main.min.js`) and restart Notebook.

Also notice that the official MathJax has updated its structure while Notebook hasn't (`v2.7`, specifically), which means the **old version** of MathJax is needed.

mgeier:
I've just created #5998, which would fix this issue.

PR #5998, opened 2021-03-03
Change MathJax fonts from STIX-Web to TeX

mgeier:
Currently, the fonts used by MathJax don't look great, switching to `TeX` font makes it look much better.

Also, `nbviewer` seems to be using `TeX`, so this way both would be consistent.

See also:

* https://stackoverflow.com/q/28909747/
* #2536

I'm not sure whether this PR uses the right/best way to do this, but it works for me when I try it locally.

In an unrelated, later change, it would of course also be great to upgrade to MathJax version 3, see #5758.

### DIFF
--- a/notebook/static/base/js/mathjaxutils.js
+++ b/notebook/static/base/js/mathjaxutils.js
@@ -32,7 +32,7 @@ define([
                     availableFonts: [],
                     imageFont: null,
                     preferredFont: null,
-                    webFont: "STIX-Web",
+                    webFont: "TeX",
                     styles: {'.MathJax_Display': {"margin": 0}},
                     linebreaks: { automatic: true }
                 },
--- a/setupbase.py
+++ b/setupbase.py
@@ -187,11 +187,11 @@ def find_package_data():
 
     for tree in trees + [
         mj('localization'), # limit to en?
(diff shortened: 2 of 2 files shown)


afshin:
@mgeier Could you please attach a couple screenshots to show the difference that this change makes in some representative notebooks?

cc: @vidartf @isabela-pf

mgeier:
It looks like in the meantime the situation has magically improved on its own (at the MathJax side I guess?).

Previously, bold symbols -- especially when decorated with `\boldsymbol{\hat{...}}` or something -- looked hideous. But now that has become much better.

The remaining differences are quite subtle.

> Could you please attach a couple screenshots to show the difference that this change makes

Here's one example that caught my eye:

    $a/b$

before:

image

after:

image

----

I still think it's worth switching, but the difference isn't that big anymore.

jtpio:
The corresponding PR in JupyterLab was closed with the following comment: https://github.com/jupyterlab/jupyterlab/pull/9907#issuecomment-1336489047

> I think this becomes obsolete once JupyterLab switches to MathJax 3 

Closing since Notebook 7 is built on JupyterLab, and now depends on MathJax 3. Thanks all!

PR #6009, opened 2021-03-16
Fix small mistake in save_hooks documentation

Islast:
There is a mistake in the example code for the `pre_save_hook`, where the method is given as a method of the `FileContentsManager` class. The `pre_save_hook` is actually a method of the `ContentsManager` class. 

This correction should make it easier for users copying and pasting from this template.

### DIFF
--- a/docs/source/extending/savehooks.rst
+++ b/docs/source/extending/savehooks.rst
@@ -38,7 +38,7 @@ A pre-save hook for stripping output::
             cell['outputs'] = []
             cell['execution_count'] = None
 
-    c.FileContentsManager.pre_save_hook = scrub_output_pre_save
+    c.ContentsManager.pre_save_hook = scrub_output_pre_save
 
 A post-save hook to make a script equivalent whenever the notebook is saved
 (replacing the ``--script`` option in older versions of the notebook):


kevin-bates (commented):
Hi @Islast - thank you for taking the time to draft this PR.  

I'm not sure the original documentation is in error.  

Since `ContentsManager` is essentially an abstract base class from which all contents manager subclasses ultimately derive, this update is saying that no matter what concrete form of `ContentsManager` will be used, instances of _that_ class will use _this_ `pre_save_hook`.  Whereas the current doc states that only instances of `FileContentsManager` (or its subclasses) will use _this_ `pre_save_hook`.

Which `ContentsManager` class is used by the server is configured via `NotebookApp.contents_manager_class`.  That said, server extensions are free to enlist the services of other `ContentsManager` subclasses where those extensions may not want a `pre_save_hook` configured for that particular class - for example.

Because the original doc entry is in the context of an example and the example is functionally correct, I think it might be best to retain the current doc.  Since the [default `ContentsManager` class is actually `LargeFileManager`](https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L1382-L1383), perhaps an update to that would be more appropriate - since one should probably target the most-leaf class to avoid side-affecting other subclasses.

Islast:
Fair enough. I had a series of errors using the the `FileContentsManager.pre_save_hook` beforehand, but since I can't reproduce them today, happy to accept your explanation. Thanks for your review.

kevin-bates:
Thanks for the update.  If the issues crop up again, let's take a closer look as to why that is.  The traitlets and configuration stuff can be a bit challenging when multiple envs are involved, etc.

Issue #1922, opened 2016-11-25
Possibility to launch notebook via .desktop file


tomspur:
There is a current feature request at Fedora (see [bugzilla bug](https://bugzilla.redhat.com/show_bug.cgi?id=1275800), were a `.desktop` file is proposed to start the notebook:

```
[Desktop Entry]
Name=Jupyter
Exec=sh -c "ipython notebook"
Terminal=false
Type=Application
Icon=jupyter-sq.png
Categories=GNOME;GTK;
StartupNotify=true
```

Do you think this is a good idea and would you like to support this also when installing this package?

minrk:
Do you mean the Fedora package, or the Python package? I think it would be great for this to be part of the fedora/debian package (note: it should be `jupyter notebook` not `ipython notebook`).

takluyver:
For reference, I have a .desktop file in nbopen, which is principally to support opening .ipynb files from a GUI file explorer (so it doesn't actually show up in the menu):

https://github.com/takluyver/nbopen/blob/master/nbopen/nbopen.desktop

tomspur:
I mean the Python package. I would prefer using a desktop file with an proposed and supported startup method by you. Maybe something like the `nbopen.desktop` file would work in this case too.
(yes, that bug and the Fedora ipython package is a bit older, it will be `jupyter` and not `ipython` at the end :))

minrk:
Adding `jupyter notebook %f` would be good for launching a notebook server in a given directory. It would be even better if it did what nbopen does, and launches IFF one doesn't already exist, and opens if not.

Amr-Ibra:
This issue is now closed by https://github.com/jupyter/notebook/pull/6010 and https://github.com/jupyter/notebook/pull/6099.

kevin-bates:
Good point - thank you @Amr-Ibra!

PR #6010, opened 2021-03-17
Add square logo and desktop entry files

befeleme:
Hi! 
Following the ancient discussion in #1922 I work on adding a desktop entry to Notebook when installed as RPM in Fedora. 
I was looking for the icon to use an thought the official [Jupyter logo](https://commons.wikimedia.org/wiki/File:Jupyter_logo.svg) would suit the best. The image, being rectangle more than square, got however unpleasantly distorted by the system. The graphics was resized so that it's 48x48px square with preserved original proportions and is easily scalable to any other square size.

If the files were included in Python package, they could be used by the other packagers as well. It'd be nice to have them coming from the upstream.


befeleme:
As was pointed out to me, environments displaying smaller icons (like Xfce) would benefit from icon without the text _jupyter_ inside, so I add another file, favicon alike.
I propose to install the files using `data_files` which ensures the installation to given location. This works well both with system- and user-wide installation.
Looking forward to your opinion.

blink1073 (approved):
Very nice, thanks!

befeleme:
@blink1073, just a quick question: when could the v. 6.3 be released? I'll help me adjust the next steps regarding the RPM package.

blink1073:
We plan to cut 6.3 early next week.  I'm making a changelog entry PR today.

PR #6018, opened 2021-03-22
Fix skipped tests & remove deprecation warnings

befeleme:
Hey,
I propose some changes to the tests:
1. Unskip tests using obsolete pytest `yield` feature. That is changed to recommended `pytest.mark.parametrize`. 
2. Remove DeprecationWarning caused by `\d` in regular expression - use recommended raw string instead.
3. Remove PytestCollectionWarning failing on recognizing TestingStopApp - problematic part is `^Test*` in the class name - renaming makes it work again.

### DIFF
--- a/notebook/tests/test_notebookapp.py
+++ b/notebook/tests/test_notebookapp.py
@@ -137,7 +137,7 @@ def test_notebook_password():
             assert nb.password != ''
             passwd_check(nb.password, password)
 
-class StopTestApp(notebookapp.NbserverStopApp):
+class StopAppTest(notebookapp.NbserverStopApp):
     """For testing the logic of NbserverStopApp."""
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


notebook/tests/test_notebookapp.py
```diff
@@ -139,7 +137,7 @@ def test_notebook_password():
             assert nb.password != ''
             passwd_check(nb.password, password)
 
-class TestingStopApp(notebookapp.NbserverStopApp):
+class StopTestApp(notebookapp.NbserverStopApp):
```
kevin-bates, line 140: Since the purpose of this is to test the "StopApp", I would suggest a rename to `StopAppTest`.

kevin-bates (commented):
Thank you @befeleme - these changes look great.  I had the comment about the stop app class name so that the test subject (StopApp) is preserved.

notebook/tests/test_notebookapp.py
befeleme: Thanks for pointing this out, fixing it.

kevin-bates (approved):
Looks good. Thank you.

PR #6019, opened 2021-03-22
Add Issue Lock and Answered Bots

afshin:
The answer bot enables us to explicitly indicate we have answered the question by giving the issue the `status:Answered` label, and then after 30 days of inactivity, the issue is scheduled to be closed after one more week, presumably answered.

cf https://github.com/jupyterlab/jupyterlab/pull/9920

The lock bot locks issues and PRs that have been closed for six months, preventing zombie thread discussions.

cf https://github.com/jupyterlab/jupyterlab/pull/9754

### DIFF
--- a/.github/answered.yml
+++ b/.github/answered.yml
@@ -1,0 +1,20 @@
+# This action automatically schedules issues to be closed that have been
+# labeled as answered if there is no activity on them for 30 days. This takes
+# care of the common usecase of an issue being answered to the best of our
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

jasongrout:
It looks like the jlab workflow is never running. Probably because the action file is not in the workflow subdirectory, which is the same case here?

PR #6020, opened 2021-03-23
Fix readthedocs environment

blink1073:


frenzymadness:
Would it make sense to add it also [here](https://github.com/jupyter/notebook/blob/master/setup.py#L129)?

blink1073:
@frenzymadness, thanks for the heads up!  I made this PR from the web UI so I can't edit other files, would you like to submit a separate PR to make that change?

frenzymadness:
Will do it in a moment.

frenzymadness:
See https://github.com/jupyter/notebook/pull/6021

PR #6026, opened 2021-03-30
Correct 'Content-Type' headers

faucct:


notebook/gateway/managers.py
```diff
@@ -401,7 +401,9 @@ async def start_kernel(self, kernel_id=None, path=None, **kwargs):
 
             json_body = json_encode({'name': kernel_name, 'env': kernel_env})
 
-            response = await gateway_request(kernel_url, method='POST', body=json_body)
+            response = await gateway_request(
+                kernel_url, method='POST', headers={'Content-Type': 'application/json'}, body=json_body
```
kevin-bates, line 405: Since `headers` can be set via configuration, `gateway_request()` (or more likely `load_connection_args()`) needs to now handle the case where a given keyword already exists in the static arguments and merge accordingly.  Up until now, that _key-level_ merge wasn't necessary.

kevin-bates (commented):
Hi @faucct - thank you for the pull request.  As noted in the comment, there will need to be some extra merge logic to account for multiple `headers` dictionaries.

notebook/gateway/managers.py
faucct: Done.

faucct:
A single check have failed, but I don't think it is related to the changes. Could you restart it, please?

blink1073:
I kicked the Selenium tests

kevin-bates:
@faucct - thank you for the contribution.  Since the notebook server is transitioning to [jupyter_server](https://github.com/jupyter-server/jupyter_server), it would be fantastic if you could make this change there as well.

PR #6041, opened 2021-04-22
Add base.handlers.get_current_request()

flupke:
This function allows to retrieve the current web request in async
contexts.

Retrieving the current context can be useful for example when
customizing Enterprise Gateway RemoteProcessProxy, to retrieve
authentication headers.

The implementation makes uses of Python 3.7's new contextvars module.
There are backports efforts for 3.6:

- https://pypi.org/project/aiocontextvars/ - does not have automatic
  asyncio support

- https://pypi.org/project/contextvars/ - which doesn't have it either.
  The tracking issue on this matter has been stale for 3 years now, and
  the conclusion of the discussion is "upgrade to a newer version of
  Python": https://github.com/MagicStack/contextvars/issues/2

Since notebook has no control on the event loop it is running into, I
preferred to drop Python 3.6 support.

### DIFF
--- a/notebook/base/handlers.py
+++ b/notebook/base/handlers.py
@@ -14,6 +14,7 @@ import sys
 import traceback
 import types
 import warnings
+import contextvars
 from http.client import responses
 from http.cookies import Morsel
 
@@ -35,6 +36,10 @@ from notebook.i18n import combine_translations
 from notebook.utils import is_hidden, url_path_join, url_is_absolute, url_escape, urldecode_unix_socket_path
 from notebook.services.security import csp_report_uri
 
+
+_current_request_var: contextvars.ContextVar = contextvars.ContextVar("current_request")
+
+
 #-----------------------------------------------------------------------------
 # Top-level handlers
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


notebook/base/handlers.py
```diff
@@ -932,6 +940,13 @@ def get(self):
         self.write(prometheus_client.generate_latest(prometheus_client.REGISTRY))
 
 
+def get_current_request():
+    """
+    Get :class:`tornado.httputil.HTTPServerRequest` that is currently being processed.
+    """
+    return _current_request_var.get(None)
```
flupke, line 947: Could you please point me to the right place in tests to cover this?

kevin-bates:
Hi @flupke. I think this pull request is probably better suited for [`jupyter_server`](https://github.com/jupyter-server/jupyter_server) at this point for the following reasons...
1. It constitutes new functionality which we're trying not to include in Notebook at this time
2. Enterprise Gateway is now (master branch) based on jupyter_server for all future releases
3. I think we should try to continue supporting 3.6 in this project for as long as possible and there may be a better argument for dropping 3.6 in jupyter server 

You should find the source bases extremely similar (although the test frameworks vary quite a bit).

As for where to apply test code, and assuming I'm understanding the change, you could probably take a look at the [API kernels tests](https://github.com/jupyter-server/jupyter_server/blob/d3b40f46654d9666a2ec5ba18b7133b0eb8a1820/jupyter_server/tests/services/kernels/test_api.py#L55) and not await the kernel start request and immediately follow it with one of these requests.

flupke:
Ok thanks, I was not aware of this change (we're still using an older version), I will come back once we update :)

PR #6052, opened 2021-05-09
Revert "do not apply asyncio patch for tornado >=6.1"

minrk:
revert #5907

While tornado 6.1 adds support for running on proactor, avoiding  an error, I think it's still better to use Selector since it avoids putting add_reader / writer in a thread.

Since ~all of our events are via add/remove_reader, keeping those on the main thread still seems preferable.


minrk:
Tests failures are on Windows, which is where I would expect behavior to change, but the same failures seem to have cropped up on master and are all terminal-related. Maybe terminado or something?

PR #6061, opened 2021-05-13
Bind the HTTPServer in start

minrk:
Binding sockets hooks them up to the running eventloop, which we should try to avoid in initialize
in case any other stages may configure the event loop policy

This is ~equivalent to https://github.com/jupyter-server/jupyter_server/pull/517 but in start instead of at the end of initialize.

I think calling listen in start is better, and easier because the bind was already a separate method call that was easy to move.


blink1073 (approved):
Thanks!

PR #6083, opened 2021-06-03
fix(terminal): enable OPTIONS method for delete

reynld:
This enables the OPTIONS request method to terminal API handlers which is required for cors to make delete requests.

### DIFF
--- a/notebook/terminal/api_handlers.py
+++ b/notebook/terminal/api_handlers.py
@@ -17,7 +17,7 @@ class TerminalRootHandler(APIHandler):
 
 
 class TerminalHandler(APIHandler):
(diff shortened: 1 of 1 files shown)


jtpio:
Closing as this would have to be implemented in the `jupyter_server_terminals` extension for Jupyter Server.

For reference the Jupyter Server extension code currently still references `GET` and `DELETE` as the supported methods:

https://github.com/jupyter-server/jupyter_server_terminals/blob/142b4a6d6998f865bd45dbfd6c7f9373062c5e27/jupyter_server_terminals/api_handlers.py#L66

So feel free to open an issue or PR there if you think this is still relevant. Thanks!

PR #6094, opened 2021-06-17
import six secure_notebook.py

RubenCantos:
importing only a part of the library would make the code faster

### DIFF
--- a/tools/secure_notebook.py
+++ b/tools/secure_notebook.py
@@ -7,7 +7,7 @@ the corresponding entries in the notebook json configuration file.
 
 """
 
-import six
+from six import u
 
 from notebook.auth import passwd
 from traitlets.config.loader import JSONFileConfigLoader, ConfigFileNotFound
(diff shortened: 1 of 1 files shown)


kevin-bates (commented):
Hi @RubenCantos - thanks for opening this PR.

Since we no longer support Python < 3 we should go ahead and remove the import of `six` (and use of `u`) altogether since it's merely a pass-thru for Python >= 3.

kevin-bates:
Hi @RubenCantos - would you mind making the suggested changes?  Thanks.

kevin-bates (commented):
Please see https://github.com/jupyter/notebook/pull/6094#pullrequestreview-686807489

RubenCantos:
> Please see [#6094 (review)](https://github.com/jupyter/notebook/pull/6094#pullrequestreview-686807489)

I did not understand [#6094 (review)](https://github.com/jupyter/notebook/pull/6094#issuecomment-888335014) ?

kevin-bates:
We shouldn't need to use `six` at all since its use was for py2 and py3 compatibility and we're fully py3 at this point.

PR #6101, opened 2021-06-30
Fix version of jupyter-packaging in pyproject.toml

frenzymadness:
The specifier `~=0.9` is equivalent to `>= 0.9, == 0.*` so there is no need to specify `<2`.

See PEP 440 for more details: https://www.python.org/dev/peps/pep-0440/#compatible-release


kevin-bates (approved):
LGTM - thank you.

PR #6107, opened 2021-07-17
Prompt for name when new folder created and open it when done

### LINKED ISSUE #3928: Prompt for folder name when creating a new folder

Honza-m:
Closes #3928 and follows recommendations in #4030.

This proposed update to the IU shows a modal prompt for folder name when a new folder is created and uses the PUT _new_ endpoint instead of the _new_untitled_ endpoint to create the folder with a name. It also opens the folder once it's successfully created in line with how a new notebook is opened immediately after creation.

### DIFF
--- a/notebook/static/services/contents.js
+++ b/notebook/static/services/contents.js
@@ -128,6 +128,33 @@ define(function(requirejs) {
         return utils.promising_ajax(this.api_url(path), settings);
     };
 
+
+    /**
+     * Creates a new empty file or directory in the specified directory path.
+     *
+     * @method new
+     * @param {String} path: the directory in which to create the new file/directory
+     * @param {Object} options:
+     *      ext: file extension to use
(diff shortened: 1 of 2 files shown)


kevin-bates:
Hi @Honza-m - Thank you for your pull request.   Since this would constitute a fairly significant change in UX behavior, it is something that we probably [shouldn't take on at this time](https://github.com/jupyter/notebook#notice).  If the same functionality is applicable in [Jupyter Lab](https://github.com/jupyterlab/jupyterlab), this feature might get better traction there.

Honza-m:
Thanks for having a look Kevin. May I suggest closing #3928 as well with a note that it's an issue that won't be addressed as Jupyter Lab is taking over?

PR #6141, opened 2021-08-18
My jupyter cannot define the py code

mdhaashim786:
If you want to open jupyter notebook with the help of cmd, then you need to go to anaconda folder and give the command : **jupyter notebook** . If you want to run the jupyter notebook file as a python file from cmd then give the command :- **jupyter nbconvert --to python filename.ipynb.**
Capture

If you made a new project and write various code. For running the new file the command is python **filename.py**


minrk:
Looks like a PR opened by mistake, since branch source and target don't make sense. Description doesn't appear to mention any issues. @mdhaashim786 if you have an issue, feel free to open one at https://github.com/jupyter/notebook/issues/new

PR #6160, opened 2021-08-27
Use default JupyterLab CSS sanitizer options for Markdown

krassowski:
Fixes https://github.com/jupyter-server/jupyter_server/issues/577

6.4.1 removed all CSS styles from HTML inside of Markdown cells after moving to a maintained version of the sanitizer. This PR restores CSS styles in Markdown cells. For a test case of:

```markdown
<span style="color: red; font-size: 30px">test</span>
```
Before:
Screenshot from 2021-08-28 17-00-57

After:
Screenshot from 2021-08-28 16-59-08

### DIFF
--- a/notebook/static/base/js/markdown.js
+++ b/notebook/static/base/js/markdown.js
@@ -106,7 +106,7 @@ define([
                     html = mathjaxutils.replace_math(html, math);
                 }
                 if(options.sanitize) {
-                    html = $(security.sanitize_html_and_parse(html));
+                    html = $(security.sanitize_html_and_parse(html, true));
                 }
             }
             callback(err, html);


krassowski:
The question is whether CSS styles should also be allowed in the output:

https://github.com/jupyter/notebook/blob/2cfff07a39fa486a3f05c26b400fa26e1802a053/notebook/static/notebook/js/outputarea.js#L692-L695

I think that they should, just like they used to be and are in JupyterLab. Any thoughts?

Zsailer:
I opened #6173 to follow-up on this question. In the meantime, I think we should merge this fix and cut a patch release to help unblock #6172.

PR #6163, opened 2021-08-30
Update conf.py

shruthi05605605:
library install before importing


kevin-bates:
[`shutil`](https://docs.python.org/3/library/shutil.html) is part of the python standard library and should not require explicit installation (nor does such a package exist).   As a result, I'm closing this pull request.  If this closure was premature, please comment as to why and we can reopen if necessary.

PR #6180, opened 2021-09-14
Uses jupyter/jupyter_client#656 to set notebook name in env.

Carreau:
Unfortunately this does requires upstream, because even if most
functions takes **kw, those in the end get passed to Popen() which does
not like unknown keywords.


jtpio:
Closing since the equivalent PR has been merged in Jupyter Server: https://github.com/jupyter-server/jupyter_server/pull/679

Thanks @Carreau!

PR #6181, opened 2021-09-15
Recover from failure to render mimetype

martinRenou:
It's easier to review the diff while disabling whitespace diffs: https://github.com/jupyter/notebook/pull/6181/files?diff=split&w=1

This is closer to JupyterLab's behavior. 

As an example, if a Jupyter widget fails to render (due to widget state not found), we display other available mimetypes instead of nothing. 

This goes in the direction of this PR: https://github.com/matplotlib/ipympl/pull/343 (add support for saving a static version of the matplotlib plot next to the widget mimetype) which works properly in JupyterLab but not in Notebook yet.

PR goes in pair with https://github.com/jupyter-widgets/ipywidgets/pull/3280

### DIFF
--- a/notebook/static/notebook/js/outputarea.js
+++ b/notebook/static/notebook/js/outputarea.js
@@ -48,7 +48,7 @@ define([
     OutputArea.config_defaults = {
         stream_chunk_size: 8192, // chunk size for stream output
     };
-    
+
     /**
      * Class prototypes
      **/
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
This looks good to me!

notebook/static/notebook/js/outputarea.js
```diff
@@ -682,36 +682,43 @@ define([
     OutputArea.prototype.append_mime_type = function (json, element, handle_inserted) {
         for (var i=0; i < OutputArea.display_order.length; i++) {
             var type = OutputArea.display_order[i];
-            var append = OutputArea.append_map[type];
```
minrk, line 685: Can the try/catch be a little narrower? Where does this typically raise, in the `toinsert = append.apply`?

martinRenou: Yes, it's `append.apply` that would throw if https://github.com/jupyter-widgets/ipywidgets/pull/3280 gets merged, I will make it narrower. Thanks for the review!

SylvainCorlay (approved):

PR #6194, opened 2021-10-06
Pin nbsphinx to 0.8.6, clean up orphaned resources

### LINKED ISSUE #6193: [DOCS] Notebook examples not rendering properly
### Description

Between version `v5.7.6` and the `stable`, the notebook examples section has become broken. For example:

- [The working version on `v5.7.6`](https://jupyter-notebook.readthedocs.io/en/5.7.6/examples/Notebook/Notebook%20Basics.html)

image

- [The broken version on `stable`](https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Notebook%20Basics.html)

image

It looks like the notebook `.ipynb` files are being parsed as raw text, rather than parsed properly by `nbsphinx`.

I tried getting a dev environment set up to debug this, but `pip install .` wasn't working because I ran into the error `error: [Errno 2] No such file or directory: 'bower'`. So opening this issue to track it.

The notebook examples are at the link below, and I suspect the fix will need to be somewhere around there, or in the Sphinx configuration:

https://github.com/jupyter/notebook/tree/master/docs/source/examples/Notebook

Originally reported in https://discourse.jupyter.org/t/error-in-jupyter-readthedocs-page-rendering/11038

kevin-bates:
It turns out that the documentation for Notebook has not been properly producing its [Notebook Examples](https://jupyter-notebook.readthedocs.io/en/master/examples/Notebook/examples_index.html#notebook-examples) section of its docs.  In looking into the issue, it can be reproduced by simply using `nbsphinx` 0.8.7 (over 0.8.6).  However, I suspect there may be more to this than that, but this is a start.  This pull request pins `nbsphinx` to 0.8.6.

While looking into this I noticed the following:
1. A warning message was produced that `docs/source/examples/Notebook/header.md` was not listed in any toctree - which is true.  However, the contents of that file are already equivalently expressed in `docs/source/examples/Notebook/examples_index.rst`, so I've removed `header.md`.
2. The "Contributing to the Jupyter Notebook" section is a straight include of `CONTRIBUTING.rst` at the repo root.  In that file, a simple verification image is referenced down in `docs/source/_static/images`.  However, while that image will be rendered when [viewed from GitHub](https://github.com/jupyter/notebook/blob/master/CONTRIBUTING.rst#verification), the same link is invalid relative to the [generated docs](https://jupyter-notebook.readthedocs.io/en/master/contributing.html#verification).  Since the image didn't really provide much information beyond what the text already describes, I chose to remove this image altogether, thereby removing the erroneous link produced in the ReadTheDocs version.

Resolves #6193

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -66,8 +66,7 @@ Verification
 
 While running the notebook, select one of your notebook files (the file will have the extension ``.ipynb``).
 In the top tab you will click on "Help" and then click on "About". In the pop window you will see information about the version of Jupyter that you are running. You will see "The version of the notebook server is:".
-If you are working in development mode, you will see that your version of Jupyter notebook will include the word "dev".
-If it does not include the word "dev", you are currently not working in development mode and should follow the steps below to uninstall and reinstall Jupyter.
+If you are working in development mode, you will see that your version of Jupyter notebook will include the word "dev". If it does not include the word "dev", you are currently not working in development mode and should follow the steps below to uninstall and reinstall Jupyter.
 
 Troubleshooting the Installation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


blink1073:
I think it is due to the multiple webhooks set up: https://readthedocs.org/dashboard/jupyter-notebook/integrations/

kevin-bates:
Interesting - that seems like a good bet.  Looking at the hooks' links, the first appears to get triggered monthly, while the 2nd and third are more frequent (and similar).  Can you determine if both the 2nd and 3rd are necessary or how they differ?

blink1073:
I'd say remove either 2 or 3, since both appear to be tied to PR builds.  That should clear up the duplicate builds.  Then, create a test PR and make sure the RTD build is still fired.

blink1073:
Seems to be working!

blink1073:
(and no duplicate build this time)

kevin-bates:
Right on.  The generated doc looks correct as well (relative to this PR).

PR #6201, opened 2021-10-11
Fully revert the pinning of nbsphinx to 0.8.6

kevin-bates:
Remove the last vestige of pinning nbsphinx to 0.8.6.  (Sorry, I should have caught this when reviewing #6200).


kevin-bates:
I'm going ahead and merging so we can get to possible other troubleshooting approaches wrt the docs builds.

Issue #6210, opened 2021-10-20
[Discuss] The future of the classic notebook interface


Zsailer:
The purpose of this issue is to discuss the future of the jupyter/notebook repository. 

**Background:** The classic Jupyter notebook interface is currently maintained by a skeleton crew of developers. On the other hand, many people still use the classic interface today. 

**Problem:** We have a widely-used piece of software with very little developer support. 

Let's have an open discussion about what to do with this project moving forward.

Zsailer:
An alternative proposal was to offer this project as an opportunity for another team/company/group to maintain this project moving forward. There might be some team with vested interest in making sure this project stays alive indefinitely. 

There was some concern raised with this idea. Handing off the project to another (possibly not Jupyter-affiliated) team puts the Jupyter brand (?, just echoing what I've heard said here) at risk.

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/the-future-of-classic-notebook-interface/11328/1

psychemedia:
*Apologies in advance for a long post... This repeats a lot of things posted elsewhere, but I think I still haven't found a way to express the growing gulf I see between the experience and expectations of highly skilled developers, and people who are very naive end-users of Jupyter tools doing very, very simple tasks that appear to be little more than (and often are very little more than) the following to a seasoned developer:*

```
for i in [1, 2, 3, 4]:
  print("hello world")
```

*At the same time, this is not to say that we shouldn't be providing such novices with environments that are, underneath, very powerful, and scalable in terms of what you can do with them. What I try to do below is explore what notebooks made possible, who they empowered, and what we are at risk of losing.*

As I've commented on this issue many times, I'll argue the position of:

- an educator, using notebooks to present material to novice learners in disciplines that may only touch tangentially on computing (for example, a newly graduated K12 teacher teaching basic programming to a class of two dozen 12 year olds; a near-retirement lecturer teaching first year undergrad mechanical engineering);
- educator coding experience: basic HTML, inline style/CSS, maybe some elementary (non-framework) Javascript (2 -30 lines enough to implement a basic function to do a single thing);
- institutional support: essentially none; maybe some grudging hosting of a JupyterHub server.

Skill level of users: JupyterLab IDE is overkill. Classic notebook UI is at the limit of UI complexity; notebook extension toolbar buttons bewilder a significant a percentage of users and make the UI unusable.

[...]

PR #6213, opened 2021-10-25
get bower dependencies from npm

minrk:
Avoids dependency on deprecated bower package manager.

This is the same strategy adopted some time ago by JupyterHub, which also used Bower. Rather than making significant changes to how everything is built, which caused problems during the webpack, this only changes how the dependencies are downloaded. A `bower-lite` script copies dependencies declared in `package.json` from node_modules to `notebook/static/components` to match the previous installation location.

Most packages are identical, but a couple of packages have slightly different layouts (jquery-ui, react, xterm). It's unclear if these relative URLs should be considered public APIs, in which case we can make do more detailed renames to match everything exactly.

Related to #6210


kevin-bates (approved):
Although I'm not qualified to review these changes, notebook builds and appears to run fine.

I think we should move forward with this so we can address https://github.com/jupyter/notebook/issues/6254, but would like someone with frontend experience to review as well.

kevin-bates:
> I think there's a real test failure, presumably related to a url somewhere or other that needs updating, but I haven't found it.

@minrk - could you please point out the test failure (tests aren't very stable at the moment and I'm not sure what might be real or fake)?  Perhaps someone could then take a look if there's a starting point.

Issue #6153, opened 2021-08-25
remove usage of ipython_genutils


Carreau:
See https://github.com/ipython/ipython_genutils/issues/18, 

ipython_genutils requires nose which is not packagable on Python 3.10 for many linux distro. 
ipython_genutils should not be necessary anymore as it's mostly a Py2/3 compat layer.

jtpio:
Closing as this is fixed in Notebook 7.

Issue #4463, opened 2019-03-09
ipython_genutils has been dropped - do not use it


opoplawski:
https://pypi.org/project/ipython_genutils/ states:

Pretend this doesn’t exist. Nobody should use it.

PR #6214, opened 2021-10-25
Partial #6153 #4463 : reduce ipython_genutils usage

penguinolog:
* remove stdlib code can be used directly or helper redundant:
  * str is already unicode
  * type casting is done in external lib
  * import aliases
* remove unused imports
* reduce amount of deprecated code:
  * `u` prefix for strings
  * `Logger.warn`

### DIFF
--- a/docs/source/public_server.rst
+++ b/docs/source/public_server.rst
@@ -137,7 +137,7 @@ You can then add the hashed password to your
 :file:`jupyter_notebook_config.py` is in your Jupyter folder in your home
 directory, ``~/.jupyter``, e.g.::
 
-    c.NotebookApp.password = u'sha1:67c9e60bb8b6:9ffede0825894254b2e042ea597d771089e11aed'
+    c.NotebookApp.password = 'sha1:67c9e60bb8b6:9ffede0825894254b2e042ea597d771089e11aed'
 
 Automatic password setup will store the hash in ``jupyter_notebook_config.json``
 while this method stores the hash in ``jupyter_notebook_config.py``. The ``.json``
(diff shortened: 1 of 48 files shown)


penguinolog:
In case of python 3.6 kill it's possible wider migration with less copy-paste

penguinolog:
@kevin-bates Can I drop python 3.6 support and use more Ipython shared code?

kevin-bates:
> Can I drop python 3.6 support and use more Ipython shared code?

I think dropping 3.6 is the right approach.  Not sure what you mean by using "more ipython shared code" however.  I thought this exercise was about _removing_ ipython dependencies.  Can you please provide examples of such shared code?

penguinolog:
> > Can I drop python 3.6 support and use more Ipython shared code?
> 
> I think dropping 3.6 is the right approach. Not sure what you mean by using "more ipython shared code" however. I thought this exercise was about _removing_ ipython dependencies. Can you please provide examples of such shared code?

Main final goal is dropping `ipython_genutils` as having problems with modern python and officially recommended to not use.
BTW it's better to make this PR again from fresh branch after #6260 due to large diff

kevin-bates:
> Main final goal is dropping ipython_genutils as having problems with modern python and officially recommended to not use.

I agree.  You had used "ipython" in a previous response when we should try to replace with _native_ python functions and why I needed some clarification.  Thank you.

> BTW it's better to make this PR again from fresh branch after #6260 due to large diff

Yes, I realized that as well last night.

jtpio:
Closing since this targets the `6.4.x` branch, which is not supported anymore.

If you think this is still needed and would like to continue the discussion, please open a new issue in https://github.com/jupyter/nbclassic.

Thanks!

PR #6215, opened 2021-10-25
Fix: CI error for python 3.6 & macOS

penguinolog:
Run actions/setup-python@v1
Error: Version 3.6 with arch x64 not found

Exclude nonexistent configurations from matrix


kevin-bates (commented):
Since Python 3.6 will be [end-of-life in a month](https://devguide.python.org/#status-of-python-branches), should we take this opportunity to replace 3.6 with 3.10 in the CI matrix instead?

penguinolog:
I'm agree with killing python 3.6, but until it not done - tests should pass

Op ma 22 nov. 2021 23:02 schreef Kevin Bates ***@***.***>:

kevin-bates:
@penguinolog, do you mind making the update to the test matrix that replaces 3.6 with 3.10?

penguinolog:
> @penguinolog, do you mind making the update to the test matrix that replaces 3.6 with 3.10?

If main maintainers agree - I'll be happy

penguinolog:
@kevin-bates removed all python 3.6 links and replaced by python 3.10 (except tests -> it's still using lowest supported version)

penguinolog:
Python 3.6 is dead. Can we move forward?

kevin-bates (approved):
The file changes look good to me, but I don't know what to make of the extraneous commits going on (not a GH expert enough to know if this is a concern). 

Would @Zsailer or @blink1073 be able to take a look?

Also, since we probably shouldn't drop official support for 3.6 in a patch release, I'm going to open an issue for dropping 3.6 and adding 3.10 "officially".

blink1073 (approved):
Thanks!  Final changes look good.  A squash merge will take care of the extraneous commits.

blink1073:
> I'm going to open an issue for dropping 3.6 and adding 3.10 "officially".

Yeah, agreed that should be a minor release.

PR #6218, opened 2021-11-01
Change to use a universal Chinese translation on certain words

jackexu:
I saw different translations of certain common words like cell and kernel so make it universal. 
Also, improve some translation through native tongue


kevin-bates (approved):
Although Google translate is not definitive, I consistly found its translation for the updated text to be closer to the target string than the original translations.  For example, where 'cell' is desired, the old translation had 'block', while the update had 'cell'.  

I'm approving the updates - thank you @jackexu - and will merge in a day or two barring a strong exception.

Issue #6164, opened 2021-08-30
Asyncio error when opening notebooks


nissankarkifm:
When I open a notebook from /tree, i get the following error in terminal: 
```c:\users\work\miniconda3\envs\playground\lib\json\encoder.py:257: UserWarning: date_default is deprecated since jupyter_client 7.0.0. Use jupyter_client.jsonutil.json_default.
  return _iterencode(o, 0)
[W 20:44:29.469 NotebookApp] Notebook NLP/3.02-obtain-model-results.ipynb is not trusted
ERROR:asyncio:Exception in callback <TaskWakeupMethWrapper object at 0x000001D12EB68DC8>(<Future finis...5ea"\r\n\r\n'>)
handle: <Handle <TaskWakeupMethWrapper object at 0x000001D12EB68DC8>(<Future finis...5ea"\r\n\r\n'>)>
Traceback (most recent call last):
  File "c:\users\work\miniconda3\envs\playground\lib\asyncio\events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
RuntimeError: Cannot enter into task <Task pending coro=<HTTP1ServerConnection._server_request_loop() running at c:\users\work\miniconda3\envs\playground\lib\site-packages\tornado\http1connection.py:823> wait_for=<Future finished result=b'GET /api/co...f5ea"\r\n\r\n'> cb=[IOLoop.add_future.<locals>.<lambda>() at c:\users\work\miniconda3\envs\playground\lib\site-packages\tornado\ioloop.py:688]> while another task <Task pending coro=<KernelManager._async_start_kernel() running at c:\users\work\miniconda3\envs\playground\lib\site-packages\jupyter_client\manager.py:336>> is being executed.
```
  

Desktop (please complete the following information):
 - OS: Windows
 - Browser Chrome
 - Version 92.0.4515.159

kevin-bates:
Hi @nissankarkifm - thanks for opening this issue.

I suspect this was introduced with the refactoring done in jupyter_client to [use async methods for the synchronous `KernelManager`](https://github.com/jupyter/jupyter_client/pull/623).   The symptoms are the same as [here](https://github.com/erdewit/nest_asyncio/issues/22) but this is a bit beyond my knowledge level. Although the kernel/notebook appears to behave fine, there's probably something hung up somewhere.  

On the bright side, this issue can be avoided by configuring Notebook to run with its `AsyncMappingKernelManager` instead - which introduces a full async stack.
```
[...]

PR #6221, opened 2021-11-03
Fix asyncio error when opening notebooks

dleen:
This is a fix for https://github.com/jupyter/notebook/issues/6164

`nest_asyncio` must be applied before any async tasks have been created
otherwise there could be tasks queued that are unpatched, and thus do
not respect nested loops. An example of an unpatched task can be seen in
the original issue:

```
<Task pending coro=<HTTP1ServerConnection._server_request_loop() running at /apps/python3/lib/python3.7/site-packages/tornado/http1connection.py:823>
```
which originates from Tornado.

A similar issue was reported in `nest-asyncio`: https://github.com/erdewit/nest_asyncio/issues/22
where the solution is to call `apply` on import so that unpatched tasks
do not get created.

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -2092,7 +2092,34 @@ class NotebookApp(JupyterApp):
 
         FIXME: if/when tornado supports the defaults in asyncio,
                remove and bump tornado requirement for py38
+
+        With the introduction of the async kernel, the existing sync kernel
+        requires the use of nested loops in order to run code synchronously.
+        This is done in `jupyter_client` using the helper util `run_sync`:
+
+        ref: https://github.com/jupyter/jupyter_client/blob/f453b51eeeff9e905c583b7da3905c0e35cfbdf0/jupyter_client/utils.py#L11
+
+        which creates a new event loop and relies on `nest_asyncio` patching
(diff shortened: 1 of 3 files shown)


notebook/notebookapp.py
```diff
@@ -2092,7 +2092,34 @@ def _init_asyncio_patch(self):
...
+        to allow nested loops. This requires that *all* potential tasks are
+        patched before executing. When only some tasks are patched it leads to
+        the following issue:
+
+        ref: https://github.com/jupyter/notebook/issues/6164
+
+        So we must call `nest_asyncio.apply()` method as early as possible. It
+        is preferable to do this in the consuming application rather than the
+        `jupyter_client` as it is a global patch and would impact all consumers
+        rather than just the ones that rely on synchronous kernel behavior.
```
kevin-bates, line 2112: Nice comment - thank you.

kevin-bates (approved):
These changes look good to me and I've confirmed the pending task error is not produced.  Thank you.

I'd prefer to wait for another review prior to merging and have requested the help of others.

blink1073 (approved):
Looks good, thank you @dleen!

dleen:
@minrk Are you able to take a look?

minrk:
Thanks!

PR #6240, opened 2021-11-23
Fix part of the problem with notebook losing trust.

Carreau:
See https://github.com/jupyter/nbformat/pull/236 and
https://github.com/jupyter/nbformat/issue/235.

This mostly affects frontends that do not always add ids to cells,
like classic notebook, and that issue is not limited to cells ids, but
also to empty cells, that may not have/need a trusted field.

But that is another issue.

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -39,6 +39,29 @@ from os.path import samefile
 
 _script_exporter = None
 
+# workaround nbformat bug:
+# nbformat save logic will mutate the notebook if it does not
+# have cells ids, so if we sign the notebook,
+# the signature of the saved notebook is changed.
+# we thus need to add cell ids to the notebook before signing.
+try:
+    from nbformat.validator import normalize
+except ImportError:
+    from nbformat.corpus.words import generate_corpus_id
+    from copy import deepcopy
+
+    def normalize(nbdict, version, version_minor):
+        nbdict = deepcopy(nbdict)
+
+        if version >= 4 and version_minor >= 5:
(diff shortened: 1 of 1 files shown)


notebook/services/contents/filemanager.py
```diff
@@ -39,6 +39,29 @@
...
+# the signature of the saved notebook is changed.
+# we thus need to add cell ids to the notebook before signing.
+try:
+    from nbformat.validator import normalize
+except ImportError:
+    from nbformat.corpus.words import generate_corpus_id
+    from copy import deepcopy
+
+    def normalize(nbdict, version, version_minor):
+        nbdict = deepcopy(nbdict)
```
kevin-bates, line 54: Could we make the `deepcopy` conditional on version? Not sure how many pre-cell-id notebooks are out there but this would save some time/space for them.

notebook/services/contents/filemanager.py
```diff
@@ -39,6 +39,29 @@
...
+
+        if version >= 4 and version_minor >= 5:
```
kevin-bates, line 56: This needs a minor adjustment so lower minor versions are included for future major versions.
```suggestion
        if (version == 4 and version_minor >= 5) or version >= 5:
```

kevin-bates (commented):
Thanks @Carreau, just had a couple of comments/suggestions.

jtpio:
Closing as this would be something to implement in Jupyter Server, if still applicable: https://github.com/jupyter-server/jupyter_server

Also the Notebook 7 trust mechanism follows what JupyterLab and Jupyter Server implement.

PR #6245, opened 2021-12-09
handle KeyError when get session

ccw630:
If gateway is enabled, `row_to_model` can delete session row when kernel culled or died on gateway. Then it raises a `KeyError` unhadled in `get_session`.

### DIFF
--- a/notebook/services/sessions/sessionmanager.py
+++ b/notebook/services/sessions/sessionmanager.py
@@ -189,7 +189,10 @@ class SessionManager(LoggingConfigurable):
 
             raise web.HTTPError(404, u'Session not found: %s' % (', '.join(q)))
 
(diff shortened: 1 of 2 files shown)


kevin-bates (approved):
LGTM - thank you.

ccw630:
Wondering if I should make a same PR to repo `jupyter_server` - what's the diffrence between those *backend* code?

kevin-bates:
Yes, that would be helpful, thank you. 
Jupyter server was derived directly from notebook but only includes backend code. That is, it is only the server portion. 
The notebook repository is essentially in a maintenance-only mode while jupyter_server entertains feature additions, etc. Front-end applications have or are migrating to jupyter_server.

PR #6251, opened 2021-12-20
Fix to remove potential memory leak on Jupyter Notebooks ZMQChannelHandler code

Vishwajeet0510:
Adding a session to the _open_sessions list only after validating kernel. This will prevent memory wastage if any requests are made with an incorrect kernel ID.

### DIFF
--- a/notebook/services/kernels/handlers.py
+++ b/notebook/services/kernels/handlers.py
@@ -251,7 +251,8 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
         if stale_handler:
             self.log.warning("Replacing stale connection: %s", self.session_key)
             yield stale_handler.close()
-        self._open_sessions[self.session_key] = self
+        if self.kernel_id in self.kernel_manager:  # only update open sessions if kernel is actively managed
+            self._open_sessions[self.session_key] = self
 
     def open(self, kernel_id):
         super(ZMQChannelsHandler, self).open()


notebook/services/kernels/handlers.py
```diff
@@ -219,6 +219,7 @@ def pre_get(self):
         # We don't want to wait forever, because browsers don't take it well when
         # servers never respond to websocket connection requests.
         kernel = self.kernel_manager.get_kernel(self.kernel_id)
+        self._open_sessions[self.session_key] = self
```
kevin-bates, line 222: Rather than remove the initialization of the `_open_sessions` entry from `_register_session()` would it make sense to check the kernel_id's presence prior to `_register_session()` instead (and keep `_register_session()` intact)?

rahul26goyal: @kevin-bates : `_register_session` is also responsible for cleaning up any existing / stale connection..so we can not skip the call completely by checking kernel's presence. 

We felt this solution is much cleaner as we populate the `_open_sessions`  after `get_kernel` which makes sure that the kernel exists.

kevin-bates: I see - good point since `get_kernel` throws if the id is not present - understood.  I would just rather not remove the one line that "registers" a session from `_register_session()`.  

What about checking the kernel manager's existence in a more passive manner within `_register_session()` and conditionally updating `_open_sessions`:
```python
        self.session_key = '%s:%s' % (self.kernel_id, self.session.session)
        stale_handler = self._open_sessions.get(self.session_key)
        if stale_handler:
            self.log.warning("Replacing stale connection: %s", self.session_key)
            yield stale_handler.close()
        if self.kernel_id in self.kernel_manager:  # only update open sessions if kernel is actively managed
            self._open_sessions[self.session_key] = self
```
while the subsequent `get_kernel()` call throws.

rahul26goyal: +1 @kevin-bates. 
@Vishwajeet0510 : please test that adding this extra check does not have corner cases.

Vishwajeet0510: Hello @kevin-bates, @rahul26goyal 

I've verified the above mentioned change and it is working fine, I will update the PR with this.

Thanks Kevin

kevin-bates (approved):
Thanks @Vishwajeet0510 and @rahul26goyal!

Vishwajeet0510:
Hey @kevin-bates 

Do you have any idea when this change will be released?

kevin-bates:
FYI - Notebook 6.4.8 is available on pypi: https://pypi.org/project/notebook/6.4.8/. Thanks for your help @Vishwajeet0510 and @rahul26goyal.

blink1073:
I added this to https://github.com/jupyter-server/jupyter_server/issues/110, if you'd like to make the same submission in the new Jupyter Server @Vishwajeet0510

kevin-bates:
Ah, yes, thank you Steve!

@Vishwajeet0510 - since EG uses JupyterServer in more recent releases, a similar change would be advised.  Thank you!

Vishwajeet0510:
Hello @kevin-bates and @blink1073 

Thanks for the quick followup and release. I will create a similar PR for jupyter_server as well.

rahul26goyal:
thanks a lot @kevin-bates for all the help in resolving this issue. :)

Vishwajeet0510:
Hello @kevin-bates 

Is it possible to back port this fix so it is available with Notebook v6.0.3? We are currently using JEG v2.1.0 and it is not compatible with Jupyter Notebook version >6.0.3. Please let us know if this would be possible.

kevin-bates:
Hi @Vishwajeet0510 - could you please elaborate on why EG 2.1.0 is incompatible with NB >6.0.3?  The only compatibility issue I'm aware of is `jupyter_client > 7` and EG has caps in place (in its more recent releases) to address that.

Vishwajeet0510:
Hi @kevin-bates 

As per my understanding, from Jupyter Notebook v6.1.0 async kernel management is supported and corresponding changes on JEG side are supported from JEG v2.1.1 onwards.

kevin-bates:
Hi @Vishwajeet0510 - since this is EG-specific and I'd like to explore things (possibly creating a longer comment chain), I've opened this discussion item in EG and hoping you and @rahul26goyal can join me there: https://github.com/jupyter/enterprise_gateway/discussions/1037

We can update this thread once things are resolved.  Thank you.

PR #6253, opened 2021-12-21
Fix deprecation warning

tornaria:
In sagemath, while doctesting `src/sage/repl/ipython_kernel/install.py`:

> DeprecationWarning: invalid escape sequence '\s'

The solution is to make the regex a raw string.


penguinolog (approved):

kevin-bates (approved):
LGTM - thank you @tornaria.

PR #6262, opened 2021-12-27
u

leetao50:
ok


kevin-bates (commented):
The conjunction `OR` is necessary for the phrase _PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES_ and should not be removed.

kevin-bates:
Hi @leetao50 - as noted in the review comment, the `OR` conjunction is necessary.  Since that was the only change in the PR, I'm going ahead with its closure.

jasongrout:
Additionally, I think it is a nonstarter to modify the standard BSD license text, even if it was grammatical. It's important for us to have a standard license, and not change its legal text.

PR #6289, opened 2022-02-22
only patch the event loop policy on old versions of tornado

graingert:


graingert:
looks like this was reverted here already: https://github.com/jupyter/notebook/pull/6052

PR #6315, opened 2022-03-18
Add support for opening a document with a different factory

jtpio:
First steps towards https://github.com/jupyterlab/retrolab/issues/290.

For example when opening a JSON file with different factories:

https://user-images.githubusercontent.com/591645/159045232-1957a135-16ee-4bde-9e6d-b43fd1b55354.mp4

This is also useful so folks can open notebooks with the text editor too:

https://user-images.githubusercontent.com/591645/159045837-297a5dc1-680a-4868-b1e7-4476e1a9df35.mp4

### TODO

- [x] Basic support for custom factories when opening a document
- [x] ~Should this work for arbitrary widgets, like the settings editor? Decide whether to do this in this PR or as follow-up~ -> deferring to another PR / issue, see this comment for a proof of concept: https://github.com/jupyter/notebook/pull/6315#issuecomment-1072621569
- [x] Should the `?factory=` URL parameter be removed from the URL after the document is opened? -> keeping it when it is not `default`, so the page can be reloaded and the document renders the same

### DIFF
--- a/packages/application-extension/src/index.ts
+++ b/packages/application-extension/src/index.ts
@@ -46,16 +46,6 @@ import { DisposableDelegate, DisposableSet } from '@lumino/disposable';
 import { Widget } from '@lumino/widgets';
 
 /**
- * The default notebook factory.
- */
-const NOTEBOOK_FACTORY = 'Notebook';
-
-/**
- * The editor factory.
- */
-const EDITOR_FACTORY = 'Editor';
-
-/**
  * A regular expression to match path to notebooks and documents
  */
 const TREE_PATTERN = new RegExp('/(notebooks|edit)/(.*)');
@@ -179,18 +169,12 @@ const opener: JupyterFrontEndPlugin<void> = {
         }
 
         const file = decodeURIComponent(path);
-        const ext = PathExt.extname(file);
+        const urlParams = new URLSearchParams(parsed.search);
+        const factory = urlParams.get('factory') ?? 'default';
         app.restored.then(async () => {
-          // TODO: get factory from file type instead?
-          if (ext === '.ipynb') {
-            docManager.open(file, NOTEBOOK_FACTORY, undefined, {
-              ref: '_noref'
-            });
-          } else {
-            docManager.open(file, EDITOR_FACTORY, undefined, {
-              ref: '_noref'
-            });
-          }
+          docManager.open(file, factory, undefined, {
+            ref: '_noref'
+          });
         });
       }
     });
--- a/packages/docmanager-extension/src/index.ts
+++ b/packages/docmanager-extension/src/index.ts
@@ -40,8 +40,9 @@ const opener: JupyterFrontEndPlugin<void> = {
         return;
       }
       const ext = PathExt.extname(path);
-      const route = ext === '.ipynb' ? 'notebooks' : 'edit';
-      window.open(`${baseUrl}${route}/${path}`);
+      const route =
+        widgetName !== 'Notebook' || ext !== '.ipynb' ? 'edit' : 'notebooks';
+      window.open(`${baseUrl}${route}/${path}?factory=${widgetName}`);
       return undefined;
     };
   }
--- a/packages/docmanager-extension/src/index.ts
+++ b/packages/docmanager-extension/src/index.ts
@@ -47,7 +47,12 @@ const opener: JupyterFrontEndPlugin<void> = {
       ) {
         route = 'notebooks';
       }
-      window.open(`${baseUrl}${route}/${path}?factory=${widgetName}`);
+      let url = `${baseUrl}${route}/${path}`;
+      // append ?factory only if it's not the default
+      if (widgetName !== 'default') {
+        url = `${url}?factory=${widgetName}`;
+      }
+      window.open(url);
       return undefined;
     };
   }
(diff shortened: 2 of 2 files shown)


jtpio:
>  Should the ?factory= URL parameter be removed from the URL after the document is opened?

Keeping it could be useful, so it can be copy pasted and sent to other people (or bookmarked). But some folks used to the classic notebook might wonder what this actually means.

jtpio:
> Should this work for arbitrary widgets, like the settings editor?

Probably this one could have special treatment.

Right now the `NotebookShell` only allows one widget to be added to the `main` area:

https://github.com/jupyter/notebook/blob/c13331705874452af00380431b51b5da39c67184/packages/application/src/shell.ts#L140-L143

An alternative could be to add it to the tree / dashboard page:

image

jtpio:
> An alternative could be to add it to the tree / dashboard page:

Some proof of concept to explore this:

https://user-images.githubusercontent.com/591645/159051661-9a5b19d7-dfe9-40c3-a83d-d88e4be997c9.mp4

Posting here for visibility, but should probably be explored more in another PR.

fcollonval:
Thanks a lot for starting @jtpio. 

I got a call yesterday with @parmentelat about using Jupytext in notebook v7 (something we cannot overlook). Starting from [v1.13.2](https://github.com/mwouts/jupytext/releases/tag/v1.13.2), Jupytext uses the document factory to display files with the notebook editor. So we definitely should check this PR works with Jupytext.

cc @mwouts

jtpio:
Thanks @fcollonval for the heads up. I think it should work with the Jupytext Notebook factory. Maybe we'll need to make sure it properly opens with the `/notebook` URL.

I installed `jupytext` with `pip install jupytext` locally and the server and lab extensions seem to be properly installed.

```
$ jupyter server extension list
Config dir: PREFIX/etc/jupyter
    jupyterlab enabled
    - Validating jupyterlab...
      jupyterlab 4.0.0a22 OK
    jupytext enabled
    - Validating jupytext...
      jupytext 1.13.7 OK
    notebook_shim enabled
    - Validating notebook_shim...
      notebook_shim  OK
```

```
$ jupyter labextension list
JupyterLab v4.0.0a22
PREFIX/share/jupyter/labextensions
        jupyterlab-jupytext v1.3.8 enabled  X (python, jupytext)
        @jupyter-notebook/lab-extension v7.0.0-alpha.1 enabled OK
```

The lab extension loads fine. However the `jupytext` server extension seems to be failing on startup: with

```
[W 2022-03-19 17:15:02.569 ServerApp] jupytext | extension failed loading with message: name 'build_jupytext_contents_manager_class' is not defined
```

jtpio:
Ah looks like it tries to import `LargeFileManager` from `notebook.services.contents.largefilemanager` here:

https://github.com/mwouts/jupytext/blob/d25c0c640ca0b2c1fb0e15b99be81bb82d2fe5ab/jupytext/contentsmanager.py#L573-L581

Which could be related since these are available in Notebook v7...

Does jupytext work with Jupyter Server / JupyterLab only? Or does it expect the classic notebook server (via the `notebook` package) to be available too?

jtpio:
> Ah looks like it tries to import `LargeFileManager` from `notebook.services.contents.largefilemanager` here:

This looks good when running with the fix proposed in https://github.com/mwouts/jupytext/pull/933:

https://user-images.githubusercontent.com/591645/159160135-0647a621-d232-415f-9bd2-efbc5aa10d87.mp4

Just need to open it under `/notebooks` to match the behavior of the classic notebook v6.

jtpio:
> I think we will be able to supported it out of the box by applying https://github.com/jupyterlab/jupyterlab/pull/11540 in Jupytext.

Maybe only the `Notebook` factory should open in `/notebooks` then? Looks like this would indeed work out of the box with this PR?

jtpio:
Documents now open with their default factory, for example when double-clicking on the item in the file browser.

When opened with a different factory, `?factory=` is kept in the URL so the browser tab can be reloaded and the document renders the same:

https://user-images.githubusercontent.com/591645/159239491-bc88b355-8fa8-4f8a-bd14-1f038300f04d.mp4

This also means that some files like JSON now open with the JSON viewer instead of the text editor by default. This differs from the behavior of the classic notebook and could maybe be customized via the settings if needed. Opening with the default JSON viewer by default is however useful because it's coherent with the behavior in JupyterLab.

jtpio:
cc @fcollonval if you would like to have a look.

Happy to get this in and iterate further. We can also make a new alpha release if that can ease development of third-party extensions like `jupytext`.

fcollonval (approved):
Thanks a lot for this one.

parmentelat:
indeed @jtpio many thanks !

fcollonval:
> Maybe only the `Notebook` factory should open in `/notebooks` then?

Looking into this briefly, the trouble for jupytext are actually for file types already defined in core (like .py or .md). I was thinking to use the `Router` to deal with path like `/notebooks/not/a-ipynb-file.ext` URL to redirect to the `/edit/not/a-ipynb-file.ext?factory=Jupytext%20Notebook` if the extension is supported by Jupytext. What do you think @jtpio ?

jtpio:
Thanks @fcollonval let's track this in https://github.com/jupyter/notebook/issues/6324.

PR #6316, opened 2022-03-18
Add the Settings Editor to the landing page

jtpio:
This adds the Settings Editor to the existing `TabPanel` on the landing page (`/tree`):

https://user-images.githubusercontent.com/591645/212135471-861f5246-e8c0-4bae-ad86-53565606c565.mp4

**Previous iteration**

https://user-images.githubusercontent.com/591645/159055329-52c9e003-716d-4f3f-8fbc-8f4a67ae2180.mp4

### TODO

- [x] Add the settings editor by default
- [x] Make it closable?
- [x] ~Expose the tab panel for other plugins: https://github.com/jupyterlab/retrolab/issues/21~


jtpio:
> Expose the tab panel for other plugins:

Maybe this would actually be a more sustainable approach for the first 7.0 release. So other types of JupyterLab widgets could be added to that tree panel, so they are all under `/tree`. Instead of trying to have a generic way to open widgets in new browser tabs?

jtpio:
> Maybe this would actually be a more sustainable approach for the first 7.0 release. So other types of JupyterLab widgets could be added to that tree panel, so they are all under /tree. Instead of trying to have a generic way to open widgets in new browser tabs?

Looks like this could indeed play nicely, especially since it would be similar to how some extra functionalities were being added to the classic notebook before, for example "Clusters" and "Nbextensions":

image

jtpio:
Maybe we can keep the settings editor hidden for now by default, so it looks more like the classic notebook.

And add it to the tab panel when opening it via the menu entry:

https://user-images.githubusercontent.com/591645/212135471-861f5246-e8c0-4bae-ad86-53565606c565.mp4

jtpio:
There is an upstream issue when using the command palette.

Tracked in https://github.com/jupyterlab/jupyterlab/issues/13760 and related PR: https://github.com/jupyterlab/jupyterlab/pull/13761

jtpio:
Now looking good for a first step. We can continue iterating in follow-up PRs.

PR #6314, opened 2022-03-18
Update to JupyterLab 4.0.0a22

jtpio:
This should help fix the failing CI on `main`.

This should also help with https://github.com/jupyterlab/retrolab/issues/358, since the new release include https://github.com/jupyterlab/jupyterlab/pull/12218.

The new release includes the fix for `ipython_genutils`:

- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.0a22
- https://github.com/jupyterlab/jupyterlab/pull/12202

PR #6318, opened 2022-03-19
Fix rendering of markdown

jtpio:
Follow-up to #6314 

- [x] Include the new `@jupyterlab/markedparser-extension` that was added in https://github.com/jupyterlab/jupyterlab/pull/11971
- [x] Refresh dependencies

Fixes the following issue on `main` (not released).

#### Before

image

#### After

image


jtpio:
Looking good on Binder:

image

PR #6338, opened 2022-04-01
Add pytest and handle warnings

blink1073:
Adapted from similar logic in `jupyterlab_server`


jtpio:
Thanks!

Looks like the test failures are related?

blink1073:
Yup, looks like I had outdated dependencies locally.

PR #6346, opened 2022-04-06
Add project URLs to setup.cfg

tlinhart:
Adding project URLs (especially source code URL) makes it easier for dependency management tools (like Renovate).


jtpio (approved):
Thanks!

Issue #6319, opened 2022-03-20
[6.4.x] CI failures on 6.4.x branch


echarles:
After reviewing the open PRs on the `6.4.x` branch, the selenium and javascript tests all fail. This is probably coming from the CI, and not from the respective PRs.

I am not sure what the status was before the creation of that 6.4.x branch. @Zsailer or @jtpio any thoughts on this?

jtpio:
@echarles the CI on this repo has been failing for many weeks / months. See this comment for reference: https://github.com/jupyter/notebook/pull/6213#issuecomment-1005852539

There hasn't been that much effort put into fixing it because:

- the repo was (and the 6.x is) in maintenance mode
- there was no plan to merge anything significant related to 6.x, except maybe https://github.com/jupyter/notebook/pull/6213

echarles:
Thx @jtpio for the info. I have updated the initial comment to track the issues and their fixes.

cc/ @kevin-bates

echarles:
For the Linux JS Tests, an issue is related to the `waitForPopup` which does not anymore do the expected job

https://github.com/jupyter/notebook/blob/232f724c64ddf84d8a21d630bde5ea2be3f4b586/notebook/tests/util.js#L30-L35

I have tried to force the popup url without luck. This may come from the webkit evolution, and how the tab opening are handled and recognized by `caspersjs`. I have tried with `slimerjs` option, downgrading my firefox to version 59 as documented on https://github.com/laurentj/slimerjs, but got segmentaion fault while running the test.

echarles:
Unless a solution can be found to let webpack generate phantomjs friendly javascript (https://github.com/jupyter/notebook/pull/6354#issuecomment-1094947104), a workaround solution would be to change the sanitizer index.js when running the CI.

echarles:
Progressing well on js-test:

-base: `green` ✅ https://github.com/jupyter/notebook/runs/5984992097?check_suite_focus=true

- services: `FAIL 82 tests executed in 70.297s, 74 passed, 8 failed, 0 dubious, 0 skipped.` on https://github.com/jupyter/notebook/runs/5984992147?check_suite_focus=true - I see errors like `ModuleNotFoundError: No module named 'IPython.kernel'` - CI runs `pip install .[test]` so this sounds strange. Any idea @blink1073 ?

[...]

PR #6354, opened 2022-04-11
Use var instead of const in sanitizer to partially fix ci js tests

echarles:
Partially fixes https://github.com/jupyter/notebook/issues/6319 for the js-tests

Phantom.js does not seem to recognize `const` keyword. We just have to change that to `var`. No sanitizer behavior is changed with this PR.

### DIFF
--- a/notebook/static/base/js/security.js
+++ b/notebook/static/base/js/security.js
@@ -16,7 +16,7 @@ define([
          * if allow_css is true (default: false), CSS is sanitized as well.
          * otherwise, CSS elements and attributes are simply removed.
          */
-         const options = {};
+         var options = {};
          if (!allow_css) {
              options.allowedStyles = {};
          }


echarles:
Once this gets merged, we can fix any remaining failing js-tests one by one. As the js-tests have not been maintained since some time, there will be a few to fix.

blink1073 (approved):
Thanks!

echarles:
I have further investigated the js-test failure. The root cause is javascript flavor which is too new (e.g. es6, with let instead of var...) for the old webkit used by phantomjs. I have tried numerous ways/configs to generate old javascript flavors (see e.g. below).

The file which gives issues is the `notebook/static/components/sanitizer/index` created from `@jupyterlab/apputils/lib/sanitizer/index`. So even with this PR which fixes the `base/security/index.js`, phantomejs has can not load correctly.

```js
  module: {
    rules: [
      {
        test: /\.m?jsx?$/,
        use: {
          loader: 'babel-loader',
          options: {
            presets: [
              ['@babel/preset-env', {
                corejs: {
                  version: 2,
                  proposals: true
                },
                spec: true,
                modules: "umd",
                useBuiltIns: 'entry',
              }]
            ],
          }
        }
      }
    ]
  }
```

blink1073:
Maybe try https://www.npmjs.com/package/string-replace-loader?

echarles:
> Maybe try https://www.npmjs.com/package/string-replace-loader?

The issue is the generation of the sanitizer/index.js file which is done one time. Then the base/security.js is injected with that file... which can not be loaded by the headless phantomjs browser used by the js-test.

I have tried to tune webpack so it creates good-old javascript flavor without success.

How do you see the usage of `string-replace-loader`? As I understand it, it replaces so javascript modules by other modules, but does not really define the javascript flavor.

BTW webpack is not use at run-time, nor test-time AFAIK. It is just used to create the sanitizer/index.js, which itself is used at run-time and test-time.

blink1073:
True, it seems like something static is needed to run after the dev build to down-convert the JS.

echarles:
> to down-convert the JS.

yeah, this is what I am looking for now, but maybe the needed downgrade is so large that modern babel can not do that anymore...

blink1073:
TypeScript also has a [target option](https://www.typescriptlang.org/tsconfig#target).  You could write a shim file that gets transpiled by TS down to es5.

blink1073:
> WDYT?

Works for me, as long as the sanitizer tests themselves run with the newer code.

echarles:
Thx a lot @blink1073 for the help and ideas to find a solution.

PR #6366, opened 2022-04-19
Initialize handler page_config from Server traitlets

bollwyvl:
This initializes `page_config` from, e.g. `ServerApp.tornado_settings.page_config_data`. These values can be subsequently overwritten by other configurables.

This corresponds to `LabHandler` in [jupyterlab_server](https://github.com/jupyterlab/jupyterlab_server/blob/4c9c972ef250c1a24a8130f265f1e3cf9d20f9f2/jupyterlab_server/handlers.py#L74).


jtpio:
Thanks!

Looks like this adds the following two items to the File menu (noticed by the UI tests):

image

Which is fine. We can always re-work the snapshots later via the settings system so they look more like the ones from the classic notebook:

image

Issue #6347, opened 2022-04-07
Build CI workflow does not run on `main`


jtpio:
Looks like the "Build" workflow does not run on the `main` branch at the moment:

![image](https://user-images.githubusercontent.com/591645/162247830-58e84b37-09a0-4479-93be-1d6bcc1668f4.png)

Also the name "Build" does not appear anymore in the list, instead it's the path to the workflow:

![image](https://user-images.githubusercontent.com/591645/162248176-b8ad6e8e-5f17-4e07-b371-4b225e3ca84b.png)

PR #6369, opened 2022-04-19
Fix build workflow on CI

jtpio:
This should fix https://github.com/jupyter/notebook/issues/6347

There was some syntax errors in the workflow that prevented the check to run on CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,6 +50,7 @@ jobs:
         pytest -vv || pytest -vv --lf
 
     - name: Integration Tests
+      run: |
         jupyter labextension list 2>&1 | grep -ie "@jupyter-notebook/lab-extension.*enabled.*ok" -
         jupyter server extension list 2>&1 | grep -ie "notebook.*enabled" -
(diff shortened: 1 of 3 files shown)


jtpio:
Tests pass locally but CI fails here:

https://github.com/jupyter/notebook/blob/60f1f01fa54f2dc278578de7cbf527e7bb28552d/tests/conftest.py#L91

Doesn't seem to find the `jupyterlab_server` test data:

image

jtpio:
> Doesn't seem to find the `jupyterlab_server` test data:

This seems to be fixed by using Python 3.10 on CI in https://github.com/jupyter/notebook/pull/6369/commits/c5f24f787240a3f9e56fd120478bc106d0c2851b (same version as used locally).

cc @blink1073 since this might affect other projects using `jupyterlab_server.test_data` in their tests.

jtpio:
CI all green with the build workflows running again.

blink1073:
> cc @blink1073 since this might affect other projects using jupyterlab_server.test_data in their tests.

That's odd, thanks for the heads up!

PR #6371, opened 2022-04-25
Update to JupyterLab 4.0.0a24

jtpio:
Bring in the latest fixes and improvements from https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.0a24

- [x] Update to the latest lab 4.0 pre-release
- [x] Remove deprecated webpack loaders


jtpio:
Looks like we are now getting this extra icon by default, even though no preferred path is set:

image

jtpio:
Looks like this was added in https://github.com/jupyterlab/jupyterlab/pull/12297.

jtpio:
So debugging this, it seems to be because in the case of notebook using a different page config than JupyterLab Server, `preferredPath` is not set:

image

jtpio:
> So debugging this, it seems to be because in the case of notebook using a different page config than JupyterLab Server, `preferredPath` is not set:

For now we can use the same logic as in JupyterLab Server (https://github.com/jupyter/notebook/pull/6371/commits/ce898be69ac2a7f5128ac569de9752e5403e4019).

We should then consider using the `LabHandler` directly.

PR #6376, opened 2022-04-26
Remove blank space at the bottom of the notebook

jtpio:
Set a lower `min-height` for the `::after` element at the bottom of the notebook.

**Before**

The notebook would take the full page height, and sometimes add a scrollbar:

image

**After**

image

This is closer to the Notebook v6 look and feel:

image


jtpio:
cc @HaudinFlorence maybe this will fix the issue you noticed when embedding a notebook in a documentation site with JupyterLite?

jtpio:
Merging as this improves the look and feel of the notebook document.

Happy to iterate more on it based on feedback. This could also be made a setting if folks prefer the blank space like in JupyterLab.

PR #6377, opened 2022-04-26
Show file checkboxes by default in the file browser

jtpio:
Enable the file checkboxes features in the file browser added in JupyterLab in https://github.com/jupyterlab/jupyterlab/pull/12299.

https://user-images.githubusercontent.com/591645/165315708-fbe89573-518e-4a7e-8177-94ed27b9b6d3.mp4

### TODO

- [x] Enable the setting by default if there is no user override
- [x] Update reference snapshots
- [x] Check if we can open multiple files (or track that in a different issue)


jtpio:
cc @gabalafou FYI this is what it looks like in Notebook v7

jtpio:
>  Check if we can open multiple files (or track that in a different issue)

This seems to be the case:

https://user-images.githubusercontent.com/591645/165344315-69aac258-c8fc-4f96-9710-2ec06254ef0a.mp4

jtpio:
A nice follow-up would be to implement something similar to the classic notebook, with action buttons showing up in the file browser toolbar:

https://user-images.githubusercontent.com/591645/165345364-4435811a-5ede-45c1-a36d-dadd0aabeb73.mp4

These actions are already accessible via the context menu, but could still be nice to have in the toolbar so they are more visible.

@gabalafou would you be interested in looking into it?

I'll open a separate issue so we can more easily track this. Maybe this would need some other changes in JupyterLab, not sure there is a signal or a way to know when a selection has changed already.

Thanks!

jtpio:
> I'll open a separate issue so we can more easily track this.

Tracked in https://github.com/jupyter/notebook/issues/6378

PR #6327, opened 2022-03-24
Adds left and right panels

jweill-aws:
Addreses #6326 by adding a left and right panel to Jupyter Notebook.

Adapted from https://github.com/jupyterlab/retrolab/pull/275 which was to address https://github.com/jupyterlab/retrolab/issues/272.

PR #6412, opened 2022-04-29
Remove the "Appearance" menu entry

jtpio:
Remove the "Appearance" menu entry for now, since it is empty.

We might want to add it back later for the left and right panels: https://github.com/jupyter/notebook/pull/6327

- [x] Disable the menu entry
- [x] Update the reference snapshots

**Before**

image

**After**

image


jtpio:
cc @afshin since this was brought during the Notebook weekly call

PR #6425, opened 2022-05-14
Use hatch backend

blink1073:
- Uses new [hatch_jupyter_builder](https://github.com/blink1073/hatch_jupyter_builder) plugin
- [hatch](https://hatch.pypa.io/latest/) is a newly minted pypa project, and appears to be a drop-in replacement for our use of `setuptools` / `jupyter-packaging`


blink1073:
We're blocked by https://github.com/pypa/hatch/issues/243.

blink1073:
Found a workaround using `force-include`

ofek:
FYI just released https://github.com/pypa/hatch/releases/tag/hatchling-v1.0.0

Also I think you can copy what you did here https://github.com/ipython/ipyparallel/pull/714/commits/e3e896ab3abb9502a564289b60543216f207b030

blink1073:
Thanks, and congratulations!

blink1073:
We can remove `setup.py` once https://github.com/jupyter-server/jupyter_releaser/pull/319 is released, then this should be good to go!

ofek:
It's undocumented until I find an auto-generator for non-Click CLIs, but if build deps are met you can do `hatchling version`

blink1073:
>  if build deps are met you can do hatchling version

Thanks!   Perhaps https://sphinx-argparse.readthedocs.io/en/stable/?

blink1073:
> Perhaps [sphinx-argparse.readthedocs.io/en/stable](https://sphinx-argparse.readthedocs.io/en/stable/)?

Ah, you are using `mkdocs`...

blink1073:
@jtpio, any idea why the snapshot tests are failing?

jtpio:
It looks like the menu and toolbar customization files are missing (defined in `schema`) for one of the failures:

image

blink1073:
Hmm, after adding the schemas the tests don't appear to start at all.

jtpio:
> Hmm, after adding the schemas the tests don't appear to start at all.

hmm strange. For the UI tests the package is first installed with the following:

```
python -m pip install -vv notebook*.whl
```

Will check locally.

jtpio:
Nice, CI is now passing with https://github.com/jupyter/notebook/pull/6425/commits/fcdb15f0106be4a5e66d5c5807d56670e67e3be4.

Looks like there are a few conflicts after the merge of https://github.com/jupyter/notebook/pull/6429.

blink1073:
All green again!

blink1073:
Moving to draft until we move the plugin to the jupyterlab org and publish it to conda-forge.

pyproject.toml
```diff
@@ -1,32 +1,109 @@
...
+"notebook/schemas/@jupyter-notebook" = "share/jupyter/lab/schemas/@jupyter-notebook"
+"jupyter-config/jupyter_server_config.d" = "etc/jupyter/jupyter_server_config.d"
+
+
+[tool.hatch.build]
+artifacts = [
+    "notebook/labextension",
+    "notebook/static",
+    "notebook/schemas"
+]
```
ofek, line 83: Do you want these in everything or just the sdist?

blink1073: Previously the files were in both so I kept it that way.  I agree it doesn't make much sense to have them in the wheel since the files are only used in the shared-data directory.  What do you think @jtpio?

ofek: For me, putting this under `[tool.hatch.build.targets.sdist]` instead brings the wheel size from ~3 MiB (see https://pypi.org/project/notebook/7.0.0a4/#files) down to ~50 KiB

blink1073: Actually, `"notebook/static"` is definitely needed, it is not used as `shared-data`, but the static assets used for the `notebook` app-specific page content.  I *think* the other two folders can be excluded, but @jtpio would know beter.

jtpio: `notebook/static` is indeed needed.

And it looks like `notebook/labextension` and `notebook/schemas` can be dropped if they are already specified in `tool.hatch.build.targets.wheel.shared-data`?

jtpio:
Ah I thought CI was already uploading the built assets as artifacts.

Maybe we could add the following step to the check release so it's easier to inspect the sdist and wheel?

```yaml
- name: Upload Distributions
  uses: actions/upload-artifact@v2
  with:
    name: notebook-jupyter-releaser-dist-${{ github.run_number }}
    path: .jupyter_releaser_checkout/dist
```

blink1073:
Good call.  I also played with my wip comparison [script](https://gist.github.com/blink1073/e77218a824af4ace9711f6de5b226092#file-hatch_comparison_script-py) and got this output now:

```
Removed_files:
notebook-7.0.0a4/jupyter-config/jupyter_server_config.d
notebook-7.0.0a4/jupyter-config/jupyter_server_config.d/notebook.json
notebook-7.0.0a4/notebook/labextension/static/remoteEntry.4fb60df0e7df1388ab11.js


Added files:
notebook-7.0.0a4/notebook/labextension/static/remoteEntry.5ea5fe7dc5a5fbc91231.js
notebook-7.0.0a4/jupyter_config.json
notebook-7.0.0a4/lerna.json
notebook-7.0.0a4/pyrightconfig.json
notebook-7.0.0a4/.eslintrc.json
```

jtpio:
Looking good!

jtpio:
> Which seems to be related to the `jupyter labextension develop` command expecting to find a `setup.py`:

This is likely to be fixed by https://github.com/jupyterlab/jupyterlab/pull/12606.

blink1073:
Yeah I think we need to use a `setup.py` shim for extensions that are targeting versions of `jupyterlab` that require it.  Once we have 3.x and 4.x versions that don't require it, the cookiecutter itself can be updated to not include it.  For the migration script, we can include it if there was already one there and eventually remove it.

blink1073:
Drat, now we have the issue that symlinks can't be created over existing files.  With `setuptools` the shared data was not written in editable mode.  It feels like we need to fix a few bugs and backport them in JupyterLab before doing any migrations.  We shouldn't *need* a `setup.py` and we shouldn't error out when overwriting symlinks.

ofek:
Hatchling resolves symlinks fyi

blink1073:
Ah, the symlink error was actually from a file in this repository, the binder is working now.  I actually think this PR is good to go now.

pyproject.toml
```diff
@@ -1,32 +1,126 @@
...
+    "bump2version",
+    "hatchling"
+]
+
+[tool.hatch.version]
+path = "notebook/_version.py"
+source = "code"
+
+[tool.hatch.build.targets.wheel.shared-data]
+"notebook/labextension" = "share/jupyter/labextensions/@jupyter-notebook/lab-extension"
```
agoose77, line 73: @blink1073 I'm not an expert on the filesystem layout for JupyterLab extensions, but should this be
```suggestion
"notebook/labextension" = "share/jupyter/labextensions/@jupyter-notebook"
```

blink1073: Yes, I agree

jupyter_config.json
```diff
@@ -1,4 +0,0 @@
-{
-  "LabApp": { "collaborative": true, "expose_app_in_browser": true },
-  "JupyterNotebookApp": { "collaborative": true, "expose_app_in_browser": true }
```
jtpio, line 3: Probably we want to keep this file, so RTC can be tested on Binder?

jtpio (approved):
Thanks!!

Looking good on Binder and also locally using the assets built by the releaser.

Left a small comment about the config for RTC, but otherwise looks good!

pyproject.toml
blink1073: The extension is nested in our case, under `@jupyter-notebook`.   The full extension name is `@jupyter-notebook/lab-extension`.

jupyter_config.json
blink1073: Restored

pyproject.toml
agoose77: Right, I'm with you. This is where the fact that `/` is allowed in NPM package names threw me for a second!

jtpio:
Ah looks like something is wrong with the Python version: https://github.com/jupyter/notebook/pull/6448

PR #6447, opened 2022-06-09
Fix docs build

jtpio:
Looks like the docs check is failing in https://github.com/jupyter/notebook/pull/6446.

Not sure what changed.


jtpio:
Looks like this fixes it:

image

blink1073:
Sphinx 5.0 started giving warnings about the language field being `None`, and we error on warnings.

PR #6464, opened 2022-07-05
Fix tests

jtpio:
Tests have started failing with the latest `tornado` release which deprecates: https://www.tornadoweb.org/en/stable/releases/v6.2.0.html#deprecation-notice

> The [IOLoop](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop) constructor is deprecated unless the make_current=False argument is used. Use [IOLoop.current](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.current) when the loop is already running instead.

Example run: https://github.com/jupyter/notebook/runs/7191744285?check_suite_focus=true

This PR attempts at fixing it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,6 @@ test = [
     "pytest>=6.0",
     "pytest-cov",
     "requests",
-    "pytest-tornasync",
     "pytest-timeout",
     "pytest-console-scripts",
     "ipykernel",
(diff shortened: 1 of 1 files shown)


jtpio:
Marking as ready for now.

Looks like the proper fix would be for `pytest-tornasync` to fix these warnings at some point: https://github.com/eukaryote/pytest-tornasync

Issue #6410, opened 2020-12-09
Expose the tree TabPanel to it can be consumed by other extensions


jtpio:
The main view on the tree page is actually a single widget, that shows both the file browser and the running sessions using the tab panel. 

![image](https://user-images.githubusercontent.com/591645/101641170-b9350400-3a31-11eb-8b12-c82f0b8a7905.png)

Ideally we should expose a plugin, so other extensions could consume it and add new tabs to this view.

PR #6478, opened 2022-07-19
Management of widgets in tree view

brichet:
The main area can only contain one widget in NotebookShell. If the main area is a TabPanel (**tree** view for instance), there is no easy way to add a widget to that TabPanel, or to search in its widgets from an extension.

This PR allows using `app.shell.add(...)` and `app.shell.activateById(...)` in the **tree** view as in JupyterLab main area.

Using a flag in NotebookShell :
- adding a widget in the main area will add it to the TabPanel.
- searching a widget in the main area will search in the TabPanel widgets.

May be related to #6316 as it transparently give access to the tab panel.

### DIFF
--- a/packages/application/src/shell.ts
+++ b/packages/application/src/shell.ts
@@ -13,7 +13,7 @@ import { Message, MessageLoop, IMessageHandler } from '@lumino/messaging';
 
 import { ISignal, Signal } from '@lumino/signaling';
 
-import { Panel, Widget, BoxLayout } from '@lumino/widgets';
+import { Panel, Widget, BoxLayout, TabPanel } from '@lumino/widgets';
 
 /**
  * The Jupyter Notebook application shell token.
@@ -73,6 +73,7 @@ export class NotebookShell extends Widget implements JupyterFrontEnd.IShell {
     rootLayout.addWidget(this._main);
 
     this.layout = rootLayout;
+    this._mainIsTabPanel = false;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks @brichet for bringing this.

Linking to https://github.com/jupyter/notebook/issues/6410 which expresses a similar idea.

packages/application/src/shell.ts
```diff
@@ -103,12 +104,29 @@ export class NotebookShell extends Widget implements JupyterFrontEnd.IShell {
     return this._menuWrapper;
   }
 
+  /**
+   * Get the status of main area.
+   */
+  get mainIsTabPanel(): Boolean {
```
jtpio, line 110: Need to think about this, but at first glance trying to know what a page or a set of extensions is trying to add sounds a bit odd.

The shell should be generic enough, for example we could imaging other custom application reusing `NotebookShell` but without the tree view (or a different one not using a `TabPanel`).

brichet: I agree that this solution is not generic enough.

> Also for more context the `NotebookShell` was initially thought as only allowing 1 widget in the main area on purpose.

Is that still relevant as there is already the tree view that would allow several widgets in main area ?

Indeed if the tree view will be the only one needing several widgets, the better way might be to expose its tab panel as said in #6410. Would the idea be to use token ?

Otherwise, another way could be to use a dock panel like in JupyterLab, but limited to one widget by default. That would keep the initial context but allows several widgets in main area if necessary.

jtpio: > Is that still relevant as there is already the tree view that would allow several widgets in main area ?

Good point. For the tree page it would make sense for widgets to be added to the tab panel. And would be natural to extension authors using `app.shell.add(widget, 'main')`.


> Would the idea be to use token ?

Yes the idea would have been to require the token to be able to add new widgets to the tab panel. Thinking more about it this might be simpler to implement, although maybe less straighforward to extension authors who would somehow need to create a dedicated plugin for this.

PR #6474, opened 2022-07-14
Make notebook 6.5.x point to nbclassic static assets

ericsnekbytes:
This makes notebook use the `static` assets of `nbclassic`, so both use the same physical files for future frontend development. Main changes (still being refined) are:

- Change the DEFAULT_STATIC_FILES_PATH in `__init__` to point to nbclassic/static folder
- Delete the `static/` folder from `notebook`
- Remove static file build/related logic from setup system (`nbclassic` will do this itself, we don't need the logic here anymore)
- Remove JS tests from CI (those tests are covered in nbclassic)

_Note that this PR involves removing static frontend JS/CSS files from notebook and obtaining those instead from nbclassic, it does not remove the handlers and other logic, those are still in place._

PR #6484, opened 2022-07-21
Selenium test updates

ericsnekbytes:
Refactor deprecated selenium `find_element_by_*` calls to new `find_element(By.*, "blah")` API.

See also [this SO post](https://stackoverflow.com/questions/69875125/find-element-by-commands-are-deprecated-in-selenium).


ericsnekbytes:
Undrafted! Trying to get the tests to complete successfully here still...

echarles:
@ericsnekbytes  Looks like the new API does the job. I see a new error. Has this been introduced by the new selenium?

```
selenium.common.exceptions.ElementClickInterceptedException: Message: Element <div class="cell code_cell unrendered selected"> is not clickable at point (683,175) because another element <div class="modal-body"> obscures it
```

ericsnekbytes:
> @ericsnekbytes Looks like the new API does the job. I see a new error. Has this been introduced by the new selenium?
> 
> ```
> selenium.common.exceptions.ElementClickInterceptedException: Message: Element <div class="cell code_cell unrendered selected"> is not clickable at point (683,175) because another element <div class="modal-body"> obscures it
> ```

Which one is this from? This looks like a page layout issue to me, I wouldn't expect it to be caused by the new version.

ericsnekbytes:
Just checked the static PR and [that same failure was there](https://github.com/jupyter/notebook/runs/7450072277?check_suite_focus=true) for Ubuntu/3.9...

echarles:
@ericsnekbytes Just merged https://github.com/jupyter/notebook/pull/6474 which has introduced a conflict here.

echarles (approved):
LGTM Thx @ericsnekbytes

Issue #6405, opened 2021-06-19
Show TOC


goerz:
First off, I'm very impressed with the new retro UI!

I was wondering if it was possible to show a table of contents, preferably in a format similar to the [classic notebook TOC2 extension](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/toc2/README.html). Presumably, in retrolab, the feature would actually be based on the jupyterlab TOC plugin that's part of the core since v3, but ideally I'd love to see all the "classic" features I mention in https://github.com/jupyterlab/jupyterlab/issues/10452

Issue #6403, opened 2021-11-03
Make left/right areas visible in RetroLab for extensions


jweill-aws:
This issue is a successor to jupyter/retrolab#257.

### Problem

Extensions in RetroLab may not appear if they need to add items to the left or right areas. These areas are not present in RetroLab and cannot be shown.

### Proposed Solution

Allow users to install extensions that show up in the left and right panels. Do not show the left and right tab bars, as is done in JupyterLab, but instead show each extension in the View menu as a cascading submenu item. The left and right panels should have an X (close) button to dismiss them. Left and right panels should appear in a horizontal grid with the main area, not overlapping the main area.

### Additional context

@jtpio showed @ellisonbg and myself a concept to let users hide tab bars and side panels on the left and right of a notebook: https://gist.github.com/jtpio/4e9a3d435d7c3950dc4150c10abbb1d7

Issue #6326, opened 2022-03-24
Left and right panels for installing JupyterLab extensions


jweill-aws:
Adapted from https://github.com/jupyterlab/retrolab/issues/272 by @ellisonbg:

Improve the notebook shell to handle the L and R areas. Add a way to install JupyterLab extensions that use the L/R area into Jupyter Notebook - such as the table of contents, running panel, debugger, etc.

PR #6487, opened 2022-07-25
Left right panel in Notebook view

brichet:
This PR follows https://github.com/jupyter/notebook/pull/6327 to add left and right panel to Notebook.

The left and right panels can be opened from the `view` menu.

Closes https://github.com/jupyter/notebook/pull/6327
Fixes https://github.com/jupyter/notebook/issues/6326
Fixes https://github.com/jupyter/notebook/issues/6403
Fixes https://github.com/jupyter/notebook/issues/6405


jtpio:
Nice thanks @brichet for picking this up :+1:

brichet:
Is there a way to retrieve the `yarn.lock` file generated during the `check_release` action ?
It seems that it changes, but I can't reproduce it locally.

jtpio:
> Is there a way to retrieve the `yarn.lock` file generated during the `check_release` action ?
> It seems that it changes, but I can't reproduce it locally.

hmm strange, if it changes during the check release then it should also change locally after running `jlpm`.

jtpio:
Just tried running `jlpm` in a fresh GitHub Codespace and it does modify the `yarn.lock`:

image

So these changes to the `yarn.lock` should be committed and pushed as well.

brichet:
> So these changes to the `yarn.lock` should be committed and pushed as well.

Right, sorry for that, indeed after cleaning everything it changes `yarn.lock`.

brichet:
Maybe we could increase the test time. This PR adds some little ui-test, but I think 10 minutest is too short.

jtpio:
> Maybe we could increase the test time. This PR adds some little ui-test, but I think 10 minutest is too short.

Hmm normally 10 minutes should be more than enough? Or do the new tests require that much more time?

Here is a screenshot of the cancelled check: https://github.com/jupyter/notebook/runs/7686750791?check_suite_focus=true

image

But on `main` they take 5 minutes to complete so still plenty of room: https://github.com/jupyter/notebook/runs/7686282650?check_suite_focus=true

image

brichet:
> Hmm normally 10 minutes should be more than enough? Or do the new tests require that much more time?

Not that much, they should be really quick, only opening left (toc) and right (notebooktools) panels.

jtpio:
Maybe there is something that makes them time out on CI (normally locally as well) with the new changes?

brichet:
The timeout is quite long, 4 minutes :
https://github.com/jupyter/notebook/blob/5f1980e8456bfa1dda87e23789c0d626f85646e2/ui-tests/playwright.config.ts#L5
When a test fail 2 times (it is retried 1 time) while awaiting a specific screenshot or element, it almost timeout the whole tests.
Is there a reason to have a 4 minutes timeout ? Galata is set to 1 minute by default.

That will not fix the tests but may allow us to get the errors instead of cancelled test.

packages/application/src/app.ts
```diff
@@ -37,6 +37,11 @@ export class NotebookApp extends JupyterFrontEnd<INotebookShell> {
         this.registerPlugin(plugin);
       }
     }
+
+    this.restored = this.shell.restored
+      .then(() => undefined)
+      .catch(() => undefined);
```
jtpio, line 43: Is this needed? Or could `this.shell.restored.then()` be directly used on the line below?

jtpio:
Looks like the `-darwin` snapshot should be removed?

image

packages/application-extension/src/index.ts
```diff
@@ -556,6 +590,114 @@ const topVisibility: JupyterFrontEndPlugin<void> = {
...
+          return false;
+        }
+
+        return currentWidget.id === (args['id'] as string);
+      }
+    });
+
+    app.restored.then(() => {
+      // Create a menu entry for left and right panel and update it with current widgets
+      if (menu) {
```
jtpio, line 676: Maybe these entries could also be added to the command palette? 

We might need to prepend the label with "Show" so it's easier to understand what it's about when the "Table of Contents" entry shows up as a palette entry?

image

packages/application/src/app.ts
brichet: Right, it's left from a copy and paste.
But we should keep `this.restored = this.shell.restored`. For example, the table of contents extension waits for `app` to be restored, but needs the Notebook panel to be ready. That's why the `shell` resolves the restored only when the main area has been added.

brichet:
> Looks like the `-darwin` snapshot should be removed?

Good catch

packages/application/src/app.ts
brichet: Fixed in https://github.com/jupyter/notebook/pull/6487/commits/52f10ede26493331a2db77fa08d0ce3e65470190

packages/application-extension/src/index.ts
brichet: Fixed in https://github.com/jupyter/notebook/pull/6487/commits/52f10ede26493331a2db77fa08d0ce3e65470190 and https://github.com/jupyter/notebook/pull/6487/commits/c6bc3ef2796f3ae686f038d12398b1989d2e8f4f

brichet:
The errors on tests seems not related (save status, kernel status...), maybe we can try running them again.

jtpio:
> The errors on tests seems not related (save status, kernel status...), maybe we can try running them again.

Yes these are a bit flaky sometimes.

jtpio:
Just tried on Binder and it looks good.

https://user-images.githubusercontent.com/591645/183838825-7babc575-e43b-445e-b1cb-ab132aa07553.mp4

At first it might feel a bit odd that the notebook is shifted to the side when the left or right area is opened. @hbcarlos had some good feedback about this on the original PR: https://github.com/jupyterlab/retrolab/pull/275#issuecomment-974317305

Posting it here for reference: 

> I think by default they should be hidden and only show up when the user triggers an action that requires something on the side panel. Also, regarding the design, we could divide the shell into three parts to keep the notebook always in the middle and then create the box panel for the side panels, the shell would be divided like in the following image:

> notebook

Probably we could try to change how the side panels are positioned in the `NotebookShell`. Maybe this could be tried out in a follow-up PR.

For reference Google Docs does shift the main page a bit when opening an add-on on the right side:

https://user-images.githubusercontent.com/591645/183842255-c965372f-f2cd-4a6b-b062-8a845e0b284f.mp4

Although their top area takes the full width, which reduces this feel of being "off" since the page is still centered with the top area.

image

brichet:
Nice, thanks for sharing, it would be better that way.
And as in Google Docs, we could use that space to display a button which close the panel.

brichet:
The debugger panel is also missing, but I think it can be added in a follow up PR as well, I'm not sure it will be as easy to include than the other panels, as it depend on current opened widget (`notebook`, `console`...).

jtpio:
> The debugger panel is also missing, but I think it can be added in a follow up PR as well. I'm not sure it will be as easy to include than the other panels

Agree this can be added in a different PR :+1: Hopefully the extension should already handle the active widget properly, otherwise we can submit fixes upstream in JupyterLab.

jtpio:
FYI @brichet I'm planning to do a pass on the changes, especially the ones on the shell.

Overall it looks good. I would like to experiment with the alternative side-panel approach mentioned above. Will report here when I have more updates.

brichet:
Thanks @jtpio.
Let me know if you want me to give this alternative side panel a try.

jtpio:
Here is what the alternative approach looks like with some local changes (to be committed and pushed):

https://user-images.githubusercontent.com/591645/185116801-0573fad4-125d-40b1-b7eb-2d7400223025.mp4

This feels a bit more similar to what Google Docs does, keeping the notebook centered with the top area.

If that looks good to you @brichet I'll push the commit to the branch directly. There are a couple of things we might want to discuss in the implementation of the shell, I'll leave some comments right after.

jtpio (commented):
Posting some comments mostly focused on the changes in `NotebookShell` for now, after looking at the code more closely and making some changes locally.

brichet:
> Here is what the alternative approach looks like with some local changes (to be committed and pushed)

Look perfect, thanks @jtpio.
Maybe we could add a `X` on panel to easily close it, but it can be done in a follow-up PR.
Fell free to push the commit to the branch. Do I have to give you permissions on my fork ?

packages/application/src/shell.ts
```diff
@@ -40,39 +50,98 @@ export class NotebookShell extends Widget implements JupyterFrontEnd.IShell {
...
-    const menuWrapper = (this._menuWrapper = new Panel());
     menuWrapper.id = 'menu-panel-wrapper';
     menuWrapper.addWidget(this._menuHandler.panel);
 
-    BoxLayout.setStretch(topWrapper, 0);
-    BoxLayout.setStretch(menuWrapper, 0);
+    BoxLayout.setStretch(this._topWrapper, 0);
+    BoxLayout.setStretch(this._menuWrapper, 0);
+
+    if (this.sidePanelsVisible()) {
```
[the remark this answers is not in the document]
brichet: I agree, that's unnecessary duplication in our case and can be blocking for other extensions.

packages/application/src/shell.ts
```diff
@@ -289,4 +529,407 @@ namespace Private {
...
+          type: 'submenu',
+          submenu: menu
+        });
+      }
+    }
+
+    _entryLabel: string;
+    _command: string;
+    _mainMenuEntry: IRankedMenu;
+    _commandRegistry: CommandRegistry;
```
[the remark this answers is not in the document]
brichet: > Maybe there is a way to construct this logic at the plugin level instead.

Right, I'll look at it.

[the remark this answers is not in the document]
brichet: I'm not sure if this is the best implementation but the logic about Menu has been moved to the plugin in https://github.com/jupyter/notebook/pull/6487/commits/00d131d8b02543abed19f017aab11b743456120e

packages/application/src/shell.ts
```diff
@@ -289,4 +491,339 @@ namespace Private {
...
+     */
+    addUpdateMenuFn(updateMenu: SideBarPanel.UpdateSideBarMenuFn): void {
+      this._updateMenu = updateMenu;
+      this.updateMenu();
+    }
+
+    /**
+     * Associate palette entries to the sidebar, and update it with the current widgets.
+     */
+    createPaletteEntry(options: SideBarPaletteOption): void {
```
jtpio, line 554: Wondering if it would also be possible to move the command palette related logic to the plugin level as well?

It looks like the palette and the shell shouldn't be too coupled to allow for greater flexibility in how the plugins are composed (or if some are disabled by end users). There could be applications without palettes or menus and having references to these in the shell implementation feels odd.

So leaving the shell handle and track the widgets with a stacked panel. And the plugin build the appropriate menu and palette entries.

packages/application-extension/src/index.ts
```diff
@@ -556,6 +597,158 @@ const topVisibility: JupyterFrontEndPlugin<void> = {
...
+        });
+      }
+
+      return disposableMenu;
+    };
+
+    app.restored.then(() => {
+      // Add the notebook tools in right area.
+      if (notebookTools) {
+        notebookShell.add(notebookTools, 'right');
```
jtpio, line 727: It looks like this shouldn't be in the `@jupyter-notebook/notebook-extension` extension rather than `@jupyter-notebook/application-extension` if we want to include it by default?

jtpio:
Looking into adding the debugger extension as well by default. This seems to be handled nicely after adding the dependencies to `app/package.json` and the plugins to `app/index.js`:

image

However the menu entry seems to be missing the title which is likely to be an issue in the debugger extension in JupyterLab.

packages/application/src/shell.ts
brichet: Yes, maybe we could add a module in the extension with this `sideBarPalette`, and pass a palette update function to the Shell, like for the Menu.

packages/application-extension/src/index.ts
brichet: I don't understand what you mean. What is in the `@jupyter-notebook/notebook-extension` ?

packages/application-extension/src/index.ts
```diff
@@ -91,6 +115,13 @@ namespace CommandIDs {
   export const resolveTree = 'application:resolve-tree';
 }
 
+/**
+ * Are the left and right panels available on the current page?
+ */
+const sidePanelsEnabled: () => boolean = () => {
+  return PageConfig.getOption('notebookPage') === 'notebooks';
+};
+
```
brichet, line 124: Currently, side panels are only available on Notebook page, but they will be available on other pages.
So perhaps this function should be renamed `isNotebookPage`, and the function `sidePanelsEnabled` will return `isNotebookPage`, at first.
This way we can include `NotebookTools` only for the notebook page, for example.

packages/application/src/shell.ts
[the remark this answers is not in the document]
fcollonval: Definitely :100: for this especially since https://github.com/jupyter/notebook/pull/6315 is using `/edit`

packages/application-extension/src/index.ts
jtpio: Here are the plugins more related to the notebook than general: https://github.com/jupyter/notebook/blob/main/packages/notebook-extension/src/index.ts

brichet: OK, the "shouldn't" misled me.
https://github.com/jupyter/notebook/pull/6487#pullrequestreview-1081591262 is useless as it was the same idea but in `@jupyter-notebook/application-extension`

jtpio: Ah sorry for the typo, I corrected it in the message above.

packages/application-extension/src/index.ts
brichet: solved with https://github.com/jupyter/notebook/pull/6487#pullrequestreview-1081113649

packages/application/src/shell.ts
```diff
@@ -289,4 +508,262 @@ namespace Private {
...
+    private _panel: Panel;
+    private _widgetPanel: StackedPanel;
+    private _current: Widget | null;
+    private _lastCurrent: Widget | null;
+    private _updated: Signal<SideBarHandler, void> = new Signal(this);
+    private _menuEntryLabel: string;
+    private _disposableMenu: IDisposable | null = null;
+    private _updateMenu: SideBarPanel.UpdateSideBarMenuFn | null = null;
+    private _addPaletteEntry: SideBarPanel.AddPaletteEntryFn | null = null;
+    private _removePaletteEntry: SideBarPanel.RemovePaletteEntryFn | null = null;
```
jtpio, line 767: Wondering if there is a way to make these more generic and not specific to the menu and command palette only?

It looks like the plugin code (in `@jupyter-notebook/application-extension`) interacting with the menu and command palette need to know when a new widget is added or removed from the shell? So maybe exposing such generic signals might be enough?

brichet:
The tests are failing because there is no error thrown when adding a widget in side panel on tree view :
https://github.com/jupyter/notebook/runs/8039561039?check_suite_focus=true#step:5:140

Now the side panels are available in any page, I think these tests should be removed, right ?

jtpio:
> Now the side panels are available in any page, these tests should be changed, right ?

yes that sounds right.

packages/application/src/shell.ts
brichet: Agree, thanks for the suggestion.
https://github.com/jupyter/notebook/pull/6487/commits/dd46b05259d0894dc81498511f44dc62b53cf80e includes the use of signals.

ui-tests/test/notebook.spec.ts
```diff
@@ -91,4 +91,47 @@ test.describe('Notebook', () => {
...
+    );
+    await page.goto(`notebooks/${tmpPath}/${notebook}`);
+
+    await waitForKernelReady(page);
+
+    await page.menu.open(menuPath);
+
+    await page.isVisible('#notebook-tools.jp-NotebookTools');
+
+    await page.isVisible('#notebook-tools.jp-NotebookTools > #add-tag.tag');
```
jtpio, line 135: Thanks for adding a test here this is going to be useful to catch potential regressions :+1: 

Wondering whether we should also add new reference snapshots to make sure the side panels render correctly?

packages/application/src/shell.ts
jtpio: Nice thanks I'll check it out but this is looking good.

ui-tests/test/notebook.spec.ts
jtpio: This could also be done in a follow-up PR.

brichet:
> Not sure what's happening with the UI tests currently failing on this PR.

It seems that the selector does not select the correct button anymore. Two more buttons with same text are added in left panel (but not displayed) from the *running-extension* package in jupyter : https://github.com/jupyterlab/jupyterlab/blob/3b83ceee09ce39ed3bf7de4e250135e5749cf3e4/packages/running-extension/src/index.ts#L78

This raises 2 questions :
- Do we need the *running-extension* package ? It seems that we need the *running* package to build the main panel widget, but not sure about the extension.
- Why isn't the left panel added to the `View` menu when the extension widget is added to the left panel ?

ui-tests/test/notebook.spec.ts
brichet: Added snapshots on *toc* and *NotebookTools* panels in https://github.com/jupyter/notebook/pull/6487/commits/7062af6cda1bbcab4d92d17f27a62db275596192

yarn.lock
```diff
@@ -3301,126 +3589,342 @@
...
+    "@lumino/disposable" "^1.10.2"
+    "@lumino/domutils" "^1.8.2"
+    "@lumino/dragdrop" "^1.14.1"
+    "@lumino/keyboard" "^1.8.2"
+    "@lumino/messaging" "^1.10.2"
+    "@lumino/properties" "^1.8.2"
+    "@lumino/signaling" "^1.10.2"
+    "@lumino/virtualdom" "^1.14.2"
+
+"@lumino/widgets@^2.0.0-alpha.1", "@lumino/widgets@^2.0.0-alpha.4":
```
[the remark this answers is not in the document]
brichet: Yes, it seems there are some conflicts in the Lumino packages. 
We should wait for the update to Lumino 2 https://github.com/jupyter/notebook/pull/6539

jtpio:
FYI @brichet I'll be pushing a couple of commits to the branch directly.

jtpio:
https://github.com/jupyter/notebook/pull/6487/commits/d5e9cec53de9fed5b120dbb0df08aa1a31e30ef8 adds the debugger sidebar to the right area:

image

jtpio:
> I was just wondering if the `SideBarHandler` and the `SideBarPalette` classes should be move to a `sidepanel.ts` module, for more readability ? I can handle it if necessary.

Probably yes :+1:  We might also want to do another pass on the exported names and methods for the shell and sidebar handlers, to make sure only the minimum is exposed (keep the API surface as small as possible).

But we can track that separately in a follow-up issue since the PR is already pretty big.

jtpio (approved):
Nice thanks @brichet for the hard work and your patience!

We can track follow-up improvements and fixes discussed in the PR in separate issues.

PR #6489, opened 2022-07-26
Automated Changelog Entry for 6.4.13 on 6.5.x

echarles:
Automated Changelog Entry for 6.4.13 on 6.5.x
```
Python version: 6.4.13
npm version: jupyter-notebook-deps: 4.0.0
```

After merging this PR run the "Full Release" Workflow on your fork of `jupyter_releaser` with the following inputs
| Input  | Value |
| ------------- | ------------- |
| Target | jupyter/notebook  |
| Branch  | 6.5.x  |
| Version Spec | minor |
| Since | 6.4.12 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,33 @@ Use `pip install pip --upgrade` to upgrade pip. Check pip version with
 
 <!-- <START NEW CHANGELOG ENTRY> -->
 
+## 6.4.13
+
+([Full Changelog](https://github.com/jupyter/notebook/compare/6.4.12...87d57658aaeccaffb5242a3b7b95702636922e8c))
+
+### Maintenance and upkeep improvements
+
+- Selenium test updates [#6484](https://github.com/jupyter/notebook/pull/6484) ([@ericsnekbytes](https://github.com/ericsnekbytes))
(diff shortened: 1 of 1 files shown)


echarles:
Closing as we should not have 6.4.13 version here.

jtpio:
> Automated Changelog Entry for 6.4.13 on 6.5.x

Is that intended? Or should it be on the `6.4.x` branch? Or a `6.5.0` pre-release?

echarles:
I wanted to cut the first 6.5.0 release from 6.5.x branch. I have run the release from 6.5.x branch and set `minor`. Should I change https://github.com/jupyter/notebook/blob/6.5.x/notebook/_version.py#L8 and set 6.5.0.dev0 there?

echarles:
... another way would be to explicitly state `Version Spec = 6.5.0` WDYT?

jtpio:
It could be that the releaser setup on `6.5.x` does not support `minor` as a version spec and defaulted to `next` / `patch`?

echarles:
> ... another way would be to explicitly state Version Spec = 6.5.0 WDYT?

I finally applied that option which worked fine https://github.com/jupyter/notebook/pull/6490

6.5.0a0 is released.

> It could be that the releaser setup on 6.5.x does not support minor as a version spec and defaulted to next / patch

Maybe. The setup of 6.5.x has not been touched and is prolly the same as the one of 6.4.x branch.

PR #6496, opened 2022-08-01
Add token on tree widget

brichet:
Fixes https://github.com/jupyter/notebook/issues/6410

The `NotebookShell` allows only one widget in the main area.
As the tree view is a `TabPanel`, it would be interesting to be able to add widgets (as it was possible in Classic Notebook).

This PR adds a token on the `NotebookTreeWidget`, the widget in main area in tree view.
Any extension will now be able to optionally try to retrieve the `INotebookTree` in the `activate` function, and add widgets in it using the `TabPanel`'s interface functions.

It should close https://github.com/jupyter/notebook/pull/6478 as a more generic implementation.


packages/tree/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@jupyter-notebook/tree",
+  "version": "7.0.0-alpha.5",
+  "description": "Jupyter Notebook - Tree ",
```
jtpio, line 4: Trailing whitespace spotted:

```suggestion
  "description": "Jupyter Notebook - Tree",
```

jtpio:
Thanks @brichet for working on this.

Diff looks good. CI is currently broken because of `ypy-websocket`, pending a new pre-release of JupyterLab 4.0.

brichet:
I called the widget and token `NotebookTree`, but I wonder if it should be something like `NotebookMain` or `NotebookRoot` to avoid confusion (for me, `tree` refers to tree files widget but I could be wrong).

jtpio:
> I called the widget and token NotebookTree

Right, this naming mostly comes from the name of the `/tree` endpoint.

Probably it's fine to keep that way for now while in alpha. We can always revisit later and maybe we could also rename the `tree-extension` as well.

packages/tree-extension/src/index.ts
```diff
@@ -27,11 +27,13 @@ import { ITranslator } from '@jupyterlab/translation';
...
-  runningIcon,
-  TabBarSvg
+  runningIcon
 } from '@jupyterlab/ui-components';
 
-import { Menu, MenuBar, TabPanel } from '@lumino/widgets';
+import { Menu, MenuBar } from '@lumino/widgets';
+
+import { NotebookTreeWidget } from '@jupyter-notebook/tree';
+import { INotebookTree } from '@jupyter-notebook/tree';
```
jtpio, line 36: Looks like these two imports could be combined?

packages/tree/src/notebook-tree.ts
```diff
@@ -0,0 +1,16 @@
+import { TabBarSvg } from '@jupyterlab/ui-components';
+
+import { TabPanel } from '@lumino/widgets';
+
+import { INotebookTree } from './token';
+
+export class NotebookTreeWidget extends TabPanel implements INotebookTree {
```
jtpio, line 7: Let's add a docstring to the class and `constructor` for consistency with the rest of the code base?

jtpio:
The UI tests checks still appear to be cancelled after restarting them:

```
yarn run v1.22.17
$ playwright test --browser firefox

Running 24 tests using 1 worker
Error: The operation was canceled.
```

brichet:
Thanks for reviewing @jtpio.

brichet:
A part of the tests are currently failing because of :
```
Failed to load toolbar items for factory FileBrowser from 
Error: Plugin `id` parameter is required for settings fetch.
```
So the tree widget toolbar is not displayed and filled with `Upload` and `New` buttons as expected.

packages/tree-extension/src/index.ts
```diff
@@ -93,10 +94,87 @@ const createNew: JupyterFrontEndPlugin<void> = {
   }
 };
 
+function activateNotebookTreeWidget(
```
jtpio, line 97: Any specific reason to move the activate function to a separate function?

Maybe we can keep inlined as part of the plugin definition for consistency with the rest of the code base?

brichet: Not really, I reverted it

jtpio:
Thanks @brichet this is looking good.

I wanted to check on Binder but it seems like it doesn't start v7 anymore, which also happens on `main`: 

image
 
Probably something changed in the setup (maybe the hatch migration), I'll have a look in a separate PR.

jtpio:
> Probably something changed in the setup (maybe the hatch migration), I'll have a look in a separate PR.

This should be fixed by https://github.com/jupyter/notebook/pull/6505.

brichet:
> > Probably something changed in the setup (maybe the hatch migration), I'll have a look in a separate PR.
> 
> This should be fixed by #6505.

The binder link is created only when the PR opens the first time I think

jtpio:
> The binder link is created only when the PR opens the first time I think

Right but it doesn't change, so we can open it again or refresh the page.

jtpio (approved):
Thanks!

PR #6497, opened 2022-08-01
Update versioning mechanism and pin jupyter_client<7

RRosio:
- Use version parsing from 6.5.x branch to construct the version tuple here and allow use of 'a', 'b' and 'rc' strings in version.
- Pin jupyter_client<7 to avoid asyncio runtime error outlined [here](https://github.com/jupyter/notebook/issues/6164).

### DIFF
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -2,12 +2,15 @@
 store the current version info of the notebook.
 
 """
+import re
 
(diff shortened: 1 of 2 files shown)


RRosio:
I realize I changed the version maybe a bit prematurely here... Should I wait to make that a separate commit or would this be okay?

echarles (approved):
LGTM Thx @RRosio

PR #6505, opened 2022-08-04
Uninstall `nbclassic` on Binder so Notebook v7 can load

jtpio:
As noted in https://github.com/jupyter/notebook/pull/6496#issuecomment-1205422753, Binder does not start v7 anymore:

image

This is currently happening on the `main` branch. Opening this PR to troubleshoot the issue.

### DIFF
--- a/binder/postBuild
+++ b/binder/postBuild
@@ -2,6 +2,10 @@
 set -euo pipefail
 
 python -m pip install -e . --force-reinstall
+
+# TODO: remove when it's possible to install nbclassic next to Notebook v7
+# without nbclassic shadowing the v7 endpoints
 python -m pip uninstall nbclassic -y
+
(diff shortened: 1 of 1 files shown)


jtpio:
Arf Binder seems to be getting some rate limiting errors at the moment, oh well let's wait then...

image

jtpio:
A guess might be it's related to the recent `nbclassic` releases which also use the same `/tree` endpoint.

jtpio:
> A guess might be it's related to the recent `nbclassic` releases which also use the same `/tree` endpoint.

Yes, explicitly uninstalling `nbclassic` (which might be coming from the base Binder / repo2docker setup) in the `postBuild` does the trick:

image

jtpio:
Merging since this fixes the Binders for previewing PRs as well.

PR #6515, opened 2022-08-19
Add a micromamba-based devcontainer

jtpio:
Add a micromamba-based devcontainer to the repo so it's easier to make contributions from the web browser with GitHub Codespaces.

### TODO

- [x] Add devcontainer files
- [ ] Fix git setup
- [ ] Add section to the contributing guide


jtpio:
Not sure we need micromamba actually. The default Codespaces image includes many tools including Python already.

Will check if the setup can be simplified.

jtpio:
Giving up for now.

Issue #6519, opened 2022-08-23
Update UI tests snapshots for nbconvert 7


jtpio:
nbconvert 7 adds new exporters which makes the UI tests fail as noticed in: https://github.com/jupyter/notebook/actions/runs/2910072295

![image](https://user-images.githubusercontent.com/591645/186123426-50a73def-e6d3-45a5-a528-671bf3acfa9b.png)

PR #6520, opened 2022-08-23
Update reference snapshots

jtpio:
Fixes https://github.com/jupyter/notebook/issues/6519


jtpio:
The remaining failing check is not related, and similar to https://github.com/jupyterlab/jupyterlab/issues/12991

PR #6522, opened 2022-08-23
Using group method to match regex items 

### LINKED ISSUE #6521: 5.7.14 breaks installation of notebook extensions.
**Describe the bug**
The version 5.7.14 for python2breaks installation of notebook extensions. Running  `jupyter contrib nbextension install --user` fails with `TypeError: '_sre.SRE_Match' object has no attribute '__getitem__'`

**To Reproduce**

1. Create a venv: 
```bash
pyenv shell 2.7.18
virtualenv -p `which python` env2
source env2/bin/activate
```
2. Install jupyter: 
```bash
pip install jupyter
pip install jupyter-nbextensions-configurator==0.4.1
pip install jupyter_contrib_nbextensions==0.5.1
```
3. Run `jupyter contrib nbextension install --user`

**Expected behavior**
Running `jupyter contrib nbextension install --user` should succeed.

**Additional context**
Reinstalling 5.7.13 (`pip install notebook==5.7.13`) resolve the issue. My understanding is that this [commit](https://github.com/jupyter/notebook/commit/986c88317ddc3aa67a2244d597604626836b4610) introduce an issue. The following code 
```python
import re
__version__ = '5.7.14'
# Build up version_info tuple for backwards compatibility
pattern = r'(?P<major>\d+).(?P<minor>\d+).(?P<patch>\d+)(?P<rest>.*)'
match = re.match(pattern, __version__)
parts = [int(match[part]) for part in ['major', 'minor', 'patch']]
if match['rest']:
    parts.append(match['rest'])
version_info = tuple(parts)
```
It returns `TypeError: '_sre.SRE_Match' object has no attribute '__getitem__'` like in the described bug. According to [this](https://stackoverflow.com/questions/52848239/python-typeerror-sre-sre-match-object-has-no-attribute-getitem) this syntax was only added with python3.6.
Replacing  the code by the following one works.
```python
import re

__version__ = '5.7.14'
# Build up version_info tuple for backwards compatibility
pattern = r'(?P<major>\d+).(?P<minor>\d+).(?P<patch>\d+)(?P<rest>.*)'
match = re.match(pattern, __version__)
parts = [int(match.group(part)) for part in ['major', 'minor', 'patch']]
if match.group('rest'):
    parts.append(match.group('rest'))
version_info = tuple(parts)
```

RRosio:
Works for Python versions < 3.6. Fixes #6521 so users can install extensions as noted by @GaspardBT

### DIFF
--- a/notebook/_version.py
+++ b/notebook/_version.py
@@ -10,7 +10,7 @@ __version__ = '5.7.14.dev'
 # Build up version_info tuple for backwards compatibility
 pattern = r'(?P<major>\d+).(?P<minor>\d+).(?P<patch>\d+)(?P<rest>.*)'
 match = re.match(pattern, __version__)
-parts = [int(match[part]) for part in ['major', 'minor', 'patch']]
-if match['rest']:
-    parts.append(match['rest'])
+parts = [int(match.group(part)) for part in ['major', 'minor', 'patch']]
+if match.group('rest'):
+    parts.append(match.group('rest'))
 version_info = tuple(parts)


echarles (approved):
LGTM. Thx @RRosio

PR #6523, opened 2022-08-24
Troubleshoot failing `macos` check on CI

jtpio:
Trying to fixing the `macos` failure on CI with mismatched Python versions.


jtpio:
The trick was to switch from `bash -l {0}` to `bash -e {0}`.

PR #6549, opened 2022-10-01
fix activate environment according to mamba documentation

buddhics:
According to mamba [documentation](https://mamba.readthedocs.io/en/latest/user_guide/mamba.html#mamba-vs-conda-clis), 

image

`mamba activate notebook` has to be changed to `conda activate notebook`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ The `jlpm` command is JupyterLab's pinned version of [yarn](https://yarnpkg.com/
 mamba create -n notebook -c conda-forge python nodejs -y
 
 # activate the environment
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @buddhics.

Normally `mamba activate` should now be supported in the latest versions. There are some mentions to it in the changelog: https://github.com/mamba-org/mamba/blob/main/CHANGELOG.md#20220211

At least it works on Linux for sure (using it). Maybe the mamba docs should be updated?

buddhics:
> 

You're correct @jtpio ,

It seems like, Mamba is still sorting out its issues with `mamba activate`. We can ignore this issue since this is working on Linux.

Issue #6532, opened 2022-09-02
No mouse pointer when hovering on the Jupyter icon


jtpio:
With the latest 7.0 pre-release hovering on the Jupyter icon in the top area does not show the mouse pointer:

![image](https://user-images.githubusercontent.com/591645/188081580-4759821a-c07d-45be-b3aa-069dc555c160.png)

https://user-images.githubusercontent.com/591645/188081975-b57caccb-005a-4311-b7f1-b3c4522f291b.mp4

I think it used to be, so probably a regression that happened at some point.

dhbsbdhdhd:
Hey! I am a first time contributor.I would like this issue to be my first contribution. Can you guide me more about the issue so that I can work on it?

MuhammadNoorani:
Is this issue open,
I would like to contribute to it, would love to collaborate and solve this issue. So, can I go ahead

AaronB3110:
Hello! I am a beginner contributor and I would love to contribute to this issue if it has not been worked on.
I would love guidance through it and work on it.

harsh788:
Hello there! I am new to the open source world and would like to contribute to some real-world projects. If this issue is not resolved yet, can I be assigned to it?

jtpio:
@AaronB3110 @harsh788 @MuhammadNoorani yes the issue is still open, feel free to work on it and open a PR, thanks!

Grimoors:
Hi! This looks like a good first issue. I am new to open source and want to try to contribute to this. May I try to work on this issue as well? (@jtpio )

jtpio:
Sure, whoever wants to work on this please feel free to open a PR, thanks!

PR #6550, opened 2022-10-03
Fix: No mouse pointer when hovering on the Jupyter icon

hawkeyes21:
Fix for issue #6532

Quick context: 
The cursor was not changing into a pointer when hovered over Jupyter icon on navbar.

https://user-images.githubusercontent.com/26214265/193533886-8450bb4b-83c0-4eef-b212-42b75f23df15.mp4


jtpio (approved):
Thanks!

PR #6579, opened 2022-10-15
Update to lerna 6

jtpio:
Update to the latest lerna version: https://github.com/lerna/lerna/releases/tag/v6.0.0

This new major version includes some task pipeline configuration by default now (via `nx`), which could be leveraged in this repo to speed up the build and enhance the developer experience: https://lerna.js.org/docs/concepts/task-pipeline-configuration

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,6 @@
 {
   "npmClient": "yarn",
+  "useNx": false,
   "version": "independent",
   "useWorkspaces": true
 }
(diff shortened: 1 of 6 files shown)


jtpio:
Looks like some changes to the releaser actions lead to `GITHUB_OUTPUT` being create on disk:

```
$ node ./buildutils/lib/release-bump.js --force --skip-commit next
> git status
On branch main
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	GITHUB_OUTPUT
```

image

lerna.json
```diff
@@ -1,5 +1,6 @@
 {
   "npmClient": "yarn",
+  "useNx": false,
```
jtpio, line 3: This seems required for now as it seems to be breaking the releaser `check-npm` step currently.

Will track that in a different issue.

PR #6581, opened 2022-10-17
Maintenance cleanup

blink1073:
- Use `Flake8-pyproject` to avoid needing a `.flake8` file
- Use mdformat instead of prettier for markdown files
- Update to releaser v2 
- Add pyproject validation
- Rename readthedocs config file with recommended name
- Other workflow cleanup


jtpio (approved):
Thanks!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,8 +1,6 @@
----
-name: Is this a bug in Notebook? Open an issue.
-about: If you're not sure, feel free to post your question on Jupyter's Discourse channel.
-labels: bug
----
+______________________________________________________________________
+
+## name: Is this a bug in Notebook? Open an issue. about: If you're not sure, feel free to post your question on Jupyter's Discourse channel. labels: bug
```
jtpio, line 3: Ah just noticed this formatting seems to be breaking the issue template.

PR #6585, opened 2022-10-17
Update `actions/checkout` and `actions/download-artifact`

jtpio:
This should help fix some of the following warnings:

image


jtpio:
Will do a follow-up to add a dependabot file and trigger these updates automatically.

PR #6588, opened 2022-10-17
Tree view enhancement

brichet:
This PR improves the tree view by:
- ~~activating by default all newly created tabs.~~
- adding a command to open the *file browser* tab. This is useful when another tab send the command `go-to-path`, now the *file browser* is activated after the path has changed.


packages/application-extension/src/index.ts
```diff
@@ -264,12 +264,14 @@ const pages: JupyterFrontEndPlugin<void> = {
       }
     });
 
-    app.commands.addCommand(CommandIDs.openTree, {
-      label: trans.__('Open Files'),
-      execute: () => {
-        window.open(`${baseUrl}tree`);
-      }
-    });
+    if (PageConfig.getOption('notebookPage') !== 'tree') {
```
jtpio, line 267: Wondering if we need this check, or simply let folks open the file browser again if they want?

brichet: In fact the `filebrowser-extension` already adds an entry in menu View:
https://github.com/jupyterlab/jupyterlab/blob/fc3f6ab716ef82944ab6c4e5d1a06418daad49dc/packages/filebrowser-extension/schema/browser.json#L24-L29
That would be a duplicated entry, one to switch to file browser tab, and one to open a new page.
This is also why I renamed the *Open Files* entry to *File Browser*, for consistency with the JupyterLab one.

packages/tree/src/notebook-tree.ts
```diff
@@ -19,4 +19,11 @@ export class NotebookTreeWidget extends TabPanel implements INotebookTree {
     });
     this.addClass('jp-TreePanel');
   }
+
+  addWidget(widget: Widget, activate = true): void {
```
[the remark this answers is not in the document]
brichet: No, it can be let to the plugin. I added it because I was thinking that creating a new tab was to activate it (except for the initial ones). I'll remove it.

packages/application-extension/src/index.ts
brichet: Thanks, I didn't know about that one.
I just realized that clicking on the same menu entry `File Browser` when the file browser is activated try to call `filebrowser:hide`, which is not currently defined in notebook. This raises an error in console.
Do you think we should defined the command `filebrowser:hide` without any action, to avoid the error ?

jtpio: Does it? I just tried with this menu entry on Binder but it does not seem to be printing anything:

image

brichet: On binder too I have this error message in the console, when the file browser is already activated before clicking on the link:



The command added by JupyterLab is `filebrowser:toggle-main`, which calls `filebrowser:activate` or `filebrowser:hide-main` depending on the `isHidden` status. Maybe it is not triggered every time for some reason.

BTW I changed the way this command is defined (using `isVisible`) in https://github.com/jupyter/notebook/pull/6588/commits/8559a4df20702e7a784f3f5aae4bb2d8f37193a0

brichet: > Do you think we should defined the command `filebrowser:hide` without any action, to avoid the error ?

We could also make the file browser tab 'closeable' and define properly the command `filebrowser:hide`.
Not sure this is expected for *Jupyter Notebook*, that may change a bit the philosophy.

packages/application-extension/src/index.ts
```diff
@@ -264,17 +272,25 @@ const pages: JupyterFrontEndPlugin<void> = {
       }
     });
 
-    app.commands.addCommand(CommandIDs.openTree, {
-      label: trans.__('Open Files'),
-      execute: () => {
-        window.open(`${baseUrl}tree`);
-      }
-    });
+    if (!app.commands.isVisible('filebrowser:toggle-main')) {
```
jtpio, line 275: I was more wondering whether it would be possible to define `isVisible` as part of the command definition below.

packages/tree-extension/src/index.ts
```diff
@@ -94,6 +94,28 @@ const createNew: JupyterFrontEndPlugin<void> = {
...
+  id: '@jupyter-notebook/tree-extension:open-file-browser',
+  requires: [INotebookTree, IFileBrowserFactory],
+  autoStart: true,
+  activate: (
+    app: JupyterFrontEnd,
+    notebookTree: INotebookTree,
+    factory: IFileBrowserFactory
+  ) => {
+    const { commands } = app;
+    commands.addCommand('filebrowser:activate', {
```
jtpio, line 110: Defining a `filebrowser:activate` command to activate the filebrowser sounds fine though :+1:

packages/application-extension/src/index.ts
jtpio: @brichet would you mind posting a summary of what the intended behavior should be, to make sure I understand correctly? Thanks!

brichet: I may be confused about that, I'll try to clarify.

There are 2 different things:

1- What I was expected at first was a command to activate the file browser in tree view from another tab. For example nbgrader brings some tabs with links which open a path in file browser, using *filebrowser:go-to-path* command.  This command already exists in tree view, it is created when importing the plugin '@jupyterlab/filebrowser-extension:browser'. But at the end, this command call the command *filebrowser:activate*, which is not registered in Notebook yet. The result is that the path is opened but the file browser tab is not activated, so it look like nothing happened for user. 
This is fixed with the command below.

2- The plugin '@jupyterlab/filebrowser-extension:browser' also bring a command *filebrowser:toggle-main*, with a menu entry called `Open Files`, which call *filebrowser:activate* or *filebrowser:hide-main* depending on the file browser status. I thought we could use this entry already existing in tree view to open the file browser tab, instead of creating a new one which may be confusing.

Maybe the good thing to do is:
- modifying the creation of the command *filebrowser:toggle-main* in JupyterLab, to register it (and create menu entry) in the plugin '@jupyterlab/filebrowser-extension:browserWidget' instead of '@jupyterlab/filebrowser-extension:browser'. this would avoid importing it in Notebook and having this menu entry by default in tree view.
- creating a menu entry to open the file browser tab in Notebook when we create the command 'filebrowser:activate'.
- creating the menu entry to open a new `tree view` page only if the previous one does not exists (as you said).

Hope it clarifies.

packages/application-extension/src/index.ts
brichet: You are right.
It can be solved by fixing it in JupyterLab, as explain below: suppress the `filebrowser:toggle-main` entry from menu and create one only for `filebrowser:activate`.

packages/application-extension/src/index.ts
```diff
@@ -262,18 +269,22 @@ const pages: JupyterFrontEndPlugin<void> = {
...
     });
+    const page = PageConfig.getOption('notebookPage');
 
     app.commands.addCommand(CommandIDs.openTree, {
-      label: trans.__('Open Files'),
+      label: trans.__('File Browser'),
       execute: () => {
-        window.open(`${baseUrl}tree`);
+        if (page === 'tree') {
+          app.commands.execute(filebrowserCommandIDs.activate);
```
jtpio, line 278: To avoid bringing a dependency on `@jupyter-notebook/tree` for `@jupyter-notebook/application-extension`, maybe we should just use the `filebrowser:activate` string here directly?

jtpio: For example this is what is done in the JupyterLab `docmanager-extension`, using a command defined in another plugin: https://github.com/jupyterlab/jupyterlab/blob/d614e293eb4b90055307828d5e81c9661e1147c9/packages/docmanager-extension/src/index.tsx#L1125

packages/tree/src/notebook-tree.ts
```diff
@@ -4,6 +4,14 @@ import { TabPanel } from '@lumino/widgets';
 
 import { INotebookTree } from './token';
 
+/**
+ * The namespace for command IDs.
+ */
+export namespace CommandIDs {
```
jtpio, line 10: With https://github.com/jupyter/notebook/pull/6588/files#r1067176298 above, these `CommandIDs` could be moved to the `tree-extension` package.

packages/application-extension/src/index.ts
brichet: Sure, perhaps we should do the same in `tree-extension:openFileBrowser` and drop the `CommandIDs` namespace in `tree`.

packages/application-extension/package.json
```diff
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@jupyter-notebook/application": "^7.0.0-alpha.10",
+    "@jupyter-notebook/tree": "^7.0.0-alpha.10",
```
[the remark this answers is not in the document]
brichet: Good catch, thanks

packages/application-extension/src/index.ts
```diff
@@ -242,6 +242,11 @@ const menuSpacer: JupyterFrontEndPlugin<void> = {
 
 /**
  * Add commands to open the tree and running pages.
+ *
+ * ## NOTES:
+ * The optional token IFileBrowserCommands is useful to ensure the corresponding
```
jtpio, line 247: Also this comment can probably be removed now.

jtpio (approved):
Thanks!

brichet:
Thanks a lot for reviews @jtpio, especially since this PR was pretty confusing at first.

jtpio:
FYI @brichet a new pre-release is now available, which includes this change: https://github.com/jupyter/notebook/releases/tag/v7.0.0a11

PR #6580, opened 2022-10-16
Ensure custom preload is correctly handled

echarles:
Follow-up of https://github.com/jupyter/nbclassic/pull/155

The custom preload needs to be handled in `page.html` and `build-main.js`.

We also bump and pin to `nbclassic==0.4.6`.

PR #6570, opened 2022-10-13
fix: pin temporary to nbclassic 0.4.5

echarles:
Pin nbclassic to 0.4.5

PR #6593, opened 2022-10-20
Depend on nbclassic 0.4.7

echarles:
This PR upgrades the nbclassic dependency to 0.4.7.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -13,7 +13,7 @@ dependencies:
 - sphinx
 - terminado
 - myst-parser
-- nbclassic==0.4.7
+- nbclassic>=0.4.7
 - pip:
   - nbsphinx
   - Send2Trash
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -13,7 +13,7 @@ dependencies:
 - sphinx
 - terminado
 - myst-parser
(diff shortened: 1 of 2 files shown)


bnavigator:
Just wanted to let you know that from a distribution point of view this pinning and bumping out of sync is really annoying.

echarles:
@bnavigator thx for your feedback. We have today a community call (https://github.com/jupyter/notebook-team-compass) where we can discuss that. You are welcome to join, or share here more context (why, conseqences, alternatives...)

echarles:
We have discussed that during the community call and the conclusion was that we should define a lower boundary instead of a strict pin. I have just pushed a commit that defines `nbclassic>=0.4.7`

bnavigator:
Thank you. I understand that sometimes it is necessary, e.g. the pinning to 0.4.5 because 0.4.6 broke it (https://github.com/jupyter/notebook/pull/6580) But when you do, please at least add a comment and explain it in your commits/pull requests. For example, #6570 is unnecessarily opaque.

echarles:
Right, I have added a small explanation on https://github.com/jupyter/notebook/pull/6570#issuecomment-1294499684

RRosio:
LGTM, thanks @echarles!

Issue #6603, opened 2022-10-28
Provide the 'serverRoot' in jupyter-config-data for tests


brichet:
When testing extensions, it can be useful to retrieve the serverRoot from `jupyter-config-data` script in DOM.
It could be added in `page_config` with a condition: https://github.com/jupyter/notebook/blob/2f26f1b7057896946423410995a1bf63273753dc/notebook/app.py#L61-L64

A simple solution without adding a new option could be to add it if the option `exposeAppInBrowser` is True (if this is an option for debugging only).

PR #6604, opened 2022-10-28
Make the serverRoot variable available in DOM 

brichet:
This PR is a proposition to fix https://github.com/jupyter/notebook/issues/6603

It make the serverRoot path available in `#jupyter-config-data` script when the option `exposeAppInBrowser` is True.

### DIFF
--- a/notebook/app.py
+++ b/notebook/app.py
@@ -61,6 +61,8 @@ class NotebookBaseHandler(ExtensionHandlerJinjaMixin, ExtensionHandlerMixin, Jup
         server_root = self.settings.get("server_root_dir", "")
         server_root = server_root.replace(os.sep, "/")
         server_root = os.path.normpath(os.path.expanduser(server_root))
+        if app.expose_app_in_browser:
+            page_config["serverRoot"] = server_root
         try:
             # Remove the server_root from pref dir
             if self.serverapp.preferred_dir != server_root:


jtpio:
Thanks @brichet.

I checked the JupyterLab (server) code and `serverRoot` does not seem to be exposed there: https://github.com/jupyterlab/jupyterlab_server/blob/904bd9535934f210ca270c9a43bf85afdfe7c3f2/jupyterlab_server/handlers.py#L89

Although it might be in another place.

What kind of testing requires accessing `serverRoot`?

brichet:
@jtpio if I'm not wrong it is exposed here: https://github.com/jupyterlab/jupyterlab_server/blob/904bd9535934f210ca270c9a43bf85afdfe7c3f2/jupyterlab_server/handlers.py#L86
BTW, maybe it should be changed as it exposes the full path without any specific setting.

The `serverRoot` is used in UI tests (playwright) that requires to switch to the root directory of the application, to run some system command:  https://github.com/jupyter/nbgrader/blob/066e793fb973b8ed03ea222a28c81950361edc94/nbgrader/tests/labextension_ui-tests/tests/test_utils.ts#L61
That was the easiest way I found to retrieve the root directory from UI tests, but there might be other ways.

jtpio:
Alright. For now I would say it's best to not expose more options to the client unless very necessary.

PR #6609, opened 2022-11-02
Fix rename_file to handle relative paths properly (related to GHSA-v7vq-3x77-87vg?)

### LINKED ISSUE #6473: Bug when renaming file (notebook 6.4.12)
**Describe the bug**
In notebook 6.4.12，we cannot rename file sometimes.

**To Reproduce**

1. Install notebook 6.4.12
2. Start notebook with `root_dir="/home/foobar"`
3. Create a directory named `abc` (or any name with length `len(root_dir) - len("Untitled") - 1`)
4. Create a notebook with default filename 'Untitled.ipynb'
5. Rename the notebook to any other filename

Then you should got the error message 

```
File "/home/xxxxxxx/.miniconda3/lib/python3.8/site-packages/notebook/services/contents/filemanager.py",
line 579, in rename_file raise web.HTTPError(400, f'Cannot rename hidden file or directory {os_path!r}')
NameError: name 'os_path' is not defined
```

**Expected behavior**

Rename the file successfully.

According to the error message, I found a typo here (`os_path` is not defined, it should be `old_path`). .https://github.com/jupyter/notebook/blob/3e5b767ddbc3526729cadbc69fe472ffc3ec2e7e/notebook/services/contents/filemanager.py#L578-L579 And I notice that the result of `is_hidden(old_path, self.root_dir)` is true, this is because the result of `"abc/Untitled.ipynb"[len("/home/foobar"):]` is `.ipynb` and it is treated as a hidden file.

It seems that this bug is introduced by this commit https://github.com/jupyter/notebook/commit/f69eb96cff7149b38bca068f4b7faaa7baf55902 .

yacchin1205:
Fix #6473 and I assume that this completely resolves [GHSA-v7vq-3x77-87vg](https://github.com/jupyter/notebook/security/advisories/GHSA-v7vq-3x77-87vg).

As @jinzhen-lin mentions in #6473, if a path of the same length as notebook_dir is given in `rename_file`, it is interpreted as an attempt to rename a file named `.ipynb`, which is misjudged as a hidden file because the first character is `.`.
This problem could be caused by old_path and new_path of `rename_file` are relative paths, but `is_hidden` is assumed to be an absolute path. Therefore, I modified the code of `rename_file` so that absolute paths are provided to `is_hidden`.

Additionally, the test was also fixed. The test passed until now, but it did not test if `rename_file` raises the expected error. Because the `FileContentsManager.new` before it raised `HTTPError` and marks it as success. (`is_hidden` is also called in `new`.)

**[Related to the security advisory?]**
The original commit seems to be https://github.com/jupyter-server/jupyter_server/commit/877da10cd0d7ae45f8b1e385fa1f5a335e7adf1f on jupyter_server. (This commit seems to handle absolute paths properly.) This appears to be related to Security Advisory [GHSA-v7vq-3x77-87vg](https://github.com/jupyter/notebook/security/advisories/GHSA-v7vq-3x77-87vg) So I consider this PR may be an important fix.

### DIFF
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -526,13 +526,13 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         os_path = self._get_os_path(path)
         rm = os.unlink
         
+        if is_hidden(os_path, self.root_dir) and not self.allow_hidden:
+            raise web.HTTPError(400, f'Cannot delete file or directory {os_path!r}')
+
         four_o_four = "file or directory does not exist: %r" % path
 
         if not self.exists(path):
             raise web.HTTPError(404, four_o_four)
-
-        if is_hidden(os_path, self.root_dir) and not self.allow_hidden:
-            raise web.HTTPError(400, f'Cannot delete file or directory {os_path!r}')
 
         def is_non_empty_dir(os_path):
             if os.path.isdir(os_path):
--- a/notebook/services/contents/filemanager.py
+++ b/notebook/services/contents/filemanager.py
@@ -575,15 +575,15 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         if new_path == old_path:
             return
 
-        if (is_hidden(old_path, self.root_dir) or is_hidden(new_path, self.root_dir)) and not self.allow_hidden:
-            raise web.HTTPError(400, f'Cannot rename file or directory {old_path!r}')
-
         # Perform path validation prior to converting to os-specific value since this
         # is still relative to root_dir.
         self._validate_path(new_path)
 
         new_os_path = self._get_os_path(new_path)
         old_os_path = self._get_os_path(old_path)
+
+        if (is_hidden(old_os_path, self.root_dir) or is_hidden(new_os_path, self.root_dir)) and not self.allow_hidden:
+            raise web.HTTPError(400, f'Cannot rename file or directory {old_path!r}')
 
         # Should we proceed with the move?
         if os.path.exists(new_os_path) and not samefile(old_os_path, new_os_path):
--- a/notebook/services/contents/tests/test_manager.py
+++ b/notebook/services/contents/tests/test_manager.py
@@ -253,76 +253,68 @@ class TestFileContentsManager(TestCase):
 
         #Test rename behavior
         #Test rename with source file in hidden directory
-        with self.assertRaises(HTTPError) as excinfo:
-            with TemporaryDirectory() as td:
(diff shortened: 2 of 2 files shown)


RRosio:
Hi @yacchin1205, thank you for submitting this PR! One point of interest that has come up while speaking with @echarles, would it be possible to update the tests to be more in line with the testing done for this functionality in jupyter-server? After this PR is merged it would be great to have these changes added to the 6.5.x branch as well!

yacchin1205:
Thank you for your reply, @RRosio . OK, I will update the tests!

yacchin1205:
@RRosio I referred to jupyter-server and found the same problem in the test of jupyter-server, so I fixed the test and found the (new) problem 😓 .
I made a Pull Request for the issue. https://github.com/jupyter-server/jupyter_server/pull/1073 .
I will wait for it to be accepted and try to fix this Pull Request as well.

yacchin1205:
Hi @RRosio , The fixes to the test on jupyter-server have also been merged, and I have modified this test accordingly. Please review it.

RRosio:
Hi @yacchin1205 thank you for implementing these updates. I have brought up this PR in the notebook community meeting and there was interest in a little more testing on our end to ensure that these updates have no other side-effects.

yacchin1205:
Hi @RRosio , The reason why `delete_file` checks for hidden files before checking for existence, I understand, is due to a problem that allows some files (e.g. `.ssh/id_rsa`) to be checked for existence. 
(So I assume that the current implementation of [rename_file](https://github.com/jupyter/notebook/blob/6.4.x/notebook/services/contents/filemanager.py#L571) also checks whether the file is hidden first.)

If you would like to add the same fix to `delete_file` of jupyter_server here, I can add it this PR. How about it?

RRosio:
Thank you for your response @yacchin1205! 
> If you would like to add the same fix to `delete_file` of jupyter_server here, I can add it this PR. How about it?

Yes, if you could please add that fix here, that would be great!

yacchin1205:
Hi @RRosio , I have committed a fix regarding `delete_file`. Please review it!

echarles:
I have tested the latest commit and confirms the hidden files are not shown on my env and can not be addressed directly.

RRosio:
Thank you for the updates @yacchin1205! This looks great! Feel free to let us know whether you would like to submit these same updates to `6.5.x` as well.

yacchin1205:
Thank you for the merge @RRosio ! I have also created a Pull Request for 6.5.x. I would appreciate it if you could check that as well.

PR #6624, opened 2022-11-16
Use pathlib instead of os.path.join in `tests/conftest.py`

### LINKED ISSUE #6622: Modernize `os` operations in `conftest.py`
Currently in `conftest.py` there are multiple uses of `os.path.join`: https://github.com/jupyter/notebook/blob/main/tests/conftest.py

Ideally it would be great to do a pass on that file to use `pathlib`: https://docs.python.org/3/library/pathlib.html

kianelbo:
Fixes #6622

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,9 +2,9 @@ import glob
 import json
 import os
 import os.path as osp
+import pathlib
 import shutil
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

PR #6626, opened 2022-11-18
Fix issue template formatting

jtpio:
As noticed in https://github.com/jupyter/notebook/pull/6581#pullrequestreview-1185091856.


jtpio:
ok so adding the extra plugins seems to be doing the trick.

PR #6648, opened 2022-11-29
Remove PDF link to Jupyter Notebook docs

jtpio:
Try fixing the `check-links` CI failure (https://github.com/jupyterlab/maintainer-tools/issues/160).


jtpio:
Looks like it fixes it.

We can consider re-adding it later if needed.

PR #6669, opened 2022-12-23
Add spelling and docstring enforcement

blink1073:


.github/workflows/build.yml
```diff
@@ -69,6 +69,9 @@ jobs:
       - uses: actions/checkout@v3
       - uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
       - run: |
+          sudo apt-get update
+          sudo apt install enchant-2  # for spelling
```
jtpio, line 73: Does it mean we would need to install that package locally if we want to build the docs manually?

jtpio: Ah ok just saw the `try except` block below, so it's probably fine and does prevent building the docs if the package is missing.

blink1073: Yes, on Windows it works automatically but on other systems you have to separately install `enchant`, so we add this block to avoid requiring the c library on posix.

jtpio (approved):
Thanks!

PR #6675, opened 2023-01-03
Fix `check_links` and  and `Test Lint` CI failures

jtpio:
Fix broken link that is causing CI to fail in https://github.com/jupyter/notebook/pull/6674.


jtpio:
OK there seems to be another failure related to a newer version of `mypy` maybe?

image

PR #6696, opened 2023-01-16
Create `auto_author_assign.yml` workflow

jtpio:
Add a workflow similar to the one in JupyterLab to auto-assign PRs to the contributors: https://github.com/jupyterlab/jupyterlab/blob/master/.github/workflows/auto_author_assign.yml


jtpio:
Back to the openssl error:

image

PR #6700, opened 2023-01-18
Help-extension split into multiple plugins

### LINKED ISSUE #6406: Split the help extension into multiple plugins
### Problem

At the moment the help plugin is all contained in a single plugin:

https://github.com/jupyterlab/retrolab/blob/752e21fde9fbd995f2ab006a5c8650f7ae55a1f0/packages/help-extension/src/index.tsx#L42-L166

Ideally the functionalities would be split into multiple plugins, similar to JupyterLab:

https://github.com/jupyterlab/jupyterlab/blob/master/packages/help-extension/src/index.tsx

This would make it easier for downstream projects or end users to disable some of the help plugins, for example the about dialog which is being replaced in JupyterLite: 

https://github.com/jtpio/jupyterlite/blob/a198d902908483f8eea36fecae858cbd14b4a078/packages/application-extension/src/index.tsx#L63-L156 

### Proposed Solution

Split plugins in the `help-extension` package like in JupyterLab:

https://github.com/jupyterlab/jupyterlab/blob/master/packages/help-extension/src/index.tsx

RRosio:
Fix #6406


jtpio:
Thanks @RRosio for working on this!

packages/help-extension/src/index.tsx
```diff
@@ -42,20 +42,13 @@ namespace CommandIDs {
 }
 
 /**
- * The help plugin.
+ * A plugin to open the about section with resources.
  */
-const plugin: JupyterFrontEndPlugin<void> = {
+const open: JupyterFrontEndPlugin<void> = {
   id: '@jupyter-notebook/help-extension:plugin',
```
jtpio, line 48: Maybe this `id` could be `@jupyter-notebook/help-extension:open` for consistency with the other plugins?

jtpio:
```suggestion
  id: '@jupyter-notebook/help-extension:open',
```

RRosio:
Thank you @jtpio! I made this change locally however when re-installing `notebook` the Help menu option no longer contains the `About Jupyter Notebook` and `Keyboard Shortcuts` sections. I was thinking maybe I need to update the schema as well?

jtpio:
Right, probably this file would need to be renamed then: https://github.com/jupyter/notebook/blob/main/packages/help-extension/schema/plugin.json

RRosio:
Thank you! I just submitted that change.

jtpio:
Thanks!

I just tried on Binder with the latest changes and it looks good:

image

jtpio (approved):
Thanks!

Issue #6671, opened 2022-12-28
[accessibility] 


brichet:
@balajialg highlighted the fact that there have been some improvements about accessibility with Notebook 7: https://github.com/berkeley-dsep-infra/datahub/issues/3885.

He provided a list of problems found by the [Axe Decque](https://www.deque.com/axe/) test tool that should be fixed.

These remaining errors have several levels of severity and come from different projects (Codemirror, JupyterLab, Lumino and Notebook).

### Critical/serious

*JupyterLab*: 
- [ ] the responsive button in toolbar that have no `aria-label`
- [ ] several insufficient color contrast 
- [ ] `aria-label` on div without `role`

*Codemirror*:
- [ ] missing `aria-label` to div with `role`
- [ ] several insufficient color contrast 

### Moderate/minor

*Notebook*
- [ ] Missing landmarks in document: https://dequeuniversity.com/rules/axe/4.4/landmark-one-main?application=AxeChrome

*Lumino*:
- [ ] wrong `role` (button) to title of accordion panel
- [ ] wrong `role` (checkbox) to toggleable items of the command palette

PR #6719, opened 2023-02-08
Resolving some accessibility issues

brichet:
This PR addresses some accessibility issues:
- adds landmarks roles
- removes duplicate id on *running session* from side panel and form main area
- adds title on the sidebar button (a later PR will be necessary for translation)
- avoids adding kernel logo image if not necessary, to avoid image without title

Related to https://github.com/jupyter/notebook/issues/6671

### DIFF
--- a/packages/application/src/panelhandler.ts
+++ b/packages/application/src/panelhandler.ts
@@ -90,6 +90,8 @@ export class SidePanelHandler extends PanelHandler {
       this.hide();
     };
     closeButton.className = 'jp-Button jp-SidePanel-collapse';
+    closeButton.title = 'Collapse side panel';
+
     const icon = new Widget({ node: closeButton });
     this._panel.addWidget(icon);
     this._panel.addWidget(this._widgetPanel);
--- a/packages/tree-extension/src/index.ts
+++ b/packages/tree-extension/src/index.ts
@@ -185,7 +185,7 @@ const notebookTreeWidget: JupyterFrontEndPlugin<INotebookTree> = {
 
     if (manager) {
       const running = new RunningSessions(manager, translator);
-      running.id = 'jp-running-sessions';
+      running.id = 'jp-running-sessions-tree';
       running.title.label = trans.__('Running');
       running.title.icon = runningIcon;
       nbTreeWidget.addWidget(running);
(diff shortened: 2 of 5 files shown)


packages/tree-extension/src/index.ts
```diff
@@ -185,7 +185,7 @@ const notebookTreeWidget: JupyterFrontEndPlugin<INotebookTree> = {
 
     if (manager) {
       const running = new RunningSessions(manager, translator);
-      running.id = 'jp-running-sessions';
+      running.id = 'jp-running-sessions-tree';
```
jtpio, line 188: Looks like the following selector should also be updated in the UI tests:

https://github.com/jupyter/notebook/blob/ce9c16a01d3c80401096ccb09a8954f2b40d95de/ui-tests/test/tree.spec.ts#L43

brichet: Thanks

packages/application/src/panelhandler.ts
```diff
@@ -90,6 +90,8 @@ export class SidePanelHandler extends PanelHandler {
       this.hide();
     };
     closeButton.className = 'jp-Button jp-SidePanel-collapse';
+    closeButton.title = 'Collapse side panel';
```
jtpio, line 93: Should this string be translated?

brichet: Yes, but I was thinking doing this in a later PR, as it involves more changes.
The `sidepanel`s are created with the `NotebookShell`, so (I guess) before the translator.
Currently the `NotebookShell` does not include a translator, so we should probably add a `translator-extension` that add a translator to the shell, in order to translate the title.
If you think there is a simpler way maybe we can include it in this PR.

jtpio: Ah right. Then we can indeed have a look separately and track this in an issue.

jtpio: Just checked and JupyterLab sets the translator here:

https://github.com/jupyterlab/jupyterlab/blob/e81eb3d5e05e63f162d729066c2f644aa6243a32/packages/translation-extension/src/index.ts#L56-L59

jtpio (approved):
Thanks!

packages/application/src/panelhandler.ts
jtpio: Opened https://github.com/jupyter/notebook/issues/6722 to track this.

Issue #6690, opened 2023-01-12
Setup GitHub Action to automatically update the Playwright snapshots


jtpio:
Currently the UI Tests snapshots must be updated manually.

JupyterLab and other projects are using a GitHub Action to do this more easily by just posting a comment on the PR, with https://github.com/jupyterlab/maintainer-tools#update-snapshots.

We should setup similar here in this repo.

jtpio:
Reopening as there are still some issues being investigated in #6724

PR #6723, opened 2023-02-09
Add workflow to update snapshots

jtpio:
Fixes https://github.com/jupyter/notebook/issues/6690

This will come handy when we make UI changes and need to update the reference snapshots.


jtpio:

jtpio:
ah we probably need the workflow to merged in `main` first.

Issue #6734, opened 2023-02-15
Update to JupyterLab 4.0.0a35


jtpio:
Items to keep track of:

- [ ] https://github.com/jupyterlab/jupyterlab/pull/14044
- [ ] https://github.com/jupyterlab/jupyterlab/pull/13925
- [ ] https://github.com/jupyterlab/jupyterlab/pull/14072
- [ ] https://github.com/jupyterlab/jupyterlab/pull/14064
- [ ] https://github.com/jupyterlab/jupyterlab/pull/14039
- [ ] https://github.com/jupyterlab/jupyterlab/pull/14096
- [ ] Try removing `skipLibCheck`:

https://github.com/jupyter/notebook/blob/e7b5c3701be4eaeb685b14c4830083d48ecd033d/tsconfigbase.json#L17

jtpio:
Moved some items to https://github.com/jupyter/notebook/issues/6752

PR #6735, opened 2023-02-15
Update to TypeScript 5

jtpio:
Similar to the upstream work in JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/13872

More info in https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/

### DIFF
--- a/buildutils/package.json
+++ b/buildutils/package.json
@@ -32,7 +32,7 @@
     "@jupyterlab/buildutils": "^4.0.0-alpha.19",
     "commander": "^6.2.0",
     "fs-extra": "^9.1.0",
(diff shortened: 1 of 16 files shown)


jtpio:
CI passes although `skipLibCheck` is still enabled for now (probably due to legacy config):

https://github.com/jupyter/notebook/blob/e7b5c3701be4eaeb685b14c4830083d48ecd033d/tsconfigbase.json#L17

We can track removing `skipLibCheck` in a separate PR, maybe while updating to next JupyterLab 4 pre-release: https://github.com/jupyter/notebook/issues/6734

PR #6737, opened 2023-02-16
Docs: add migrate to notebook7 page

echarles:
Contribues to https://github.com/jupyter/nbclassic/issues/214

### DIFF
--- a/docs/source/migrate_to_notebook7.rst
+++ b/docs/source/migrate_to_notebook7.rst
@@ -1,0 +1,73 @@
+.. _htmlnotebook:
+
+NbClassic and Notebook
+======================
+
+NbClassic in the Jupyter Ecosystem
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Read more details about the changes currently taking place in the
+Jupyter Ecosystem in this `team-compass issue`_.
+
+Any extension developed for Notebook < 7 or NbClassic will not be
+compatible with Notebook 7 and upwards.
+
+**NbClassic and Notebook 7**
+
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks @echarles for starting this :+1:

docs/source/migrate_to_notebook7.rst
```diff
@@ -0,0 +1,73 @@
+.. _htmlnotebook:
+
+NbClassic and Notebook
```
jtpio, line 3: Since this will be part of the Notebook docs, I'm wondering whether the title should say something like "Updating to Notebook 7" instead?

Some Notebook users might not know what NBClassic is.

Especially since it will be visible in the navigation bar at the top of the page:

image

jtpio: Which means there could be a first section on this page summarizing [JEP 79](https://jupyter.org/enhancement-proposals/79-notebook-v7/notebook-v7.html) with new features, the change of the extension system, switch to jupyter server.

And maybe the sections below should be called "Compatibility with older versions" or "Using the Classic Notebook UI".

echarles: Thx @jtpio I have update the migrate_to_notebook7.rst page based on your feedbacks. The idea here is to bootstrap that page and further enrich it.

jtpio:
Looks like the docs need a lint pass to fix the CI check.

echarles:
> Looks like the docs need a lint pass to fix the CI check.

CI is back green. I think this is good to merge as first iteration of the doc.

jtpio (approved):
Thanks!

PR #6728, opened 2023-02-10
Add .mo and .json files for translations

frenzymadness:
See the discussion in https://github.com/jupyter/notebook/issues/6717

Shipping just .po files is useless because users would need to generate .mo files from them to make the translation work. Much better would be to update the release process to make this automatic, regenerate .mo and .json files, and remove .po files when building the wheel but because notebook v 7 will use a different system with language packs (where this automation is implemented) it does not really make sense to make this fix more complicated than it is.

I know that the .mo files are binary and therefore not good for being tracked in the git but this seems to be acceptable as a short term solution for notebook 6.

Also, the files were for some reason ignored via .gitignore file but nl translation has them.

PR #6744, opened 2023-02-22
Update kernel translation

JasonWeill:
Updates ja_JP translation as mentioned in #6728.


RRosio:
Thank you @JasonWeill!

Issue #6688, opened 2023-01-11
Customize layout from settings


jtpio:
Support for customizing the layout via the settings was added in JupyterLab in https://github.com/jupyterlab/jupyterlab/pull/11614.

It's available in the latest JupyterLab 4.0 pre-releases.

With it, end users should be able to configure where widgets should be placed in the application shell. For example to be able to open terminals in the right area instead of a new tab.

For now it seems to be tied to the `@jupyterlab/application-extension:shell` plugin, which is not loaded in Notebook 7.

jtpio:
This might help with https://github.com/jupyter/notebook/issues/6386.

jtpio:
In its settings, JupyterLab makes it possible to define the layout for `single` and `multiple` (document): https://github.com/jupyterlab/jupyterlab/blob/e0b4de1eb37c9dd37c11e5319ed00cb8d37b8d92/packages/application-extension/schema/shell.json#L32-L42

```json
"properties": {
  "single": {
    "$ref": "#/definitions/layout",
    "default": {
      "Linked Console": { "area": "down" },
      "Inspector": { "area": "down" },
      "Cloned Output": { "area": "down" }
    }
  },
  "multiple": { "$ref": "#/definitions/layout", "default": {} }
}
```

For Notebook 7, maybe it could be interesting to offer the ability to customize per page.

PR #6747, opened 2023-02-24
Add the extension manager

jtpio:
Add the extension manager to the Notebook 7 interface:

https://user-images.githubusercontent.com/591645/221237823-9dd6b830-8066-4ee6-87b9-cc39ba9e87f3.mp4

### Changes

- [x] Add `@jupyterlab/extensionmanager-extension
- [ ] Should it be accessible from all the pages, or just from the tree view next to the other tabs?

image


jtpio:
Leaving it as is for now (opening on the left area) for simplicity (just reusing the upstream plugin without modifications).

Need to check if there could still be a way to reuse the upstream plugin without changes, but place the extension manager next the Files and Running tabs. Maybe with https://github.com/jupyter/notebook/issues/6688.

PR #6766, opened 2023-03-13
Clean up interface switcher plugin in preparation for release

afshin:
This PR makes the `INotebookTracker` required instead of checking for it at runtime. It also adds an ellipsis `…` character to the interface button/dropdown to indicate it is a two-step UI control.


jtpio:
Thanks @afshin.

> It also adds an ellipsis `…` character to the interface button/dropdown to indicate it is a two-step UI control.

Maybe we could use the same caret icon as in the file browser for the "New" dropdown?

https://github.com/jupyter/notebook/blob/072a5d6b0c0e4a7b36fcdb44d9428a33034fb021/packages/tree-extension/src/index.ts#L86-L89

image

afshin:
That's a good idea for consistency.

afshin:

jtpio:
Kicking CI.

jtpio:
The snapshot update shows that the label is now slightly higher: 

image

Maybe we could try to re-align it like it was before, but this could also be done in a follow-up PR.

jtpio:
Looks like there were no changes to push when updating the snapshots? https://github.com/jupyter/notebook/actions/runs/4405807628/jobs/7717182164

image

afshin:
Chromium fails but Firefox passes ... do the snapshots only come from Firefox?

I visually can confirm my changes are visible in Chrome.

jtpio:
> do the snapshots only come from Firefox?

No it's for both chromium and firefox. We can try again.

jtpio:

jtpio:
> I visually can confirm my changes are visible in Chrome.

hmm the snapshot update still doesn't pick them up. I also tried building this branch in a fresh environment and this is what it looks like in Firefox (not aligned with the kernel name):

image

packages/lab-extension/style/base.css
```diff
@@ -9,3 +9,7 @@
   align-items: center;
   justify-content: center;
 }
+
+.jp-InterfaceSwitcher .lm-MenuBar-itemIcon {
```
jtpio, line 13: Something like this might do the trick?

```suggestion
.jp-InterfaceSwitcher .lm-MenuBar-itemIcon svg {
```

afshin: Yeah you're right, it does render differently when it's a more specific selector in Firefox.

afshin:

Issue #6552, opened 2022-10-04
Improve the About dialog


jtpio:
The look of the current "About" dialog could be improved:

![image](https://user-images.githubusercontent.com/591645/193758807-639a9464-880b-4947-be4a-c9b1bc166c89.png)

As a comparison this is what the about dialog in JupyterLab looks like:

![image](https://user-images.githubusercontent.com/591645/193759193-3dd12c0f-ea9b-46c6-9f77-fece0e5bbbfc.png)

Maybe the notebook dialog could be aligned to look more like the one in JupyterLab, and also display a link to the contributor list.

jtpio:
This is where the About dialog is created:

https://github.com/jupyter/notebook/blob/e8e0f8f32172ab834ce9005277c32ad04fadf8cd/packages/help-extension/src/index.tsx#L119-L161

krassowski:
The Notebook v6 About dialog was even more informative:

![Screenshot from 2022-10-04 10-20-01](https://user-images.githubusercontent.com/5832902/193783359-0b3ce353-de42-4558-ab23-dade5e94ad2e.png)

In particular, showing jupyter-server version (and maybe also the Python version it is running on) might be worth considering.

I think I remember that the kernel info was a separate menu item in JupyterLab at some point but I cannot find it any more - either a regression or I am wrong and it was never implemented.

davinderpalrehal:
Is someone working on this issue? If not I would like to have a crack at it.

kostyafarber:
@davinderpalrehal you still working on this?

davinderpalrehal:
@kostyafarber plans are to work on this over the weekend, is it time sensitive? If it's needed urgently I don't mind if someone else picks it up, my intention was to have the PR open by Monday.

kostyafarber:
Oh no go ahead! Just wanted to make sure if you weren't working on it I could pick it up. Thanks.

davinderpalrehal:
Sounds good, thanks @kostyafarber.

davinderpalrehal:
Hey @kostyafarber , I couldn't finish setting up the repo on my laptop and get it up and running, if you are still free please proceed with picking this issue up.

kostyafarber:
No worries @davinderpalrehal I'll try check it out

dohrisalim:
Is this issue still open? @kostyafarber @jtpio

kostyafarber:
If you think there are further improvements to be made and the maintainers agree, please go ahead!

[...]

PR #6769, opened 2023-03-14
Edited index.tsx to display python version 

akshithakumbam:


jtpio:
Thanks @Akshi22.

Would you mind providing a bit more context for this change? And link to any related issue if any.

Akshi22:
helo @jtpio
From the comments on #6552 , I saw that python version can be added to the about. I have made the necessary changes, I believe. Can you please check.

krassowski:
The implementation attempted here is not going to work, because the JavaScript in Jupyter Notebook is frontend-only. The Python version would need to be retrieved from the server via a REST API (which may or may not need defining a new endpoint).

jtpio:
Closing as stale

PR #6775, opened 2023-03-15
Replace the use of `toArray` by `Array.from`

### LINKED ISSUE #6774: Replace the use of `toArray` by `Array.from`
#6736 inadvertently brought the `toArray` utility from `@lumino/algorithm`:

https://github.com/jupyter/notebook/blob/44eb35e2e584b37c8c6fcbbbeb9f840ed510c29d/packages/notebook-extension/src/trusted.tsx#L21

With Lumino 2, we should now be able to use `Array.from` instead.

Which means this import can be removed here:

https://github.com/jupyter/notebook/blob/44eb35e2e584b37c8c6fcbbbeb9f840ed510c29d/packages/notebook-extension/src/trusted.tsx#L7

tarunsamanta2k20:
Fixes #6774 

Please don't close the. Ask me once for changes.

### DIFF
--- a/packages/notebook-extension/src/trusted.tsx
+++ b/packages/notebook-extension/src/trusted.tsx
@@ -4,8 +4,6 @@ import { Notebook, NotebookActions } from '@jupyterlab/notebook';
 
 import { ITranslator } from '@jupyterlab/translation';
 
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

Issue #6784, opened 2023-03-22
Stub toolbar visible at top of text files


JasonWeill:
There is a weird remnant of a minimized toolbar at the top of text files in notebook. Looks like we are pulling this over from lab without any modification to make it look right. Another difficult thing is that the line number column has the same background as the rest of the page, making it look odd.

This was found by @ellisonbg during a bug bash.

![image](https://user-images.githubusercontent.com/93281816/226790526-4344f62a-42e0-42be-9e83-152b6d9b4c8d.png)

jtpio:
Thanks for reporting.

I believe this comes from the default toolbar of the `MainAreaWidget` being empty. And is likely similar in JupyterLab as well. This was briefly discussed at a recent Notebook weekly call (cc @afshin).

Probably there should be some styling to hide it if is empty (although some extensions might add toolbar buttons to it).

PR #6791, opened 2023-03-23
Hides 'micro' toolbars in notebook

### LINKED ISSUE #6787: Extra white inner border and stub toolbar in terminal
In Notebook 7.0.0a17, the terminal has extra white inner border in light mode and white stub toolbar at the top. There is more space between the black rectangle and the bottom of the page than for the text editor, but maybe we need a bit more - but that spacing should match for the editor and terminal.

This was found by @ellisonbg during a bug bash.

Screen Shot 2023-03-20 at 3 34 57 PM

JasonWeill:
Fixes #6787 and #6784.

Hides the `jp-Toolbar-micro` stub toolbars in Jupyter Notebook. These are used as spacers below the JupyterLab tab bar, but this tab bar is not used in Jupyter Notebook, so they should not be visible.

Before:

image

image

After:

Note that the thin white micro-toolbar above the main widget is gone.

image

image


jtpio:
> this tab bar is not used in Jupyter Notebook, so they should not be visible.

But they could be used by third-party extensions though. For example they could add buttons to the toolbar to refresh the terminal:

image

packages/terminal-extension/style/base.css
```diff
@@ -0,0 +1,5 @@
+/* Hide the stub toolbar that appears above terminals */
+
+.jp-Toolbar-micro {
```
jtpio, line 3: This selector should probably take into account whether the toolbar has items in it or not.

jtpio: Actually maybe this could be implemented in JupyterLab directly:

image

Although in JupyterLab is looks like strange because there is the dock panel.

JasonWeill: @jtpio Per the code below, `jp-Toolbar-micro` is _only_ applied when a toolbar has no visible widgets.

https://github.com/jupyterlab/jupyterlab/blob/a3a0e3e7abf75e53810ed7ff8037c439cbb07697/packages/ui-components/src/components/toolbar.tsx#L58-L64

JasonWeill:

packages/terminal-extension/style/base.css
jtpio: Ah that's good to know. Then maybe this simple CSS change will just work and be enough.

jtpio: We could also put that under UI tests (either here or in JupyterLab) to make sure we don't hide the toolbar in case there is a widget in it.

JasonWeill:

jtpio:
Not sure there is any snapshot for files and terminals at the moment.

packages/terminal-extension/style/base.css
jtpio, line 3: Should this be a more specific selector?

For example `jp-MainAreaWidget > .jp-Toolbar-micro`?

JasonWeill: Moved to the application extension's CSS.

jtpio (approved):
Thanks!

Issue #6806, opened 2023-03-30
Shorten "Interface" menu and options


JasonWeill:
See also #6792.

Rename the "Interface" dropdown in the notebook toolbar to "Open in". Reduce the width of options: for example, reduce "Jupyter Notebook" to "Notebook" and reduce "JupyterLab" to "Lab." If possible, display the longer interface name in a tooltip.

JasonWeill:
It looks like in Lumino menus, the name of the command ("Open with Jupyter Notebook", for example) must be used as the name of the menu item; there is no provision to override this yet.

PR #6809, opened 2023-03-31
Renames Interface menu to "Open in", shortens menu options

JasonWeill:
Fix for #6806.

Renames the "Interface" menu to "Open in" and renames the commands to just the name of the desired interface.

In Notebook view:

image

In Lab view:

image

Note that the commands have particularly short names because Lumino doesn't yet allow custom menu item labels; see https://github.com/jupyterlab/lumino/issues/570 .

### DIFF
--- a/packages/lab-extension/src/index.ts
+++ b/packages/lab-extension/src/index.ts
@@ -44,6 +44,7 @@ namespace CommandIDs {
 interface ISwitcherChoice {
   command: string;
   commandLabel: string;
+  commandCaption?: string;
   buttonLabel: string;
   urlPrefix: string;
 }
@@ -78,7 +79,7 @@ const interfaceSwitcher: JupyterFrontEndPlugin<void> = {
     };
     const menubar = new MenuBar(overflowOptions);
     const switcher = new Menu({ commands });
-    switcher.title.label = trans.__('Interface');
+    switcher.title.label = trans.__('Open in');
     switcher.title.icon = caretDownIcon;
     menubar.addMenu(switcher);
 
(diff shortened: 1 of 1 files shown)


JasonWeill:

jtpio:
Thanks @JasonWeill for looking into this.

Would it be possible to keep the longer form when the command is displayed in the command palette? With this change on Binder it doesn't seem possible to open with another interface via the command palette anymore:

image

But it shows entry with `JupyterLab` only:

image

krassowski:
Just as a hopefully helpful note: this can be implemented by using functional label testing for `args['isPalette']` instead of hard-coded label. For example, see the [terminal `terminal:create-new` command](https://github.com/jupyterlab/jupyterlab/blob/ee06f972d274fc0d7b3422af937b6e7a7d84894a/packages/terminal-extension/src/index.ts#L347-L350).

JasonWeill:
> Just as a hopefully helpful note: this can be implemented by using functional label testing for `args['isPalette']` instead of hard-coded label. For example, see the [terminal `terminal:create-new` command](https://github.com/jupyterlab/jupyterlab/blob/ee06f972d274fc0d7b3422af937b6e7a7d84894a/packages/terminal-extension/src/index.ts#L347-L350).

@krassowski I just pushed a change that adds a separate "short label" for use in the menu, making the longer "label" used in the command palette instead. When I built my changes locally, though, I still see the short label in both places. Can you please take a look?

andrii-i:
@krassowski Attempting to implement approach described by you in notebook PR #6847, in [this](https://github.com/jupyter/notebook/pull/6847/commits/7c9e09d116875ee8b8dd06191e3752d74510d65d#diff-b6f83512c3b30af618335d9b7c7601533a34ec4dd41550a23a8f18f0745d6c6fR97-R98) commit but I'm missing a lot of details and could really use a consultation if/when you would have some time. Do I need to have a separate `addCommands` function for this to work?

jtpio:
Closing as fixed by https://github.com/jupyter/notebook/pull/6847, thanks!

Issue #6833, opened 2023-04-05
Header part not reflecting Jupyter Notebook Top Area default setting


yumyumqing:
**Describe the bug**
In settings editor, `Jupyter Notebook Top Area` is only reflecting changes from `User Preferences` but not `Settings Overrides`.

When I start the jupyter notebook with default settings overrides, the menu does pick up the `visible: false` setting, but the notebook is still showing header part.
If I toggle the setting manually by clicking the box, the change does reflect in the notebook header part.
But then if I click `Restore to Defaults`, the notebook header is not picking up the change again.

**To Reproduce**

1. Set up dev environment as https://github.com/jupyter/notebook/blob/main/CONTRIBUTING.md#setting-up-a-development-environment
2. Create a `overrides.json` file in `settings` dir in [JupyterLab Application Directory](https://jupyterlab.readthedocs.io/en/stable/user/directories.html#application-directory). 
If using `mamba` as step 1, the dir should look something like `$HOME/mambaforge/envs/notebook/share/jupyter/lab/settings`.
```
{
  "@jupyter-notebook/application-extension:top": {
    "visible": false
  }
}
```
3. Build and launch `jupyter notebook` in dev mode
4. Go to the browser page opened by `jupyter notebook`
5. Click `Settings` - `Settings Editor` - `Jupyter Notebook Top Area`
6. See the box is `not` checked
7. In a new tab, open a notebook in notebook mode, and the `header` is still showing
8. Back to the `setting` page and `check` the box and `uncheck` again
9. Refresh the notebook page, see the `header` is `not` showing now
10. Back to the `setting` page and `check` the box and then click `Restore to Defaults`
11. Refresh the notebook page, see the `header` is `still` showing

**Expected behavior**
In my opinion, I think the `header` part should reflect the default settings (if there is no `User Preference`) for 2 reasons:
1. The user might get confused when a unchecked box can lead to 2 different behaviors
2. For tools like `jupyterlite`, that uses this plugin but does not support `User Preference` override, it would be hard for the users to hide the `header` part by default.
(I think `jupyterlite` uses browser cache for `User Preference` settings and does not allow overrides at build time. Please correct me if I am wrong on this.)

**Screenshots**
<img width="920" alt="Screenshot 2023-04-05 at 8 11 38 PM" src="https://user-images.githubusercontent.com/21135321/230076595-f915afdf-e373-4e03-8eb3-3c9c794582e6.png">
[...]

PR #6836, opened 2023-04-06
Fix top bar visibility not picking up settings overrides (#6833)

yumyumqing:
* The header part reflects both user preference and default settings overrides.
* Toggle 'Show Header' turns off the automatic adjustment.
* Before this commit behavior:
  * Automatically adjust header visibility to screen size at first
  * If manually check/uncheck the `visible` box in the settings editor, `not` automatically adjust header visibility.
  * If manually toggle `Show Header` in menu->view, `not` automatically adjust header visibility.
  * If overrides default settings at runtime, it doesn't reflect on the UI.
* After this commit behavior:
  * Automatically adjust header visibility to screen size at first
  * If manually `uncheck` the `adjustToScreen` box in the settings editor, `not` automatically adjust header visibility.
  * If manually toggle `Show Header` in menu->view, `not` automatically adjust header visibility.
  * If overrides default settings at runtime, it does reflect on the UI.
* Related issue: https://github.com/jupyter/notebook/issues/6833

### DIFF
--- a/packages/application-extension/schema/top.json
+++ b/packages/application-extension/schema/top.json
@@ -18,10 +18,11 @@
   },
   "properties": {
     "visible": {
-      "type": "boolean",
+      "type": "string",
+      "enum": ["yes", "no", "automatic"],
       "title": "Top Bar Visibility",
-      "description": "Whether to show the top bar or not",
-      "default": true
+      "description": "Whether to show the top bar or not, yes for always showing, no for always not showing, automatic for adjusting to screen size",
+      "default": "automatic"
     }
   },
   "additionalProperties": false,
--- a/packages/application-extension/src/index.ts
+++ b/packages/application-extension/src/index.ts
@@ -482,22 +482,29 @@ const topVisibility: JupyterFrontEndPlugin<void> = {
       execute: () => {
         top.setHidden(top.isVisible);
         if (settingRegistry) {
-          void settingRegistry.set(pluginId, 'visible', top.isVisible);
+          void settingRegistry.set(
+            pluginId,
+            'visible',
+            top.isVisible ? 'yes' : 'no'
+          );
         }
       },
       isToggled: () => top.isVisible,
     });
 
-    let settingsOverride = false;
+    let adjustToScreen = false;
 
     if (settingRegistry) {
       const loadSettings = settingRegistry.load(pluginId);
       const updateSettings = (settings: ISettingRegistry.ISettings): void => {
-        const visible = settings.get('visible').composite;
+        // 'visible' property from user preferences or default settings
+        let visible = settings.get('visible').composite;
         if (settings.user.visible !== undefined) {
-          settingsOverride = true;
-          top.setHidden(!visible);
+          visible = settings.user.visible;
         }
+        top.setHidden(visible === 'no');
+        // adjust to screen from user preferences or default settings
+        adjustToScreen = visible === 'automatic';
       };
 
       Promise.all([loadSettings, app.restored])
@@ -517,7 +524,7 @@ const topVisibility: JupyterFrontEndPlugin<void> = {
     }
 
     const onChanged = (): void => {
-      if (settingsOverride) {
+      if (!adjustToScreen) {
         return;
       }
       if (app.format === 'desktop') {


yumyumqing:
Here are some expected behaviors in different cases:

**No default settings override + no user preference override:**
  - This is the same as the earlier default behavior.
  - **Some key variables output:**
    > [TOP]: visible/composite =>  true
    [TOP]: visible/user =>  undefined
    [TOP]: setHidden =>  false
    [TOP]: adjustToScreen/composite =>  true
    [TOP]: adjustToScreen/user =>  undefined
    [TOP]: noAdjust =>  false

  - **Behavior:** 
    - 2 boxes both checked with no `Restore to Defaults` link.
      Screenshot 2023-04-06 at 10 06 38 PM

    - `adjustToScreen` setting overrides the `visible` setting, with the header `showing on large` screen and `not showing on small` screen.
      Screenshot 2023-04-06 at 10 07 09 PM
      Screenshot 2023-04-06 at 10 07 26 PM

**Default settings override: visible -> false, adjustToScreen -> false + no user preference override:** 
  - **Some key variables output:**
    > [TOP]: visible/composite =>  false
[TOP]: visible/user =>  undefined
[TOP]: setHidden =>  true
[TOP]: adjustToScreen/composite =>  false
[TOP]: adjustToScreen/user =>  undefined
[TOP]: noAdjust =>  true

  - **Behavior:** 
    - 2 boxes both not checked with no `Restore to Defaults` link.
      Screenshot 2023-04-06 at 10 12 31 PM

    - `Not` automatically adjusting the header visibility, the header is `not` showing on large or small screen.
      Screenshot 2023-04-06 at 10 12 49 PM
      Screenshot 2023-04-06 at 10 13 02 PM

(Some other cases, eliminating the outputs and screenshots for easier read.)
1. Default settings override: `adjustToScreen -> false` + no user preference override
    - 1 box checked and 1 not checked with no `Restore to Defaults` link.
    - `Not` automatically adjusting the header visibility, the header is `showing` on large and small screen.
2. No default settings override + user preference override: `visible -> false`, `adjustToScreen -> false`
    - Both boxes `not` checked with a `Restore to Defaults` link.
    - `Not` automatically adjusting the header visibility, the header is `not` showing on large or small screen.
3. No default settings override + user preference override: `adjustToScreen -> false`
    - 1 box `checked` and 1 box `not` checked with a `Restore to Defaults` link.
    - `Not` automatically adjusting the header visibility, the header is `showing` on large and small screen.
4. More other cases not listed here.

yumyumqing:
Hi @jtpio just a friendly reminder on this PR :) Have you looked at it and/or is there anything I need to add? Thanks!

jtpio:
@yumyumqing sorry for the late reply!

It looks good on Binder and should indeed fix the main usability issue reported in https://github.com/jupyter/notebook/issues/6833. 

Looking at the code and the new setting I was wondering if there was a way to combine the two options into a single one. But maybe it's simpler and more explicit to keep the two options separated.

yumyumqing:
Hi @jtpio thanks for your response! 

If we would like to keep the exact same old variables as before to keep users' current overriding files work, I couldn't think of a better way to merge it in 1 variable.
But if we are willing to break some of the current behavior, I think we can turn `TopBarVisibility` into a string variable, and then user can input `yes` for always showing, `no` for always not showing, `automatic` for adjusting to size. And we can default to `automatic` to match the current default behavior.

Do you have any advice on that? Thanks!

jtpio:
> I think we can turn `TopBarVisibility` into a string variable, and then user can input `yes` for always showing, `no` for always not showing, `automatic` for adjusting to size. And we can default to `automatic` to match the current default behavior.

That sounds like a good idea. JupyterLab uses `enums` like the one proposed here for some settings. It's probably fine to "break" the current behavior now since it's a beta, and details like this one are still subject to changes. We can also add an more visible comment in the changelog afterwards to highlight this if needed.

yumyumqing:
Hi @jtpio I have updated the PR according to our discussion. Could you please help take a look? Thanks a lot!

jtpio (approved):
Thanks!

PR #6840, opened 2023-04-17
fix import of warnings.warn

minrk:
`warn` was never part of the traitlets API, and changed in https://github.com/ipython/traitlets/pull/838


jtpio:
cc @RRosio @echarles

echarles (approved):
LGTM

PR #6841, opened 2023-04-17
Load custom CSS

RRosio:
- [x] Placeholder custom CSS file
- [x] Link stylesheet in template files
- [x] Add handler for static file 
- [x] Update scripts involving static files

~~**To Fix**
Current error is `Refused to apply style from 'http://localhost:8888/custom/custom.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.`~~


jtpio:
Thanks @RRosio for looking into this!

> Current error is `Refused to apply style from 'http://localhost:8888/custom/custom.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.`

Maybe it's because the handler gives a 404? (which would be an HTML page). What does the network tab of the dev tools say?

RRosio:
> Maybe it's because the handler gives a 404? (which would be an HTML page). What does the network tab of the dev tools say?

Hi @jtpio! Thank you for taking a look here! Yes that is exactly the issue I overlooked! I updated the href for the file to `href="{{page_config['fullStaticUrl'] | e}}/custom/custom.css"` and it now grabs the template CSS file I added in `/notebook/static/custom` successfully, but does not apply the styles from the CSS file in my `.jupyter/custom` directory. I have been trying to understand how that CSS file in my jupyter config paths is supposed to override the template CSS file in `/notebook/static/custom`, but am not sure about it yet. I can't seem to find much about how that configuration is set...

jtpio:
> is supposed to override the template CSS file in `/notebook/static/custom`, but am not sure about it yet

My guess is that it is added alongside the existing template CSS file? So that it doesn't need to override, but instead provide additional CSS rules.

jtpio:
Maybe it's because `ExtensionAppJinjaMixin` is missing? 

According to the docs: https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html#jinja-templating-from-frontend-extensions

> This adds a `<name>_jinja2_env` setting to Tornado Web Server’s settings that will be used by request handlers.

jtpio:
Hmm that's strange. I triggered a new CI run on https://github.com/jupyter/notebook/pull/6887 and the docs build seems to be passing there.

docs/source/config_overview.md
```diff
@@ -26,6 +26,10 @@ and editing settings is similar for all the Jupyter applications.
 > - [traitlets](https://traitlets.readthedocs.io/en/latest/config.html#module-traitlets.config)
 >   provide a low-level architecture for configuration.
 
+### Disabling Custom CSS
```
jtpio, line 29: As a follow-up (could be another PR), maybe it could be interesting to document the feature more (for example in the user guide), to show an example of such `custom.css` file.

jtpio:
There is indeed some logic in the `hatch-jupyter-builder` plugin to ensure some target files have are available, but not sure these would be the cause of the issue above:

https://github.com/jupyter/notebook/blob/245bc87ecf56b5a86be0595bc86f709110340859/pyproject.toml#L165-L168

jtpio:
Just tried from this branch and it seems to be working well!

I created `.jupyter/custom/custom.css` with the following content for testing:

```css
#top-panel-wrapper, #jp-top-bar {
  background-color: blue!important;
}
```

And then simply refreshed the page to see the changes:

image

notebook/templates/tree.html
```diff
@@ -7,6 +7,10 @@
   {% block favicon %}
   <link rel="icon" type="image/x-icon" href="{{ base_url | escape }}static/favicons/favicon.ico" class="favicon">
   {% endblock %}
+
+  {% if custom_css %}
+  <link rel="stylesheet" type="text/css" href="{{ base_url | escape }}custom/custom.css">
+  {% endif %}
```
jtpio, line 13: Should this also be added to the `consoles.html` page?

https://github.com/jupyter/notebook/blob/main/notebook/templates/consoles.html

jtpio (approved):
Thanks!

jtpio:
Ah looks like there is a small conflict to fix. Otherwise this should be good to go :+1:

jtpio:
Thanks!

Issue #6832, opened 2023-04-05
v7: "Close tab" and "Close and Shutdown notebook" should be reworked.


Carreau:
They both close the widgets/lumino tabs, and not the browser tab. 

The first one seem unnecessary, user can close browser tab.

The second make sens to shutdown the kernel, but should close the browse tab. 

<img width="465" alt="Screen Shot 2023-04-05 at 13 58 52" src="https://user-images.githubusercontent.com/335567/230073967-b89e2437-6efd-4c1b-94e2-6686ea4483fc.png">

Carreau:
I'm going to add that "Kernel > Shutdown All kernels" make little sens in the menu if you have a single kernel / notebook per page...

jtpio:
Right. There should be an audit of the menu entries during the beta.

This is also related to https://github.com/jupyter/notebook/issues/6398.

For reference it should be possible to hide some of these menu entries via the settings, for example like:

https://github.com/jupyter/notebook/blob/6e60c72eadb1caf1efa396742f0ed2f50a9ac67b/packages/application-extension/schema/menus.json#L24

jtpio:
@Carreau maybe we could close this issue in favor of https://github.com/jupyter/notebook/issues/6398?

Carreau:
Up to you, I think #6398 is a high level goal, while this one have explicit menus to fix, which might be easier to tackle one by one by new contributors. Up to you. Maybe make this a subtask of https://github.com/jupyter/notebook/issues/6398 ?

timmerk3:
Hey, can I take this issue? I'm a new contributor so I'd like to try it out.

jtpio:
Sure thanks @timmerk3. Feel free to comment here if you need any help.

PR #6843, opened 2023-04-17
Remove "Close Tab" menu entry

timmerk3:
Disables the "Close Tab" menu entry from the file dropdown, brought up in #6832. The user should be the one manually closing the tab.


jtpio:
Thanks @timmerk3 for looking into it.

packages/application-extension/schema/menus.json
```diff
@@ -13,6 +13,13 @@
           {
             "command": "notebook:trust",
             "rank": 20
+          },
+          {
+            "type": "submenu",
```
jtpio, line 18: Not sure it should be `submenu` here, since the entry appears as a regular single entry?

Also when trying this PR on Binder the menu entry still seems to be displayed:

image

jtpio:
FYI @timmerk3 I pushed a small change to fix the config.

jtpio (approved):
Thanks!

Issue #6800, opened 2023-03-24
Accessibility Issues Needing Addressing for WCAG 2.1 compliance (As of Version 7.0.0a15)


manfromjupyter:
I know Jupyter Notebook v7 just announced huge accessibility improvements recently available in alpha but I wanted to throw my hat into the ring with remaining pieces before we celebrate fully that it's fully accessible. Lot of this is super low hanging fruit with potential huge impact.

## Summary
**Blind Users** – minorly functional but with great annoyances. `Grade C`
**Low Vision** –  accessible, just one minor annoyance. `Grade A-`
**Color Blind** – completely supported on light theme, `Grade A+`. `A- on dark theme`
**Ambulatory** - mostly usable, just a few blockers for supplementary actions. `Grade B`
**Deaf/Hard of Hearing** - completely supported. `Grade A+`
**Cognitive** - completely supported. `Grade A+`
**Seizures** - completely supported. `Grade A+`

Related to https://github.com/jupyterlab/jupyterlab/issues/9399 but `NotebookV7` obviously much better and what should be recommended to persons with disabilities and just needs a few small fixes to be fully supporting of Low Vision, Color Blind, and Ambulatory users.

## TLDR Issues 
- [ ] Blind users with screenreader read everything in German no matter what (JS sets lang as an invalid type: "default")
- [ ] Using just keyboard, user can't escape Terminal or Markdown block on New Terminal and New Markdown pages once they get inside them (perhaps also other non-notebook pages as well)?
- [ ] Using just keyboard, user can't get to Running tab on /tree page
- [ ] Keyboard-only users cannot modify notebook titles. Needs tabindex="0" on the notebook title link that triggers modal to rename.
- [ ] Keyboard-only user focus on Jupyter logo is just a vertical line not bounding box around the element they are focused on.
- [ ] On focus and hover on the notebook title on just the dark theme, this needs a more distinct color change. As is (change from dark gray to lighter gray) is not enough of a difference to visually demonstrate it's currently focused/it does something.
- [ ] No skip link found (keyboard-only users have to tab 20 times every time to get to main region to do work).
- [ ] Main page (tree page)'s table is not a real table; it's `<div>`s within `<div>`s which works for visual users but not blind users/screen readers.
- [ ] Keyboard-only users (ambulatory), do not see the hoverable/tooltip of the icons in heading and in code cells when focusing on the elements.
- [ ] Hovering over something with a hoverable popup/tooltip does not let you dismiss it with the ESC key.
[...]

PR #6844, opened 2023-04-19
Add Skip Link to Notebook

seirani:
Closes one of the tasks on #6800

Problem: "No skip link found (keyboard-only users have to tab 20 times every time to get to main region to do work)."
Changes to Code: Added a SkipLink widget

To Fix: Fails test case checking that all areas are empty by default. Skip to Link is only functional for Notebook page, not for Tree header page (where I wasn't sure what the desired "main" content to skip to should be)

Without using the skip link: User has to click through every single tab in the menu before reaching the edit feature of the first cell.

https://user-images.githubusercontent.com/92682603/232935674-0873aba7-3441-41b0-8689-5ab468577d14.mov

With the functional skip link: User tabs to "Skip to Main" first, then can click enter to go directly to the first cell edit.

https://user-images.githubusercontent.com/92682603/232935712-12b4173d-0389-4220-b185-6a763a8d3875.mov

### DIFF
--- a/packages/application/src/shell.ts
+++ b/packages/application/src/shell.ts
@@ -448,6 +448,6 @@ export namespace Private {
     }
 
     private _skipLinkWidget: Widget;
-    private _isDisposed: boolean = false;
+    private _isDisposed = false;
   }
 }
--- a/packages/application/src/shell.ts
+++ b/packages/application/src/shell.ts
@@ -116,7 +116,7 @@ export class NotebookShell extends Widget implements JupyterFrontEnd.IShell {
     rootLayout.addWidget(hsplitPanel);
 
     this.layout = rootLayout;
-        
+
     // Added Skip to Main Link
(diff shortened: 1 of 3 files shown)


jtpio:
Thanks @seirani!

Looks like there are a few lint issues that should be fixed when running `jlpm run prettier` normally.

jtpio:
> To Fix: Fails test case checking that all areas are empty by default. Skip to Link is only functional for Notebook page, not for Tree header page (where I wasn't sure what the desired "main" content to skip to should be)

Likely we could update the tests to remove the `top` area from this list:

https://github.com/jupyter/notebook/blob/787339c09e27cbee1b17799f372126f085abe045/packages/application/test/shell.spec.ts#L32-L35

And maybe check the `top` area has the skip link widget by default?

jtpio (approved):
Thanks!

PR #6847, opened 2023-04-19
Change "Interface" menu to "Open in...", shorten names of menu items inside

### LINKED ISSUE #6806: Shorten "Interface" menu and options

andrii-i:
- Change "Interface" menu to "Open in...", shorten command names inside the menu, use longer command names for command palette. Fixes #6806, supersedes #6809

Screenshot 2023-04-19 at 9 52 47 PM
Screenshot 2023-04-19 at 9 58 14 PM


jtpio:
Thanks @andrii-i for looking into this!

andrii-i:
Thank you @jtpio. I'm still working on adding conditional option to open in nbclassic if nbclassic is installed but current state of this PR is self-contained and complete for task 1. In case Notebook 7 release would be to happen today, this can be merged as-is

andrii-i:
Dynamically detecting nbclassic would require server modification so I will take care of it in a separate PR.

andrii-i:
Bot please update playwright snapshots

jtpio:
Ah looks like the bot might be broken, but not sure why: https://github.com/jupyter/notebook/actions/runs/4866690782/jobs/8678479167

jtpio:
@andrii-i actually one of the snapshot failures is relevant. The menu entry should be like the command palette and say "Open in JupyterLab":

image

packages/lab-extension/src/index.ts
```diff
@@ -90,22 +91,33 @@ const interfaceSwitcher: JupyterFrontEndPlugin<void> = {
...
+      commands.addCommand(paletteCommand, {
+        label: (args) => (args.noLabel ? '' : commandPaletteLabel),
+        caption: commandLabel,
+        execute,
         isEnabled,
       });
 
       if (palette) {
-        palette.addItem({ command, category: 'Other' });
+        palette.addItem({ command: paletteCommand, category: 'Other' });
```
jtpio, line 120: We should update the menu entries as well. For example by using the palette command here:

https://github.com/jupyter/notebook/blob/1ea8e1e3aa02dbad11ea31b9122d3b4e52d9cd26/packages/lab-extension/schema/interface-switcher.json#L11-L20

andrii-i: updated naming conventions and menu entries, now they use "Open In [...]" strings

andrii-i:
@jtpio managed to resolve all conflicts, it's all green now 🟢. Manually finding and replacing snapshots is definitely not very optimal and takes time.

packages/lab-extension/src/index.ts
```diff
@@ -90,22 +91,33 @@ const interfaceSwitcher: JupyterFrontEndPlugin<void> = {
...
-          if (!current) {
-            return;
-          }
-          window.open(`${urlPrefix}${current.context.path}`);
-        },
+        execute,
+        isEnabled,
+      });
+
+      commands.addCommand(descriptionCommand, {
```
jtpio, line 112: Wondering if there is a way to use the `isPalette` trick mentioned in https://github.com/jupyter/notebook/pull/6809#issuecomment-1492958716.

To avoid adding "proxy" commands to the commands registry.

jtpio: Done in https://github.com/jupyter/notebook/pull/6847/commits/52d12b593809bdcde13edeff5163b26c978d3f9e

jtpio:
FYI @andrii-i I pushed a small change to avoid the use of an extra command to handle different labels.

jtpio (approved):
Thanks!

andrii-i:
Thank you @jtpio

packages/lab-extension/src/index.ts
andrii-i: So that's where you provide args! Thank you @jtpio

PR #6850, opened 2023-04-24
Added extra information and user-action summaries to 7 migration docs.

ericsnekbytes:
See https://github.com/jupyter/notebook/issues/6851, this PR adds extra information to the migrate_to_notebook7 doc:

- A summary of user-actions to take
- Additional context
- Formatting/document structure updates


echarles:
Thx @ericsnekbytes I have gone across your changes and they make sense to me.

docs/source/migrate_to_notebook7.md
```diff
@@ -1,27 +1,62 @@
...
+released out of beta, along with all its new features, like realtime
+collaboration, debugger, and theming.
+
+For users who need to use extensions or other customizations, you have a couple
+of options:
+
+- Look for Notebook 7 compatible versions of the extensions you already use,
+  and [find replacements for those that are not available]
+
+- If you NEED to maintain compatibility with the Classic Notebook for extensions
```
jtpio, line 31: Maybe the all caps is not needed here?

```suggestion
- If you need to maintain compatibility with the Classic Notebook for extensions
```

docs/source/migrate_to_notebook7.md
```diff
@@ -1,27 +1,62 @@
...
+  or other customizations that are critical to your workflows, you can switch to
+  [nbclassic], the new home for users who need to maintain compatibility with the
```
jtpio, line 33: The "new" and "maintain" words in this sentence give the somewhat misleading idea that nbclassic will be maintained in the long run?

NBClassic is supposed to help with the transition to Notebook 7, but shouldn't be seen as an "alternative" to Notebook 7.

docs/source/migrate_to_notebook7.md
```diff
@@ -1,27 +1,62 @@
...
+  old notebook interface
+
+## Why things are changing
```
jtpio, line 36: Maybe something like the following?

```suggestion
## Why a new major version?

jtpio (commented):
Left a couple of comments.

docs/source/migrate_to_notebook7.md
ericsnekbytes: Yeah it's a subtle thing to communicate. We want people who need it to know it's there and to use it, without giving the impression it's a permanent thing. I'll add a bit/re-work to add some clarity. Thanks!

jtpio: I posted a comment in the team compass issue to see if the EC can provide guidance for the documentation as well, since they will be looking at the blog post: https://github.com/jupyter/notebook-team-compass/issues/24#issuecomment-1521772403

docs/source/migrate_to_notebook7.md
ericsnekbytes: I think this one is worth keeping, because it simplifies the language and reduces jargon (related to semantic versioning / "major version", which not everyone is going to know about). Maybe there's a plain-language compromise, like "Why the big version change?" Is it the "major-version-number-is-being-incremented-to-7" idea that you want to communicate here?

docs/source/migrate_to_notebook7.md
ericsnekbytes: I have updated this portion with a recent commit, and we can modify as needed after the EC takes a look.

docs/source/migrate_to_notebook7.md
ericsnekbytes: @jtpio Is "Why the big version change?" good?

jtpio: Why not use "major"? "big" sounds ambiguous: what does big mean?

Otherwise maybe "Why a new version" could also work?

ericsnekbytes: Big was supposed to communicate that sweeping changes are taking place in plain-language (without the "major version" jargon and the odd-sounding/confusing phrasing that it necessitates...."Why a new major version" sounds odd, but "Why a major new version" does not...[probably related to points noted in this BBC article](https://www.bbc.com/news/blogs-trending-37285796)...).

I've pushed an update to just "Why a new version?" which I think is good enough, let me know if you've got any other suggestions.

jtpio:
Looks like it needs a lint pass to fix the CI check (`jlpm run prettier` should do it).

jtpio:
Thanks!

Issue #6553, opened 2022-10-05
Toolbar is often empty when opening a Notebook


jtpio:
Sometimes the notebook toolbar moves most of the items in the toolbar popup, leaving only the popup opener visible.

This is one of the causes for the flaky UI tests ([example run](https://github.com/jupyter/notebook/actions/runs/3192850297)):

https://user-images.githubusercontent.com/591645/194165361-234a2c8e-f9e6-40d3-b098-0376cd87d91d.mp4

But this sometimes also happens when opening a notebook or refreshing the page:

![image](https://user-images.githubusercontent.com/591645/194166181-6b1d013c-327b-40a5-b41f-934b8e798dda.png)

There seems to be a race condition somewhere. Or the toolbar resize logic is being triggered too early in the process of loading the notebook widget and adding it to the shell.

jtpio:
This looks like a real issue / race condition.

Part of the fix should include the removal of this workaround in the UI tests added in https://github.com/jupyter/notebook/pull/6487:

https://github.com/jupyter/notebook/blob/62c0fa84d2ea58afd5227b9c5cbc0657be243940/ui-tests/test/mobile.spec.ts#L21-L23

and:

https://github.com/jupyter/notebook/blob/62c0fa84d2ea58afd5227b9c5cbc0657be243940/ui-tests/test/mobile.spec.ts#L44-L46

andrii-i:
I'm having a hard time reproducing this locally. Does this happens locally (often) or this is more of a testing problem? If it's the latter, should this issue be moved to 7.x and not be a blocker?

jtpio:
Right, as noticed in https://github.com/jupyter/notebook/pull/6887#issuecomment-1581521852, the issue now seems to be more about some items being cropped rather than items missing from the toolbar:

![image](https://github.com/jupyter/notebook/assets/591645/0dbe2401-0d7e-4229-b8a9-e29d374a1cfe)


However it seems to be reproducible when testing with a mobile layout via the dev tools and refreshing the page several times:

**items are "cropped"**

![image](https://github.com/jupyter/notebook/assets/591645/7b262065-41d3-45be-a13d-ed49655f2fb6)

**expected behavior**

![image](https://github.com/jupyter/notebook/assets/591645/a3fd0391-f222-49cf-abf3-c070e4e41b2c)

jtpio:
[...]

PR #6887, opened 2023-05-25
Remove viewport workaround in the UI tests

jtpio:
Fixes https://github.com/jupyter/notebook/issues/6553

### DIFF
--- a/ui-tests/test/mobile.spec.ts
+++ b/ui-tests/test/mobile.spec.ts
@@ -9,7 +9,7 @@ import { test } from './fixtures';
 
 import { hideAddCellButton, waitForKernelReady } from './utils';
 
-test.use({ autoGoto: false, viewport: { width: 512, height: 768 } });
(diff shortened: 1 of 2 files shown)


ui-tests/test/mobile.spec.ts
```diff
@@ -9,7 +9,7 @@ import { test } from './fixtures';
 
 import { hideAddCellButton, waitForKernelReady } from './utils';
 
-test.use({ autoGoto: false });
+test.use({ autoGoto: false, viewport: { width: 600, height: 800 } });
```
jtpio, line 12:
```suggestion
test.use({ autoGoto: false, viewport: { width: 524, height: 800 } });
```

jtpio:
Looks like the toolbar items are visible, but some are cropped:

image

jtpio:
For some reasons, rebasing on the latest `main` makes the test pass.

It could be because the issue was fixed in a JupyterLab 4.1 pre-release.

PR #6899, opened 2023-06-04
Configurable `default_url` for JupyterNotebookApp

paulgb:
The `default_url` parameter on `JupyterNotebookApp` is currently hard-coded to `"/tree"`. This makes it configurable.

This behaviour is consistent with the behavior of `LabApp`, and allows users to start a notebook that redirects to a particular notebook by default.


jtpio (approved):
Thanks!

Issue #6697, opened 2023-01-16
Find text only searches part of file


PAGWatson:
If I open a long .py file in Notebook 7.0.0a10 and try to search for text with Ctrl+F, text only seems to be found if it appears within the portion of the file being viewed on screen, and it says there are zero instances otherwise. If I try it in jupyterlab (4.0.0a32), the text is found as long as it exists somewhere in the file, so this seems to be an issue specific to Notebook. It would be desirable to be able to search the whole of python scripts, text files etc.

PAGWatson:
Sure. As an example, one of my scripts has a function "get_color_cycle", and if I search for this in the searchbar brought up by Ctrl+F when viewing the top of the file, then it returns zero results:

![image](https://user-images.githubusercontent.com/6704648/212890690-e4ca8bc6-3739-459c-bddb-089967224c4a.png)

If I scroll down to show the function in the browser window and search again, the text is found:

![image](https://user-images.githubusercontent.com/6704648/212890989-4849ac51-dcca-4a41-8e0f-8a8d727722c7.png)

My guess was that the notebook is only loading the file contents near the location that is being viewed or something? For small text files and scripts, I'd guess it's generally better to load the whole thing.

jtpio:
Ah interesting thanks for sharing the screenshots.

> If I try it in jupyterlab (4.0.0a32), the text is found as long as it exists somewhere in the file

In JupyterLab, does the Ctrl-F trigger the browser search popup like in Notebook 7? Or the JupyterLab search / replace tool?

Higgs32584:
So I looked over this bug to see if I could replicate it. I was able to replicate it, but the functionality is more nuanced. So the HTML search bar does not pick up on the text in the python file, but the localized search bar does. I guess it is a matter of if we want the HTML search bar to pick up on the python code. 



![case#3](https://user-images.githubusercontent.com/55243596/222810705-49b24399-dbf5-4eae-a44c-6963e870d048.png)
on built into the application, where
![screenshot#2](https://user-images.githubusercontent.com/55243596/222810725-42899810-5a82-4afe-bd77-5efa62830d43.png)

fcollonval:
Small comment: the browser search will always fail as the text editor (i.e. codemirror 6) is only rendering the text content that fits in the viewport.

There are two solutions:
- most desirable: use the jupyter search (as implemented in JupyterLab)
- less desirable: use the codemirror 6 search extension

jtpio:
[...]

PR #6905, opened 2023-06-07
Fix for "find searches only part of the file" bug

ericsnekbytes:
As suggested in https://github.com/jupyter/notebook/issues/6697, this PR enables the Jupyter Document Search shortcut as a default behavior, so that users will get it by default instead of the browser find tool (Lab and Notebook 7 now only render visible-on-page content in the DOM, so the browser find is not going to work for file content outside the visible/viewport range).


jtpio:

packages/documentsearch-extension/schema/notebookShellWidgetListener.json
```diff
@@ -6,7 +6,7 @@
       "command": "documentsearch:start",
       "keys": ["Accel F"],
       "selector": ".jp-mod-searchable",
-      "disabled": true
+      "disabled": false
```
jtpio, line 9: Since these settings only include the keyboard shortcut override, maybe we could actually remove the whole file?

jtpio:
Thanks @ericsnekbytes for working on this!

I left a small suggestion above to clean up unused settings files, but otherwise should be good to go.

jtpio:
cc @yuvipanda for awareness since you had originally commented on this on the RetroLab issue: https://github.com/jupyterlab/retrolab/pull/294#issuecomment-988712849

As explained in https://github.com/jupyter/notebook/issues/6697, the motivation for reverting this change is the new virtual implementation of the notebook in JupyterLab (and in Notebook 7). Some cells might not be visible on the page until they are in the viewport.

If we keep the shortcut disabled, then the browser search bar is still triggered on `Ctrl-F` but it might not be possible to find some text in the notebook document.
However with this PR, hitting `Ctrl-F` opens the JupyterLab search bar and we can search the entire notebook.

So with this PR it's mostly a matter of changing the default. Users who still want the browser search (even though it would not work on hidden elements) would have to disable the shortcut via the settings.

docs/source/notebook.md
```diff
@@ -332,6 +332,16 @@ to remember are the following:
...
 
+## Searching
+
+Jupyter Notebook has an intelligent built-in search plugin for finding text within a
+notebook or other document, which uses the {kbd}`Ctrl-F` shortcut by default.
+
+Your browser's `find` function will give unexpected results because it doesn't have
+access to the full content of a document, but you can still use your browser find
+function from the browser menu if you want, or you can disable the intelligent search
+shortcut using the advanced settings editor. TODO: DETAILS (investigate settings bug?)
```
jtpio, line 343: Should we track this TODO in an isssue?

ericsnekbytes: @jtpio I've verified that the settings JSON schema that shows up in the advanced settings editor is different than your sample video. The issue is that I don't know the expected schema for the JSON object being edited, and thus how to explain how to disable the shortcut to a user.

There's also the matter of explaining to users what a JSON object is and what a key/value pair is, on top of where those go in the schema to properly disable the shortcut...this is more content that should be communicated in a docs entry. At this point I'm wondering if simply acknowledging the behaviors of intelligent-search and browser-search (which is already done in the docs in this PR) is enough for now...we can merge this in and worry about docs for disabling later. Thoughts?

jtpio: > simply acknowledging the behaviors of intelligent-search and browser-search (which is already done in the docs in this PR) is enough for now...we can merge this in and worry about docs for disabling later.

Yes, and this is likely something to address in JupyterLab actually.

But it would be nice to not leave the `TODO` in the docs before merging.

ericsnekbytes: That's done, I'm opening a new PR for Lab too and will link it here.

ericsnekbytes: Same docs updates for JupyterLab are at: https://github.com/jupyterlab/jupyterlab/pull/14667

docs/source/notebook.md
```diff
@@ -332,6 +332,16 @@ to remember are the following:
...
 
+## Searching
+
+Jupyter Notebook has an intelligent built-in search plugin for finding text within a
+notebook or other document, which uses the {kbd}`Ctrl-F` shortcut by default.
+
+Your browser's `find` function will give unexpected results because it doesn't have
+access to the full content of a document, but you can still use your browser find
+function from the browser menu if you want, or you can disable the intelligent search
+shortcut using the advanced settings editor.
```
andrii-i, line 343: Should Advanced Settings Editor be capitalized as in UI? 

```suggestion
shortcut using the Advanced Settings Editor.
```

Screenshot 2023-06-09 at 4 13 45 PM

This also reminds me that we use JSON Settings Editor and Advanced Settings Editor terms interchangeably in different spots but this is out of scope of this PR.

andrii-i (commented):
Thank you for working on this @ericsnekbytes. This looks good to me. Please find a suggestion on capitalization for consistency below.

jtpio (approved):
Thanks!

Issue #5770, opened 2020-09-24
Document how to remove the download option


pmeakins:
Hi,

Is there any settings to reduce the functionality in the UI, i particularly want to prevent UPLOAD/DOWN, to stop the notebook being able to expose data back to a local filesystem and then using the download feature to lift the data out from Jupyter.

Any suggestions.

pmeakins:
Hi, I'm still looking for some advice on locking down features in Jupyter notebooks, is it possible to completely remove DOWNLOAD capability from Notebooks and from Lab.  Become a hughe issue - with users decrypting content - copy back and downloading

kevin-bates:
I'm sorry for the delayed response.  I was hoping someone with front-end dev knowledge would have responded by now.  

Looking in the server handlers, I don't see a way this can be configured from the server and I suspect such a disablement would be too far-reaching.  Since this functionality is inherent in the general operation of the notebook application, I doubt it can be configured other than removing the actual option from the menu - which may still leave paths by which the same outcomes could occur via other actions.  You might try gleaning some hints from the [front-end configuration docs](https://jupyter-notebook.readthedocs.io/en/stable/frontend_config.html#configuring-the-notebook-frontend).

If someone knows how to disable/hide various menu options could you please respond?  Thank you.

bollwyvl:
As long as you want users to be able to load a file, or run a cell, it's
not going to be possible to prevent an even mildly interested malicious
user from exfiltrating data that is shown on the screen. F12, network tab,
and you know everything you need, and kernel messages can be automated with
short js console commands.

The jupyter voila project offers a somewhat more locked-down experience, as
it turns off the contents service and disables "unsafe" kernel messages
(execute_request). But even that is irrelevant if a malicious user has the
ability to put a notebook in a place and use it, or can control the
environment the kernel runs in.

Hulva:
https://github.com/jupyter/notebook/issues/5001#issuecomment-1078615798

jtpio:
>  i particularly want to prevent UPLOAD/DOWN,

As a heads-up for that particular case, it will be possible to disable the download plugins in the next major version of Notebook (v7).

Adding to the `7.0` milestone for consideration so this can be documented.

jtpio:
This could be added to the documentation, to show how to disabled items in the menus.

The docs could also link to the JupyterLab documentation: https://jupyterlab.readthedocs.io/en/latest/user/interface_customization.html#menu-bar

PR #6909, opened 2023-06-08
Document how to manage plugins and extensions

jtpio:
Fixes https://github.com/jupyter/notebook/issues/5770


jtpio:
This is quite minimal, but could be expanded upon in follow PRs.

Issue #6904, opened 2023-06-07
Notebook v7 JupyterHub Integration


tuncbkose:
In looking at the [v7.0.0b3](https://github.com/jupyter/notebook/releases/tag/v7.0.0b3) release, I didn't notice any JupyterHub integrations. 

In JupyterLab, there is the `hub-extension` package. Are there any plans of including that in v7 or providing similar functionality?

jtpio:
Thanks @tuncbkose for raiding raising this question.

I believe Notebook 7 is already being used with JupyterHub in some deployments so it should already work with JupyterHub.

For reference the `hub-extension` was added in https://github.com/jupyterlab/retrolab/pull/148 (when Notebook 7 was still RetroLab). And it is still part of Notebook 7 here:

https://github.com/jupyter/notebook/blob/245bc87ecf56b5a86be0595bc86f709110340859/app/package.json#L57

jtpio:
@tuncbkose would you be able to check if Notebook 7 is working as expected on JupyterHub?

tuncbkose:
@jtpio Could you take a look at a convenient time or restore the Needs Triage label?

andrii-i:
Moving to 7.0.x for further follow up after Notebook 7 release. Please feel free to move it back if necessary.

jtpio:
Taking a closer look, the JupyterLab plugin expects `hubPrefix` to be defined here to add the menu entries:

https://github.com/jupyterlab/jupyterlab/blob/86326b258138b9d0413de77089c83d5ec95c6a4e/packages/hub-extension/src/index.ts#L49-L52

It also passes the the `hubPrefix` option as part of the page config here:

https://github.com/jupyterlab/jupyterlab/blob/86326b258138b9d0413de77089c83d5ec95c6a4e/jupyterlab/labapp.py#L827-L848

But Notebook 7 uses the same approach here:

https://github.com/jupyter/notebook/blob/bf42a0c797a1ce49361e896cc7b7af3ac865370d/notebook/app.py#L61-L72

So normally the `hubPrefix` should be available on the frontend and be activated.

jtpio:
I tried aligning the JupyterHub config in https://github.com/jupyter/notebook/pull/6918 but it looks like the `hub_prefix` option is not set correctly when Notebook 7 is started first, which means the menu items are not visible.

[...]

PR #6918, opened 2023-06-13
Align JupyterHub config with JupyterLab

jtpio:
As noticed in https://github.com/jupyter/notebook/issues/6904#issuecomment-1588677062, the JupyterHub config is not handled exactly the same way in Notebook 7 and JupyterLab.

This change aligns Notebook 7 to use the same logic for handling the JupyterHub config.


andrii-i (approved):
Looks good

Issue #6914, opened 2023-06-12
cannot open a jupytext notebook in nb7


parmentelat:
**Describe the bug**

The jupytext extension does not work as expected in nb7

**To Reproduce**

1. install jlab4 as released in pip
2. install nb7 with `pip install -U --pre notebook`
3. install jupytext with `pip install jupytext`
4. install the jupytext labextension with
  `pip install git+https://github.com/mwouts/jupyterlab-jupytext.git@jupyterlab4`
5. create a simple jupytext notebook named for instance `foo-nb.md`
6. under jlab, check that you can open the notebook: right click → "Jupytext Notebook"
7. perform the same operation from the regular notebook tree browser: the "Jupytext Notebook" option is present but does not open the file as a notebook, it shows a text editor instead

**Expected behavior**
1. we should be able to open a jupytext notebook from the regular notebook file browser; 
2. (also when opening a URL like localhost:8888/notebooks/path/file.md we need this to open a notebook and not a text editor, but for that to work the former behaviour 1. must first work)

**Screenshots**

![image](https://github.com/jupyter/notebook/assets/4519227/3f764548-74e9-4ec5-9334-1af96299bad6)
 
**Desktop (please complete the following information):**

- OS: \[e.g. iOS\] MacOS & linux
- Browser \[e.g. chrome, safari\] - chrome
- Version \[e.g. 22\]

* jlab 4.0.2
* notebook 7.0.0b4
* as of this writing, https://github.com/mwouts/jupyterlab-jupytext.git@jupyterlab4 is on commit d451f5c
 
**Additional context**

Please see also https://discourse.jupyter.org/t/generic-question-about-jlab4-and-nb7/19846

jtpio:
Thanks @parmentelat for reporting.



> Steps to reproduce the behavior:

Looks like we also need to install `jupytext` with `pip install jupytext`? Not sure `jupyterlab-jupytext` installs `jupytext` automatically?

After following the instructions above and the additional `pip install jupytext` it seems like it's possible to open the markdown file with the notebook factory via the context menu:

https://github.com/jupyter/notebook/assets/591645/392196d7-2961-456f-a79c-ab059f521b02

But the behavior seems to be similar in JupyterLab right? By default double-clicking on the markdown file or opening it with the default "Open" context menu shows the markdown source:

https://github.com/jupyter/notebook/assets/591645/f3127393-4e8b-4e3d-aee6-94313fd4ee48

jtpio:
[...]

PR #6923, opened 2023-06-15
Fix default viewers override

jtpio:
Troubleshooting https://github.com/jupyter/notebook/issues/6914

Notebook 7 should respect the `defaultViewers` settings overrides when opening a document.

Testing with Jupytext with the following:

- `pip install jupytext`
- `pip install git+https://github.com/mwouts/jupyterlab-jupytext.git@jupyterlab4`

### DIFF
--- a/packages/application-extension/src/index.ts
+++ b/packages/application-extension/src/index.ts
@@ -183,7 +183,7 @@ const opener: JupyterFrontEndPlugin<void> = {
     router: IRouter,
     docManager: IDocumentManager
   ): void => {
-    const { commands } = app;
+    const { commands, docRegistry } = app;
 
     const command = 'router:tree';
     commands.addCommand(command, {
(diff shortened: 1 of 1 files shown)


jtpio:
Still troubleshooting this.

jtpio:
Looking better:

https://github.com/jupyter/notebook/assets/591645/6070024d-5a2f-42d5-bb41-be3d36e51d3a

Although not sure why there is no toolbar when opening the markdown file with the `Jupytext Notebook` factory. This also seems to be an issue with JupyterLab:


https://github.com/jupyter/notebook/assets/591645/293f0bae-59a6-4e96-ae1a-2cf144f73791

jtpio:
cc @parmentelat for awareness, since this PR should already help in case you would like to try it out.

As noted above, it seems like there is an issue with the toolbar even with JupyterLab. Maybe this could be addressed as part of the JupyterLab 4 compatibility happening in https://github.com/mwouts/jupyterlab-jupytext/pull/3

packages/application-extension/src/index.ts
```diff
@@ -195,10 +203,37 @@ const opener: JupyterFrontEndPlugin<void> = {
...
 
-        const file = decodeURIComponent(path);
-        const urlParams = new URLSearchParams(parsed.search);
-        const factory = urlParams.get('factory') ?? 'default';
         app.started.then(async () => {
+          const file = decodeURIComponent(path);
+          const urlParams = new URLSearchParams(parsed.search);
+          let defaultFactory = docRegistry.defaultWidgetFactory(path).name;
+
+          // Explicitely get the default viewers from the settings because
```
jtpio, line 211: Maybe we can track that in an issue so it can be discussed.

andrii-i (approved):
Main needs to be merged in, otherwise looks good. Thank you @jtpio

Issue #6934, opened 2023-06-19
Browser tab does not close when a notebook is shutdown v7.0.0.rc.0


gutow:
**Describe the bug**

Non-blocking, but issue to consider in 7.0.0.rc.0:

Choosing `File>Close and Shutdown Notebook` closes and shuts down the notebook, but it does not close the tab/window as I would expect. Tested on MacOS using FireFox and Chrome > v114.

**To Reproduce**

1. Open a notebook in Notebook 7.0.0.rc.0.
1. Click on the file menu and select "Close and Shutdown Notebook".

The notebook will close and shutdown, but the browser window/tab will stay open.

**Expected behavior**
Expect the notebook to shutdown and the tab/window it was in to close.

**Screenshots**

**Desktop (please complete the following information):**

- OS: MacOS
- Browser Chrome and Firefox > v114
- Version Notebook v7.0.0.rc.0

**Additional context**

Issue #1159, opened 2016-03-02
"Close and Halt" not closing the page in master


fperez:
On current master (216793eb), "close and halt" doesn't close the tab, it only kills the kernel.  I'm pretty sure this is a recent regression, though I haven't bisected it.

takluyver:
Works for me in both Firefox and Chrome.

minrk:
```javascript
if (user === 'fperez') {
    all_the_buggs.trigger();
}
```

@fperez does it make a difference if you save immediately before hitting it?

fperez:
No kidding!!!

But yes, it does make a difference: if I save and close/halt immediately, it works (i.e. kernel killed and tab closed). If I try to close/halt a long-open nb, it kills the kernel but doesn't close the tab.

minrk:
That suggests it's to do with the dirty flag, which, when set, should show you a dialog to confirm navigation away from the page. Do you see any dialog when you hit close & halt? If not, when you make some changes and close immediately, does it let you, or does it prompt that you are going to lose unsaved changes?

fperez:
If I have unsaved changes it does prompt me. The only misbehavior I'm seeing is the failure to close the browser tab after the kernel shuts down.

minrk:
Can you check browser(s) version(s) where this happens?

Just to reiterate: you do get the confirmation dialog, click "Leave this page", and then you stay on the page? That dialog is provided by `window.onbeforeunload`, which I *thought* meant we weren't even allowed to keep you on the page if we wanted to.

fperez:
It's tricky to replicate, b/c it seems to only happen with notebooks that have been left open for a while, so they've been checkpointed already.

1. If I have recent changes, then the 'do you want to save' dialog does pop up, and if I say "leave", I do go away as you describe.

1. If I've saved manually, then the close happens as expected.

But I've seen multiple instances (so far on Chrome) where I'm left with a kernel-less but open notebook...

Anyway, if it's an "fperez special", let's ignore it for now. I'll keep trying to find the conditions to trigger it...

minrk:
Is there any chance you have accidentally clicked "Stay on this Page" instead of "Leave this Page"? Because that's exactly what would happen: kernel is shutdown, but page close aborted, so you have a page open with no kernel.

fperez:
No, it happened multiple times, without the dialog coming out in those
cases. It simply shut down the kernel and stayed on the page.

minrk:
[...]

Issue #6398, opened 2021-10-30
Notebook v7 menus should match the v6 ones whenever possible


fperez:

andrii-i:
As I mentioned in another thread, we have 2 [release blocking issues](https://github.com/jupyter/notebook/issues?q=is%3Aissue+is%3Aopen+label%3A%22tag%3ARelease+Blocker%22) left and I wanted to align on what has to be done to release Notebook 7. Ease of transition and need to not invalidate Notebook v6 (educational content) are big parts of JEP 79.  Would you say we have achieved sufficient consistency for the release and can address remaining points post-release?

andrii-i:
- Moving to 7.0.x for further follow up after Notebook 7 release. Related comment: https://github.com/jupyter/notebook/issues/6307#issuecomment-1587710396

fperez:
Thanks for checking, @andrii-i! I have to admit I've been busy with other things and haven't looked in a while - I've also fully transitioned to JLab even for all my undergraduate teaching since I last wrote here, so now I'm much less of a reasonable judge :)

I trust the team has done a huge amount of work on this, so we're probably at the point where it's best to make a release, announce it, and plan on responding to the (inevitable) small issues that people always find once a package goes into much wider usage.  But I see the list of known blockers is now down to zero, and I'm not sure waiting much more is beneficial.

There's another reason to release soon, I think: it's early summer, and it's a good time for educators who are some of the big Notebook constituents, to test the materials they may use in the fall semester, still having time for the cycle of feedback and potential changes that may arise from that.

In summary, my 10,000 ft perspective is that we're probably better off moving forward with a release at this point, planning on perhaps a few point updates, so that by August (start of the fall teaching season in many countries), there's a 7.0.x (for a hopefully small value of x) that meets just about everybody's needs.

andrii-i:
Thank you for the feedback @fperez!

dbpengra:
I also support making the menus consistent between the new and old versions of the notebook.  

Another problem I have is that the Help menu in the new version does not provide links to the usual pile of online documentation for Python and all of the major packages (NumPy, Matplotlib, etc.).  These were/are very helpful!

goelakash:
Want to take a stab at this. Rather than a single sweeping audit, I think it makes sense to split it menu by menu:

**A. File menu** — add `Download` and a `Download as` submenu of nbconvert exporters from inside an open notebook. This is the specific gap called out in the original post and still the case today.

[...]

PR #6937, opened 2023-06-20
Close the browser tab when clicking on "Close and Shut Down Notebook"

jtpio:
Fixes https://github.com/jupyter/notebook/issues/6934
Fixes https://github.com/jupyter/notebook/issues/1159

This mimics the behavior of the Classic Notebook:

https://github.com/jupyter/notebook/assets/591645/a86bd386-59ba-493a-a033-70b4d0eaee6f

In Notebook 7:

https://github.com/jupyter/notebook/assets/591645/37cfa27a-bb40-4e0f-aecb-ef093e1c9d4f

### Changes

- [x] Add new semantic command to the `closeAndCleaners`
- [x] Move the close and cleaners down in the file menu. Related to https://github.com/jupyter/notebook/issues/6398
- [x] Keep the same "Close and Halt" name in the menu entry as in the classic UI
- [x] Update reference snapshots
- [x] Add UI test to check the browser tab is closed when clicking on the menu item


jtpio:

jtpio:
cc @gutow for awareness

packages/notebook-extension/src/index.ts
```diff
@@ -126,6 +128,40 @@ const checkpoints: JupyterFrontEndPlugin<void> = {
...
+    menu: IMainMenu,
+    translator: ITranslator | null
+  ) => {
+    const { commands } = app;
+    translator = translator ?? nullTranslator;
+    const trans = translator.load('notebook');
+
+    const id = 'notebook:close-and-halt';
+    commands.addCommand(id, {
+      label: trans.__('Close and Halt'),
```
andrii-i, line 150:
```suggestion
      label: trans.__('Close Notebook and Shutdown Kernel'),
```

ui-tests/test/notebook.spec.ts
```diff
@@ -155,4 +155,24 @@ test.describe('Notebook', () => {
     const imageName = 'notebooktools-right-panel.png';
     expect(await panel.screenshot()).toMatchSnapshot(imageName);
   });
+
+  test('Clicking on "Close and Halt" should close the browser tab', async ({
```
andrii-i, line 159:
```suggestion
  test('Clicking on "Close Notebook and Shutdown Kernel" should close the browser tab', async ({
```

ui-tests/test/notebook.spec.ts
```diff
@@ -155,4 +155,24 @@ test.describe('Notebook', () => {
...
+    tmpPath,
+  }) => {
+    const notebook = 'simple.ipynb';
+    await page.contents.uploadFile(
+      path.resolve(__dirname, `./notebooks/${notebook}`),
+      `${tmpPath}/${notebook}`
+    );
+    await page.goto(`notebooks/${tmpPath}/${notebook}`);
+
+    const menuPath = 'File>Close and Halt';
```
andrii-i, line 170:
```suggestion
    const menuPath = 'File>Close Notebook and Shutdown Kernel';
```

andrii-i (approved):
This looks and works well, thank you for working on this @jtpio.

I believe there is spaceto improve user experience further beyond adding parity with nbclassic:
1. It's not clear what exactly "Close and Halt" and "Shut Down" items do and what is the difference between them. Please find suggestions below on renaming "Close and Halt" into "Close Notebook and Shutdown Kernel". "Shut Down" should then be renamed into "Shut Down Server". I think this can be done within this PR but can also be taken care of separately.
2. "Close and Halt"  terminates kernel of the notebook and "Shut Down" terminates server with all kernels accordingly. When this happens all unsaved data is lost. Therefore these buttons are destructive and can lead to data loss. To remedy this, we could ask a user to confirm his action via modal/popup window. We should then also add "..." to the end of both options to indicate that selecting the item will lead to further options or require additional input. It would probably be best to do it in a separate PR after Notebook 7 launch, please let me know if you agree with the approach and I can create an issue to track it.

jtpio:
> 1. It's not clear what exactly "Close and Halt" and "Shut Down" items do and what is the difference between them. Please find suggestions below on renaming "Close and Halt" into "Close Notebook and Shutdown Kernel". "Shut Down" should then be renamed into "Shut Down Server". I think this can be done within this PR but can also be taken care of separately.

Good question. JupyterLab uses "Close and Shut Down Notebook", while for the Classic Notebook it's "Close and Halt". This PR went with "Close and Halt" so it looks the same as the classic notebook and also taking into consideration #6398 (for example if folks have screenshots showing these menu entries in the course materials).

However for consistency with JupyterLab it would be fine to go with "Close and Shut Down Notebook" yes, no strong opinion.


> 2\. To remedy this, we could ask a user to confirm his action via modal/popup window

Normally there is already a dialog (see the second screencast):

image

andrii-i:
> However for consistency with JupyterLab it would be fine to go with "Close and Shut Down Notebook" yes, no strong opinion.

From my point of view, "Close and Shut Down Notebook" is significantly more clear in explaining its function. "Close and Shut Down Notebook" still describes what happens with kernel better vs for example "Close and Halt Notebook" as "halt" implies pausing. Also we already use "Shut Down" for process termination so with "halt" we would be introducing a second term for the same action. We could also check in with other contributors during the call today.

> Normally there is already a dialog (see the second screencast)

Yes, somehow missed this.

packages/notebook-extension/src/index.ts
jtpio: If going with this suggestion then we'll likely need to use `Shut Down` instead of `Shutdown` for consistency with the other menu entries.

jtpio:
> From my point of view, "Close and Shut Down Notebook" is significantly more clear in explaining its function. "Close and Shut Down Notebook" still describes what happens with kernel better vs for example "Close and Halt Notebook" as "halt" implies pausing. Also we already use "Shut Down" for process termination so with "halt" we would be introducing a second

Yes that sounds fine too. Users familiar with the `Close and Halt` entry from the classic notebook would probably be not too confused if it's now name `Close and Shut Down Notebook` now anyway.

jtpio:

jtpio:
FYI @andrii-i I updated the PR to take your suggestions in consideration.

jtpio:
Merging as the comments above have been addressed.

Thanks @andrii-i for the review!

PR #6946, opened 2023-06-23
Fix opening files from the CLI

### LINKED ISSUE #6914: cannot open a jupytext notebook in nb7

jtpio:
Investigate ways to fix #6914

### DIFF
--- a/notebook/app.py
+++ b/notebook/app.py
@@ -234,7 +234,7 @@ class JupyterNotebookApp(NotebookConfigShimMixin, LabServerApp):
     app_version = version
     extension_url = "/"
     default_url = Unicode("/tree", config=True, help="The default URL to redirect to from `/`")
-    file_url_prefix = "/notebooks"
+    file_url_prefix = "/tree"
     load_other_extensions = True
     app_dir = app_dir
     subcommands: dict = {}
@@ -329,13 +329,6 @@ class JupyterNotebookApp(NotebookConfigShimMixin, LabServerApp):
             # if the serverapp set one
             page_config["token"] = ""
 
-        self.handlers.append(
-            (
-                rf"/{self.file_url_prefix}/((?!.*\.ipynb($|\?)).*)",
-                web.RedirectHandler,
-                {"url": ujoin(self.serverapp.base_url, "/edit/{0}")},
-            )
-        )
         self.handlers.append(("/?", RedirectHandler))
         self.handlers.append(("/tree(.*)", TreeHandler))
         self.handlers.append(("/notebooks(.*)", NotebookHandler))


jtpio:
So the idea if that we should be able to open files from the CLI by using the existing `/tree/<file>` redirection logic.

jtpio:
cc @parmentelat if you want to try with the built artifacts: https://github.com/jupyter/notebook/actions/runs/5356963835?pr=6946

jtpio:
For reference `file_url_prefix` is documented here in Jupyter Server: https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html#anatomy-of-an-extensionapp

> file_url_prefix: the prefix URL added when opening a document directly from the command line. For example, classic Notebook uses /notebooks to open a document at http://localhost:8888/notebooks/path/to/notebook.ipynb.

notebook/app.py
```diff
@@ -234,7 +234,7 @@ class JupyterNotebookApp(NotebookConfigShimMixin, LabServerApp):
     app_version = version
     extension_url = "/"
     default_url = Unicode("/tree", config=True, help="The default URL to redirect to from `/`")
-    file_url_prefix = "/notebooks"
+    file_url_prefix = "/tree"
```
jtpio, line 237: So the idea of using `/tree` is that there is already a redirect handler to open either with `/edit` or `/notebooks`:

https://github.com/jupyter/notebook/blob/7f1a2f49ef77abdca2dd22c42a4fcfb8e6890a11/notebook/app.py#L127-L160

PR #6965, opened 2023-07-10
Fix Python 3.12 compatibility

frenzymadness:
.joinpath with an empty argument no longer works as before thanks to this change in CPython: https://github.com/python/cpython/commit/cea910ebf14d1bd9d8bc0c8a5046e69ae8f5be17#diff-2e741d925220d74a9cc04cda1314d3649d9d189c0efc7db18e5387a51225b61c

This change mimics the old internal behavior by using _paths attribute directly.

I'm not saying it's perfect but it's the shortest solution I was able to come up with when building notebook 7rc2 with Python 3.12 in Fedora rawhide.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -91,7 +91,7 @@ def make_notebook_app(  # noqa PLR0913
     )
 
     # Copy the schema files.
-    test_data = str(files("jupyterlab_server.test_data").joinpath(""))
+    test_data = str(files("jupyterlab_server.test_data")._paths[0])  # type: ignore
     src = pathlib.PurePath(test_data, "schemas", "@jupyterlab")
     dst = pathlib.PurePath(str(schemas_dir), "@jupyterlab")
     if os.path.exists(dst):


frenzymadness:
I've reported the issue in CPython upstream: https://github.com/python/cpython/issues/106614

jtpio:
Thanks @frenzymadness for looking into this.

Looks like the lint failure might be relevant.

frenzymadness:
I'm not sure what to do here and I'm at EuroPython now so I don't have much time to look into it. The problem is that documentation says that `files` returns `Traversable` and it's true that `Traversable` does not have `_path` attribute but in this specific case, `files` returns `MultiplexedPath` which is a subclass of `Traversable` and it has that attribute. This is also true in Python 3.11 so it's nothing coming with the new release.

jtpio:
Since it's just the `lint` check failing, maybe we can add a comment to ignore the check?

The tests seem to be passing on `3.8` and `3.11`.

jtpio:
Also maybe we could start testing on Python 3.12 on CI too? (and add it to the matrix)

frenzymadness:
This is not really a big deal so if you are okay with skipping the check, so I am.  I dug a little bit deeper and opened one more issue for importlib.resources where a change in typing might fix the problem here: https://github.com/python/importlib_resources/issues/286

jtpio:
I guess it would be fine yes, also this line is only for the tests.

frenzymadness:
Rebased and added a comment to ignore the typing issue.

jtpio:
Should we also try adding `3.12` to the test matrix on CI?

frenzymadness:
> Should we also try adding `3.12` to the test matrix on CI?

Good idea but the configuration is kinda complex so I'm not sure I'm able to do it properly.

jtpio:
Sure, I'll have a look in a separate PR.

jtpio (approved):
Thanks!

PR #6971, opened 2023-07-19
Fix hatch version when going from rc

jtpio:
As noticed in https://github.com/jupyter/notebook/actions/runs/5600904438/jobs/10243994486


jtpio:
Just double checked this change didn't introduce a regression when making a patch release with `next`.

For example this recent commit on `main` correctly bumps to `7.0.1` as expected: https://github.com/jupyter/notebook/commit/dc2082c84c3f29ecf31f3a05e9b51cf401d5dddd

image

PR #6972, opened 2023-07-20
The future is now

frenzymadness:
so we should update the Readme.


jtpio (approved):
Thanks!

jtpio:
The `check-links` CI failure is not related and should be fixed by https://github.com/jupyter/notebook/pull/6973

PR #7005, opened 2023-08-09
Add CI check to install with homebrew on `macos`

jtpio:
Troubleshoot https://github.com/jupyter/notebook/issues/6974

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,15 +181,15 @@ jobs:
           path: ./dist
       - name: Install the prerequisites
         run: |
-          python -m pip install -U pip wheel
+          python3 -m pip install -U pip wheel
       - name: Install the package
         run: |
           cd dist
-          python -m pip install -vv 'notebook*.whl'
+          python3 -m pip install -vv 'notebook*.whl'
       - name: Validate environment
         run: |
-          python -m pip freeze
-          python -m pip check
+          python3 -m pip freeze
+          python3 -m pip check
       - name: Validate the install
         run: |
           jupyter labextension list
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,6 +167,38 @@ jobs:
           jupyter notebook --version
           jupyter notebook --help
 
+  install-brew:
+    needs: [build]
+    runs-on: macos-latest
+    timeout-minutes: 15
+    steps:
+      - name: Install Python
+        run: |
+          brew install python
+      - uses: actions/download-artifact@v3
+        with:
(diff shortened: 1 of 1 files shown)


jtpio:
cc @defjaf

Here are the logs for the homebrew installation, which passes on CI: https://github.com/jupyter/notebook/actions/runs/5809598910/job/15748922126?pr=7005

If you would like to double check to see if you have something similar locally.

jtpio:
And the rest of the logs? https://github.com/jupyter/notebook/actions/runs/5809598910/job/15748922126?pr=7005

defjaf:
I admit it's pretty hard to parse what I've got...

So I attach a file with the console output from pip. Note that this is starting from having jupyterlab 3.6.3 and notebook 6.5.5 installed, so they are upgraded to 4.0.4 and 7.0.3 (but it didn't automatically upgrade other stuff). It does seem to have put a bunch of files below `/opt/homebrew/share/jupyter/` but nothing new in `/opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/share/jupyter/`. And I still get the schema console errors and the UI problems.

[pip-install-notebook==7 output.txt](https://github.com/jupyter/notebook/files/12303627/pip-install-notebook.7.output.txt)

defjaf:
@jtpio Looking more carefully, even from the beginning of the logs, the CI is using `/Library/Frameworks/Python.framework/...` whereas my homebrew install is using `opt/homebrew/...` — it looks like the CI build isn't using the homebrew python?

defjaf:
> Maybe we could check whether it's possible to install under /opt/homebrew on CI.

Well, I think the point is that the CI process isn't using the homebrew python and so not automatically picking up the homebrew paths like `sys.prefix`. Can you write the script with a hard-coded python executable `/opt/homebrew/bin/python3`?

defjaf:
> Or can you write the script with a hard-coded python executable `/opt/homebrew/bin/python3`?

Of `/usr/local/bin/python3` if the CI machines are intel.

defjaf:
Seems like the CI machines are running on Intel, and that the same issues do not arise in that case, according to https://github.com/jupyter/notebook/pull/7006.

defjaf:
see https://github.com/jupyter/notebook/pull/7006#issuecomment-1682181317 for information about running in a venv.

DTiming24:
Nice CI addition for the Homebrew install path. Even though it's a workflow-only change, this should make macOS breakages much easier to catch before they land on users.

jtpio:
Closing as per https://github.com/jupyter/notebook/issues/6974#issuecomment-5324717307

PR #7039, opened 2023-09-11
Adopt sp-repo-review

blink1073:
Adopt [pypi.org/project/sp-repo-review](https://pypi.org/project/sp-repo-review/), which has a standard set of checks for packages in the scientific python ecosystem, as seen in [learn.scientific-python.org/development](https://learn.scientific-python.org/development/).

### DIFF
--- a/notebook/templates/edit.html
+++ b/notebook/templates/edit.html
@@ -1,43 +1,39 @@
-<!doctype html>
+<!DOCTYPE html>
 <html>
-  <head>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <title>{{page_config['appName'] | e}} - Edit</title>
-    {% block favicon %}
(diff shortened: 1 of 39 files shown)


notebook/templates/edit.html
```diff
@@ -1,39 +1,43 @@
...
+    <link
+      rel="icon"
+      type="image/x-icon"
+      href="{{ base_url | escape }}static/favicons/favicon-file.ico"
+      class="favicon"
+    />
+    {% endblock %}
+  </head>
+  <body data-notebook="edit">
+    {# Copy so we do not modify the page_config with updates. #} {% set
```
jtpio, line 17: Looks like we might want to keep the previous line breaks here for readability.

blink1073: Done

jtpio (approved):
Thanks!

PR #7046, opened 2023-09-14
fix import of warnings.warn (#6840)

holzman:
* fix import of warnings.warn

never part of the traitlets API

* remove yanked codecov


holzman:
Should we forward-port this patch to 6.5.x?

blink1073:
Thanks @holzman, I'm closing this in favor of #7051, which has a related fix as well.

PR #7070, opened 2023-09-22
ci: set minimal permissions to workflows

### LINKED ISSUE #7069: CI: Set minimal permissions on GitHub Workflows
### Problem

I noticed that some of your GitHub Workflows don't specify their permissions (as [ui-tests.yml](https://github.com/jupyter/notebook/blob/main/.github/workflows/ui-tests.yml)), and others define some potentially dangerous write permissions (i.e., usually `contents: write`) that might not be required. It's recommended that you always define minimal permissions to your workflows, as secures you against erroneous or malicious behaviours from external jobs you call from them. It's specially important for the case they get compromised, for example, and it's a recommendation by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

### Proposed Solution

I'd recommend that you review the write permissions of your workflows and substitute them by read permissions whenever  possible. Ideally you would set a top-level read-only permission on all workflows, so that they would be inherited by any job that does not define job-level permissions. For any job that requires write permissions, they can be defined locally as job-level permissions.

As those would be changes of few lines of code, I'll take the liberty to send a PR setting the required minimal permissions as I understand by the workflows. Then you can evaluate the suggestion more easily and correct me where I'm mistaken. 

#### Context
I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://opensource.googleblog.com/2023/04/googles-open-source-security-upstream-team-one-year-later.html)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

diogoteles08:
Fixes #7069

Mind that I'm not really familiar with the `jlpm` commands and with some of the jupyter workflows called on some actions, so I recommend that you double-check if I haven't removed any write permission that is actually needed =)

### DIFF
--- a/.github/workflows/playwright-update.yml
+++ b/.github/workflows/playwright-update.yml
@@ -5,13 +5,14 @@ on:
     types: [created, edited]
 
 permissions:
-  contents: write
-  pull-requests: write
+  contents: read
 
 jobs:
   update-snapshots:
     if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, 'update playwright snapshots') }}
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # Required by actions/update-snapshots 
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/playwright-update.yml
+++ b/.github/workflows/playwright-update.yml
@@ -12,7 +12,7 @@ jobs:
     if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, 'update playwright snapshots') }}
     runs-on: ubuntu-latest
     permissions:
-      pull-requests: write # Required by actions/update-snapshots 
+      pull-requests: write # Required by actions/update-snapshots
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/auto_author_assign.yml
+++ b/.github/workflows/auto_author_assign.yml
@@ -6,10 +6,12 @@ on:
     types: [opened, reopened]
 
 permissions:
-  pull-requests: write
+  contents: read
 
 jobs:
   assign-author:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
     steps:
       - uses: toshimaru/auto-author-assign@v1.6.2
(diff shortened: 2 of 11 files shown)


.github/workflows/playwright-update.yml
```diff
@@ -5,13 +5,14 @@ on:
...
-  contents: write
-  pull-requests: write
+  contents: read
 
 jobs:
   update-snapshots:
     if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, 'update playwright snapshots') }}
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # Required by actions/update-snapshots
```
krassowski, line 15:
```suggestion
      pull-requests: write # Required by actions/update-snapshots
```

krassowski (commented):
Thank you! `jlpm` is just a version-pinned alias for `yarn`. Underneath a lot is delegated to https://github.com/jupyterlab/maintainer-tools so we probably shall review it at some point too.

jtpio (approved):
Thanks!

.github/workflows/playwright-update.yml
```diff
@@ -5,13 +5,14 @@ on:
     types: [created, edited]
 
 permissions:
-  contents: write
-  pull-requests: write
+  contents: read
```
jtpio, line 8: Wondering if this change could be the reason for the snapshot update workflow to not work anymore:

https://github.com/jupyter/notebook/actions/runs/6492244850/job/17630837598

image

diogoteles08: Hey! Yep, that seems like it. Sorry about that :/
It was difficult to interpret the needed permissions from the internal jupyterlab workflows, I only could [find that it required](https://github.com/jupyterlab/maintainer-tools/blob/main/.github/actions/update-snapshots/action.yml#L11) `pull-requests: write`
Saw that you already updated this to concede `contents: write` permission as well, that should work =)

Issue #7041, opened 2023-09-13
Fix keyboard shortcut regressions in new Notebook interface


jph00:
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

### Problem

@fperez committed that in moving over to a lab-based approach for the new notebook that there wouldn't be a regression in keyboard accessibility, and, more generally, the very strong developer experience in notebook classic. However, after upgrading, I'm already seeing key pieces missing. Therefore I'm opening this issue to attempt to track issues as I see them.

### Keyboard shortcuts

- Some of the shortcuts from the Collapsible Headings extension have been added - thank you! However, the ability to hold `shift` whilst pressing `right-arrow` has been lost -- this ought to select from the current cell to the end of the section
- ctrl-shift-q is shown in the menu as halting and closing the current notebook, which worked in classic, but it doesn't actually do anything for me (latest MacOS / Chrome)
- `o` no longer toggles showing/hiding the output of a cell, and I don't see any way to add this
- I can't see a way to add a keyboard shortcut for "run all above current cell" (or ditto for below cell)
- In general, there does not seem to be a way, at least in the GUI, to add keyboard shortcuts to all menu items.

### Proposal

Ideally, everything that can be achieved with the mouse would be surfaced as an element that could be tied to a keyboard shortcut via the GUI.

jph00:
I'd also like to make a request. @fperez in #6210 said:

- Notebook 7 would be released only after ensuring that it really meets the needs of current users as stated above. Critically, this means putting in the necessary work to port/adapt critical extensions like nbgrader, RISE and TOC to work on this future nb7 from Day 1.
- Notebook 6 would remain in minimal maintenance mode at least for a while. Details TBD.

I hope that nb 6 will remain in maintenance mode until enthusiastic nb 6 users have agreed that nb 7 has reached the high bar that v6 achieved! :D

jph00:
Thanks for pointing out that existing issue @RRosio.

I'll continue adding regressions here as I find them -- hope that's OK!

Some others that are causing me some trouble:

- A number of lab shortcuts on Mac OS override existing browser shortcuts, at least in Chrome. In particular, cmd-f no longer brings up the browser Find functionality, but instead brings up Lab's. However the Lab functionality does not search in cell outputs, so is not a substitute
- The undo buffer is no longer per-cell, so hitting cmd-z now (surprisingly, to me) changes cells other than the current one
[...]

Issue #6596, opened 2022-10-21
[Request] Keyboard shortcut parity with notebook classic


jtpio:
### Discussed in https://github.com/jupyter/notebook/discussions/6591

<div type='discussions-op-text'>

<sup>Originally posted by **lemontheme** October 18, 2022</sup>
Excited to learn that future versions of notebook will be based off the same core as JupyterLab. I'm creating this post as a user who has tried to embrace JupyterLab but, despite all its great extra features, has always returned to notebook classic. 

The reason: JupyterLab's keyboard shortcuts different enough that they frustrate many years of muscle memory built up in notebook classic. Sure, I can configure them to what I'm used to, but that means additional work each time I want to set up a (sometimes one-off) Jupyter environment somewhere. 

Having just tried the Notebook 7 alpha, I can already tell the shortcuts are more like those in JupyterLab. I'm curious, does anyone know why JupyterLab didn't just keep the same shortcuts as before? It's always striked me as such a weird UX decision.

As for the question in my title, I'd like to float the idea of reversing that move in Notebook 7. Sooner or later, notebook classic users will be upgrading to notebook 7 without realizing the important changes beneath the surface. Doubtless many of them will be less than excited to learn they need to learn new shortcuts to get their work done faster.

Or is it just me?</div>

jtpio:
This article might also be relevant: https://www.dataschool.io/jupyter-notebook-keyboard-shortcuts/

jtpio:
cc @Carreau since you opened a couple of recent issues about keyboard shortcuts

Carreau:
> I'm curious, does anyone know why JupyterLab didn't just keep the same shortcuts as before? It's always striked me as such a weird UX decision.

That's a complicated story with likely some things that I would prefer to talk only face to face. You will be at JupyterCon right ? We can try to discuss this.

jtpio:
Ah just to make it clear it was not me posting this message but @lemontheme. The discussion from https://github.com/jupyter/notebook/discussions/6591 was converted into an issue.

I have updated the top comment to disambiguate this.

Carreau:
Ah sorry, regardless. If someone wish to have some context, it's better to talk in person.

tonyfast:
adding some keyboard parity issues from lab

* https://github.com/jupyterlab/jupyterlab/issues/14641
* https://github.com/jupyterlab/jupyterlab/issues/14548
* https://github.com/jupyterlab/jupyterlab/issues/14738

PR #7075, opened 2023-09-26
[WIP] Adding classic shortcuts extension

RRosio:
- [ ] Enable/Disable mechanism
- [ ] Delete line shortcut

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -14,6 +14,7 @@
     "@codemirror/view": "~6.15.3",
     "@jupyter-notebook/application": "~7.0.4",
     "@jupyter-notebook/application-extension": "~7.0.4",
+    "@jupyter-notebook/classic-shortcuts-extension": "~7.0.4",
     "@jupyter-notebook/console-extension": "~7.0.4",
     "@jupyter-notebook/docmanager-extension": "~7.0.4",
     "@jupyter-notebook/documentsearch-extension": "~7.0.4",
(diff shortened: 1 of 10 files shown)


jtpio:
Thanks @RRosio for looking into this!

This would help some folks migrate from the classic notebook to Notebook 7.

Wondering if this new extension should only be available in Notebook 7 though? Or could it be that some JupyterLab users would also like to have this set of keyboard shortcuts available? If so, maybe it should be its own extension? (for example in https://github.com/jupyterlab-contrib)

jtpio:
Relevant issues:

- https://github.com/jupyterlab/jupyterlab/issues/14738
- https://github.com/jupyter/notebook/issues/6596
- https://github.com/jupyter/notebook/issues/7041

jtpio:
> Initially I was thinking that maybe it would not be an extension that we would necessarily expect to see in JupyterLab because there was no previous indication (to my knowledge) that JupyterLab would provide parity with the Notebook v6 user experience

True. Maybe the main advantage would be to have a coherent set of shortcuts between JupyterLab 4 and Notebook 7.

> I appreciated with this extension in the Notebook 7 codebase that there would be easy access to enable with a few minimal clicks.

Normally the number of clicks or other UX considerations should not depend on whether an extension lives in a different repo? Unless it requires an extra `pip install` to install it.

> I guess my biggest motivation to develop the extension outside Notebook 7 codebase would be whether it should be used with JupyterLab, but I am not sure that there is a precedent for that? Would you have a bias toward one option or the other?

So similar to the custom CSS feature, maybe the best would indeed be to develop the extension in the notebook repo like in this PR, and include it as part of the Notebook application but not the JupyterLab application.

If some users show interest to have the extension enabled in JupyterLab as well, we could convert the extension to a regular JupyterLab extension but still develop it in the `notebook` repo. Like the `lab-extension` package in this repo: https://github.com/jupyter/notebook/tree/main/packages/lab-extension

jtpio:
Adding to the `7.1` milestone as a nice-to-have feature: https://github.com/jupyter/notebook/pull/7075

jtpio:
@RRosio just checking in to see if you would like to have this for the 7.1 release?

I would like to publish a first 7.1 beta release soon, to try to stay in sync with the JupyterLab release cycle.

Otherwise no problem to postpone it for later and consider that for `7.2` (or a separate extension to start with).

PR #7090, opened 2023-10-05
Update index.ts to fix the issue

navdeepgill14:
### Issue Description
This pull request addresses the issue of 404 errors occurring when requesting favicons while executing notebook cells in Notebook 7, particularly when using Binder or a custom base URL. The problem originates from the hard-coded favicon URLs in the `index.ts` file.

### Proposed Changes
To resolve this issue, I have updated the favicon URLs in the `index.ts` file to be relative to the root of the application rather than hard-coded absolute paths. This change ensures that the correct favicon paths are generated, regardless of the base URL configuration.

### How This Fixes the Issue
By making the favicon URLs relative, they will adapt to the current application's base URL dynamically. This means that whether the Notebook is hosted on Binder or uses a custom base URL, the favicons will be correctly served without resulting in 404 errors.

### Changes Made
- Modified `index.ts` to use relative paths for favicon URLs.
- Removed hard-coded absolute paths.

### Testing
I have tested these changes in the following scenarios:
1. Locally hosted Notebook 7.
2. Notebook 7 on Binder with various base URLs.
3. Notebook 7 with a custom base URL configuration.

In each scenario, the favicons were successfully loaded without any 404 errors.

### Screenshots

Screen Shot 2023-10-05 at 1 30 54 pm

### Checklist
- [x] Updated `index.ts` with relative favicon URLs.
- [x] Removed hard-coded absolute paths.
- [x] Tested changes in various scenarios.
- [x] Added appropriate documentation/comments.
- [x] Screenshots for after.
- [x] Reviewed and addressed any linting/formatting issues.

### Related Issue
This pull request fixes issue #[7076].

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,7 @@ repos:
         args: ['-L', 'sur,nd']
         exclude: |
           (?x)^(
+              yarn.lock|
               binder/example.ipynb|
               docs/source/examples/images/FrontendKernel.graffle/data.plist|
           )$


jtpio:
Thanks @navdeepgill14 for looking into this :+1: 

> Notebook 7 on Binder with various base URLs.

Looks like this is still an issue on Binder? This is what it looks like when trying the Binder for this PR: https://mybinder.org/v2/gh/navdeepgill14/notebook/main?urlpath=tree

image

packages/notebook-extension/src/index.ts
```diff
@@ -411,8 +411,9 @@ const tabIcon: JupyterFrontEndPlugin<void> = {
   requires: [INotebookTracker],
   activate: (app: JupyterFrontEnd, tracker: INotebookTracker) => {
     // the favicons are provided by Jupyter Server
-    const notebookIcon = ' /static/favicons/favicon-notebook.ico';
-    const busyIcon = ' /static/favicons/favicon-busy-1.ico';
+    const baseURL = window.location.origin;
```
jtpio, line 414: Maybe this should be `PageConfig.getBaseUlr()` instead?

PR #7098, opened 2023-10-09
Ignore yarn.lock for codespell

jtpio:
Fix the recent CI failure as noticed on other PRs:

image


jtpio:
Looks like it fixes it:

image

PR #7103, opened 2023-10-12
Update to JupyterLab 4.0.7

jtpio:
Update to https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.7

PR #7104, opened 2023-10-12
Fix typings check on CI

jtpio:
Try fix the typings CI failure noticed in https://github.com/jupyter/notebook/pull/7103

Probably related to an upstream release.


jtpio:
Now passing:

image

Issue #7076, opened 2023-09-27
404 when loading notebook favicons


jtpio:
When using Notebook 7 on Binder (or with a different base url), requesting favicons when executing notebook cells give 404 error:

![image](https://github.com/jupyter/notebook/assets/591645/2f829dd6-f90f-4c70-824e-196d3898ba1d)

This is likely related to the following logic:

https://github.com/jupyter/notebook/blob/b397a3b834e6cf2e79e40ba9e1068abf97611192/packages/notebook-extension/src/index.ts#L413-L415

navdeepgill14:
Hello,

I'm interested in working on this issue. Could you please assign it to me? I'll be happy to take on the task and work towards a resolution.

Thank you!
Navdeep

navdeepgill14:
Hello everyone,

I'm excited to share that I've successfully resolved this issue! 🎉

Here are the steps I took to address the problem:

1. Updated the `index.ts` file to use relative paths for favicon URLs.
2. Removed the hard-coded absolute paths, allowing for dynamic URL generation.
3. Thoroughly tested these changes in various scenarios, including local hosting and Binder with different base URLs, and confirmed that the 404 errors no longer occur.

The corresponding pull request has been created and is ready for your review. You can view the details of the changes and test it on your end to ensure that it resolves the issue as expected.

**[Update index.ts to fix the issue](https://github.com/jupyter/notebook/pull/7090)**

I kindly request you to review the pull request at your earliest convenience. If it works as intended on your end and resolves the issue, please approve it, and I will proceed to close this issue.

Your feedback and collaboration are greatly appreciated.

Thank you!

Best regards,
Navdeep

jayeshsingh9767:
is this issue fixed? if not I will Take this up. 
Also, @jtpio could u please guide me about how to recreate this issue at my local environment.

jtpio:
Well there was an attempt at it linked in the comment just above: https://github.com/jupyter/notebook/pull/7090

But the contributor seems to have closed it since. So the issue is not fixed yet.

navdeepgill14:
Hi @jayeshsingh9767 and @jtpio, 
I want to provide an update on the status of this issue. I initially closed it, but it's not because I've abandoned it. I had to close it temporarily because I was working on another issue, and both of these were interrelated. As a result, I created two PRs on the same branch, and one had to be temporarily closed.

[...]

PR #7109, opened 2023-10-17
Updated fav-icon Base URL from JupyterLab PageConfig.

jayeshsingh9767:
We were Getting 404 because base URL of binder(different base url) was used for fetching fav-icons I have update code to use Base URL from JupyterLab PageConfig. 
This closes issue #7076

### DIFF
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -415,8 +415,8 @@ const tabIcon: JupyterFrontEndPlugin<void> = {
   activate: (app: JupyterFrontEnd, tracker: INotebookTracker) => {
     // the favicons are provided by Jupyter Server
     const baseURL = PageConfig.getBaseUrl();
-    const notebookIcon = ' /static/favicons/favicon-notebook.ico';
-    const busyIcon = ' /static/favicons/favicon-busy-1.ico';
+    const notebookIcon = `${baseURL}/static/favicons/favicon-notebook.ico`;
+    const busyIcon = `${baseURL}/static/favicons/favicon-busy-1.ico`;
 
     const updateBrowserFavicon = (
       status: ISessionContext.KernelDisplayStatus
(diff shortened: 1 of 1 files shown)


packages/notebook-extension/src/index.ts
```diff
@@ -422,10 +426,10 @@ const tabIcon: JupyterFrontEndPlugin<void> = {
       ) as HTMLLinkElement;
       switch (status) {
         case 'busy':
-          link.href = busyIcon;
+          link.href = baseURL + busyIcon;
```
jtpio, line 429: Should this string concatenation be moved above where `notebookIcon` and `busyIcon` are defined?

jtpio: Also it could be combined with `URLExt.join()` (some example uses in this repo).

jayeshsingh9767: Thanks for the Quick review. This will be more readable. have Updated the PR.

jayeshsingh9767:
@jtpio could u please guide me about why **Build / Test Lint (pull_request)** is failing?, Logs of the job are pointing to file [/notebook/blob/main/notebook/app.py](https://github.com/jupyter/notebook/blob/main/notebook/app.py).

jtpio:
The lint failure looks related to an upstream change, maybe this new release? https://github.com/jupyter-server/jupyter_server/releases/tag/v2.8.0

Similar to https://github.com/jupyter/notebook/pull/7104, we can fix that separately.

jtpio:
FYI @jayeshsingh9767 I've updated the branch to include https://github.com/jupyter/notebook/pull/7104, which should hopefully help fix the CI.

jtpio:
The new failure can likely be fixed by running `jlpm prettier` at the top-level (normally done automatically via the `pre-commit` hook though).

packages/notebook-extension/src/index.ts
```diff
@@ -38,6 +38,9 @@ import { Widget } from '@lumino/widgets';
 
 import { TrustedComponent } from './trusted';
 
+import { PageConfig, URLExt } from '@jupyterlab/coreutils';
```
jtpio, line 41: These imports can be grouped with the other imports above.

jayeshsingh9767: Got it will update the PR.

jtpio (approved):
Thanks!

Just checked with a different base URL and it seems to be working fine:


https://github.com/jupyter/notebook/assets/591645/c5ef8b77-ee95-449b-a3d3-246c1a374a47

jayeshsingh9767:
Thanks @jtpio

PR #7153, opened 2023-11-14
CI: Hashpin github-actions with dangerous permissions

### LINKED ISSUE #7119: Hash-pin GitHub Actions that have access to sensitive permissions and secrets
Hi, I'm Diogo and I raised the issue #7069 suggesting you to set minimal permissions to your workflows. Now I'm coming back to  suggest a modification that would provide extra safety for the workflows that yet require dangerous permissions (e.g., `contents: write`).

### Problem

Some of your workflows (e.g. [auto_author_assign.yml](https://github.com/jupyter/notebook/blob/main/.github/workflows/auto_author_assign.yml) or [playwright-update.yml](https://github.com/jupyter/notebook/blob/main/.github/workflows/playwright-update.yml)) are using dangerous permissions  permissions while running external dependencies pinned only by tag. At [publish-release.yml](https://github.com/jupyter/notebook/blob/main/.github/workflows/publish-release.yml) they're not called with dangerous permissions but have access to important secrets (although at this case I see that it's a jupyter-lab action, so I'd understand if you prefer to just trust this one). Those patterns could be dangerous because if any of those actions get hijacked (and at the end they're all repositories and are susceptible to attacks like any other), an attacker could change the code that your tags point to, gaining access to your secrets and/or write permissions to your code.

### Proposed Solution

A simple solution for this problem would be to hash-pin those sensitive actions, pointing the actions to the very specific commit of that release. It follows and example of the change:
```
- uses: r-lib/actions/pr-fetch@v2 
would become
- uses: r-lib/actions/pr-fetch@11a22a908006c25fe054c4ef0ac0436b1de3edbe # v1.3.1
```

It shouldn't require any more maintenance, because dependabot is also able to keep them updated. It would update them at the same pace you're already used to, and it would still keep a comment with the human-readable version used =).

### Conclusion
Let me know if you are interested in this change and I'd be happy to send a PR solving this.

Thanks!

diogoteles08:
Closes #7119


diogoteles08:
Hey @krassowski, I've updated the PR after a version update dated after my initial PR, and resolved the merge conflicts.

It seems like the "Enforce PR label" is still blocking the merge, let me know if there is anything I can do to help solving it.

Cheers,

krassowski:
Will we get clever dependabot updates, or will we need to manually update the pins?

jtpio:
Should this be discussed in a team compass repo maybe? For example https://github.com/jupyterlab/team-compass?

These actions are used by many Jupyter projects, and if going with these pins then there is no reason for it to be specific to the `notebook` repo, but would likely be relevant to other repos as well? For example https://github.com/jupyterlab/jupyterlab, https://github.com/jupyterlab/jupyterlab_server, https://github.com/jupyter-server/jupyter_server, and more.

.github/workflows/playwright-update.yml
```diff
@@ -40,12 +40,12 @@ jobs:
...
 
       - name: Base Setup
-        uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+        uses: jupyterlab/maintainer-tools/.github/actions/base-setup@4b05d5f62ffa18bed92d556309c86a95554401b9 # v1
 
       - name: Build
         uses: ./.github/actions/build-dist
 
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
```
jtpio, line 48: So we can't trust GitHub's own actions anymore?

diogoteles08: I don't think we should just "not trust github", the company has huge credibility and should be more trustful than random or less-known workflows. But we must recognize that the security of their actions is held differently than the security of any of their other infrastructures. Their main infrastructure is closed-source and should have their own security (I can't really tell, but we hope (?)); but their actions are just open source repositories (see https://github.com/actions/download-artifact), and could be vulnerable to risks like any others.

In their [own documentation](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), Github mentions that using tag-versioning can be risky even while only using actions from trusted authors.

And besides that, if we're already hash-pinning the other actions that are run in sensitive context, there is also a "why not" factor to also hash-pin the sensitive ones from GitHub.

diogoteles08:
> Will we get clever dependabot updates, or will we need to manually update the pins?

You'll receive clever dependabot updates! It would update them at the same pace you're already used to, and it would still keep a comment with the human-readable version used =).

If you wish, we can also configure it to update all github actions in a single monthly PR, for example.


> Should this be discussed in a team compass repo maybe? For example https://github.com/jupyterlab/team-compass?

Let me know if you want my help raising this discussion anywhere else. I'd be happy to help =)

jtpio:
In the meantime there has been some discussion about this and other project in the Jupyter ecosystem have also adopted zizmor for pinning GitHub Actions.

So this should now be addressed via:

- https://github.com/jupyter/notebook/pull/7880
- https://github.com/jupyter/notebook/pull/7938

Thanks @diogoteles08 for initially proposing this!

Issue #7137, opened 2023-11-02
Change the interface switcher widget


brichet:
### Problem

The interface switcher widget is currently a `MenuBar` widget (from Lumino) with only one first level item (`Open with...`).
The `MenuBar` widget catches the left and right arrow keys event, to move from one menu item to an other (in this case nothing happen since there is only one item).
This behavior breaks attempt to improve accessibility in toolbar by using arrow keys to navigate between items (https://github.com/jupyterlab/jupyterlab/pull/15021#issuecomment-1790795589).

### Proposed Solution

Using a select component, more appropriate for a toolbar.

PR #7166, opened 2023-11-30
Bump to JupyterLab 4.1.0a4

jtpio:
Update to the freshly released `4.1.0a4`: https://github.com/jupyterlab/jupyterlab/releases/tag/v4.1.0a4


jtpio:
Ouch a couple of breakages in that one it seems...

Breaking change in Galata?

image


Icons are now placed on the left?

image

Some extra label:

image

brichet:
> I believe this is mostly https://github.com/jupyterlab/jupyterlab/pull/15021 CC @brichet 

Yes probably, I'll look at it @jtpio

jtpio:
Bumping the `@jupyterlab/galata` helps get past the `await page.filebrowser.refresh()` error, which makes sense as the selector was changed in https://github.com/jupyterlab/jupyterlab/pull/15021.

Now some tests are failing when trying to do file manipulations via the file browser toolbar:

https://github.com/jupyter/notebook/blob/ef1f8db5c22b61107ef600e10f867db66be98566/ui-tests/test/filebrowser.spec.ts#L20-L83

[d3efbeff9e3336758817295039be2073bc8d89e4.webm](https://github.com/jupyter/notebook/assets/591645/d2ca458b-90ba-45c2-88e4-363c0a8ef752)

jtpio:
So the remaining issue now are:

**Refresh File List button label**

Probably related to the switch to `jupyter-ui-toolkit`?

image

**Left icon for the interface switcher toolbar button**

If this is the way to go or if there are reasons to do this w.r.t accessibility then it's fine to leave it this way (we'll just update the reference snapshots)

image

---

Some other tests are failing due to snapshot comparison because of some minor style update, but we can just update the reference snapshots.

brichet:
> Refresh File List button label

Also a selector to change at 
https://github.com/jupyter/notebook/blob/ef1f8db5c22b61107ef600e10f867db66be98566/packages/tree/style/base.css#L45

to `jp-button[data-command='filebrowser:refresh'] .jp-ToolbarButtonComponent-label`

brichet:
There will probably be some other side effect, like the padding left of the interface switcher icon
https://github.com/jupyter/notebook/blob/ef1f8db5c22b61107ef600e10f867db66be98566/packages/lab-extension/style/base.css#L21-L25

which should be 
```css
.jp-nb-interface-switcher-button
  > .jp-ToolbarButtonComponent
  > svg {
  padding-left: 3px;
}
```

brichet:
> **Some CSS inconsistencies on the file action buttons (wrong height)**

That one is more tricky, since it's not only about style but also about accessibility.

Currently these `FileActions` buttons are included in a widget, adding a dom node to the main toolbar.
Therefore, the buttons cannot be focused like the other toolbar button.

We should include these buttons as direct child of the filebrowser toolbar.

jtpio:
OK that makes sense, thanks @brichet for providing these details :+1:

krassowski:
There is also https://github.com/jupyter/notebook/issues/7137 about the interface switcher that was created during review.

@brichet would you have time to add the example style changes that extension authors may need to make to the extension migration guide section? https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#use-of-ui-toolkit-for-toolbar-and-toolbarbuttoncomponent

jtpio:
Closing in favor of https://github.com/jupyter/notebook/pull/7172.

PR #7183, opened 2023-12-13
[7.0.x] Install stable JupyterLab 4.0 in the releaser hook

jtpio:
Investigate CI failures:

- [x] Use stable JupyterLab 4.0 in the releaser hook to avoid https://github.com/jupyterlab/jupyterlab/issues/15503
- [x] ~Fix typings~ -> defered to backporting some other PRs

image


jtpio:
Gonna try to backport some other maintenance PRs to fix the typing check.

PR #7190, opened 2023-12-14
Update ruff config

blink1073:
Use updated recommendations for Ruff [config](https://learn.scientific-python.org/development/guides/style/#ruff), and the new ruff [`docstring-code-format`](https://astral.sh/blog/ruff-v0.1.8#formatting-code-snippets-in-docstrings) that replaces `blacken-docs`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,15 +21,9 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.27.2
+    rev: 0.27.3
     hooks:
       - id: check-github-workflows
-
-  - repo: https://github.com/adamchainz/blacken-docs
-    rev: '1.16.0'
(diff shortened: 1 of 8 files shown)


jtpio:
Thanks!

The UI tests failure might be relevant? Unless it is caused by a new release of an underlying package?

**actual (extra home icon)**

image

**expected**

image

blink1073:
I addressed that failure, but it looks like there is a pre-existing race condition, because this firefox test failed with the same error as this chromium test did: https://github.com/jupyter/notebook/actions/runs/7198343749/job/19607743050

jtpio:
Thanks!

Is this the same error? Now the issue seems to be related to the trusted indicator not being displayed correctly:

**Actual**

image

**Expected**

image

But this is likely not related to this PR, as it also happens in https://github.com/jupyter/notebook/pull/7191 now.

jtpio (approved):
Thanks!

Issue #7248, opened 2024-02-09
Add an upper bound on the `jupyterlab` dependency


jtpio:
Currently Notebook 7.0.x depends on `jupyterlab >=4.0.2,<5`:

https://github.com/jupyter/notebook/blob/99db9423e34e4c1afc0f0ddbb01e86d4099bcd75/pyproject.toml#L35

However as noticed in https://github.com/jupyter/notebook/issues/7243, this may cause some issues when a version of Notebook built on the 4.0.x JupyterLab packages pulls a new minor version of JupyterLab, in this case 4.1.0.

Maybe we should consider aligning the versions more closely. For Notebook 7.0.x that would mean depending on `jupyterlab >=4.0,<4.1`, and also update the JS dependencies to use `~` instead of `^`, for example switch to `"@jupyterlab/application": "~4.0.0"`.

PR #7249, opened 2024-02-09
Add upper bound on `jupyterlab<4.1`

jtpio:
Fixes https://github.com/jupyter/notebook/issues/7248


jtpio:
We could skip the switch to `~` in this PR (mentioned in #7248), but do it for `main` (currently tracking `7.1.0` and JupyterLab 4.1.0)

Issue #7216, opened 2024-01-21
UI issue in the "tree view" if directory called "notebooks" is in the path.


avloss:
## Description

Getting a pesky warning in the "tree" view, if path to the directory includes "notebooks"(with that exact spelling).

## Reproduce

1. Create a path that includes "notebooks", and has sub-directories. ("http://localhost:8888/tree/test/notebooks/test")
2. Navigate to that path via UI tree view.
3. Click Refresh
4. Get a pesky warning:

```
File Load Error for test_dire
Invalid response: 404 Not Found
```
5. Renaming directory "notebooks" to **LITERALLY ANYTHING** (including capitalising any of the letters) removes the issue.

## Expected behavior

No warning happens
<img width="1152" alt="jupyter-tree-warning" src="https://github.com/jupyter/notebook/assets/17853685/38d502d3-c223-4817-8cc0-90aabb53cabe">

jtpio:
Thanks @avloss for reporting the issue :+1: 

Just to double check, which version of Notebook 7 are you using?

avloss:
> Thanks @avloss for reporting the issue 👍
> 
> Just to double check, which version of Notebook 7 are you using?

notebook==7.0.7

SATOSHIFNAKAMOTO:
Hello all 👋,

I've been able to reproduce the "tree view" issue on my end. After navigating to the /test/notebooks/test directory in the Jupyter Notebook UI, I clicked the browser's refresh button (specifically in Firefox, not the Jupyter in-app refresh), and I encountered the "File Load Error for test" stating "test is a directory, not a file".

Here's the detailed setup I'm working with:

Jupyter Notebook version: 7.1.0a2
Operating System: macOS 13.6.3 (22G436)
Browser: Firefox 121.0.1 (64-bit)
Attached is a screenshot showing the error message that appeared. It seems that the browser refresh is being misinterpreted, leading to this unexpected error. Has anyone had a similar experience, or does anyone have insights on what might be causing this behavior?

Best,
Satoshi 中本哲史

<img width="672" alt="Screenshot 2024-01-24 at 10 09 00 AM" src="https://github.com/jupyter/notebook/assets/135542247/d4c5550c-ff9e-47bc-b8b7-12c1751f8487">

itsmevichu:
@jtpio I have raised a PR [#7253](https://github.com/jupyter/notebook/pull/7253) please look into it.

PR #7253, opened 2024-02-10
UI issue in the "tree view" #7216

itsmevichu:
**Issue No.** [#7216](https://github.com/jupyter/notebook/issues/7216) 

### **Problem**
The problem arises whenever these keywords, namely {“notebooks”, “edit”, ”consoles”, ”terminals”}, are detected in the path. This leads to undesired behaviour when directories with these keywords are present.

Some of our plugins actively monitor the URL path for the above keywords to execute corresponding commands. When these plugins come across the path with matching keywords(directories) during file browsing(tree view), they are forced to execute the associated command, which causes the issue.

### **Solution**
1. Since our file browsing consistently include the "tree" keyword in the path, I've explicitly handled the situation by ensuring that the plugins do not execute when the path contains the keyword "tree".
2. In another scenario, when the folder structure is like /terminals/temp.ipynb, opening it redirects the path to /notebooks/terminals/temp.ipynb(also applies for other file formats); this unintentionally triggers the terminal-extension plugin. I alse have addressed this specific case explicitly to prevent such occurrences (This also applies to consoles).

Thank you for considering my contribution. And I welcome any feedback or suggestions for further improvements.


packages/application-extension/src/index.ts
```diff
@@ -200,11 +200,18 @@ const opener: JupyterFrontEndPlugin<void> = {
   ): void => {
     const { commands, docRegistry } = app;
 
+    const ignoreTreePattern = new RegExp('/tree/(.*)');
```
jtpio, line 203: Curious if this approach works when a user creates a folder called "tree"?

itsmevichu: @jtpio Thanks for pointing this out, the above code will not work for the scenario that you've mentioned. I have revised the code with a new approach. Please have a look into it.

packages/application-extension/src/index.ts
```diff
@@ -200,16 +200,24 @@ const opener: JupyterFrontEndPlugin<void> = {
...
         const parsed = args as IRouter.ILocation;
         const matches = parsed.path.match(TREE_PATTERN) ?? [];
+
         const [, , path] = matches;
         if (!path) {
           return;
         }
 
+        const pathSegments: string[] = parsed.path.split(pathSegmentPattern);
+        if (pathSegments.length > 1 && pathSegments[0].match(ignoredPattern)) {
```
jtpio, line 217: Wondering if following a similar logic as in JupyterLab (disposing the command after routing) could help keep things simpler and avoid special cases using patterns? 

https://github.com/jupyterlab/jupyterlab/blob/cdca0a20ba84af21922c390c46e086e50120d4cc/packages/application-extension/src/index.tsx#L938-L939

goelakash:
This looks like it's been superseded by #7903 which fixes the same issue (#7216) with a simpler regex anchor change and includes a regression test. Can this one be closed?

DTiming24:
This looks like a sensible wiring fix across the app, console, and terminal extensions. If there are other tree-view entry points, a quick smoke test for those panes would help confirm the behavior stays consistent everywhere.

Issue #7271, opened 2024-02-23
Add GitHub workflow to automatically update to newer JupyterLab versions


jtpio:
To ease maintenance, it would be useful to have a script that checks what the latest version of JupyterLab is, and automatically opens a PR that bumps to the newer packages.

Some ideas and possible config options for that script:

- run as a cron, and / or via a workflow dispatch on demand
- support updating to a specific version, `latest` (stable) or `next` (pre-release)
- this might also require being able to specify a branch, as multiple versions of notebook (stable and pre-release) can be developed at the same time

JupyterLab Desktop has something similar: https://github.com/jupyterlab/jupyterlab-desktop/pull/311

jtpio:
In some cases there is more work to do than just updating the versions (for example integrating some upstream plugins more natively in Notebook). But at least this script would help with the repetitive parts.

itsmevichu:
@RRosio @jtpio Can we close this issue?

jtpio:
Right, there is now a workflow, but it currently needs to be triggered manually (I've been using it from the fork for the past few months).

krassowski:
The approach from https://github.com/jupyterlab/jupyterlab-desktop/blob/master/.github/workflows/sync_lab_release.yml should work here too, it requires setting up a new app.

PR #7281, opened 2024-02-27
Workflow to update JupyterLab dependencies automatically

itsmevichu:
**Issue**- [#7271](https://github.com/jupyter/notebook/issues/7271)

**Description:** 
This workflow can be used to update all the jupyterlab dependecies to the latest release. This can also be manually triggered, which takes the JupyterLab version as an input.

**Things to re-visit:**
-  Github secrets, user, email - to create branch and raise PR.
-  Executing "yarn install" post dependency updates results in a substantial amount of modifications, with over 15,000 additions and 15,000 deletions. - Done 
- Scheduling time for cron job.

The created PR looks like - [Sample pull-request](https://github.com/itsmevichu/notebook/pull/10)

Feel free to adapt the workflow based on your requirements.


jtpio:
Thanks @itsmevichu for looking into this!

jtpio:
> Executing "yarn install" post dependency updates results in a substantial amount of modifications, with over 15,000 additions and 15,000 deletions.

Maybe it's because it's using the default yarn (version 1), while the repo here uses yarn 3 via `jlpm`. So we would probably need to install `jupyterlab` and use `jlpm`.

jtpio:
Thanks @itsmevichu.

Looks like this needs a lint pass.

itsmevichu:
@jtpio The lint check is failing because of the coverage for the scripts. Is it possible for us to exclude the "scripts" folder from coverage checks?

itsmevichu:
@jtpio I am fetching the latest versions of the lab dependencies from the below URL - https://raw.githubusercontent.com/jupyterlab/jupyterlab/v4.1.4/jupyterlab/staging/package.json.

The package "@jupyterlab/testutils" is not present in the above package.json. As a result, only that specific package won't get updated by this PR.

jtpio:
> @jtpio The lint check is failing because of the coverage for the scripts. Is it possible for us to exclude the "scripts" folder from coverage checks?

Yes I think so. An alternative could be to write the scripts in TypeScript and put them alongside other scripts here: https://github.com/jupyter/notebook/tree/main/buildutils/src



> The package "@jupyterlab/testutils" is not present in the above package.json. As a result, only that specific package won't get updated by this workflow.

Right it's fine, we only need to update the packages used here in the notebook repo.

itsmevichu:
@jtpio I have converted the scripts to typescript and included them in the buildutils. 

**Pending tasks:**
1. To refer the appropriate secrets. https://github.com/jupyter/notebook/blob/2154f0f039e9a5f3b8c42f94934dc8f6023b7ecd/.github/workflows/upgrade-juypterlab-dependencies.yml#L67-L68
2. Cron job timings.
https://github.com/jupyter/notebook/blob/2154f0f039e9a5f3b8c42f94934dc8f6023b7ecd/.github/workflows/upgrade-juypterlab-dependencies.yml#L5

itsmevichu:
@jtpio Are any changes needed in this PR? If not, we could simply utilize this workflow, as it would minimize the effort required to upgrade dependencies with each JupyterLab release.

krassowski:
It would be really handy to have this in!

buildutils/src/upgrade-lab-dependencies.ts
```diff
@@ -0,0 +1,101 @@
...
+    const existingPackageJson = JSON.parse(fs.readFileSync(filePath, 'utf-8'));
+
+    const newDependencies = {
+      ...newPackageJson.devDependencies,
+      ...newPackageJson.resolutions,
+    };
+
+    updateDependencyVersion(existingPackageJson, newDependencies);
+
+    fs.writeFileSync(filePath, JSON.stringify(existingPackageJson, null, 2));
```
krassowski, line 46: I tried that just now and noticed that it is missing a final new line so maybe:

```suggestion
    fs.writeFileSync(filePath, JSON.stringify(existingPackageJson, null, 2) + '\n');
```

?

krassowski:
It looks like it missed one package, `"@jupyterlab/testutils": ~4.2.0`

itsmevichu:
> @jtpio I am fetching the latest versions of the lab dependencies from the below URL - https://raw.githubusercontent.com/jupyterlab/jupyterlab/v4.1.4/jupyterlab/staging/package.json.
> 
> The package "@jupyterlab/testutils" is not present in the above package.json. As a result, only that specific package won't get updated by this workflow.

I noticed the same, but it seems that the @jupyterlab/testutils information is not present in our source package.json.

buildutils/src/upgrade-lab-dependencies.ts
itsmevichu: I think adding a newline isn’t necessary since our current package.json doesn’t have one, and even though if we have one our lint-check would flag it. Correct me If I am wrong.

krassowski: So we need a single new line and lack of it is actually flagged by precommit linter. We do not need/want a trailing new line, just the last line to end with new line symbol.

krassowski:
> I noticed the same, but it seems that the @jupyterlab/testutils information is not present in our source package.json.

I think it is under `devDependencies`.

itsmevichu:
> > I noticed the same, but it seems that the @jupyterlab/testutils information is not present in our source package.json.
> 
> I think it is under `devDependencies`.

I've looked through, it appears to be missing from devDependencies. Can I proceed adding it in the respective jupyterlab's package.json?

krassowski:
See this commit: https://github.com/jupyter/notebook/pull/7447/commits/f7f2ead7eb1575b42b0f438acdbb467949b642db

jtpio:
Thanks @itsmevichu and sorry for the late reply!

Looks like we could try this workflow for the next updates and iterate on it if needed :+1: 

The lint failure seems related, I'll push a fix right after.

The other CI failure don't seem related and should hopefully be fixed soon: https://github.com/jupyter/notebook/pull/7539#issuecomment-2553290617

jtpio (approved):
Thanks!

Let's try it!

jtpio:
Looks like it managed to create and push the branch, but failed to open the PR: https://github.com/jupyter/notebook/actions/runs/12435633780/job/34721856783

jtpio:
Opened https://github.com/jupyter/notebook/issues/7545 to track follow-up improvements.

PR #7291, opened 2024-03-08
Update to JupyterLab 4.1.5 packages

jtpio:


jtpio:
Seeing the following behavior with the notebook toolbar items, which might be new:

[notebook-7-toolbar-items.webm](https://github.com/jupyter/notebook/assets/591645/de010174-ac78-4c75-a143-26cb14ca02f1)

jtpio:
> Seeing the following behavior with the notebook toolbar items, which might be new:

cc @brichet could this be related to https://github.com/jupyterlab/jupyterlab/pull/15843?

brichet:
@jtpio It is probably related to the PR you mentioned, but I don't know why yet.
There is also an error in the console:

brichet:
The error seems to come from https://github.com/jupyterlab/jupyterlab/blob/864754e0b4274df82ca93296fb9bf7f80ff9f710/packages/ui-components/src/components/toolbar.tsx#L640.
This could mean that an item has not been added in the toolbar using the `insertItem()` method (https://github.com/jupyterlab/jupyterlab/blob/864754e0b4274df82ca93296fb9bf7f80ff9f710/packages/ui-components/src/components/toolbar.tsx#L241).

brichet:
> Maybe it's related to the interface switcher extension from https://github.com/jupyter/notebook/blob/main/packages/lab-extension/src/index.ts (I think that would be the main difference between lab and notebook for the notebook toolbar). Although this one should go through the regular toolbar system via the settings normally.

I tried to disabled it but have the same issue. But I agree it seems to be the only widget added AFAIK.

jtpio:
Also this does not seem to be an issue with JupyterLab when testing from this PR (JupyterLab also shows the interface switcher widget in the notebook toolbar).

brichet:
> Also this does not seem to be an issue with JupyterLab when testing from this PR (JupyterLab also shows the interface switcher widget in the notebook toolbar).

I think I found the culprit on Jupyterlab, probably [this call](
https://github.com/jupyterlab/jupyterlab/blob/864754e0b4274df82ca93296fb9bf7f80ff9f710/packages/ui-components/src/components/toolbar.tsx#L489)

If we replace `void this._onResize();` with `void this._resizer.invoke();` it works well.

My understanding is that the toolbar is rendered more quickly in Notebook than in Jupyterlab (probably because there is less to render).
Therefore this function is called for each items added to the toolbar at load time, which cause confusion in the computation of the widget to display or not in the toolbar. If we display the widgets (to compute) in console before [this line](https://github.com/jupyterlab/jupyterlab/blob/864754e0b4274df82ca93296fb9bf7f80ff9f710/packages/ui-components/src/components/toolbar.tsx#L640) some of them are undefined.



This does not occur with the fix above.

I'll open a PR on Jupyterlab.

jtpio:
Cool, thanks for investigating this!

jtpio:
CI now passing with the fix shipped in JupyterLab 4.1.5. Thanks @brichet!

PR #7299, opened 2024-03-18
update RISE extension installation instructions

afeld:
Removed out-of-date information. Pointed to that README rather than copying instructions here, since they may get a bit more complicated/nuanced (see https://github.com/jupyterlab-contrib/rise/pull/78).


jtpio (approved):
Thanks!

Issue #7231, opened 2024-02-01
CSS issue when the Notebook windowing mode is set to `full`


jtpio:
Testing on the latest `main`, which includes the recent changes that landed in JupyterLab w.r.t windowing mode, it looks like the CSS is off:

![image](https://github.com/jupyter/notebook/assets/591645/97ecf393-55e4-491d-bccc-d411a091bd69)

This is likely related to an underlying DOM structure change, which affects the CSS customizations in Notebook.

jtpio:
Adding to `7.1.x` since `full` is not the default windowing mode.

jtpio:
Marking as a "good first issue" if someone would like to have a look:

The windowing mode can be set to `full` via the settings editor:

![image](https://github.com/jupyter/notebook/assets/591645/57523248-d096-4c42-9295-12e67d2bfa3d)

This will also likely require tweaking some CSS depending on the DOM structure: https://github.com/jupyter/notebook/blob/main/packages/notebook-extension/style/base.css

kgaurav8026:
Hi @jtpio,
I’m interested in contributing to JupyterLab, and I’d like some additional context regarding the issue related to cell alignment. From what I understand, the issue pertains to the misalignment of the cell box when windowing. Specifically, should the cell box be centered on the screen?

jtpio:
Yes it should be centered like when using the default mode:

![image](https://github.com/jupyter/notebook/assets/591645/9ad11bc5-991a-484d-aac9-2b32cc7a2419)

v-lozko:
Hello, would it be possible to have this assigned to me?

Thanks

kgaurav8026:
Hi @v-lozko , i am finding it difficult to fix the bug.  So you can start working on it if you want. 
Thanks

kgaurav8026:
@v-lozko Also, if you are able to merge Kindly tell me the approach as i am keen on understanding how to tackle such issues in the future.

v-lozko:
hello @jtpio, i can't seem to recreate the issue. I was a bit busy the past couple days and finally able to get to it and with full windowing mode enabled when i create a new notebook it's centered.

v-lozko:
Hello @jtpio i can't seem to replicate the issue. This is the first issue i'm dealing with in Jupyter, but it definitely seems like i set up the environment correctly, and having changed my settings to full i do not have the CSS issue that you shared. it looks fine.

jtpio:
[...]

PR #7312, opened 2024-04-03
Update to JupyterLab 4.2.0b0

jtpio:
Update to https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0b0


jtpio:
Now seeing the following error in the dev tools console, which seems related to https://github.com/jupyterlab/jupyterlab/pull/14683 :

image

jtpio:
cc @afshin do you know what the main DOM differences are between the `defer` and `full` modes? Gonna look into updating the CSS to fix https://github.com/jupyter/notebook/issues/7231.

For now the browser default scrollbar is still missing.

krassowski:
> [Default to defer for now](https://github.com/jupyter/notebook/pull/7312/commits/2bba503cd053b88f1e3bf2c8d6b7fced4cb7d637)

I think this is fine. There is less of a user benefit for Notebook than for Lab from using full windowed mode because users may have 10 long notebooks in Lab opened in a single browser tab, but not in Notebook.

jtpio:
Is there a way to set the windowing mode without going through the settings?

Otherwise this also changes the windowing mode used in JupyterLab, as the settings are shared between notebook and lab. There is a similar issue with Notebook setting the file browser checkboxes by default, which currently also show up in lab (https://github.com/jupyterlab/jupyterlab/issues/14623).

krassowski:
It looks like you would need to override the `notebookConfig` in factory: https://github.com/jupyterlab/jupyterlab/blob/6152e3a5f94d728ccd2727d6d8413f12ddafb669/packages/notebook-extension/src/index.ts#L1867-L1898

An alternative would be to use setting transforms to change the default value temporarily.

jtpio:
> An alternative would be to use setting transforms to change the default value temporarily.

Looks like that would mean using a setting transform for the `@jupyterlab/notebook-extension:tracker`, which is the one defining the `windowingMode`? https://github.com/jupyterlab/jupyterlab/blob/6152e3a5f94d728ccd2727d6d8413f12ddafb669/packages/notebook-extension/schema/tracker.json#L816
But the setting file does not seem to have `jupyter.lab.transform` set to `true`.

jtpio:
Opened https://github.com/jupyter/notebook/issues/7318 to have a look later to not block this PR.

If anyone would like to help with this or https://github.com/jupyter/notebook/issues/7231 that would be appreciated, thanks!

PR #7324, opened 2024-04-14
Add --user option to pip install command

raphaelgpalma:
this adjustment was made to address installation issues observed during compilation on my local machine

by adding --user , we avoid the need for administrative permissions when installing the package. This makes it more practical for users to install the package without requiring elevated privileges


krassowski:
Possibly a better approach would be to recommend installing into a virtual environment, which does not require elevated privileges. Also there are some caveats to consider, see https://jupyterlab.readthedocs.io/en/latest/getting_started/installation.html#pip

raphaelgpalma:
I agreed, thank you for the guidance!

PR #7337, opened 2024-04-22
Fix CSS for `full` windowing mode

jtpio:
Extract part of https://github.com/jupyter/notebook/pull/7321.

Fixes https://github.com/jupyter/notebook/issues/7231
Closes https://github.com/jupyter/notebook/pull/7301


jtpio:
Moving forward with this, so there is some kind of fix available.

But the full windowing mode definitely deserves a second look so it can be enabled by default, maybe as part of https://github.com/jupyter/notebook/pull/7321 or https://github.com/jupyter/notebook/issues/7318.

PR #7343, opened 2024-04-26
Ignore `check-wheel-content` duplicate file error

RRosio:
In attempting to make a release for `6.5.x`, during `jupyter-releaser check-python`, `check-wheel-content` gives a duplicate file error: https://github.com/jupyter/notebook/actions/runs/8851516368/job/24308219469

### DIFF
--- a/notebook/bundler/tests/resources/another_subdir/test_file.txt
+++ b/notebook/bundler/tests/resources/another_subdir/test_file.txt
@@ -1,1 +1,1 @@
-Test file used to test globbing.
+Used to test globbing.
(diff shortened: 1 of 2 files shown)


krassowski:
Jumping across `actions/setup-python` versions should be mostly painless, there were not many breaking changes with the default options.

jtpio:
Maybe we could also just ignore that duplicate file error for now, to be able to move forward with the release?

For reference this was also an issue with the JupyterLab extension cookiecutter, which still has it ignored here:

https://github.com/jupyterlab/extension-cookiecutter-ts/blob/4111b5d2fb7981ec4af3c4694b7b7ae6134bf497/%7B%7Bcookiecutter.python_name%7D%7D/pyproject.toml#L87-L88

PR #7333, opened 2024-04-19
Update to JupyterLab 4.2.0rc0

jtpio:
Update to:

- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0b2
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0b3
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0rc0

PR #7346, opened 2024-04-27
Pin on `macos-12` on CI for now

jtpio:
Fixes https://github.com/jupyter/notebook/issues/7345


jtpio:
The UI tests failure are probably related to something else and not related to this change, to be investigated as part of https://github.com/jupyter/notebook/pull/7333.

Issue #7221, opened 2024-01-22
Blank page if `local_names` are non trivial


daskol:
It does not work is `c.ServerApp.local_hostnames = ['localhost', 'jupyter']` in config. In brower console one can find this.

```
6774.bundle.js:1 
        
        
       Failed to load resource: the server responded with a status of 404 (Not Found)
tree:29 Refused to execute script from 'http://jupyter:8888/static/notebook/6774.bundle.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
bundle.js:1 Uncaught (in promise) ChunkLoadError: Loading chunk 6774 failed.
(error: http://jupyter:8888/static/notebook/6774.bundle.js)
    at v.f.j (bundle.js:1:49164)
    at bundle.js:1:1260
    at Array.reduce (<anonymous>)
    at v.e (bundle.js:1:1238)
    at 37559 (bundle.js:1:101)
    at v (bundle.js:1:516)
    at bundle.js:1:49784
    at bundle.js:1:49858
6621.bundle.js:1 
        
        
       Failed to load resource: the server responded with a status of 404 (Not Found)
tree:1 Refused to execute script from 'http://jupyter:8888/static/notebook/6621.bundle.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```

daskol:
It's ArchLinux. `jupyter` hostname is an alias for 127.0.0.1.

```
$ jupyter --version
Selected Jupyter core packages...
IPython          : 8.20.0
ipykernel        : 6.29.0
ipywidgets       : 8.1.1
jupyter_client   : 8.6.0
jupyter_core     : 5.7.1
jupyter_server   : 2.12.5
jupyterlab       : 4.0.11
nbclient         : 0.9.0
nbconvert        : 7.14.2
nbformat         : 5.9.2
notebook         : 7.0.7
qtconsole        : not installed
traitlets        : 5.14.1
```

jommgoncalves:
Same issue here. Same versions.

justinfinkle:
I have a similar problem. The page won't load on chrome (it does on firefox and safari) and the server warns`404 GET /static/notebook/6774.bundle.js`. For what it's worth forcing a refresh of the jupyter notebook page on chrome resolved the issue, but I only was able to try this once.

dargor:
[...]

PR #7354, opened 2024-05-03
Add webpack config for building prod

### LINKED ISSUE #7373: Nothing shown in the browser Safari when notebook is upgrade to 7.2.0
I have upgraded the notebook and Jupyter Lab to the latest versions this morning. But when the browser pops up, only blank is seen.

jtpio:
Follow the JupyterLab config: https://github.com/jupyterlab/jupyterlab/blob/main/dev_mode/webpack.prod.config.js

To check:

- https://github.com/jupyterlab/jupyterlab/pull/9776

Fixes #7373 
Fixes https://github.com/jupyter/notebook/issues/7221

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "webpack",
-    "build:prod": "webpack --mode=production",
+    "build:prod": "webpack --config ./webpack.prod.config.js",
     "clean": "rimraf build && jlpm run clean:static",
     "clean:static": "rimraf -g \"../notebook/static/!(favicons)\"",
     "watch": "webpack --config ./webpack.config.watch.js"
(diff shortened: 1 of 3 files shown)


jtpio:
Postponing to `7.3.0`, as this likely involves using the `html-webpack-plugin` to properly inject the script tag with the cache busting hash string. We might also want to take the opportunity to restructure the build so each page has its own app and bundle, similar to the approach taken in JupyterLite:

- https://github.com/jupyterlite/jupyterlite/tree/main/app/tree
- https://github.com/jupyterlite/jupyterlite/tree/main/app/consoles
- https://github.com/jupyterlite/jupyterlite/tree/main/app/notebooks

Instead of the current approach which currently builds a single bundle, and activate a specific list of plugins on each page.

jtpio:
Closing in favor of https://github.com/jupyter/notebook/pull/7513

Issue #7311, opened 2024-04-02
Flickering notebook toolbar items when resizing the window


jtpio:
Resizing the browser changes the position of the notebook toolbar items:

[resize-notebook-toolbar.webm](https://github.com/jupyter/notebook/assets/591645/8d868454-a8bc-4e76-aa8d-5b7eca5c704b)

However this does not seem to be an issue with JupyterLab:

[resize-jupyterlab-notebook.webm](https://github.com/jupyter/notebook/assets/591645/6ee4b475-88b4-47a3-a838-8ddce8e0c921)

brichet:
Thanks @jtpio for the ping.
I can't reproduce it locally, but can you try resizing using a sidebar instead of the development tools of the browser ?
I'm asking that to know if this can be due to the computation of the items widths.

> Computing the size of the toolbar items is quite long because it requires them to be displayed.
> To avoid this computation, this widths are stored on first computation, and reused to set the position of the items https://github.com/jupyterlab/jupyterlab/blob/634cc3a8ccedf5ea79987f5bf9faa0bede83935f/packages/ui-components/src/components/toolbar.tsx#L701
> The idea is to compute the widths only when the zoom change, but the only way I found to know if the zoom has changed is by using the ratio between the `window.outerWidth` and  `window.innerWidth` https://github.com/jupyterlab/jupyterlab/blob/634cc3a8ccedf5ea79987f5bf9faa0bede83935f/packages/ui-components/src/components/toolbar.tsx#L517. 
> But this ratio also changes when using the dev tools.

jtpio:
> I can't reproduce it locally

It was found while testing https://github.com/jupyter/notebook/pull/7301 on Binder with Firefox 124.

brichet:
It seems that only the widgets on the right of the toolbar are affected. 
Wondering if they are supposed to be there, and if this is not related to the issue.

jtpio:
I think these items come from this extension: https://github.com/manics/jupyter-offlinenotebook

Last time I checked they were added via the toolbar API directly, and not using the settings system (which could be the reason?)

jtpio:
https://github.com/jupyter/notebook/issues/7128  might be related.

brichet:
Yes, it might related, if the flickering is due to some kind of floating position of items.
I can reproduce the wrong positions with the first release of Notebook 7, `7.0.0`, but it didn't flicker.

PR #7361, opened 2024-05-13
Update `@jupyterlab/galata`

jtpio:
Update `@jupyterlab/galata` to match the latest JupyterLab 4.2.0 release.

Might need some tricks from https://github.com/jupyterlab/jupyterlab/pull/15621

Also make some changes to try to reduce the UI tests flakyness.


jtpio:

.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
     rev: 'v2.2.6'
     hooks:
       - id: codespell
-        args: ['-L', 'hart,noteable']
+        args: ['-L', 'hart,noteable', '--skip', "*.spec.ts"]
```
jtpio, line 32: Looks like `afterAll` makes the pre-commit check fail: https://results.pre-commit.ci/run/github/33653601/1715676598.F47qygP7RKuJagZupcWLmA

image

Adding `afterAll` to the list of ignored words (via `-L`) does not seem to help, but skipping all `.spec.ts` files does. Might be worth investigating more in a separate issue or PR.

jtpio:
For reference it looks like we still have some flakiness with the notebook toolbar sometimes missing the overflow button menu:

image

cc @brichet who looked into the notebook toolbar recently (as part of https://github.com/jupyter/notebook/issues/7311) and may know if that might also be happening in JupyterLab.

jtpio:
> For reference it looks like we still have some flakiness with the notebook toolbar sometimes missing the overflow button menu:

We can track that in a separate issue.

PR #7363, opened 2024-05-13
Fix emoji in create issue screen

JasonWeill:
Fix incorrect encoding for 🤔

When filing a bug, 🤔 appears as `\U0001F914`.

image


jtpio (approved):
Thanks!

Issue #7370, opened 2024-05-14
Flaky UI tests


jtpio:
It looks like we still have some flakiness with the notebook toolbar sometimes missing the overflow button menu:

![image](https://github.com/jupyter/notebook/assets/591645/7c5c635f-622c-45e2-b0ed-6d08dcc60172)

cc @brichet who looked into the notebook toolbar recently (as part of https://github.com/jupyter/notebook/issues/7311) and may know if that might also be happening in JupyterLab.

_Originally posted by @jtpio in https://github.com/jupyter/notebook/issues/7361#issuecomment-2110267357_

brichet:
Thanks @jtpio for the ping.

Wondering if this is only a matter of time for the button to be visible, and if we want to fix it (if it's inconvenient to use) or only fix the test (waiting for the button to be visible).
Did you notice the missing button outside the tests?

jtpio:
> Did you notice the missing button outside the tests?

Usually not, so it might indeed just be an issue with the tests.

There is some logic to wait for the kernel status animations to be gone before taking the screenshot, which may not be enough: 

https://github.com/jupyter/notebook/blob/08fe5c5df12182178280bad5d2fbae02b3486146/ui-tests/test/mobile.spec.ts#L58-L59

But maybe there could indeed also be a check to wait for that button to be visible, just to be sure.

brichet:
I can take a look at the test and we'll see if it's enough.

brichet:
Probably related to https://github.com/jupyter/notebook/pull/6731, see https://github.com/jupyter/notebook/pull/7371 for details.

jtpio:
Re-opening to keep track of updating to a new JupyterLab version, and check if it fixes the flakiness.

jtpio:
From https://github.com/jupyter/notebook/pull/7378 (which should include https://github.com/jupyterlab/jupyterlab/pull/16335), it looks like this time it is the button to toggle the virtual scrollbar that is sometimes showing up:

| Actual | Expected |
| :---: | :---: |
| ![image](https://github.com/jupyter/notebook/assets/591645/15e2bd95-cae6-4811-a96d-433cb7e483a2) | ![image](https://github.com/jupyter/notebook/assets/591645/44ca4782-0d03-4b1f-825f-4aa837c86ba4) |

brichet:
Opened https://github.com/jupyterlab/jupyterlab/pull/16409, it fixes this issue locally, during tests and usage.
[...]

PR #6731, opened 2023-02-14
Hide the debugger button on mobile

jtpio:
Fixes https://github.com/jupyter/notebook/issues/6556

image

PR #7371, opened 2024-05-15
Fix flaky test on mobile

brichet:
Wait for the opener button to be visible in notebook toolbar.

Fixes https://github.com/jupyter/notebook/issues/7370

### DIFF
--- a/ui-tests/test/mobile.spec.ts
+++ b/ui-tests/test/mobile.spec.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { IJupyterLabPage, expect, galata } from '@jupyterlab/galata';
+import { expect, galata } from '@jupyterlab/galata';
 
 import { test } from './fixtures';
 
@@ -68,6 +68,9 @@ test.describe('Mobile', () => {
       await hideAddCellButton(notebook);
     }
 
+    await expect(
+      notebook.locator('.jp-Toolbar-responsive-opener')
+    ).toBeVisible();
     expect(await notebook.screenshot()).toMatchSnapshot('notebook.png');
     await notebook.close();
   });


jtpio:
Thanks @brichet for looking into this :+1: 

According to the Playwright test report, the toolbar button does not seem to be visible:

[responsive-opener.webm](https://github.com/jupyter/notebook/assets/591645/7e0a48b4-40c5-4b31-93f2-fcd11bcab016)

brichet:
Yes, I saw that.
But I'm not able to reproduce it locally, the button always shows up when I test it...

brichet:
> But I'm not able to reproduce it locally, the button always shows up when I test it...

I finally managed to reproduce it locally, it seems there is again an issue with the toolbar, in this very specific case...

jtpio:
Is it also reproducible in JupyterLab?

brichet:
Not that I know, but probably, it seems to be a computation issue related to the width.
I managed to reproduce it with a window width between 521px and 526px on first load.
Probably some other width ranges can be affected too, and it depends on the items displayed in the toolbar. For example, it works well with the same range if the `interface-switcher` is disabled.

brichet:
After some more tests, it may be related to the debugger icon.
When the `popup-opener` doesn't show up, the debugger icon take its place in the DOM, with a width of 0px.

jtpio:
Ah so maybe it's related to https://github.com/jupyter/notebook/pull/6731.

The motivation behind that change was to free some space in the notebook toolbar on mobile, also since the debugger is unlikely to be used on a small screen (although it can still be opened via the menu if needed).

brichet:
> Ah so maybe it's related to #6731.

Thanks for the link, I was confused about the behavior of this button, which only appeared when the window was wide... :smile:

Indeed, I can't reproduce this issue without the rule added in https://github.com/jupyter/notebook/pull/6731.
Let's close this PR and find another way to hide the button.

PR #7374, opened 2024-05-16
Duplicate notebook menu option

### LINKED ISSUE #7368: Add a "duplicate notebook" option that keeps the old and new workbooks open
> The option to duplicate the current notebook doesn't exist any more. I'm guessing "Save as" is the replacement. However, after using this, whilst the newly named notebook pops up in a new tab correctly, the old tab no longer works -- the notebook disappears and the error "kernel unknown" appears. The desired behavior is to have both the new and old notebooks running in separate kernels in separate windows/tabs.

Originally reported by @jph00 in https://github.com/jupyter/notebook/issues/7041#issuecomment-1726991197.

> In JupyterLab there is "Duplicate Notebook" option which was recently exposed in command palette and in the tab context menu [jupyterlab/jupyterlab#12870](https://github.com/jupyterlab/jupyterlab/pull/12870). It appears to be an unintentional omission that it was left out in Notebook, it would be a good first issue to restore it by adding `docmanager:duplicate` to menus and command palette where needed (which might be as simple as just changing a few JSON files, see [jupyterlab/jupyterlab#12870](https://github.com/jupyterlab/jupyterlab/pull/12870))
> 
> There is also a button in the tree view:
> 
> image

Reply posted by @krassowski in https://github.com/jupyter/notebook/issues/7041#issuecomment-1727666531.

JasonWeill:
Fixes #7368. In the "File" menu in Jupyter Notebook, adds a new "Duplicate" command that both duplicates the current file and opens the duplicate file in a new window/tab.

duplicate-command


jtpio:
Nice, thanks!

jtpio:

JasonWeill:
@jtpio If this is OK to merge, can you please merge it? I don't have permission to do this in the `jupyter` org.

jtpio:
Since this adds a new user facing string in the menu, I've set this PR to target the `7.3.0` milestone and created the [`7.2.x` branch](https://github.com/jupyter/notebook/tree/7.2.x), so we can move forward with this and other 7.3 related changes.

jtpio (approved):
Thanks!

PR #7387, opened 2024-05-30
Support custom page_data_hook function

bluestealth:
Adds support for `page_config_hook` which already exists in `LabHandler` https://github.com/jupyterlab/jupyterlab_server/pull/220


jtpio (approved):
Thanks!

PR #7392, opened 2024-06-07
Remove pseudoelement obstructing the cell collapser

### LINKED ISSUE #6987: JN 7 → Code Folding starts at the first line regardless the type
Hi, 

Jupyter Notebook 7:

I would like to propose to be able to fold any cell by the first line of text regardless its type.

Reasoning of it is that in many scenarios we have a cell with certain description and funcionality in some type of representation, flow. 

Once user of a Notebook reads instructions, it is not necessary to keep the whole cell unfolded, actually it is very beneficial to allow to fold the whole cell keeping only the first row (likely a Title of a cell).

E.g. cell worthy to read and then fold it to see only a title "The Generator X"
Cell:
```
# The Generator X
# This function produces...
A = FunctionX()

# Callable Output
#...
```

Best,
Michal

krassowski:
It was meant to add some shadows but currently it does not

Fixes #6987

### DIFF
--- a/packages/notebook-extension/style/base.css
+++ b/packages/notebook-extension/style/base.css
@@ -151,19 +151,6 @@ body[data-notebook='notebooks']
   box-shadow: 0px 0px 12px 1px var(--jp-shadow-umbra-color);
 }
 
-body[data-notebook='notebooks']
-  .jp-Notebook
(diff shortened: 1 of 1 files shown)


jtpio:
@meeseeksdev please backport to 7.2.x

PR #7423, opened 2024-07-11
Update to JupyterLab `4.3.0rc0`

jtpio:
Update to the latest pre-releases:

- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.0a2
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.0b0
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.0b1
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.0b2
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.0b3
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.0rc0

Part of https://github.com/jupyter/notebook/issues/7441


jtpio:

jtpio:
Looks like the following error might be relevant:

image


Also need to check the effects of https://github.com/jupyterlab/jupyterlab/pull/16446 on Notebook.

jtpio:
> Looks like the following error might be relevant:
> 
> image

Just noticed an upstream change that might be relevant: https://github.com/jupyterlab/jupyterlab/pull/16523

cc @krassowski who may know more about this. Wondering if this is because we had to fetch the docmanager settings manually at some point due to some race conditions:

https://github.com/jupyter/notebook/blob/43b8cceb692ebb182854f790b8a1dee5293329ad/packages/application-extension/src/index.ts#L223-L246

krassowski:
Nothing jumps at me here, but some ideas:
- does adding  notebook needs to wait on `await settingRegistry.ready // @ts-ignore` help?
- does it make sense to wrap most of the code in `settings.changed(() => { /* register if not registered */ })`?

jtpio:
The error seems to be coming from here in JupyterLab:

https://github.com/jupyterlab/jupyterlab/blob/7bae8e3238967a0a776028150b49a4903759bc90/packages/docmanager-extension/src/index.tsx#L337

image

krassowski:
Also seeing this on https://github.com/jupyter/notebook/pull/7447. I do not think this is 4.3.x specific issue

krassowski:
I see the same error on released Jupyter Notebook 7.2.0 using https://gist.github.com/jtpio/befb8ebf2b630b4748e2538f79877022

image

I now think this should not be a blocker for merging this one as there is no regression in JupyterLab 4.3 nor 4.2.x as this was just overlooked previously.

jtpio:
Thanks @krassowski for looking into this :+1: 

I guess we could just ignore this for now then.

Spinning back a dev install for this branch, it looks like there is still an issue with the background color though:

image

image

krassowski:
> an issue with the background color

Does `body` or other container need to get `.jp-ThemedContainer` (ref: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#css-styling / https://github.com/jupyterlab/jupyterlab/pull/16519)?

jtpio:
Done in https://github.com/jupyter/notebook/pull/7423/commits/40387bbef7387b5a79103b6edeae0f5e59b9a4e9.

Although not seeing any change for now:

image

jtpio:
Is there a way to reduce the size of the "File size" column? I guess it's related to https://github.com/jupyterlab/jupyterlab/pull/16646 ?

image

jtpio:
> Although not seeing any change for now:

This change of background color is also picked up in the UI tests, so it's not a local dev environment issue only:

image

jtpio:
> This change of background color is also picked up in the UI tests, so it's not a local dev environment issue only:

It looks like `.jp-ThemedContainer` was taking precedence over the custom rule defined here in Notebook:

https://github.com/jupyter/notebook/blob/43b8cceb692ebb182854f790b8a1dee5293329ad/packages/application/style/base.css#L16

image

jtpio:
Arf, looks like there are more CSS regressions to fix, for example the menus are now gray and there is no styling on hover:

image

packages/application/style/base.css
```diff
@@ -13,6 +13,13 @@
 body {
   margin: 0;
   padding: 0;
+}
+
+/*
+  Override the default background
+  See https://github.com/jupyterlab/jupyterlab/pull/16519 for more information
+*/
+.jp-ThemedContainer {
```
krassowski, line 22: I think this is because of this change. I think the right fix is to just adjust the selector to `body.jp-ThemedContainer`:

```suggestion
body.jp-ThemedContainer {
  margin: 0;
  padding: 0;
```

krassowski: This will also make embedding notebook in other websites easier (when using a full javascript rather than iframe) as the styles will not conflict :)

jtpio: Yes, tried that before but it didn't seem to be enough:

image

jtpio:
> Is there a way to reduce the size of the "File size" column? I guess it's related to [jupyterlab/jupyterlab#16646](https://github.com/jupyterlab/jupyterlab/pull/16646) ?

I guess it's because Notebook 7 does not provide support for workspaces and restoring the file browser? Maybe we should hardcode some default values that match the previous defaults for now?

packages/application/style/base.css
jtpio: Looks like `#main.jp-ThemedContainer` also had to be defined.

jtpio:
Or maybe this will be made easier with https://github.com/jupyterlab/jupyterlab/pull/16751, so we can just specify custom CSS directly instead.

jtpio:
The file browser looks great now after the `4.3.0b3` update:

image

Thanks @krassowski @JasonWeill for working on this!

jtpio:

jtpio:
Looks like the "File Size" column header is cut even though there is enough space:

[file-size-column.webm](https://github.com/user-attachments/assets/4f2fbf1c-8d4f-4a57-b246-80668b9a3e95)

Maybe we can leave it like this for now to be able to make some progress on the `7.3.0` pre-release. Also maybe https://github.com/jupyterlab/jupyterlab/pull/16749 will help with this.

jtpio:

jtpio:

jtpio:
Arf, it looks like the file browser columns sizes are still a bit unpredictable? Which seems to be introducing flakiness to the UI tests:

[filebrowser-flakiness.webm](https://github.com/user-attachments/assets/f17bd6cd-dbbb-4ddb-87eb-21c9a7031aa2)

jtpio:
@krassowski @JasonWeill is there something we are missing here to get consistent and predictable widths for the different file browser columns?

PR #7446, opened 2024-08-26
Redirect paths from the notebooks route to the tree route if they are directories

### LINKED ISSUE #7445: notebooks/<path> directory urls show error instead of redirecting to tree/<path> urls
### Problem

Paths in notebooks/<path> url routes that are directories and not notebooks will show an error instead of redirecting to tree/<path> urls.

error

This behavior differs from Classic Notebook (notebook<7) where directory path links under notebooks/<path> simply redirect to a tree/<path> link and display the directory nicely.

### Proposed Solution

Paths in notebooks/<path> urls that are directories should redirect nicely to tree/<path> urls. Do not worry about handling cases like other file types.

andyscho:
Resolves #7445 by checking if `notebooks/<path>` paths are directories before attempting to load them as notebooks.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -3,7 +3,7 @@ import os
 import pytest
 from tornado.httpclient import HTTPClientError
 
-from notebook.app import JupyterNotebookApp, TreeHandler
+from notebook.app import JupyterNotebookApp, NotebookHandler, TreeHandler
 
 
 @pytest.fixture()
@@ -31,6 +31,16 @@ async def test_notebook_handler(notebooks, jp_fetch):
         # Check that the lab template is loaded
         html = r.body.decode()
         assert "Jupyter Notebook" in html
+
+    redirected_url = None
+
+    def redirect(self, url):
+        nonlocal redirected_url
+        redirected_url = url
+
+    NotebookHandler.redirect = redirect
+    await jp_fetch("notebooks", "jlab_test_notebooks")
+    assert redirected_url == "/a%40b/tree/jlab_test_notebooks"
 
 
 async def test_tree_handler(notebooks, notebookapp, jp_fetch):
--- a/notebook/app.py
+++ b/notebook/app.py
@@ -203,8 +203,16 @@ class NotebookHandler(NotebookBaseHandler):
     """A notebook page handler."""
 
     @web.authenticated
-    def get(self, path: str | None = None) -> t.Any:  # noqa: ARG002
-        """Get the notebook page."""
+    async def get(self, path: str = "") -> t.Any:
+        """Get the notebook page. Redirect if it's a directory."""
+        path = path.strip("/")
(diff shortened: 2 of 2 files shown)


jtpio:
Nice, thanks @andyscho for looking into this!

tests/test_app.py
```diff
@@ -32,6 +32,16 @@ async def test_notebook_handler(notebooks, jp_fetch):
...
 
+    redirected_url = None
+
+    def redirect(self, url):
+        nonlocal redirected_url
+        redirected_url = url
+
+    NotebookHandler.redirect = redirect
+    await jp_fetch("notebooks", "jlab_test_notebooks")
+    assert redirected_url == "/a%40b/tree/jlab_test_notebooks"
```
jtpio, line 43: Thanks for adding this test!

In addition to this test, maybe it could be interesting to add a UI test to mimic real user interactions: https://github.com/jupyter/notebook/tree/main/ui-tests

But we can track that in a separate issue :+1:

jtpio (approved):
Thanks!

germandb:
Hi @jtpio and @andyscho, in my company, we use a single-user server with namespaces for each user. Something like 
> "/user/{podName.Value}/{userNamespace}/notebooks/"+ jupyterDocumentPath

But the content manager is personalized to store the notebooks in a remote Azure container registry, so the jupyterDocumentPath, in our case, includes info that we need to access the ACR. For example it could look like this:

> 17/document/7935/FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF/test.ipynb

We just updated to 7.3 and found that our solution stopped working because the URL redirects to the tree view.

This is the log of the pod running Jupyter:
```
[I 2025-03-13 16:41:54.629 ServerApp] 302 GET /user/82c20951-9587-4691-812d-7cd49b1a42f5/orionhubnsuser.0.30/notebooks/17/document/7935/FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF/New%20Document%205177.ipynb?theme=theme-light -> /user/82c20951-9587-4691-812d-7cd49b1a42f5/orionhubnsuser.0.30/tree/17/document/7935/FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF/New%20Document%205177.ipynb (82c20951-9587-4691-812d-7cd49b1a42f5@192.168.65.3) 2.46ms
[I 2025-03-13 16:41:54.656 ServerApp] 200 GET /user/82c20951-9587-4691-812d-7cd49b1a42f5/orionhubnsuser.0.30/tree/17/document/7935/FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF/New%20Document%205177.ipynb (82c20951-9587-4691-812d-7cd49b1a42f5@192.168.65.3) 16.78ms
```
Can you think of any solution for this?

PR #7447, opened 2024-08-26
Upgrade JupyterLab dependencies to v4.2.5

krassowski:
- grabs a number of bug fixes from JupyterLab 4.2.x releases:
  - https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.3
  - https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.4
  - https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.5
- generated using scripts from https://github.com/jupyter/notebook/pull/7281


krassowski:
The CI is all good, but I see some errors in console and problems with `full` windowing mode. I am not sure if these are regressions or not - it would be great to check before releasing.

krassowski:
I think we may want a bit more time to investigate, especially the windowing issue. I am not sure if we want to merge this one as is and iterate or wait but I think we should not release yet.

krassowski:
@RRosio after a bit more testing on my end I do not see a reason to wait with this one - the error discussed in https://github.com/jupyter/notebook/pull/7423 is also occurring in the released 7.2.0 version and the notebooks work well both in windowing mode and without it

PR #7453, opened 2024-09-05
Fix SyntaxWarning from update message

living180:
The use of backslashes in the ASCII-art "Update" banner results in an "invalid escape sequence" SyntaxWarning under Python 3.12.  Fix by changing the banner string to a raw string (which ignores backslashes).

### DIFF
--- a/notebook/notebookapp.py
+++ b/notebook/notebookapp.py
@@ -185,7 +185,7 @@ class NotebookWebApplication(web.Application):
         if settings['autoreload']:
             log.info('Autoreload enabled: the webapp will restart when any Python src file changes.')
 
-        print("""
+        print(r"""
   _   _          _      _
  | | | |_ __  __| |__ _| |_ ___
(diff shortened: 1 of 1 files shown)


JasonWeill:
@living180 Thank you for opening this pull request! Notebook 6 code is now maintained as part of nbclassic, or Classic Notebook. Looks like this style of string is being used in nbclassic as well:

https://github.com/jupyter/nbclassic/blob/abe28bc4fe5d3e026c68c01248047656f58db635/nbclassic/notebookapp.py#L328-L333

Can you please open an issue in nbclassic and open a pull request in that repo instead? Thanks again for your contribution.

living180:
Done - https://github.com/jupyter/nbclassic/pull/297.  Thanks for the input!

jtpio:
Closing as handled in https://github.com/jupyter/nbclassic/pull/297

Thanks all!

PR #7474, opened 2024-10-03
Update README.md

hackit-coder:
Steps to host Jpyter as web interface on Google, Mozilla etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ jupyter notebook
 **Access Jupyter Notebook Web Interface**
 
 1. Open a web browser (e.g., Google Chrome, Mozilla Firefox).
-2. Type: `http://localhost:8888` in the address bar.
+2. Type: `http://localhost:'host number'` in the address baralong with the login token issued.
 3. Press Enter.
 
 ### Running in a remote installation
--- a/README.md
+++ b/README.md
@@ -79,8 +79,10 @@ jupyter notebook
 ```
 **Access Jupyter Notebook Web Interface**
 
+It will open automatically by default but if does not then copy do the following:
+
 1. Open a web browser (e.g., Google Chrome, Mozilla Firefox).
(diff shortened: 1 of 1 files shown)


hackit-coder:
Hello anyone there

krassowski (changes requested):
These instructions are incorrect for a number of reasons:
- the port 8888 may be taken and a different port would be used instead
- usually there is no need to open a browser - the command above will open the default browser
- `localhost` may not work on specific systems and instead using 127.0.0.1 may be needed
- the URL above will not include the token needed to login and this token will be different each time

Each PR should address an issue. This PR is not linked to any issue. In future I would encourage you to:
- open an issue first to discuss potential enhancements
- if you are looking for a way to contribute, looking through the issues and picking up something tagged as good first issue :)

hackit-coder:
Ok changed it a bit

hackit-coder:
@krassowski can you check

README.md
```diff
@@ -77,6 +77,11 @@ Launch with:
 ```bash
 jupyter notebook
 ```
+**Access Jupyter Notebook Web Interface**
+
+1. Open a web browser (e.g., Google Chrome, Mozilla Firefox).
+2. Type: `http://localhost:'host number'` in the address baralong with the login token issued.
+3. Press Enter.
```
krassowski, line 84: `baralong`?

krassowski (commented):
I am still not convinced that this is an improvement over status quo. These instructions can be still misleading and usually none of these steps are necessary as a browser with the auto-generated temporary redirect file (which is also safer than typing in a token) will automatically open.

Maybe describing what would typically happen (that browser would open etc) and that if it does not then user should open and copy-paste URL could be more useful?

hackit-coder:
Updated it in the previous commit, it' bar along not baralong my bad!

jtpio:
Closing as stale.

PR #7528, opened 2024-12-03
Bump Python version used on CI

jtpio:
Update the Python version used on CI to the range 3.9 - 3.13.

Looking into fixing the failing checks noticed in https://github.com/jupyter/notebook/pull/7527.


jtpio:
> Looking into fixing the failing checks noticed in #7527.

Ah it's probably because the `macos-12` images are not supported anymore: https://github.com/actions/runner-images/issues/10721

PR #7530, opened 2024-12-04
Adds jupyter-ui-toolkit packages in shared scope

brichet:
This PR adds the 2 packages from [jupyter-ui-toolkit](https://github.com/jupyterlab-contrib/jupyter-ui-toolkit) in the shared scope of Notebook.

These packages have been included in jupyterlab as singleton in https://github.com/jupyterlab/jupyterlab/commit/3d525a2c24524fe1431c3227e62057c9137940ed.

Without this change, an extension depending on one of these packages would not work.

Thanks @jtpio for your help on this.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -12,8 +12,6 @@
   "resolutions": {
     "@codemirror/state": "~6.4.1",
     "@codemirror/view": "~6.28.3",
-    "@jupyter/react-components": "^0.16.6",
-    "@jupyter/web-components": "^0.16.6",
     "@jupyter-notebook/application": "~7.3.0",
     "@jupyter-notebook/application-extension": "~7.3.0",
     "@jupyter-notebook/console-extension": "~7.3.0",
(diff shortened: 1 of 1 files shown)


app/package.json
```diff
@@ -12,6 +12,8 @@
   "resolutions": {
     "@codemirror/state": "~6.4.1",
     "@codemirror/view": "~6.28.3",
+    "@jupyter/react-components": "^0.16.6",
+    "@jupyter/web-components": "^0.16.6",
```
jtpio, line 16: Maybe we can move the `@jupyter/ydoc` line below next to these new packages, so it's easier to notice (and somewhat sorted) while looking at the resolutions?

jtpio (approved):
Thanks!

Left a minor comment, otherwise looks good :+1:

app/package.json
brichet: Good catch, I moved the new entries instead to follow the ASCII order.

brichet:
Since the corresponding PR in jupyterlab has been published in 4.1, we should probably backport this one to `7.2.x` and `7.1.x` branches.

jtpio:
I guess it would be fine to backport to `7.2.x` since `7.3.0` was released today. Not sure it's worth it for `7.1.0` though, unless someone actively requests it, and since it probably affects just a few extensions?

jtpio:
@meeseeksdev please backport to 7.2.x

brichet:
I can do it

jtpio:
ah meeseek was now able to do the backport, hopefully the manual backport is not too painful since it's just a few lines.

PR #7567, opened 2025-01-23
Fix undefined error when checkpoints is empty

jdavid:
```python
      const checkpoints = await context?.listCheckpoints();       <== if checkpoints is an empty array
      if (!checkpoints) {                                         <== this will be false
        return;
      }
      const checkpoint = checkpoints[checkpoints.length - 1];     <== then checkpoint will be undefined
      node.textContent = trans.__(
        'Last Checkpoint: %1',
        Time.formatHuman(new Date(checkpoint.last_modified))      <== here will trigger an undefined error
      );
```

We found the problem while integrating jupyterlite with a FastAPI project.

### DIFF
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -124,7 +124,7 @@ const checkpoints: JupyterFrontEndPlugin<void> = {
       context?.fileChanged.connect(onChange);
 
       const checkpoints = await context?.listCheckpoints();
-      if (!checkpoints) {
+      if (!checkpoints || !checkpoints.length) {
         return;
       }
       const checkpoint = checkpoints[checkpoints.length - 1];


jtpio (approved):
Thanks!

jtpio:
@meeseeksdev please backport to 4.3.x

jtpio:
> We found the problem while integrating jupyterlite with a FastAPI project.

Do you need this change in JupyterLite soon? It may take a little while to propagate there as this would first need to be available in a Notebook release.

Normally this issue should not have too much of an impact, except for some errors in the dev tools console?

jdavid:
In our case the error breaks the notebook, but I don't need the fix quickly, we already patch the file.
Thanks for merging!

jtpio:
FYI @jdavid this is now available in `7.3.3`: https://github.com/jupyter/notebook/releases/tag/v7.3.3

jdavid:
Thanks!

Issue #7565, opened 2025-01-17
Display "Click to add a cell" button as a part of notebook, not background


andrii-i:
### Problem

Currently "Click to add a cell" is rendered as a part of background / outer notebook area (grey in the JupyterLab light theme) under the drop shadow from the notebook  / inner notebook area (white in the JupyterLab light theme). 

To me it seems that the way how button is rendered in JupyterLab, as a part of notebook / inner notebook area (white in the JupyterLab light theme) is easier to read / parse. As an added bonus, this would make design language used in lab and notebook more consistent. 

Notebook:
<img width="1192" alt="Image" src="https://github.com/user-attachments/assets/c18dcac1-5442-4143-9f6f-b8ebc90de6cb" />

JupyterLab:
<img width="1046" alt="Image" src="https://github.com/user-attachments/assets/9e364fb0-2cf3-4981-adcc-c96d5b0ee33e" />

### Proposed Solution

Render "Click to add a cell" button as a part of notebook / inner notebook area (white in the JupyterLab light theme).

jtpio:
Thanks @andrii-i.

Isn't rendering the button as part of the notebook going to add some artificial white space at the bottom of the notebook, to accommodate for the button when hovering?

andrii-i:
@jtpio thank you for looking into this. It would, I was thinking that white space at the bottom of inner notebook area could be made larger to accommodate the button when hovering in a way that there would be no need to add / "pop-in" any artificial white space.

PR #7587, opened 2025-02-18
Fix for issue #7565 on Click to add a cell button not in background

aj202088:


rsaditya01:
Welcome @aj202088,
Thanks for looking into the issue and contributing to Jupyter Notebook.

On pulling your changes in local and testing, the updated notebook UI looks different from the desired output.

(A) Notebook UI after change -
image

(B)Desired Notebook UI (mentioned in the issue)-
image

The notebook UI in (A), keeps the "Click to add cell" button in the footer area, the button appear all the time below the notebook area and the button spans to entire width of 'inner notebook area (white area)'.
While the desired output is that the "Click to add cell button" can be placed inside 'inner notebook area'(as mentioned in issue description #7565), width should be equivalent to Notebook Cell width and it should appear only "on hover" over that button (current functionality).
Could we look into making some changes in "inner notebook area" and placing the the button there with minor alteration in css.

PR #7602, opened 2025-03-10
Add devcontainer support, backed by pixi

jtpio:
Looking into reducing friction for contributing to the repo, so folks can use GitHub Codespaces if they would like to.

Also add basic support for pixi, although not officially documented in the contributing guide yet.


jtpio:
Getting this one in to be able to use for small tasks. Will iterate on the setup if more things need to be fixed or improved.

PR #7608, opened 2025-03-13
Add pixi step to the update workflow

jtpio:
To keep the `pixi.lock` file in sync too.


jtpio:
Used this branch for https://github.com/jupyter/notebook/pull/7614, which seemed to be behaving as expected ✅

PR #7618, opened 2025-03-25
Update to JupyterLab v4.4.0rc0

jtpio:
New JupyterLab release [v4.4.0rc0](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.4.0rc0) is available. Please review the lock file carefully.


jtpio:
Need to check why the kernels section is at the bottom now:

image

jtpio:
> Need to check why the kernels section is at the bottom now:

This may have just been a one-off thing as this seems to be fine now:

image


But maybe some read issue or race condition to investigate at some point.

PR #7624, opened 2025-04-04
Default `singleClickNavigation` back to `false`

jtpio:
Fixes https://github.com/jupyter/notebook/issues/7538

See https://github.com/jupyter/notebook/issues/7538 for more context.

- [x] Update default value for `singleClickNavigation`
- [x] Mention in the user facing changelog

PR #7627, opened 2025-04-09
Patch handling of singleClickNavigation

### LINKED ISSUE #7538: 'navigate files and folders with single click'  only effective in jupyter lab
## Description
upgrade to 7.3.1，'navigate files and folders with single click' not checked, but it's still affecting the behavior.

This option only effective in jupyter lab.
 
Image

jtpio:
Follow-up to https://github.com/jupyter/notebook/pull/7624

Fixes https://github.com/jupyter/notebook/issues/7538

### DIFF
--- a/.yarn/patches/@jupyterlab-filebrowser-npm-4.4.0-0616cbcbf2.patch
+++ b/.yarn/patches/@jupyterlab-filebrowser-npm-4.4.0-0616cbcbf2.patch
@@ -1,0 +1,24 @@
+diff --git a/lib/listing.js b/lib/listing.js
+index 07b699656bf1780aada21fb50c3c7748086e20e5..42c34c27f8c88993210ef1ce422a05967c2e5aeb 100644
+--- a/lib/listing.js
++++ b/lib/listing.js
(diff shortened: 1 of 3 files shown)


jtpio:
@afshin this seems to be fixing the issue, when using Firefox (138.0b5):

**On main / 7.4.0**

https://github.com/user-attachments/assets/ace4f1fb-b54a-4499-a935-7f38290085a1

**With this PR**

https://github.com/user-attachments/assets/56cefc65-a17c-4582-82a5-fe158b7323b0

RRosio:
I tried these changes locally and they resolve the issue for me as well.

jtpio:
Nice, thanks @RRosio for checking!

Do you want to submit the fix upstream in JupyterLab? Unless @afshin was planning to do it?

afshin:
I wasn't planning to, so if someone else is happy to, from my point of view testing is the actual work here, I was just guessing 🤓

jtpio:
ok I opened https://github.com/jupyterlab/jupyterlab/pull/17475 then, so we can have the fix available in JupyterLab 4.4.1 and update Notebook with the new version.

PR #7628, opened 2025-04-11
fix: Update lower bound of required python to 3.9

honnix:
Python 3.8 is no longer supported by notebook, so this lower bound needs to be changed.

```shell
$ pip3.8 index versions notebook
WARNING: pip index is currently an experimental command. It may be removed/changed in a future release without prior warning.
notebook (7.4.0)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "notebook"
 description = "Jupyter Notebook - A web-based notebook environment for interactive computing"
 readme = "README.md"
 license = { file = "LICENSE" }
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 authors = [
     { name = "Jupyter Development Team", email = "jupyter@googlegroups.com" },
 ]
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @honnix for looking into this.

It's probably because of `ruff` which now uses 3.9 as the base version. Would you be able to try running the `ruff` linter to check if it's able to auto-fix these?

honnix:
> Would you be able to try running the `ruff` linter to check if it's able to auto-fix these?

Hi @jtpio . Thank you for the quick reply. I changed ruff config to ignore `UP006` in this PR. I think it definitely makes sense to fix everything (hopefully ruff fix do that automatically), but I'm not sure it should be in this PR. I guess having a dedicated PR to fix that might be better, assuming it's gonna be a big change.

honnix:
This is now good to go. I can follow this up fixing stuff using ruff.

jtpio (approved):
Thanks!

Makes sense to fix some of the `ruff` rules in a follow-up PRs. Maybe some of them are not needed anymore?

honnix:
> Maybe some of them are not needed anymore?

That I don't know. I'm not familiar with the code base at all.

PR #7633, opened 2025-04-14
chore: Fix UP006

honnix:
This is a follow-up PR for #7628 where UP006 was ignored temporarily.

It turns out there aren't many UP006 violations (I enabled `--unsafe-fixes` (https://docs.astral.sh/ruff/rules/non-pep585-annotation/) for `ruff` and ran `pre-commit run --all-files ruff`, which should apply to the whole code base IIUC).

@jtpio PTAL. Thanks.

### DIFF
--- a/notebook/app.py
+++ b/notebook/app.py
@@ -39,7 +39,7 @@ from ._version import __version__
 
 HERE = Path(__file__).parent.resolve()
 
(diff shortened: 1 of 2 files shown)


jtpio (approved):
Thanks!

PR #7698, opened 2025-08-11
Fix open file with “#” in the filename

### LINKED ISSUE #7659: Cannot open file with “#” in the filename in version 7.4.1
> After I upgrade my jupyter notebook to version 7.4.1, in traditional notebook mode, cannot open file with “#” in the filename, and there is a “file not found” error. It seems that the full filename isn’t passed when there is “#” in the filename.
>
> But the same file can be opened in the jupyterlab mode, and in some very old version, there is no such problem.

Originally reported by winoenix in https://discourse.jupyter.org/t/cannot-open-file-with-in-the-filename-in-version-7-4-1/35426/2, reproduced by another Jupyter contributor as per the linked thread

itsmevichu:
Closes #7659


jtpio (approved):
Thanks!

PR #7713, opened 2025-08-22
Open the help in the down area (pager)

### LINKED ISSUE #6692: Open the help in a bottom panel
The classic notebook (v6) opens the help panel at the bottom of the page, for example when using the following code snippet:

```py
import math
math.pi?
```

image

JupyterLab inlines the help in the output:

image

Jupyter Notebook 7 does the same as JupyterLab since it uses the same notebook component:

image

Opening this issue to track this feature parity difference. Not sure this needs to be addressed for Notebook 7 but at least it's tracked somewhere.

jtpio:
Fixes https://github.com/jupyter/notebook/issues/6692

To offer a behavior similar to the classic notebook:

https://github.com/user-attachments/assets/0e7a2507-69c6-4cf8-a566-684ed40cca52

The idea would be to have this behavior by default in Notebook 7.7, with a setting to disable it and have the pager data be displayed in the cell output like in JupyterLab:

Documentation: https://jupyter-notebook--7713.org.readthedocs.build/en/7713/pager.html

- [x] Open the pager in the down area
- [x] Setting to disable this behavior and default to the JupyterLab behavior
- [x] If disabled, allow opening the "Contextual Help" and click around like in JupyterLab
- [x] UI test
- [x] Update snapshots
- [x] Docs


jtpio:
So, I think this PR should be ready for a first round of reviews.

As a quick summary, it brings back the behavior of the pager just like the classic notebook. This has been missing since the early Jupyter Notebook 7 releases (released a bit more than two years ago), and there's been some feedback that this was still missing.

I've also added some UI tests to cover some cases, and also added a new page in the docs so that folks can opt out of this new behavior and return to the JupyterLab behavior if they want to.

If folks would like to try it, that would be great, let me know what you think. I'm also happy to iterate on a different approach if you think the current one could be improved.

andrii-i (commented):
Thank you for this contribution @jtpio, it's a nice addition to the Notebook UI, especially for users used to Nbclassic. Please find some comments and suggestions below.

packages/notebook-extension/src/index.ts
```diff
@@ -749,6 +770,245 @@ const editNotebookMetadata: JupyterFrontEndPlugin<void> = {
...
+
+            if (pagePayload && pagePayload.data) {
+              // Remove the 'page' payload from the message to prevent it from also appearing in the cell's output area
+              content.payload = content.payload.filter(
+                (item) => item.source !== 'page'
+              );
+              if (content.payload.length === 0) {
+                // If no other payloads remain, delete the payload array from the content.
+                delete content.payload;
+              }
```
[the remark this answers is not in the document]
jtpio: > For example, @jupyterlab/outputarea [copies](https://github.com/jupyterlab/jupyterlab/blob/11a9a7c46cffa2b63fd53a12ce0d2bbf1286590d/packages/outputarea/src/widget.ts#L775) the message.

I looked a bit more into this. While we could set up handlers to check the output area model, information about the message gets lost when it's added to the output:

```ts
const output: nbformat.IOutput = {
  output_type: 'display_data',
  data: (page as any).data as nbformat.IMimeBundle,
  metadata: {}
};
model.add(output);
```

And there doesn't seem to be any easy way to know if an output is the result of pager information.

image

jtpio:
Thanks @andrii-i for the review 👍 

Putting this back as a draft to revisit the handling of the kernel message. I will get back to it later, but if someone would like to pick it up, please feel free!

jtpio:
The implementation should now be much cleaner than the previous attempt, thanks to https://github.com/jupyterlab/jupyterlab/pull/18179.

cc @MUFFANUJ @brichet if you would like to take a look

packages/application/src/shell.ts
```diff
@@ -305,7 +312,15 @@ export class NotebookShell extends Widget implements JupyterFrontEnd.IShell {
...
-          this._downPanel.show();
+          const tabIndex = this._downPanel.tabBar.titles.findIndex(
+            (title) => title.owner.id === id
+          );
+          if (tabIndex >= 0) {
+            this._downPanel.currentIndex = tabIndex;
+          }
+          if (this._downPanel.isHidden) {
+            this._showDownPanel();
+          }
```
brichet: Should we show the down panel only if  `tabIndex >= 0` ?

packages/application-extension/src/settingconnector.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+/**
+ * A data connector for fetching settings.
+ *
+ * #### Notes
+ * This connector is based on the default JupyterLab setting connector,
+ * and additionally overrides the default values of some settings.
+ */
+export class SettingConnector
```
brichet, line 47: This is really nice to override default settings.

I wonder if we could extends the [SettingConnector](https://github.com/jupyterlab/jupyterlab/blob/45c8e5b774e8648a1ba72c5897d17674acd2a9bb/packages/apputils-extension/src/settingconnector.ts#L21) from Jupyterlab directly, to avoid duplicating code.

That would simplify the `list` method for example:
```ts
  const { ids, values} = await super.list(query);
  return { ids, values: values?.map(Private.overrideDefaults) };
```

brichet (commented):
Thanks @jtpio, it looks good overall.

It looks like the helper is not scrollable. But it seems to be the case in Jupyterlab too, we should probably fix it upstream.

packages/application-extension/src/settingconnector.ts
jtpio: It would be great, although currently this would require depending on the `@jupyterlab/apputils-extension` package (an extension package) while the pattern is to usually import from the base packages. And also importing it from `lib`.

Maybe there are some improvements we could do upstream to improve that, like exporting `SettingConnector` from `@jupyterlab/apputils`?

jtpio: Opened https://github.com/jupyterlab/jupyterlab/pull/19132

jtpio: Maybe we could still move forward with this PR and do the cleanup in the next iteration when https://github.com/jupyterlab/jupyterlab/pull/19132 is available in a release?

brichet (approved):
Thanks, it looks good to me.

jtpio:
Opened https://github.com/jupyter/notebook/issues/7981 to keep track of the future cleanup.

jtpio:
Thanks @brichet for the review!

PR #7727, opened 2025-09-20
Add JupyterCon banner and add Jupyter colors

choldgraf:
This PR does a couple of things:

- Adds an announcement banner for JupyterCon
- Links a CSS file from the Jupyter documentation that we can use for jupyter colors in our docs


jtpio (commented):
Thanks!

Looks like it needs a quick lint pass, otherwise sounds like a good idea 👍

choldgraf:
OK I think I figured it out and it should work now, let's see if tests pass

Zsailer (approved):

Zsailer:
Thanks!

Issue #3170, opened 2017-12-29
Feature request: open a notebook without running kernel


ibobak:
We are working with notebooks which use PySpark kernel.    Very often we need just to look inside the notebook, but we don't need to run it.  But if I click on the notebook, it starts a kernel as well, but PySpark kernel is very heavy and it consumes resources:   after waiting for several minutes until the notebook starts, I am doing "shutdown kernel".

It would be very fine if there was a button in the notebook list "View without running kernel".

gppeixoto:
+1 for this feature, looks like it hasn't gotten much love for almost a year. Any updates on this?

slidenerd:
I triple this!

grisaitis:
Use cases
- browsing notebooks with no intent of running them
- editing non-kernel-related content, like markdown
- running notebooks with different kernels (e.g. kernel versions (e.g. py 3.7 vs 3.6), package versions)
- running multi-kernel notebooks

UX ideas:
- a "read only" option when right clicking on a notebook on the main page (file browser view)
- a "open without kernel" option

I'm surprised this isn't already possible. Maybe a plug-in implements it?

rocconnick:
+1

Zaky7:
+1

thincal:
+1, editing / viewing without a runtime just like the Google Colab.

nareshsankapelly:
+1, this will be a useful feature.

amangarg96:
+1
Would help a lot, specially when using PySpark kernels as it takes some time to launch

yucai:
+1, surprise it is not ready yet. Sometimes, just want to view the content.

JavrelWork:
+ 1 would be usefull

prgx-erodri02:
Agree with @grisaitis 
There are so many use cases for opening a notebook without the intent to run any cell. It would be great to have an "Open with no Kernel" option in the context menu

JavrelWork:
How about always open a notebook without a kernel until you try to run it, then fire up the kernel?
This could be optional setting for opening behaviour

thavidu:
+1 would love this

ostrokach:
One option for editing a notebook without starting a kernel is to use [nteract](https://nteract.io/). This does not fully address this issue, but may be helpful to some people.

LukaPitamic:
+ 1 

my case:
workspace with 3 notebooks, where only one is really run, two others are imported in the first one as code

ibobak:
[...]

PR #7728, opened 2025-09-22
Update to JupyterLab v4.5.0a4

jtpio:
New JupyterLab release [v4.5.0a4](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.5.0a4) is available. Please review the lock file carefully.


jtpio:
Postponing the integration of the new plugin for opening a notebook without starting a kernel, as there seems to be some issues currently (would need to investigate more).

We can track that in https://github.com/jupyter/notebook/issues/3170.

jtpio:
Moving forward so we can cut a new 7.5 pre-release.

PR #7747, opened 2025-10-17
Add the layout restorer

### LINKED ISSUE #6990: JN 7.0. → Remembering layout of View panels upon saving and reopening a notebook
Hello,

I would like to propose that new JN 7.0 notebook keeps info about last state of user's layout - e.g. let's say that notebook's View setup is left panel Table of Contents, right panel is Debugger - and once saved with this layout it will be activated upon opening.

Or at least add it to Settings as an option, pls - whethe to respect saved layout upon opening.

Best,
Michal

brichet:
Adds the layout restorer, to restore the state of the side/down panels.
It includes the current widgets opened in each panel, and the size of the panels.

[record-2025-10-17_20.16.35.webm](https://github.com/user-attachments/assets/0e8c7ed0-b1a7-4db9-9af7-ceef32aa68d0)

Fixes #6990 

### User facing change
- save and restore the layout, only for the Notebook view, since it is the only one with side and bottom panels
- use only one workspace (unlike Jupyterlab), to avoid confusion. This means that changing the layout in one Notebook will affect the layout of other Notebook if they are reloaded. This workspace file is save in the jupyterlab workspace directory (*.jupyter/lab/workspaces*) and is named `nb-default`.

### Code changes
- add the `RestoreLayout` and `SaveLayout` function to the shell, which uses some dedicated function to "dehydrate" and "rehydrate" the shell
- adds 2 plugins:
  - `'@jupyter-notebook/application-extension:layout'`, which handle the shell layout restoration
  - `'@jupyter-notebook/apputils-extension:state'`, which was previously activated from jupyterlab. Providing our own `IStateDB` allow to avoid some side effect, that the tab title update (which does not work correctly in Notebook, see https://github.com/jupyter/notebook/pull/7747#issuecomment-3425653114)


brichet:
**EDIT** this error has been fixed

There is a conflict on the title of the browser tab with this PR.

### BEFORE THE PR

[record-2025-10-21_11.29.54.webm](https://github.com/user-attachments/assets/5a54cc50-d680-47ac-9fc7-c0ca5d818c39)

### AFTER THE PR

[record-2025-10-21_11.30.28.webm](https://github.com/user-attachments/assets/2805fc4d-30cb-49b0-a101-18cbefe2a31c)


The tab title is not updated correctly anymore, probably because of https://github.com/jupyterlab/jupyterlab/blob/21f9c73298d7a450dba9927d404c87aae3011f7a/packages/apputils-extension/src/index.ts#L462-L469.
The code above wasn't called before this PR because there was no window resolver. It is required in this PR to allow saving the layout.

Maybe we should provide the `IStateDB` token in Notebook too.

brichet:
> Maybe we should provide the `IStateDB` token in Notebook too.

Added in https://github.com/jupyter/notebook/pull/7747/commits/c1b056e3930efed62c9852be078c1704be6ffec0

jtpio (commented):
Many thanks @brichet for working on this!

Just did a first pass on the diff without testing locally for now, and overall the changes look good 👍 

Left a couple of small comments before checking it locally.

packages/application/src/shell.ts
```diff
@@ -456,26 +480,158 @@ export class NotebookShell extends Widget implements JupyterFrontEnd.IShell {
...
+            this._downPanel.tabBar.insertTab(position, title);
+          }
+        });
+      }
+
+      if (currentWidget) {
+        const index = this._downPanel.stackedPanel.widgets.findIndex(
+          (widget) => widget.id === currentWidget.id
+        );
+        if (index) {
```
jtpio: Looks like this could evaluate to `false` if `index` is 0?

brichet:
I converted it to draft to explore layout restoration on other views than Notebook, and workspace reset in the UI.

Issue #7745, opened 2025-10-14
Use npm trusted publishers


jtpio:
Sibling issue: https://github.com/jupyterlab/jupyterlab/issues/17993

### Proposed Solution

- [x] On https://www.npmjs.com, for each package, set up the trusted publisher with the following information:
  - Organization or user*: **jupyter**
  - Repository*: **notebook**
  - Workflow filename*: **publish-release.yml**
  - Environment name: **release**
- [x] Wait for https://github.com/jupyterlab/maintainer-tools/pull/259 to be merged and released
  - [x] Use https://github.com/jupyter/notebook/pull/7749 while waiting for https://github.com/jupyter-server/jupyter_releaser/issues/617
- [x] Delete the `NPM_TOKEN` secret from the repo
- [x] Try make a new release: https://github.com/jupyter/notebook/actions/runs/18644636505/job/53149452930
- [x] Enable the strictest security option on npm: "Require two-factor authentication and disallow tokens (recommended)"
- [x] Remove `jupyter-release-bot` from the `@jupyter-notebook` org on npm

jtpio:
https://github.com/jupyter/notebook/releases/tag/v7.5.0b1 was released with the npm trusted publishers:

<img width="1936" height="962" alt="Image" src="https://github.com/user-attachments/assets/11911700-79e3-480b-bde3-8b93f5304120" />

jtpio:
> - Enable the strictest security option on npm: "Require two-factor authentication and disallow tokens (recommended)"
>  - Remove jupyter-release-bot from the @jupyter-notebook org on npm

Those items are now done.

jtpio:
https://github.com/jupyter/notebook/releases/tag/v7.6.0a1 was released after completing all the items listed above.

PR #7749, opened 2025-10-20
Use Node.js 24 / npm 11 to publish releases

jtpio:
Temporary use node 24 here until the npm 11 issue in the jupyter releaser is fixed and released: https://github.com/jupyter-server/jupyter_releaser/issues/617

To be able to try https://github.com/jupyter/notebook/issues/7745


jtpio:
Let's try that for the next pre-release.

PR #7774, opened 2025-11-24
Fix missing debugger display provider in Notebook

arjxn-py:
Notebook was not loading the `@jupyterlab/debugger-extension:display-registry` plugin added in https://github.com/jupyterlab/jupyterlab/pull/18043. This plugin became part of the debugger extension set in JupyterLab 4.5.0, but Notebook’s plugin list did not include it. As a result, debugger sources were missing proper display names.


jtpio:
Updating this branch since https://github.com/jupyter/notebook/pull/7785 updated to JupyterLab 4.5.1.

jtpio (approved):
Looks good, thanks!

Image

Issue #7534, opened 2024-12-05
`activateById` toggle side panel instead of expanding it


brichet:
Please answer the following questions for yourself before submitting an issue
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Description

Using the shell function `activateById` on a side panel widget may result to the opposite behavior, if the widget was already opened.

It may be related to https://github.com/jupyter/notebook/pull/7178.
AFAIU the PR, if the id is provided and the current widget has this ID (it is opened), the `expand` will collapse the panel.

## Reproduce

Hard to reproduce, it needs a extension that provide a sidepanel, and which call the `shell.activateById` on this sidepanel when it opened.
See context to reproduce it.

## Expected behavior

The sidepanel to remain opened.

## Context

Found when testing [jupyterlab_chat](https://github.com/jupyterlab/jupyter-chat).
This [line](https://github.com/jupyterlab/jupyter-chat/blob/967df3054b305275163a5c7d0de47a804b90471c/python/jupyterlab-chat/src/index.ts#L521) is used to ensure that the sidepanel is open, but the sidepanel is collapsed instead.

cc. @haok1402

brichet:
https://github.com/jupyter/notebook/pull/7178 has been include to fix https://github.com/jupyter/notebook/issues/7147.

But these changes have led to some unexpected behavior, as the table of contents is not initially intended to be hidden by the menu, in Jupyterlab.

My opinion is that this PR should be reverted, and the menu action changed upstream in Jupyterlab, to toggle the panel instead of only activate it.

PR #7790, opened 2025-12-17
Add a scratch-pad console to the notebook

### LINKED ISSUE #7707: Scratchpad-like workflow
### Problem

So the idea would be to reuse the JupyterLab code consoles to enable the scratchpad-like workflow. That means opening a temporary console next to the notebook via a keyboard shortcut, ideally, and being able to execute code using the same kernel. 

### Proposed Solution

Currently in Notebook 7, consoles are opened in their own browser tab. So there could be a setting to open code consoles next to a notebook instead, for example in the right area.

### Additional context

For reference, the original classic notebook extension: https://github.com/minrk/nbextension-scratchpad

Image

brichet:
This PR allows to open a console next to a Notebook (scratch-pad like), sharing the same kernel.

image

there are several way to open it from a Notebook view:
- from the menu `menu>new>Scratch-pad console`
- from the commands palette
- from the shortcut `Ctrl + B`

Fixes #7707 

Remaining tasks:
- [x] allow the console commands to run in the side panel (https://github.com/jupyterlab/jupyterlab/pull/18238)
- [x] do not toggle the side panel after the command is executed (related to https://github.com/jupyter/notebook/issues/7534)
- [x] add tests


brichet:
**EDIT**: _this comment is now solved, the PR has been rebased with #7799_

> do not toggle the side panel after the command is executed

When the command is executed, it seems that a `activateById` is called on the shell. In Notebook, this commands hide the side panel if it is already display (see https://github.com/jupyter/notebook/issues/7534 for more context).

We can see in the screen recording below that the command is executed, though, but the console needs to be reopened.

[record-2025-12-18_16.03.53.webm](https://github.com/user-attachments/assets/a09b60a2-5e15-4e59-9fd1-c38108557e4b)

app/package.json
```diff
@@ -226,7 +226,6 @@
         "@jupyter-notebook/notebook-extension": true,
         "@jupyter-notebook/terminal-extension": true,
         "@jupyterlab/application-extension": [
-          "@jupyterlab/application-extension:commands",
```
brichet: This extension is not enabled anymore, except for the `tree` view, because it already uses the `Ctrl + B` shortcut.
AFAIK the commands in this extension may not be necessary for the Notebook view (not sure it is necessary at all, most of them requires a `LabShell` to work).

brichet:
Reopening it, it has been closed by error by the description of #7799

packages/console-extension/src/index.ts
```diff
@@ -53,23 +76,58 @@ const opener: JupyterFrontEndPlugin<void> = {
...
+          // Some commands needs to be patched to be enabled.
+          // TODO: fix it upstream in jupyterlab.
+          app.commands
+            .listCommands()
+            .filter((id) => COMMANDS_TO_PATCH.includes(id))
+            .forEach((id) => {
+              (app.commands as any)._commands.get(id).isEnabled = () =>
+                tracker.currentWidget !== null &&
+                tracker.currentWidget.node.contains(document.activeElement);
+            });
```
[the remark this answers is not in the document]
brichet: Fixed.

packages/console-extension/src/index.ts
```diff
@@ -92,9 +129,121 @@ const redirect: JupyterFrontEndPlugin<void> = {
   },
 };
 
+/**
+ * Open consoles in the side panel.
+ */
+const scratchPadConsole: JupyterFrontEndPlugin<void> = {
+  id: '@jupyter-notebook/console-extension:scratch-pad',
```
jtpio: Should we rename to `scratchpad-console`? To match the command id below. And we would need to update the name of the settings too then.

Also `scratchpad` sounds more common for the spelling.


```suggestion
  id: '@jupyter-notebook/console-extension:scratchpad-console',
```

brichet:
Thanks @jtpio for the review, I updated the PR according to your comments.

> Wondering if we could use a more descriptive title in the menu entry when the scratchpad is open:

The menu entry uses the widget title, and the one from console is updated when a [property or kernel changes](https://github.com/jupyterlab/jupyterlab/blob/cc62c19e86b3df065ed27519def9465fbac36225/packages/console/src/panel.ts#L104-L105). If we use the same signal in Notebook, it will conflict and we can't know which one will be triggered first. An option could be to add an flag in the constructor of the console (in Jupyterlab), to disable the title update.

> Also when testing locally there seems to be an issue about not being able to change the prompt position when the scratchpad is open

In https://github.com/jupyterlab/jupyterlab/pull/18238, we added a way to enable the commands when the console is not in main area, but missed this use case. It uses the fact that the current console contains the `activeElement`. The menu to change the prompt position is attached to the root of the DOM, not to the node of the console. We should fix it upstream too (or make some kind of patch in Notebook, but I don't think we should go with this solution).

brichet:
> The menu entry uses the widget title, and the one from console is updated when a [property or kernel changes](https://github.com/jupyterlab/jupyterlab/blob/cc62c19e86b3df065ed27519def9465fbac36225/packages/console/src/panel.ts#L104-L105). If we use the same signal in Notebook, it will conflict and we can't know which one will be triggered first. An option could be to add an flag in the constructor of the console (in Jupyterlab), to disable the title update.

I opened https://github.com/jupyterlab/jupyterlab/pull/18331 to fix it.

> In https://github.com/jupyterlab/jupyterlab/pull/18238, we added a way to enable the commands when the console is not in main area, but missed this use case. It uses the fact that the current console contains the activeElement. The menu to change the prompt position is attached to the root of the DOM, not to the node of the console. We should fix it upstream too (or make some kind of patch in Notebook, but I don't think we should go with this solution).

I opened https://github.com/jupyterlab/jupyterlab/pull/18332 to fix it.

@jtpio do you think that we should wait for these PR to be merged and release before ? If so, we should set this PR as draft for now.

jtpio:
Ah thanks for opening the PRs upstream!

Yeah it's ok if this gets fixed later. We can track that in an issue and fix it when updating to the next JupyterLab 4.6.0a2.

jtpio (approved):
Thanks!

PR #7824, opened 2026-02-06
docs: Fix broken links in troubleshooting and migration docs

RamiNoodle733:
This PR fixes two broken documentation links:

1. troubleshooting.md: The jupyter_client link was incorrectly pointing to jupyter/jupyter_core instead of jupyter/jupyter_client

2. migrate_to_notebook7.md: The [JEP 79] reference was missing the link reference, causing a broken link (Markdown reference links are case-sensitive)

These are minor documentation fixes that improve the user experience when navigating the docs.


jtpio (approved):
Thanks!

PR #7834, opened 2026-02-22
ci: declare explicit token permissions for changelog publishing

Rohan5commit:
## Summary
Add explicit `GITHUB_TOKEN` permissions to `.github/workflows/publish-changelog.yml`:

- `contents: read`

## Why
The workflow currently relies on implicit token defaults. Declaring explicit read-only scope improves least-privilege posture and aligns with other workflows in this repository that already set explicit permissions.


goelakash:
Trivial security hardening, clearly correct and security best-practice, can be merged.

PR #7843, opened 2026-02-24
Prevent scratchpad console title to update

### LINKED ISSUE #7805: Fix scratchpad console when bumping jupyterlab to 4.6.0a2
> > The menu entry uses the widget title, and the one from console is updated when a [property or kernel changes](https://github.com/jupyterlab/jupyterlab/blob/cc62c19e86b3df065ed27519def9465fbac36225/packages/console/src/panel.ts#L104-L105). If we use the same signal in Notebook, it will conflict and we can't know which one will be triggered first. An option could be to add an flag in the constructor of the console (in Jupyterlab), to disable the title update.
> 
> I opened https://github.com/jupyterlab/jupyterlab/pull/18331 to fix it.
> 
> > In https://github.com/jupyterlab/jupyterlab/pull/18238, we added a way to enable the commands when the console is not in main area, but missed this use case. It uses the fact that the current console contains the activeElement. The menu to change the prompt position is attached to the root of the DOM, not to the node of the console. We should fix it upstream too (or make some kind of patch in Notebook, but I don't think we should go with this solution).
> 
> I opened https://github.com/jupyterlab/jupyterlab/pull/18332 to fix it.

_Originally posted by @brichet in https://github.com/jupyter/notebook/issues/7790#issuecomment-3744881240_

brichet:
Prevent the scratchpad console title to update.

## References

Fixes #7805

***Note**
The second issue of #7805 (about enabling the buttons to move the console prompt) has also been fixed when bumping jupyterlab to v4.6.0a3.*

[output.webm](https://github.com/user-attachments/assets/d7e9b1e2-4677-4f51-b7ef-e8fb474112b2)

## Code changes

Add a flag in the `'console:create'` commands, available after https://github.com/jupyterlab/jupyterlab/pull/18331, released in jupyterlab v4.6.0a3.

## User-facing changes

The title of the console is not updated in the menu `View>Right Sidebar`.

None

### DIFF
--- a/ui-tests/test/notebook.spec.ts
+++ b/ui-tests/test/notebook.spec.ts
@@ -237,6 +237,10 @@ test.describe('Notebook', () => {
 
     await waitForKernelReady(page);
 
+    // Wait for the first cell to be active
+    const firstCell = page.locator('.jp-Cell').first();
+    await expect(firstCell).toHaveClass(/jp-mod-active/);
+
     // run the two cells
     await page.keyboard.press('Shift+Enter');
     await page.keyboard.press('ControlOrMeta+Enter');


jtpio:
Ah forgot about that one before the release. Thanks!

brichet:
The failing test may be relevant, it fails reliably, but I can't reproduce it locally, and have no clue why it fails.
It seems that the notebook cells are not executed with `'Shift+Enter'` or `'ControlOrMeta+Enter'`.

jtpio:
> The failing test may be relevant, it fails reliably, but I can't reproduce it locally, and have no clue why it fails. It seems that the notebook cells are not executed with `'Shift+Enter'` or `'ControlOrMeta+Enter'`.

Restarted the check and it now passes.

It seems this check also fails on other PRs and `main` sometimes, so it's probably not related to this change.

PR #7882, opened 2026-03-31
Drop `lerna`

jtpio:
## References

Similar to https://github.com/jupyterlite/jupyterlite/pull/1843

Trying to see if we can do without `lerna` to help simplify the dev setup and remove many dependencies from the `yarn.lock`.

Closes #7828
Closes #7831

## Code changes

- [x] Drop `lerna` and `nx`
- [x] Fix custom script for bumping versions
- [x] Check if we need the `yarn workspaces` plugin -> **Let's do without for now**

## User-facing changes

None

None


jtpio:
Should be good to go now.

Will test everything works as expected with the next PRs and releases and adjust if needed.

PR #7893, opened 2026-04-21
Use `@jupyter/builder`

jtpio:
## References

Similar to https://github.com/jupyterlab/jupyterlab/pull/18723

## Code changes

Depend on `jupyter-builder` and  `@jupyter/builder` for building the application.

## User-facing changes

None

TBD

### DIFF
--- a/packages/lab-extension/package.json
+++ b/packages/lab-extension/package.json
@@ -56,6 +56,7 @@
   },
   "devDependencies": {
     "@jupyter/builder": "^0.0.7",
+    "@jupyterlab/builder": "~4.6.0-alpha.4",
     "rimraf": "^3.0.2",
     "typescript": "~5.9.3"
   },
--- a/packages/lab-extension/package.json
+++ b/packages/lab-extension/package.json
@@ -31,15 +31,15 @@
   ],
   "scripts": {
     "build": "jlpm run build:labextension:dev",
-    "build:labextension": "jupyter labextension build .",
-    "build:labextension:dev": "jupyter labextension build --development True .",
+    "build:labextension": "jupyter-builder build .",
+    "build:labextension:dev": "jupyter-builder build --development True .",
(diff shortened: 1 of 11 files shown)


jtpio:
This failure:

image

Appears to be caused by:

```
  File "/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/site-packages/jupyter_builder/federated_extensions.py", line 30, in <module>
    from tomllib import load  # Python 3.11+
ModuleNotFoundError: No module named 'tomllib'
```

packages/lab-extension/package.json
```diff
@@ -55,6 +55,7 @@
     "@lumino/disposable": "^2.1.5"
   },
   "devDependencies": {
+    "@jupyter/builder": "^0.0.7",
     "@jupyterlab/builder": "~4.6.0-alpha.4",
```
jtpio, line 59: TODO: remove

jtpio: Ah we still need it apparently it seems:

```
/home/runner/.local/share/hatch/env/virtual/notebook/yXCDlSUQ/hatch-build/lib/python3.12/site-packages/jupyter_builder/debug_log_file_mixin.py:55: UserWarning: ValueError: Extensions require a devDependency on @jupyterlab/builder@^4.6.0-alpha.3
```

jtpio: To investigate as part of https://github.com/jupyterlab/jupyter-builder/issues/81 before removing the dependency.

jtpio:
cc @Darshan808 @krassowski does this look like a good first step already?

It follows the same approach as the JupyterLab PR. Probably https://github.com/jupyterlab/jupyter-builder/issues/81 will also help remove the remaining dependency on `@jupyterlab/builder` later (can be tracked separately).

krassowski (approved):
Make sense to me

PR #7895, opened 2026-04-23
Update to `jupyter-builder==0.0.8`, drop `@jupyterlab/builder`

jtpio:
## References

Follow-up to https://github.com/jupyter/notebook/pull/7893

## Code changes

- [x] Update to https://github.com/jupyterlab/jupyter-builder/releases/tag/v0.0.8
- [x] Drop `@jupyterlab/builder`

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -192,7 +192,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install -U "jupyter-builder>=0.0.8"
+          python -m pip install -U "jupyter-builder>=0.0.8,<0.1.0a0"
           jlpm
 
       - name: Measure bundle sizes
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -192,7 +192,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install -U "jupyter-builder>=0.0.7"
+          python -m pip install -U "jupyter-builder>=0.0.8,<0.1.0a0"
```
jtpio, line 195: Maybe these pre-releases could be yanked on PyPI if not used anymore?

cc @Darshan808 @krassowski if you have an opinion, otherwise it's fine to keep a pin here in the meantime

Darshan808: Agreed. We should yank those.

PR #7903, opened 2026-04-30
fix(application): anchor opener route regex for tree paths

### LINKED ISSUE #7216: UI issue in the "tree view" if directory called "notebooks" is in the path.

The-Peacemaker:
## References

Fixes #7216

## Code changes

- Anchor the opener route regex in the application extension so it only matches true notebook/editor routes:
  - from `/(notebooks|edit)/(.*)`
  - to `^/(notebooks|edit)/(.*)$`
- Add a UI regression test in `ui-tests/test/tree.spec.ts` covering tree navigation/reload when the path contains a directory literally named `notebooks`.

## Problem / root cause

Tree URLs such as `/tree/test/notebooks/test` were being incorrectly matched by the opener route regex because it was not anchored. That caused the router to treat part of a tree URL as a notebook route and attempt to open a file, leading to `File Load Error: Invalid response: 404 Not Found`.

## Solution

Use an anchored regex so the opener only handles paths that begin with `/notebooks/` or `/edit/`.

## User-facing changes

Users can navigate and refresh tree paths containing directories named `notebooks` without spurious file load errors.

None.

### DIFF
--- a/packages/application-extension/src/index.ts
+++ b/packages/application-extension/src/index.ts
@@ -73,7 +73,7 @@ import { Menu, Widget } from '@lumino/widgets';
 /**
  * A regular expression to match path to notebooks and documents
  */
-const TREE_PATTERN = new RegExp('/(notebooks|edit)/(.*)');
+const TREE_PATTERN = new RegExp('^/(notebooks|edit)/(.*)$');
 
 /**
  * A regular expression to suppress the file extension from display for .ipynb files.
--- a/ui-tests/test/tree.spec.ts
+++ b/ui-tests/test/tree.spec.ts
@@ -53,6 +53,28 @@ test('Should redirect from notebooks route to tree route for directories', async
   expect(url.pathname).toEqual(`/tree/${dir}`);
 });
 
+test('should not show a file load error when path contains notebooks', async ({
+  page,
+  tmpPath,
+}) => {
+  const nestedPath = `${tmpPath}/test/notebooks/test`;
+  await page.contents.createDirectory(`${tmpPath}/test`);
+  await page.contents.createDirectory(`${tmpPath}/test/notebooks`);
+  await page.contents.createDirectory(nestedPath);
+
+  await page.goto(`tree/${nestedPath}`);
+  await page.waitForSelector('.jp-FileBrowser-crumbs >> text=/test/');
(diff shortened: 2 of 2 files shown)


goelakash:
I reproduced the existing bug locally:

```
mkdir -p /tmp/jupyter-test/test/notebooks/test
cd /tmp/jupyter-test
jupyter notebook --no-browser --ServerApp.token='' --port=8889
```

Navigating to http://localhost:8889/tree/test/notebooks/test shows "File Load Error for test" -> the app tries to open test as a notebook file instead of showing the directory listing.

The issue is the unanchored regex:

const TREE_PATTERN = new RegExp('/(notebooks|edit)/(.*)');

This matches /(notebooks|edit)/ anywhere in the URL. So /tree/test/notebooks/test matches the substring /notebooks/test in the middle, and the router thinks you're trying to open a file called test.

The fix anchors it with ^ so only URLs that actually start with /notebooks/ or /edit/ are matched — tree paths with a directory named notebooks are no longer intercepted.

The unit-test also looks like it reproduces to catch the same issue. LGTM (UI test failure looks unrelated).

The-Peacemaker:
Hi @jtpio, this PR is ready for review. All CI checks pass except
`enforce-label` which requires a triage label. Could you please
add the `bug` label? Thanks!

The-Peacemaker:
Friendly ping @jtpio — rebased onto latest main, no conflicts.

All checks green except enforce-label (needs bug triage label).

Fix anchors TREE_PATTERN to ^/(notebooks|edit)/(.*)$ so /tree/.../notebooks/... no longer misroutes, + UI regression test in tree.spec.ts. @goelakash already reproduced + LGTM.

Fixes #7216. Could you add the label / review when you get a chance? Thanks!

PR #7906, opened 2026-05-05
Ignore .jupyter directory (local config files)

arllemfarias:
This directory is automatically created by the Jupyter system to store user-specific settings, security tokens, and server configurations. It should not be tracked as it contains environment-specific data.


goelakash:
@arllemfarias What commands/operations create the `.jupyter` dir? Seems relevant to provide that information

PR #7911, opened 2026-05-09
Bump `@jupyter/eslint-plugin` to `v1`

Darshan808:
## Code changes

- Bump jupyter eslint plugin to `v1`
- Fix all failing lint issues

## User-facing changes
None

None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -241,7 +241,6 @@ jobs:
           pipx run doc8 --max-line-length=200 docs/source *.md
           python -m pip install -U "jupyter-builder>=0.0.8,<0.1.0a0"
           jlpm
-          jlpm build:lib
           jlpm eslint:check
           jlpm prettier:check
           jlpm build:utils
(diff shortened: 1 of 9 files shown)


Darshan808:
Not sure why these false positives are showing up in CI.
looking into it.

.github/workflows/build.yml
```diff
@@ -241,6 +241,7 @@ jobs:
           pipx run doc8 --max-line-length=200 docs/source *.md
           python -m pip install -U "jupyter-builder>=0.0.8,<0.1.0a0"
           jlpm
+          jlpm build:lib
```
jtpio: Wondering what now requires adding the build command?

Darshan808: There was an issue with `@jupyter/eslint-plugin` which was failing the lint check in [ci](https://github.com/jupyter/notebook/actions/runs/25597991909/job/75147105965) but it got fixed in `v1.0.1` and we can remove this patch.

jtpio (approved):
Thanks!

Issue #6322, opened 2022-03-21
Add more UI tests


jtpio:
We should add a couple of new UI tests, ideally alongside changes in PRs.

This issue tracks some checks we should add to:

- [ ] check support for other factories: https://github.com/jupyter/notebook/pull/6315
- [x] check Markdown is rendered correctly: https://github.com/jupyter/notebook/pull/6318
- [ ] check the file name searcher can be added to the file browser toolbar via the settings: https://github.com/jupyter/notebook/pull/6757
- [ ] check micro toolbars are not rendered for terminals and files: https://github.com/jupyter/notebook/pull/6791 
- [ ] snapshots for the file editor: https://github.com/jupyter/notebook/pull/6804

This replaces the issue opened in RetroLab: https://github.com/jupyterlab/retrolab/issues/228

goelakash:
@jtpio Raised #7930 and #7931 to address some of the checklist items.

PR #7930, opened 2026-05-24
Add UI test verifying no cell toolbar in file editor

goelakash:
## References
#6322 (checklist item: "check micro toolbars are not rendered for terminals and files")

## Code changes
Adds a Playwright test that opens a file in the text editor and asserts that no `.jp-cell-toolbar` element is rendered.

## User-facing changes
None - this is a test.

None.


jtpio (approved):
Thanks!

PR #7931, opened 2026-05-24
Add UI test for opening file with editor factory

goelakash:
## References
#6322 (checklist item: "check support for other factories")

## Code changes
Adds a Playwright test that opens a `.ipynb` file via the `/edit/` route and verifies it renders in the text editor (CodeMirror) rather than the notebook widget.

## User-facing changes
None - this is a test.

None.

### DIFF
--- a/ui-tests/test/editor.spec.ts
+++ b/ui-tests/test/editor.spec.ts
@@ -8,6 +8,7 @@ import { test } from './fixtures';
 import { expect } from '@jupyterlab/galata';
 
 const FILE = 'environment.yml';
+const NOTEBOOK = 'empty.ipynb';
 
 test.use({ autoGoto: false });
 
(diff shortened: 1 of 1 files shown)


ui-tests/test/editor.spec.ts
```diff
@@ -50,6 +50,19 @@ test.describe('Editor', () => {
     expect(url).toContain(newName);
   });
 
+  test('Should open a file in the text editor via the edit route', async ({
+    page,
+    tmpPath,
+  }) => {
+    const file = `${tmpPath}/${FILE}`;
```
jtpio: Is this testing the opening a notebook with a different factory?

Looks like the test file is a `yml` file, not a notebook?

https://github.com/jupyter/notebook/blob/db7b8ab0a1322b9ff79476323ab45ad5ff613e1e/ui-tests/test/editor.spec.ts#L10

goelakash: Good point - i think i reused the wrong file. Let me look for a notebook file.

goelakash: I think i got confused - the /edit/ route is for text files. To open a notebook with the
  editor factory, we would need a different mechanism, not just the /edit/ URL.

jtpio: The `/edit` route supports the `?factory=` query argument, for example `/edit/binder/example.ipynb?factory=Editor`

goelakash: @jtpio I see - in that case I think it would be best if I updated the test for the notebook case.

goelakash:
@jtpio all checks are passing now, hopefully this can be merged

jtpio (approved):
Thanks!

PR #7934, opened 2026-05-28
Fix flaky UI tests with proper kernel and execution waits

goelakash:
## References
#6322 (general UI test reliability)

## Code changes
- **links.spec.ts**: Added `waitForKernelReady` before clicking rendered notebook links that open in new tabs. The popup event was timing out because the markdown cell content hadn't rendered yet.
- **notebook.spec.ts (autoscroll test)**: Replaced `waitForSelector('.jp-NotebookCheckpoint')` with `waitForKernelReady` before executing cells. The checkpoint indicator doesn't guarantee the kernel is ready to accept execution.
- **notebook.spec.ts (keyboard shortcut test)**: Wait for each cell's output to appear after execution before toggling visibility with `O`. Previously, the toggle was sent before the kernel had finished producing output.

## User-facing changes
None - these are test fixes.

None.

## Test plan
- [x] Ran previously-flaky tests locally with `--repeat-each=3` (9/9 passed)
- [x] Full UI test suite passes locally (35 passed, 0 failed)

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -47,12 +47,24 @@ jobs:
       - name: Install the package
         run: |
           cd dist
-          python -m pip install -vv notebook*.whl
+          python -m pip install notebook*.whl
 
       - name: Install the test dependencies
         run: |
           cd ui-tests
           jlpm
+
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/ui-tests.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       - name: Install the package
         run: |
           cd dist
-          python -m pip install -vv notebook*.whl
+          python -m pip install notebook*.whl
```
goelakash, line 50: this forced pip to log all possible versions, which causes tests to slow down in case of version mismatch i believe

PR #7943, opened 2026-06-06
Rename File menu items to match Classic Notebook v6 (#6398)

goelakash:
Renames five File menu items to match the v6 wording. See https://github.com/jupyter/notebook/issues/6398#issuecomment-4637268360 for the full diff and per-row reasoning.

- Duplicate → Make a Copy…
- Save and Export Notebook As → Download as
- Close and Shut Down Notebook → Close and Halt

Before
Screenshot 2026-06-06 at 3 18 48 PM

Now
Screenshot 2026-06-06 at 3 22 40 PM


goelakash:
@jtpio requesting a label and rerunning the CI, thanks!

goelakash:
Hi! Could a maintainer please apply an `enhancement` (or appropriate) triage label and re-run failed jobs? The 4 failing checks all look unrelated to this PR's changes:

- `enforce-label` — needs a label applied (out of my reach).
- `check_links` — GitHub returned 502 on a contributors-graph URL in the existing CHANGELOG.
- `ui-tests (chromium)` — the `mobile.spec.ts` snapshot drift and `links.spec.ts` popup timeout that have been flaky on `main` for a while.
- `tests_check` — meta-job, fails downstream of the above.

Thanks!

krassowski:
> Close and Shut Down Notebook → Close and Halt

I think a change is needed here, because we are not actually shutting down a notebook but shutting down a kernel. What was the functionality of `Close and Halt` in v6? Was it shutting down the notebook application altogether or closing the notebook and shutting down the kernel?

krassowski:
> Duplicate → Make a Copy…

The design guide says to use … when there is a follow-up action. There is no follow up action here.

Notebook v7 was out for 3 years now. If there was no legacy of notebook v6 would we choose "Make a Copy" over duplicate?

It looks to me like the action implemented is "Duplicate", not "Make a copy". See the distinction discussed in https://discussions.apple.com/thread/1866218 and per Google AI overview:

image

Clicking this menu item will open the duplicated file immediately.

krassowski:
> Save and Export Notebook As → Download as

Again, I mentioned the genesis of this naming in the issue https://github.com/jupyter/notebook/issues/6398.

krassowski:
@goelakash I really your contributions and I think this was a good issue to explore. Genuinely thank you!

I just wonder, if we were to make these changes, should we consider what is happening at a bit deeper level and adjust _both_ the implementation and the naming so that user is not surprise? To me changing labels only to match an older version without changing the subtility of implementation is not necessarily worth it at this point.

jtpio:
> I just wonder, if we were to make these changes, should we consider what is happening at a bit deeper level and adjust _both_ the implementation and the naming so that user is not surprise? To me changing labels only to match an older version without changing the subtility of implementation is not necessarily worth it at this point.

The main motivation for aligning the menu items with the classic notebook was to ease the transition to the new interface (for example with existing screenshots in course materials, so the existing ones would still be relevant).

But since Notebook 7 was released almost 3 years ago already. I've also been wondering if this may still be needed or if users have generally adapted to the slightly different menu items since?

goelakash:
Thanks for the comments guys, I think I mostly agree with the callouts that newer menus are generally an improvement.
@jtpio @krassowski  Actually I was not aware that v7 is a few years old, should've gone back and seen the release timeline.
In either case - if this PR doesn't address a problem - I'm fine with closing it as is.
Lmk

PR #8028, opened 2026-08-11
Avoid mutating shared server flags

jtpio:
## References

Minor server side code cleanup

## Code changes

- [x] Avoid mutating shared server flags

## User-facing changes

Limited.

None

### DIFF
--- a/notebook/app.py
+++ b/notebook/app.py
@@ -8,7 +8,6 @@ import typing as t
 from pathlib import Path
 
 from jupyter_client.utils import ensure_async  # type:ignore[attr-defined]
-from jupyter_core.application import base_aliases
 from jupyter_core.paths import jupyter_config_dir
 from jupyter_server.base.handlers import JupyterHandler
 from jupyter_server.extension.handler import (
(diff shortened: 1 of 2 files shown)


notebook/app.py
```diff
@@ -271,7 +268,7 @@ class JupyterNotebookApp(NotebookConfigShimMixin, LabServerApp):  # type:ignore[
         """,
     )
 
-    flags: Flags = flags  # type:ignore[assignment]
+    flags: Flags = dict(flags)  # type:ignore[assignment]
```
brichet, line 271: @jtpio on my side `--expose-app-in-browser` doesn't work anymore, but  `--JupyterNotebookApp.expose_app_in_browser=True` still works.

```bash
$ jupyter notebook --expose-app-in-browser
usage: jupyter-notebook [-h] [--debug] [--show-config] [--show-config-json] [--generate-config] [-y] [--allow-root] [--no-browser] [--autoreload] [--script] [--no-script] [--log-level ServerApp.log_level] [--config ServerApp.config_file] [--ip ServerApp.ip] [--port ServerApp.port]
                        [--port-retries ServerApp.port_retries] [--sock ServerApp.sock] [--sock-mode ServerApp.sock_mode] [--transport KernelManager.transport] [--keyfile ServerApp.keyfile] [--certfile ServerApp.certfile] [--client-ca ServerApp.client_ca] [--notebook-dir ServerApp.root_dir]
                        [--preferred-dir ServerApp.preferred_dir] [--browser ServerApp.browser] [--pylab ServerApp.pylab] [--gateway-url GatewayClient.url]
                        [extra_args ...]
jupyter-notebook: error: unrecognized arguments: --expose-app-in-browser
```
I wonder if this change is not the culprit, the flags may not be known by the `serverapp` anymore, as this in now a local copy instead of the upstream dict.

jtpio: Ah yes, this is very likely the cause thanks for catching that. Will look into this, and ideally we could also add a couple of tests to avoid such regressions.

PR #8034, opened 2026-08-15
Bump `@jupyter/eslint-plugin` to `^1.1.0`

Darshan808:
## References 
https://github.com/jupyterlab/eslint-plugin/releases/tag/v1.1.0

## Description
Updates the linter to `^1.1.0` and fixes all the lint issues that were surfaced.

`newMenu` case isn't an actual leak but the linter flagged it because the handoff to `viewMenu.addItem()` was inside `if (menuToAdd)`.
Fix is to materialize the widget list up front and bail early if it's empty, so the `addItem` call is unconditional. Drops the `menuToAdd` flag too, and we no longer build a menu we'd throw away.


jtpio (approved):
Thanks!

PR #8038, opened 2026-08-17
Only show the Editor Syntax Highlighting menu in the file editor

### LINKED ISSUE #6825: v7: No option selectable in Editor Syntax highlighting.
I'm not sure if it's a bug or not, either something is wrong or the menu should be removed ?

Screen Shot 2023-04-05 at 11 23 27

jtpio:
## References

Fixes https://github.com/jupyter/notebook/issues/6825

## Code changes

- [x] Only show the Editor Syntax Highlighting menu in the file editor
- [x] Test

## User-facing changes

The menu is now hidden on the notebook page:

image

It sill shows up as expected for text editors:

image

None


brichet:
Thanks @jtpio for working on this.

There a plugin that disables some menu items depending on the page.
https://github.com/jupyter/notebook/blob/597c3cfc855f701a2b94858de67e10602ffabe7a/packages/application-extension/src/index.ts#L302

I wonder if we could disable the menu entry from there, instead of running `hideSyntaxHighlightingMenu` for each fetched plugin.

jtpio:
Right, this plugin was initially mostly meant to disable entire menus depending on the page in a programmatic manner. If we want to disable individual entries we probably need to react on the setting registry `pluginChanged` (or similar).

> instead of running `hideSyntaxHighlightingMenu` for each fetched plugin.

The existing `overrideDefaults` was already running for each plugin right? This new `hideSyntaxHighlightingMenu` is mostly adding a page config lookup and a string comparison to check the correct plugin.

jtpio:
Otherwise we need to have similar plugins as the existing one, in different packages so they only activate on the pages where their extension is included, and use the declarative settings system to disable individual entries.

brichet:
> The existing overrideDefaults was already running for each plugin right? This new hideSyntaxHighlightingMenu is mostly adding a page config lookup and a string comparison to check the correct plugin.

Yes, the difference is that `overrideDefaults` is generic, and it can apply to several plugins, while `hideSyntaxHighlightingMenu` is specific to one menu only. This is why I was wondering if there were a more specific way to handle it.

brichet (approved):

PR #8039, opened 2026-08-18
Enable opening notebooks without a kernel

### LINKED ISSUE #3170: Feature request: open a notebook without running kernel

jtpio:
## References

Fixes #3170 

Build on top of the upstream JupyterLab changes.

## Code changes

- [x] Allow opening notebooks without a kernel
- [x] Test

## User-facing changes

https://github.com/user-attachments/assets/6003b2d0-9bc8-4162-9494-a5cd8c14c79b

None

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -328,6 +328,9 @@
           "@jupyterlab/filebrowser-extension:settings",
           "@jupyterlab/filebrowser-extension:share-file"
         ],
+        "@jupyterlab/notebook-extension": [
+          "@jupyterlab/notebook-extension:open-with-no-kernel"
+        ],
         "@jupyter-notebook/tree-extension": true,
         "@jupyterlab/running-extension": [
           "@jupyterlab/running-extension:plugin"
--- a/packages/application-extension/src/index.ts
+++ b/packages/application-extension/src/index.ts
@@ -273,7 +273,7 @@ const opener: JupyterFrontEndPlugin<void> = {
             urlParams.get('kernel') === 'none'
               ? { shouldStart: false, shouldReuse: false }
               : undefined;
-          docManager.openOrReveal(
+          const widget = docManager.openOrReveal(
             file,
             factory,
             undefined,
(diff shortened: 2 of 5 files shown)


Daniele-Cangi:
Hi @jtpio, I tested this branch locally and found two edge cases around carrying the no-kernel state between the tree page and the notebook tab.

The first is that `shouldStart === false` is currently enough to serialize `kernel=none`, while the receiving page expands that back to `{ shouldStart: false, shouldReuse: false }`. This changes the semantics of `notebookStartsKernel=False`: it should prevent auto-starting a kernel, but still allow reconnecting to an existing session.

The second is after explicitly opening with **Notebook (no kernel)**. If a kernel is then started, `kernel=none` remains in the URL. Reloading the page therefore applies the no-reuse preference again instead of reconnecting to the running session.

I added regression coverage and a small fix on top of the current PR head (`e2ef70022df07f2b15c2b1420dc124e4896148ef`):

https://github.com/Daniele-Cangi/notebook/commit/40217c3cb0381a381b927ead4eb240f1db69c66c

The two regression tests fail on the original PR head and pass with the patch. The full Chromium `filebrowser.spec.ts` suite also passes (9/9). The reload case additionally verifies that it reconnects to the same session ID and that only one server session remains for the notebook.

Happy to open this as a small PR against your feature branch if that would be useful, or you can cherry-pick the commit directly.

jtpio:
Nice thanks @Daniele-Cangi for helping on that front!

Daniele-Cangi:
You’re welcome! The commit is available if you’d like me to open a PR against the feature branch

jtpio:
The commit should be included in that branch now (https://github.com/jupyter/notebook/pull/8039/commits/abb1e713abdee8fe0a373e47b06cb0e35d967c97), thanks!

jtpio:
@Daniele-Cangi does the current state look good to you? If yes happy to merge so it's available in the next 7.7 pre-release for testing.

Daniele-Cangi:
Yes, I replayed the checks on the current head (bb3efbe). Both no-kernel regression tests pass, including session reuse and URL/history preservation across reloads, and the full Chromium filebrowser.spec.ts suite passes 9/9. The latest simplification does not introduce regressions compared with the original patch. Looks good to merge for the 7.7 pre-release.

jtpio:
Good thanks, let's get this one in then so it's available in the next pre-release.

PR #8045, opened 2026-08-29
Add New Folder to File > New menu (#8044)

### LINKED ISSUE #8044: [Enhancement Request] `File -> New -> New Folder` menu item
Good day!

`New` menu does contain `New Folder` item.

`File -> New` _doesn't_.

The Enhancement Request:
to implement `File -> New -> New Folder` menu item.

yogiwaran2003:
## References
Closes #8044

## Code changes
The `filebrowser:create-new-directory` command already existed and was wired into the file browser toolbar's "New" dropdown (in `packages/tree-extension/src/index.ts`), but it was never registered with the top menu bar's `File > New` submenu (`IMainMenu.fileMenu.newMenu`).

This PR adds `IMainMenu` as an optional dependency to the `createNew` plugin in `packages/tree-extension/src/index.ts` and registers the existing `filebrowser:create-new-directory` command to `mainMenu.fileMenu.newMenu`, so it now appears alongside Console, Notebook, Terminal, Text File, Markdown File, and Python File.

No new commands were created — this only adds a menu registration for an existing command.

## User-facing changes
"New Folder" now appears as an item in `File > New`, in addition to the existing toolbar "New" dropdown in the file browser. It's registered with a high rank so it appears last in the list, consistent with where JupyterLab places it in its own File menu.

Before: `File > New` only listed Console, Notebook, Terminal, Text File, Markdown File, and Python File — no way to create a folder from this menu.
Screenshot from 2026-08-27 15-32-42

After: `File > New > New Folder` creates a new folder in the current directory, matching the existing toolbar behavior.
Screenshot from 2026-08-27 15-55-50

None. This is an additive change — no existing commands, APIs, or menu items were modified or removed.

### DIFF
--- a/packages/tree-extension/src/index.ts
+++ b/packages/tree-extension/src/index.ts
@@ -70,17 +70,20 @@ namespace CommandIDs {
  * Plugin to add extra commands to the file browser to create
  * new notebooks, files, consoles and terminals
  */
+import { IMainMenu } from '@jupyterlab/mainmenu';
+
 const createNew: JupyterFrontEndPlugin<void> = {
   id: '@jupyter-notebook/tree-extension:new',
   description:
     'Plugin to add extra commands to the file browser to create new notebooks, files, consoles and terminals.',
   requires: [ITranslator],
-  optional: [IToolbarWidgetRegistry],
+  optional: [IToolbarWidgetRegistry, IMainMenu],
   autoStart: true,
   activate: (
     app: JupyterFrontEnd,
     translator: ITranslator,
-    toolbarRegistry: IToolbarWidgetRegistry | null
+    toolbarRegistry: IToolbarWidgetRegistry | null,
+    mainMenu: IMainMenu | null
   ) => {
     const { commands, serviceManager } = app;
     const trans = translator.load('notebook');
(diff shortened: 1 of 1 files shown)


yogiwaran2003:
The `check_links` job failed on an unrelated pre-existing link in `docs/source/migrating/custom-themes.md` (`https://github.com/topics/jupyterlab-theme`), with a `429: Too Many Requests` response — this looks like GitHub rate-limiting the CI runner rather than a genuinely broken link. This file wasn't touched by this PR. All other required checks (`coverage`, `install`, `test_lint`, `test_docs`, `test_minimum_versions`, `test_prereleases`) passed.

I don't have permission to re-run the workflow from my end — could a maintainer re-trigger the `check_links` job when convenient? Happy to make any other changes needed in the meantime.

packages/tree-extension/src/index.ts
```diff
@@ -135,6 +138,16 @@ const createNew: JupyterFrontEndPlugin<void> = {
         }
       );
     }
+
+    // Add "New Folder" to the top File > New menu.
+    // (Notebook, Console, Terminal, and Text File are already registered
+    // there by other plugins/extensions — this adds the missing entry.)
+    if (mainMenu) {
+      mainMenu.fileMenu.newMenu.addItem({
```
jtpio, line 146: Instead of directly using `IMainMenu` here, maybe we should look into adding this new entry via the setting system just like other plugins:

https://github.com/jupyter/notebook/blob/main/packages/console-extension/schema/scratchpad-console.json#L4-L24

This allows users to override the preference and hide the entry if they want to, which is more flexible.

jtpio (changes requested):
Thanks for looking into this.

Sounds good overall, but would be worth looking into using the same approach as other plugins.

PR #8071, opened 2026-09-22
Remove pixi in favor of the standard Python and JavaScript tooling

jtpio:
## References

The main motivation for Pixi was mostly support for codespaces: https://github.com/jupyter/notebook/pull/7602

However this adds quite a bit of weight and complexity to the repo for day-to-day development, and may overall not be worth it.

## Code changes

- [x] Drop the pixi setup
- [x] Keep support for codespaces with a simpler approach 

## User-facing changes

None for users.

Contributors should still be able to use pixi if they want to.

None


jtpio:
Support for codespaces is still preserved, for example after creating a new codepace from that branch:

image

